PR #92, opened 2024-06-20
fix: concurrency issues with fail and addr

Freyskeyd:
# Description

`cargo test` were sometimes failing due to concurrency issue that are covered and backed by `nextest`.
This PR fixes the issue regarding the `addr` already in used discovered by @iljakuklic
It also fixes a concurrency issue with `fail`.

Why it does not fail on the CI ?
We're using `nextest`, which is a more advance and more performant test runner. Clusterisation/concurrency are handled differently between `cargo-test` and `nextest`.

### What to do

We can either enforce having both `nextest` and `cargo-test` tested on the CI or advertise that we only support nextest.

- Supporting both can be slower on a day to day as we'll have to replicate some logic provided by nextest or ignore some tests.
- Not providing `cargo-test` full support can lead users to not understand why some tests are failing.

My two-cents is that we could define clearly (in the README) that tests should be run with `nextest`, in order to not slow down the development and keep the important feature of nextest.

## PR Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,7 +52,7 @@ jobs:
             cache-on-failure: true
 
       - name: Test
-        run: cargo nextest run --workspace
+        run: cargo nextest run --workspace && cargo test --doc --workspace
 
   pre-e2e:
     name: E2E Tests | Docker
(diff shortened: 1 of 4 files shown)


iljakuklic:
Works well for me (with `nextest`)

muursh (approved):
Looks fine to me.

>My two-cents is that we could define clearly (in the README) that tests should be run with nextest, in order to not slow down the development and keep the important feature of nextest.

Agreed on this btw.

Freyskeyd:
> Looks fine to me.
> 
> > My two-cents is that we could define clearly (in the README) that tests should be run with nextest, in order to not slow down the development and keep the important feature of nextest.
> 
> Agreed on this btw.

Will create a ticket to upgrade the README or even create a CONTRIBUTING doc

Issue #104, opened 2024-06-11
Use Initial State + Certificates as inputs of the pessimistic proof


hadjiszs:
For now, the pessimistic proof input is a collection of `Batch` (which we renamed as `Certificate` to avoid conflicting with the CDK `Batch`) that comes with their own `LocalBalanceTree` and `LocalExitTree`.

**What we want instead:**
- Inputs: `State` + `Vec<Certificate>` as defined [here](https://potential-couscous-4gw6qyo.pages.github.io/protocol/workflow_centralized.html)

This is necessary for the integration in `agglayer-rs` which will provide to the prover the two current global trees along with the Certificates of the current Epoch.

PR #124, opened 2024-07-01
feat: adapt the pessimistic proof to take initial state + certificate

hadjiszs:
# Description

Intermediate iteration to adapt the API of the pessimistic proof to take the initial state of the sending chain + one Certificate in order to generate the leaf proof.

Closes https://github.com/agglayer/agglayer/issues/104

## PR Checklist:


crates/pessimistic-proof/Cargo.toml
```diff
@@ -18,4 +18,5 @@ sp1-derive = { git = "https://github.com/succinctlabs/sp1", tag = "v1.0.8-testne
 
 [dev-dependencies]
 hex = "0.4.3"
+lazy_static = "1.4.0"
```
Freyskeyd, line 21:
```suggestion
lazy_static.workspace = true
```

crates/pessimistic-proof/src/certificate.rs
```diff
@@ -1,71 +1,35 @@
...
     pub origin_network: NetworkId,
-    /// Initial local exit tree
-    pub prev_local_exit_tree: LocalExitTree<Keccak256Hasher>,
     /// Initial local exit root
     pub prev_local_exit_root: Digest,
-    /// Initial balance tree
-    pub prev_local_balance_tree: BalanceTree,
     /// Set of bridge exits
     pub bridge_exits: Vec<BridgeExit>,
+    /// Set of imported bridge exits
```
Freyskeyd, line 14: Could we have more doc about those imported_bridge_exits?

crates/pessimistic-proof/src/global_state.rs
Freyskeyd, line 1: why does this file is named `global_state` while not having a single `Global` struct ? Shouldn't it be `local_state` ?

crates/pessimistic-proof/src/global_state.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Eventually, this state will be entirely tracked by the networks themselves.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct LocalNetworkState {
+    /// Commitment to the [`BridgeExit`].
+    pub exit_tree: LocalExitTree<Keccak256Hasher>,
+    /// Commitment to the balance for each token.
+    pub balance_tree: BalanceTree,
+    /// Commitment to the imported [`BridgeExit`].
+    #[allow(dead_code)]
+    pub nullifier_tree: BTreeMap<NetworkId, Vec<BridgeExit>>,
```
Freyskeyd, line 24: Can we have a dedicated struct or type for the `NullifierTree` ?

crates/pessimistic-proof/src/global_state.rs
```diff
@@ -0,0 +1,67 @@
...
+impl LocalNetworkState {
+    /// Apply the [`Certificate`] on the current [`State`].
+    /// Returns the commitment on the resulting state if successful.
+    pub fn apply_certificate(
+        &mut self,
+        certificate: Certificate,
+    ) -> Result<(ExitRoot, BalanceRoot), ProofError> {
+        // Apply on Exit Tree
+        {
+            let computed_root = self.exit_tree.get_root();
```
Freyskeyd, line 36: Could we not compute the exit root each time but having the `local_exit_root` as a value of `LocalNetworkState` ?

crates/pessimistic-proof/src/global_state.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        // Apply on Balance Tree
+        {
+            for bridge_exit in certificate.bridge_exits {
+                self.balance_tree.withdraw(bridge_exit.token_info, bridge_exit.amount);
+            }
+
+            // Check whether the sender has some debt
+            if self.balance_tree.has_debt() {
+                return Err(ProofError::HasDebt {
```
Freyskeyd, line 57: At this point, there is no rollback of the leaf added above ?

crates/pessimistic-proof/src/global_state.rs
```diff
@@ -0,0 +1,67 @@
...
+                    network: certificate.origin_network,
+                });
+            }
+        };
+
+        // TODO: Apply on the nullifier tree
+
+        Ok((self.exit_tree.get_root(), self.balance_tree.hash()))
```
Freyskeyd, line 65: If we put the `local_exit_root` as a field of this, we do not need to return it.

crates/pessimistic-proof/src/global_state.rs
hadjiszs: I am removing this placeholder, I'll add all the nullifier tree related things in another PR

crates/pessimistic-proof/src/global_state.rs
hadjiszs: This would be applied on a temporary/copied state: the one received by the prover. If the prover ends up on such error, then the state on the sending side remains untouched because of the wrong state transition

I guess that eventually we could have differences between the native and the prover computation, in which case we could try some rollback mechanisms on the native computation if we think it to be faster to check whether a certificate is correct before submitting it to the prover, for now I believe that simple copy is fine

crates/pessimistic-proof/src/global_state.rs
hadjiszs: This is the code that will be run by the prover, so we want to ensure that the initial state is properly pegged to a given initial root, so in this computation we do want to compute the initial root based on the initial local exit tree

crates/pessimistic-proof/src/global_state.rs
hadjiszs: yep indeed :ok_hand: I have renamed it

crates/pessimistic-proof/src/certificate.rs
hadjiszs: Added more docs

crates/pessimistic-proof/src/global_state.rs
hadjiszs: Those are the public outputs of the sp1 program committing to the resulting state (that won't be the full `LocalNetworkState`)

crates/pessimistic-proof/src/certificate.rs
```diff
@@ -1,71 +1,44 @@
...
     pub prev_local_exit_root: Digest,
-    /// Initial balance tree
-    pub prev_local_balance_tree: BalanceTree,
-    /// Set of bridge exits
+    /// The set of bridge exits emitted by the origin network.
     pub bridge_exits: Vec<BridgeExit>,
+    /// The set of imported bridge exits for which the origin network is the target.
+    pub imported_bridge_exits: Vec<BridgeExit>,
+    /// The signature that commits to the state transition.
+    pub signature: (),
```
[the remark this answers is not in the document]
hadjiszs: Yeah true, I just removed it and will introduce that properly on a dedicated PR, same as for the nullifier tree

crates/pessimistic-proof-program/src/main.rs
```diff
@@ -1,13 +1,14 @@
...
 
 sp1_zkvm::entrypoint!(main);
 
 pub fn main() {
-    let certificates = sp1_zkvm::io::read::<Vec<Certificate>>();
+    let initial_state = sp1_zkvm::io::read::<LocalNetworkState>();
+    let certificate = sp1_zkvm::io::read::<Certificate>();
 
-    let new_roots = generate_full_proof(&certificates).unwrap();
+    let new_roots = generate_leaf_proof(initial_state, certificate).unwrap();
```
Freyskeyd, line 11: The `unwrap` is ok in this context?

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -0,0 +1,62 @@
...
+                self.balance_tree.withdraw(bridge_exit.token_info, bridge_exit.amount);
+            }
+
+            // Check whether the sender has some debt
+            if self.balance_tree.has_debt() {
+                return Err(ProofError::HasDebt {
+                    network: certificate.origin_network,
+                });
+            }
+        };
```
Freyskeyd, line 56: This code block can be reduced to this, I think:

`TokenInfo` can also derive `Copy`

```suggestion
        // Apply on Exit Tree
        let computed_root = self.exit_tree.get_root();
        if computed_root != certificate.prev_local_exit_root {
            return Err(ProofError::InvalidLocalExitRoot {
                got: computed_root,
                expected: certificate.prev_local_exit_root,
            });
        }

        certificate.bridge_exits.iter().for_each(|bridge_exit| {
            self.exit_tree.add_leaf(bridge_exit.hash());
            self.balance_tree.withdraw(bridge_exit.token_info, bridge_exit.amount);
        });

        // Check whether the sender has some debt
        if self.balance_tree.has_debt() {
            return Err(ProofError::HasDebt {
                network: certificate.origin_network,
            });
        }
```

crates/pessimistic-proof/src/proof.rs
```diff
@@ -1,66 +1,27 @@
...
 
-    if !debtors.is_empty() {
-        return Err(ProofError::NotEnoughBalance { debtors });
-    }
+/// Proves that the given [`Certificate`] can be applied on the given [`LocalNetworkState`].
+pub fn generate_leaf_proof(
+    initial_network_state: LocalNetworkState,
+    certificate: Certificate,
+) -> Result<LeafProofOutput, ProofError> {
+    let mut network_state = initial_network_state;
```
Freyskeyd, line 22: This function can be reduced to this:
```suggestion
pub fn generate_leaf_proof(
    mut initial_network_state: LocalNetworkState,
    certificate: Certificate,
) -> Result<LeafProofOutput, ProofError> {
	initial_network_state.apply_certificate(certificate)
}
```

I guess we can't just call the `apply_certificate` directly in the program, right?

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Commitment to the balance for each token.
+    pub balance_tree: BalanceTree,
+}
+
+impl LocalNetworkState {
+    /// Apply the [`Certificate`] on the current [`State`].
+    /// Returns the commitment on the resulting state if successful.
+    pub fn apply_certificate(
+        &mut self,
+        certificate: Certificate,
```
Freyskeyd, line 27: I think that `Certificate` can be passed as ref

crates/pessimistic-proof/src/proof.rs
hadjiszs: I prefer to separate like this as for now the statement `generate_leaf_proof` is nothing more than applying the certificate but that will be extended right after

crates/pessimistic-proof/src/local_state.rs
hadjiszs: done

Freyskeyd (approved):

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -0,0 +1,53 @@
...
+            self.exit_tree.add_leaf(bridge_exit.hash());
+            self.balance_tree.withdraw(bridge_exit.token_info, bridge_exit.amount);
+        });
+
+        // Check whether the origin network has some debt
+        if self.balance_tree.has_debt() {
+            return Err(ProofError::HasDebt {
+                network: certificate.origin_network,
+            });
+        }
```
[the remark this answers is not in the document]
hadjiszs: >This errors out after all the withdrawals have been applied. Is that intentional? 

Yes that's related to this point:  https://github.com/agglayer/agglayer/pull/124#discussion_r1662644972

One subtlety here is that this code is part of the proving statement executed by the prover, the proof is about this execution, basically the proof is "I have executed this code with these inputs and without error/panic".

The state given as input is small enough to be simply copied for now (which will be necessarily the case to forward the data to the external prover service).

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub balance_tree: BalanceTree,
+}
+
+impl LocalNetworkState {
+    /// Apply the [`Certificate`] on the current [`State`].
+    /// Returns the commitment on the resulting state if successful.
+    pub fn apply_certificate(
+        &mut self,
+        certificate: &Certificate,
+    ) -> Result<(ExitRoot, BalanceRoot), ProofError> {
```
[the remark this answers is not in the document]
hadjiszs: Probably under a proper type yeah in the coming PRs, they will be more exhaustive as well

crates/pessimistic-proof/tests/leaf_proof.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl From<Amounts> for BalanceTree {
+    fn from(v: Amounts) -> Self {
+        let eth = (ETH.clone(), Deposit(U256::from(v.eth)).into());
+        let usdc = (USDC.clone(), Deposit(U256::from(v.usdc)).into());
+        BalanceTree::from(vec![eth, usdc])
+    }
+}
+
+fn make_tx(_from: u32, to: u32, token: &TokenInfo, amount: u32) -> BridgeExit {
```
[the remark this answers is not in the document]
iljakuklic: > The `origin_network` in `TokenInfo` isn't necessarily the sender, typically whenever you send tokens that are not native to your network.

Ok, makes sense

crates/pessimistic-proof/tests/leaf_proof.rs
```diff
@@ -0,0 +1,107 @@
...
+struct Amounts {
+    eth: u32,
+    usdc: u32,
+}
+
+impl From<Amounts> for BalanceTree {
+    fn from(v: Amounts) -> Self {
+        let eth = (ETH.clone(), Deposit(U256::from(v.eth)).into());
+        let usdc = (USDC.clone(), Deposit(U256::from(v.usdc)).into());
+        BalanceTree::from(vec![eth, usdc])
```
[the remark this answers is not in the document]
hadjiszs: Yeah good point, that's only used in this single test but I'll change that in this PR

iljakuklic (approved):

PR #144, opened 2024-07-08
feat: adding aggregator-notifier

Freyskeyd:
# Description

This PR introduces a new crate `agglayer-aggregator-notifier` that is responsible for notifying the aggregator during an epoch packing.

## PR Checklist:


crates/agglayer-aggregator-notifier/src/sp1/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+impl<I> super::AggregatorProver<I> for SP1<LocalProver>
+where
+    I: Serialize,
+{
+    fn prove(&self, to_pack: Vec<I>) -> BoxFuture<'_, Result<Proof, anyhow::Error>> {
+        let mut stdin = SP1Stdin::new();
+        stdin.write(&to_pack);
```
hadjiszs, line 37: I see that the elf file update hasn't been reverted, is it expected that the aggregator-notifier tests remain green while it relies on the previous elf interface?

Freyskeyd: yes the ELF is updated because of the switch of sp1 version

hadjiszs: Ok so the aggregator-notifier tests doesn't aim at failing in case that the proving calls fail/panic then right?

Freyskeyd: The ELF isn't up-to-date with the current `pessimistic-proof`, it means that the tests are compatible with the ELF but not with the definition in the pessimistic-proof. The goal of the #146 is to prevent this to hapenning again

hadjiszs (approved):

PR #146, opened 2024-07-10
chore: adding jobs to check ELF

Freyskeyd:
# Description

This PR adds a check for the ELF file status based on changes of the `pessimistic-proof` related crates.

## PR Checklist:

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,7 +15,6 @@ concurrency:
 jobs:
   coverage:
     name: SonarCloud and Linter
-    if: "! contains(toJSON(github.event.commits.*.message), '[skip-ci]')"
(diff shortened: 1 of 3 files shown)


iljakuklic (approved):

Freyskeyd:
> Thinking about this... Should we somehow nail down the sp1 toolchain version being used here? It currently just downloads the latest version of the toolchain using the bash script and uses it to build the riscv elf. If the toolchain is updated, the generated elf binary will change (and this check will fail) without us touching any code at all.

I think it is a good thing, if the code breaks with the latest stable toolchain it means that something is wrong or outdated.
SP1 doesn't expose clearly the version and toolchain compatibility tho, I think that's something that we can work on later if we feel the need to pin version.

PR #148, opened 2024-07-10
test: add a simple signer test

iljakuklic:
# Description

Adding a very basic signer test of questionable value. It only checks the wallet signer for now but should be easy to extend to check GCP KMS and other signers too.

## PR Checklist:

### DIFF
--- a/crates/agglayer-signer/src/tests.rs
+++ b/crates/agglayer-signer/src/tests.rs
@@ -1,0 +1,43 @@
+use ethers::types::{Eip1559TransactionRequest, H160};
+
+use super::*;
+
+async fn check_signer_works(signer: &ConfiguredSigner) {
+    let txn = Eip1559TransactionRequest {
+        from: Some(signer.address()),
+        to: Some(H160([0x11; 20]).into()),
+        gas: None,
+        value: Some(1_000_000_000_u64.into()),
+        data: None,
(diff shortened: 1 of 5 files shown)


crates/agglayer-signer/src/tests.rs
```diff
@@ -0,0 +1,43 @@
+use ethers::types::{Eip1559TransactionRequest, H160};
+
+use super::*;
+
+async fn check_signer_works(signer: &ConfiguredSigner) {
```
Freyskeyd, line 5: Why put this in a dedicated function and not directly in the test function?

Freyskeyd (approved):

crates/agglayer-signer/src/tests.rs
iljakuklic: The idea is to have separate tests for different kinds of signers (wallet, gcp kms, etc) but reuse the test logic.

Freyskeyd: Okey. at some point we could rely on rs-test fixture then

iljakuklic: > Okey. at some point we could rely on rs-test fixture then

I ported the test to rstest to prepare for this.

Freyskeyd (approved):

PR #150, opened 2024-07-11
[DRAFT] Local pessimistic proof generation

bpfarmer:
# Description

Please do not merge! This PR is in-progress and designed to give an idea of the changes required to move from global to local pessimistic proof generation. This is highly beneficial as it removes the requirement for networks to sync with global state, or for all AggLayer nodes to maintain complete Local Balance Trees for every chain.

## Additions and Changes

Added the Nullifier Set, ImportedBridgeExit, added additional public inputs to the proof, renamed Certificate to BatchHeader and added additional fields to the struct, and modified the local state computation used in the proof.

Because we're exposing the imported local exit roots as public inputs, we're preparing for the introduction of fast interop between networks.

### New feature (non-breaking change which adds functionality)

Local PP generation, fast tinerop.

## PR Checklist:


crates/pessimistic-proof/src/local_state.rs
```diff
@@ -17,34 +18,50 @@ pub struct LocalNetworkState {
     pub exit_tree: LocalExitTree<Keccak256Hasher>,
     /// Commitment to the balance for each token.
     pub balance_tree: BalanceTree,
+    /// Commitment to the Nullifier Set for the local network, tracks claimed assets on foreign networks
+    pub nullifier_set: NullifierSet,
 }
 
 impl LocalNetworkState {
-    /// Apply the [`Certificate`] on the current [`State`].
+    /// Apply the [`BatchHeader`] on the current [`State`].
```
hadjiszs, line 26: Totally fine to rename it to something else. Initially this structure was referred as `Batch` here as well. We moved away from the `Batch` naming as it conflicts with the `Batch` on the CDK side.

What the CDK submits are actually aggregation of `Batch`. Therefore we renamed it to something else.
Either we force the CDKs to submit proofs for each and every batch individually, or for consistency we cannot use the term `BatchHeader`.

crates/pessimistic-proof-program/src/main.rs
```diff
@@ -1,14 +1,35 @@
...
 
-use pessimistic_proof::{certificate::Certificate, generate_leaf_proof, LocalNetworkState};
+use pessimistic_proof::{batch_header::BatchHeader, generate_leaf_proof, LocalNetworkState};
 
 sp1_zkvm::entrypoint!(main);
 
 pub fn main() {
     let initial_state = sp1_zkvm::io::read::<LocalNetworkState>();
-    let certificate = sp1_zkvm::io::read::<Certificate>();
+    let batch_header = sp1_zkvm::io::read::<BatchHeader>();
```
hadjiszs, line 9: >This PR is in-progress and designed to give an idea of the changes required **to move from global to local pessimistic proof generation**

As I understand it, the API of the PP in this PR remains the same as it is today in `main`. The current PP in the `main` branch is already for the local pessimistic proof (and not for the previous global proof approach): it takes one `LocalNetworkState` (the local state of one CDK) and one `Certificate` (aka `BatchHeader`) and ensures that the `Certificate` can be applied on top of the `LocalNetworkState` of the sending network.

The proto on the nullifier that you added in the `pessimistic-proof` will be helpful as we focused on the state management so far. I'll probably cherry-pick some of this once we get on it :ok_hand: 

>This is highly beneficial as it removes the requirement for networks to sync with global state, or for all AggLayer nodes to maintain complete Local Balance Trees for every chain.

I believe that one point of misalignment/confusion might be on where the `LocalNetworkState` should live, and by extension who should compute the proof.

To give back a bit of history on the spec decision, we had basically two options as proposed by Carlos hackmds:

Option 1:
- Each `LocalNetworkState` lives in their own CDK stack, but it'd require:
  - one component (or modif of an existing one) on the CDK stack to store and track the new needed data structures (e.g., nullifier tree)
  - one component (or modif of an existing one) that does the job of interacting with the SP1 infra and behaving accordingly (probably in Rust)

Option 2:
- The AggLayer service host the `LocalNetworkState` for each Network as an **intermediate step** and will handle the interactions with the SP1 prover infra

We went for the option 2 as mentioned in the rustdoc of the `LocalNetworkState` here:
https://github.com/agglayer/agglayer/blob/3f7062a58e21615bae38816758d5b2ea18a2114b/crates/pessimistic-proof/src/local_state.rs#L11-L13
    
The nice thing with the option 2 is that the preliminary development and definition of the state for the pessimistic proof lives just aside of the pessimistic proof itself which gives great flexibility for the first iteration. Once the functionality gets established, it'd totally make sense to split it up properly as currently planned.

crates/pessimistic-proof/src/local_state.rs
bpfarmer: Hmm - understood, I think that maybe it's fine to say that a set of batches is still a batch within the context of the PP. 

When you mention forcing a CDK to submit proofs for each batch individually, do you mean the PP or the validity proof?

I get that there's some specific metadata that's used to denote a batch in the CDK implementation, but I **think** that it's fine to treat some span of batches as a single batch from the perspective of the PP. This means that from the perspective of the PP,  the batch_header for a set of batches submitted by a CDK chain would be the union of all `imported_lers`, `imported_bridge_exits` and `bridge_exits`.

crates/pessimistic-proof-program/src/main.rs
bpfarmer: > As I understand it, the API of the PP in this PR remains the same as it is today in `main`

Yep, it remains the same in that we're taking some data in a batch header and applying it to the local network state. However, the `BatchHeader` and `LocalNetworkState` have changed substantially.

> The proto on the nullifier that you added in the pessimistic-proof will be helpful as we focused on the state management so far. I'll probably cherry-pick some of this once we get on it 👌

I'll finish the impl (mostly with help from Will B as I'll likely be on paternity leave shortly).

> To give back a bit of history on the spec decision, we had basically two options as proposed by Carlos hackmds:

Yep understood, I had initially proposed the fully local version, where proof generation is moved off of the AL in a hackmd. It doesn't matter to me where this info is stored or whether there's a separate service now or in the future. 

My priority with this PR is to get to the point where the API for PP generation matches what we want it to be, both from the perspective of fully local proof generation and preparing for fast interop.

crates/pessimistic-proof/src/local_state.rs
bpfarmer: Could also rename to something like BatchSetHeader if we're worried specifically about collisions with the CDK implementation.

hadjiszs: >When you mention forcing a CDK to submit proofs for each batch individually, do you mean the PP or the validity proof?

Ah I meant for whichever update, but in the current stage specifically for the PP.

>Could also rename to something like BatchSetHeader if we're worried specifically about collisions with the CDK implementation.

Hm but then that's not really an header as it contains the bridge exits among other payload not present in batch headers, we specifically moved away from the term `Batch` as in practice it was bringing confusions in the discussions, often we had to specify as you did by saying in the "agglayer perspective" or in the "pp perspective", ideally we want a term that does not refer already to something different to avoid having to precise the context for each occurrence.

Totally ok with moving away from `Certificate` but I do not have better naming alternative in mind right now unfortunately

bpfarmer:
> Hm but then that's not really an header as it contains the bridge exits
among other payload not present in batch headers

Yep if you look in the code there are a bunch of TODO's to move certain
fields out of the `BatchHeader` because they don't really belong there. The
reason that I like the header is that it will make it easy to use it in a
bunch of different places (public inputs for execution proofs, consensus
proofs, etc) beyond the PP. I think that `BatchSetHeader` works as it
indicates that it represents some state transition that is the result of
one or more batches.

PR #151, opened 2024-07-11
feat: adding default config and switch format

### LINKED ISSUE #83: Update config file format
## Description

*This PR can be mentored or part of a pairing session*

The config file format is currently using `toml` to express and define configuration.

This ticket is about finding the better solution to gather our configuration definition.
The goal is to not have something complex at first sight, but also something that we can use to express deep configuration definition.

`YAML` could have been an option to have more visual depth of the config, that *may* be something to explore.

### Ideas

- Define a max depth for the config (3 maybe, 4 max ?)
- Define if we want to support multiple format
- Define naming convention for keys (more after)

### Naming convention

Currently, the configuration is defining keys with both `PascalCase` and `lowercase`.

We have that:

```toml
[auth.gcpkms]
ProjectId = ""
Location = ""
Keyring = ""
KeyName = ""

[outbound.rpc.settle]
max_retries = 10
retry_interval = 1
confirmations = 3
```

I would suggest switching to either `snake_case` or `kebab-case`, I may find `kebab-case` more readable, but here's examples:

```toml
# snake_case

[auth.gcp_kms]
project_id = ""
location = ""
keyring = ""
key_name = ""

[outbound.rpc.settle]
max_retries = 10
retry_interval = 1
confirmations = 3
```

```toml
# kebab-case

[auth.gcp-kms]
project-id = ""
location = ""
keyring = ""
key-name = ""

[outbound.rpc.settle]
max-retries = 10
retry-interval = 1
confirmations = 3
```

## Acceptance

- [ ] Having a non-breaking configuration file (this ticket can be split into multiple ones)
- [ ] Having an agreement on the config format of the keys
- [ ] Having tests covering the non-breaking changes
- [ ] Update the `kurtosis-cdk` accordingly

### LINKED ISSUE #89: Add a command to the CLI to create a default configuration for the agglayer
## Description

In order to help users, but also facilitate our tests, we could provide a way to generate the configuration file, based on the default `serde`  values. It could also help us to make non breaking configuration changes in the future by reading and updating a configuration file in place.

This ticket needs discussion.

## Acceptance

- [ ] Having a CLI command that generate the default configuration file

Freyskeyd:
# Description

This PR is adding a `default` mode to the configuration, a CLI command to generate a default `config` and it switches the format from `PascalCase/lowercase` to `kebab-case`. It is related to the discussion #147 

There are backward compatibility tests in the config crate to cover most of the config used currently.
More can be added to prevent any regression, and when we're confident on the deprecation of the old format, we can clean them up.

This PR closes #83 and closes #89 

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/Cargo.toml
+++ b/crates/agglayer-config/Cargo.toml
@@ -16,6 +16,7 @@ tracing.workspace = true
 url = { workspace = true, features = ["serde"] }
 
 [dev-dependencies]
+pretty_assertions = "1.4.0"
 serde_json = { workspace = true }
 toml.workspace = true
 
--- a/crates/agglayer-config/src/log.rs
+++ b/crates/agglayer-config/src/log.rs
@@ -1,23 +1,24 @@
 use std::{fmt::Display, path::PathBuf};
 
-use serde::{Deserialize, Deserializer};
+use serde::{Deserialize, Deserializer, Serialize};
 use tracing_subscriber::{fmt::writer::BoxMakeWriter, EnvFilter};
 
 /// The log configuration.
-#[derive(Deserialize, Debug, Clone, Default)]
-#[serde(rename_all = "PascalCase")]
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
 pub struct Log {
     /// The `RUST_LOG` environment variable will take precedence over the
     /// configuration log level.
-    #[serde(default)]
+    #[serde(default, alias = "Level")]
     pub level: LogLevel,
+    #[serde(alias = "Outputs")]
     pub outputs: Vec<LogOutput>,
-    #[serde(default)]
+    #[serde(default, alias = "Format")]
     pub format: LogFormat,
 }
 
 /// The log format.
-#[derive(Deserialize, Debug, Default, Clone, Copy)]
+#[derive(Serialize, Deserialize, Debug, Default, Clone, Copy, PartialEq, Eq)]
 #[serde(rename_all = "lowercase")]
 pub enum LogFormat {
     #[default]
@@ -26,7 +27,7 @@ pub enum LogFormat {
 }
 
 /// The log level.
-#[derive(Deserialize, Debug, Default, Clone, Copy)]
+#[derive(Serialize, Deserialize, Debug, Default, Clone, Copy, PartialEq, Eq)]
 #[serde(rename_all = "lowercase")]
 pub enum LogLevel {
     Trace,
@@ -67,7 +68,7 @@ impl From<LogLevel> for EnvFilter {
 /// specify the output location as a string, which is then parsed into the
 /// appropriate enum variant. If the string is not recognized to be either
 /// `stdout` or `stderr`, it is assumed to be a file path.
-#[derive(Debug, Clone, Default)]
+#[derive(Serialize, Debug, Clone, Default, PartialEq, Eq)]
 pub enum LogOutput {
     #[default]
     Stdout,
--- a/crates/agglayer/tests/snapshots/config__config_display.snap
+++ b/crates/agglayer/tests/snapshots/config__config_display.snap
@@ -1,0 +1,41 @@
+---
+source: crates/agglayer/tests/config.rs
+expression: result
+---
+[full-node-rpcs]
+
+[log]
+level = "info"
+outputs = []
+format = "pretty"
+
+[rpc]
+port = 9090
+host = "0.0.0.0"
+
+[outbound.rpc.settle]
+max-retries = 3
+retry-interval = 7
+confirmations = 1
+
+[l1]
+ChainID = 1337
+NodeURL = "http://zkevm-mock-l1-network:8545/"
+RollupManagerContract = "0xb7f8bc63bbcad18155201308c8f3540b07f84f5e"
+
+[auth.Local]
+private-keys = []
+
+[telemetry]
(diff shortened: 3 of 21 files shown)


crates/agglayer-config/src/log.rs
```diff
@@ -67,7 +67,7 @@ impl From<LogLevel> for EnvFilter {
 /// specify the output location as a string, which is then parsed into the
 /// appropriate enum variant. If the string is not recognized to be either
 /// `stdout` or `stderr`, it is assumed to be a file path.
-#[derive(Debug, Clone, Default)]
+#[derive(Serialize, Debug, Clone, Default, PartialEq, Eq)]
```
iljakuklic, line 70: I see this one has a custom `Deserialize` impl but derived `Serialize` impl. These are not consistent with each other, i.e. the serialize/deserialize roundtrip will not work as expected in some cases.

crates/agglayer/tests/snapshots/config__config_display.snap
```diff
@@ -0,0 +1,41 @@
...
+private_keys = []
+
+[telemetry]
+PrometheusAddr = "0.0.0.0:3000"
+
+[epoch.TimeClock]
+EpochDuration = 5
+
+[shutdown]
+runtime_timeout = 5
```
iljakuklic, line 36: Quite a few of these are still not kebab case. I see a mixture of pascal, snake, kebab. Am I missing something? As far as I can tell, this is not related to the backwards compat test.

crates/agglayer-config/Cargo.toml
```diff
@@ -16,6 +16,7 @@ tracing.workspace = true
 url = { workspace = true, features = ["serde"] }
 
 [dev-dependencies]
+pretty_assertions = "1.4.0"
```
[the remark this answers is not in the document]
Freyskeyd: I think not, or at least not for a single use like this one. And there is no such things as workspace dev-dependencies :/

crates/agglayer-config/src/log.rs
Freyskeyd: The `Serialize` derive is ok because the default value is an empty hashmap

iljakuklic: > The `Serialize` derive is ok because the default value is an empty hashmap

But the serialize impl supports writing out non-default values too, at least in principle. That might cause issues if something like "dump current config" functionality is implemented.

crates/agglayer-config/Cargo.toml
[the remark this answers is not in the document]
iljakuklic: > I think not, or at least not for a single use like this one. And there is no such things as workspace dev-dependencies :/

It think you can throw all kinds of dependencies to workspace dependencies to get consistent versioning. Whether it depends a regular dependency or dev dependency is determined at the crate level.

The following quote from [the Cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace) also seems to suggest so:

> Dependencies in the [dependencies], [dev-dependencies], [build-dependencies], and [target."...".dependencies] sections support the ability to reference the [workspace.dependencies] definition of dependencies.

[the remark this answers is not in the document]
Freyskeyd: Yes I know, but this dependency is only used for the config crate. We do not need to put it on the workspace dependency and I want to keep the root `Cargo.toml` fairly small.

crates/agglayer-config/src/log.rs
Freyskeyd: We'll not do any `dump current config` functionality for now and this part will certainly be removed as it is fairly useless.

crates/agglayer-config/Cargo.toml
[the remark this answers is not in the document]
iljakuklic: > Yes I know, but this dependency is only used for the config crate. We do not need to put it on the workspace dependency and I want to keep the root `Cargo.toml` fairly small.

Ok, that's fair

crates/agglayer-config/src/log.rs
iljakuklic: Understood. I guess I just don't like having inconsistent impls around in the code.

I think it could be addressed by adding the `rename_all` attribute as with the other stuff and annotating the `File` arm with `#[serde(untagged)]`. That should remove the need for the custom impl completely, unless I am missing some other issue.

crates/agglayer/tests/snapshots/config__config_display.snap
```diff
@@ -0,0 +1,41 @@
...
+RollupManagerContract = "0xb7f8bc63bbcad18155201308c8f3540b07f84f5e"
+
+[auth.Local]
+private-keys = []
+
+[telemetry]
+PrometheusAddr = "0.0.0.0:3000"
+
+[epoch.TimeClock]
+EpochDuration = 5
```
iljakuklic, line 33: Some keys here are still using `PascalCase`.

crates/agglayer/tests/snapshots/config__config_display.snap
```diff
@@ -0,0 +1,41 @@
...
+
+[shutdown]
+runtime-timeout = 5
+
+[certificate-orchestrator]
+input-backpressure-buffer-size = 1000
+
+[certificate-orchestrator.prover.SP1Local]
```
iljakuklic, line 41: Also here (`SP1Local`)

iljakuklic (approved):
Looks good apart from the few missing field renames.

PR #164, opened 2024-07-16
feat: SMT data structure

wborgeaud:
Add a generic `Smt` struct that can be used e.g. for the nullifier tree.
This is a standard implementation that could be optimized in various ways, but it should already be good enough for the nullifier use case (checking non-inclusion proofs and updating the root).

## PR Checklist:


crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+pub(crate) enum SmtError {
+    #[error("trying to insert a key already in the SMT")]
+    KeyAlreadyPresent,
+    #[error("trying to generate a Merkle proof for a key not in the SMT")]
+    KeyNotPresent,
+    #[error("trying to generate a non-inclusion proof for a key present in the SMT")]
+    KeyPresent,
+}
+
+/// A node in an SMT
```
muursh, line 26:
```suggestion
/// A node in an SMT.
```

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+where
+    H: Hasher,
+    H::Digest: Serialize + DeserializeOwned,
+{
+    pub fn hash(&self) -> H::Digest {
+        H::merge(&self.left, &self.right)
+    }
+}
+
+/// An SMT consistent with a zero-initialized Merkle tree
```
muursh, line 70:
```suggestion
/// An SMT consistent with a zero-initialized Merkle tree.
```

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Smt<H, const DEPTH: usize>
+where
+    H: Hasher,
+    H::Digest: Copy + Eq + Hash + Serialize + DeserializeOwned,
+{
+    /// The SMT root
```
muursh, line 78:
```suggestion
    /// The SMT root.
```

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+    #[serde_as(as = "_")]
+    root: H::Digest,
+    /// A map from node hash to node
```
muursh, line 81:
```suggestion
    /// A map from node hash to node.
```

muursh (commented):
It looks broadly good. I'll do a deeper review this evening after my calls.

muursh (approved):
No major issues from me. I'm not 100% convinced by the HashMap but there's nothing here I'll push back against.

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+
+        Ok(new_hash)
+    }
+
+    pub fn insert<K>(&mut self, key: K, value: H::Digest) -> Result<(), SmtError>
+    where
+        K: ToBits<DEPTH>,
+    {
+        let new_root = self.insert_helper(self.root, 0, &key.to_bits(), value)?;
+        self.root = new_root;
```
iljakuklic, line 209: Do  the previous roots (and other nodes not reachable from the current root by following left/right children) get cleaned up somehow or do they stay in the node map indefinitely?

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+    KeyPresent,
+}
+
+/// A node in an SMT.
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Node<H>
+where
+    H: Hasher,
+    H::Digest: Serialize + DeserializeOwned,
```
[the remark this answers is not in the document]
wborgeaud: So I agree in principle, however here I think it would negatively impact the code's readability and succinctness to remove this (because we'd need to add a custom impl of `(De)Serialize` when `H::Digest: Serialize + DeserializeOwned` instead of a clean derive macro).
Also, I don't think we'd ever want to work with non-serializable hashes, so that's a way to enforce it in the code.

But tbh I don't think we actually need these structs (especially `Smt`) to be (de)serializable, so at some point we could just remove these derive macros altogether.

crates/pessimistic-proof-utils/src/smt.rs
wborgeaud: They just stay in the map indefinitely. That's usually not a big deal when there's not a ton of writes in the SMT.
In production, we would actually need these SMTs to be stored in persistent databases and there we would handle the removal of unused nodes.
I view this `Smt` implementation as more useful for testing/prototyping, so I think it's fine for now to not bother with this.

crates/pessimistic-proof-utils/src/smt.rs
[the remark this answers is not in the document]
muursh: > I think it would negatively impact the code's readability

Agreed entirely

crates/pessimistic-proof-utils/src/smt.rs
iljakuklic: > I view this `Smt` implementation as more useful for testing/prototyping, so I think it's fine for now to not bother with this.

Fair enough. A comment pointing this out would be useful though as it's not necessarily obvious right away.

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+        let child_hash = if bits[depth] {
+            self.insert_helper(node.right, depth + 1, bits, value)
+        } else {
+            self.insert_helper(node.left, depth + 1, bits, value)
+        }?;
+        if bits[depth] {
+            node.right = child_hash;
+        } else {
+            node.left = child_hash;
+        }
```
Freyskeyd, line 196: Could something like that works?

```suggestion
        let node_place = if bits[depth] { &mut node.right } else { &mut node.left }; 
        *node_place = self.insert_helper(*node_place, depth + 1, bits, value)?;
```

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+
+        Ok(new_hash)
+    }
+
+    pub fn insert<K>(&mut self, key: K, value: H::Digest) -> Result<(), SmtError>
+    where
+        K: ToBits<DEPTH>,
+    {
+        let new_root = self.insert_helper(self.root, 0, &key.to_bits(), value)?;
+        self.root = new_root;
```
Freyskeyd, line 209: I think you don't need the extra assignement
```suggestion
        self.root = self.insert_helper(self.root, 0, &key.to_bits(), value)?;
```

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+                            &self.empty_hash_at_height[DEPTH - i - 1],
+                            &self.empty_hash_at_height[DEPTH - i - 1]
+                        ),
+                        "The SMT is messed up"
+                    );
+                    return Ok(SmtNonInclusionProof { siblings });
+                }
+            };
+            siblings.push(if bits[i] { node.left } else { node.right });
+            hash = if bits[i] { node.right } else { node.left };
```
Freyskeyd, line 262: You can reduce the boilerplate to this, I guess: 
```suggestion
            hash = if bits[i] { node.right } else { node.left };
            siblings.push(hash);
            
```

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+            } else {
+                H::merge(&entry, &sibling)
+            };
+        }
+
+        Some(entry)
+    }
+}
+
+impl ToBits<32> for u32 {
```
Freyskeyd, line 372: Just for my understanding, why implementing it just for 32?

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+    #[test]
+    fn test_non_inclusion_proof_and_update_empty() {
+        test_non_inclusion_proof_and_update(0)
+    }
+
+    #[test]
+    fn test_non_inclusion_proof_and_update_nonempty() {
+        let num_keys = thread_rng().gen_range(1..100);
+        test_non_inclusion_proof_and_update(num_keys)
+    }
```
Freyskeyd, line 558: With `rstest` you can reduce this to one function: 
```suggestion

    #[rstest]
    #[case(0)]
    #[case(thread_rng().gen_range(1..100))]
    fn test_non_inclusion_proof_and_update(#[case] num_keys: usize) {
        let mut smt = Smt::<H, DEPTH>::new();
        let kvs: Vec<(u32, _)> = (0..num_keys).map(|_| (random(), random())).collect();
        check_no_duplicates(&kvs);
        for (key, value) in kvs.iter() {
            smt.insert(*key, *value).unwrap();
        }
        let key: u32 = random();
        assert!(
            kvs.iter().position(|(k, _)| k == &key).is_none(),
            "Check your rng"
        );
        let proof = smt.get_non_inclusion_proof(key).unwrap();
        assert!(proof.verify(key, smt.root, &smt.empty_hash_at_height));
        let value = random();
        let new_root = proof
            .verify_and_update(key, value, smt.root, &smt.empty_hash_at_height)
            .unwrap();
        smt.insert(key, value).unwrap();
        assert_eq!(smt.root, new_root);
    }
```

crates/pessimistic-proof-utils/Cargo.toml
```diff
@@ -14,6 +14,8 @@ serde_with = { version = "3" }
 tiny-keccak = { git = "https://github.com/sp1-patches/tiny-keccak", branch = "patch-v2.0.2", features = [
     "keccak",
 ] }
+thiserror = "1.0.61"
```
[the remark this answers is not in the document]
wborgeaud: Nice! 🤯

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+{
+    pub fn hash(&self) -> H::Digest {
+        H::merge(&self.left, &self.right)
+    }
+}
+
+/// An SMT consistent with a zero-initialized Merkle tree.
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Smt<H, const DEPTH: usize>
```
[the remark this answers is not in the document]
wborgeaud: I can add expand the "SMT" in the comment, but I'd rather stick with `Smt` as I feel like we always refer to them using the abbreviation.

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+
+    pub fn is_empty(&self) -> bool {
+        self.root
+            == H::merge(
+                &self.empty_hash_at_height[DEPTH - 1],
+                &self.empty_hash_at_height[DEPTH - 1],
+            )
+    }
+
+    pub fn get<K>(&self, key: K) -> Option<H::Digest>
```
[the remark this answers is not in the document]
wborgeaud: It's not used in this PR, but is in #168. Will add a comment though.

crates/pessimistic-proof-utils/src/smt.rs
```diff
@@ -0,0 +1,559 @@
...
+        if depth == DEPTH {
+            return if hash != self.empty_hash_at_height[0] {
+                Err(SmtError::KeyAlreadyPresent)
+            } else {
+                Ok(value)
+            };
+        }
+        let node = self.tree.get(&hash);
+        let mut node = node.copied().unwrap_or(Node {
+            left: self.empty_hash_at_height[DEPTH - depth - 1],
```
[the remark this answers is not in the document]
wborgeaud: I can add an `assert!(depth < DEPTH)`, but yes the idea is that `depth` starts at 0 and is incremented below, and at the top of this function we check `depth==DEPTH`, so at this point we can never have `depth >= DEPTH`.

crates/pessimistic-proof-utils/src/smt.rs
wborgeaud: I think so, very nice!

crates/pessimistic-proof-utils/src/smt.rs
wborgeaud: Just to give an example for this PR. In #168 it's implemented for more interesting structs.

crates/pessimistic-proof-utils/src/smt.rs
wborgeaud: Nice! @iljakuklic also mentioned this could be cleaned using `proptest/quickcheck`: https://github.com/agglayer/agglayer/pull/163#discussion_r1684061906
I will look into this in a separate PR.

hadjiszs (approved):

PR #166, opened 2024-07-16
feat: adding ProofSigner config and behavior

Freyskeyd:
# Description

This PR adds the `ProofSigner` config which allow a CDK to define an alternative signer to the `SignedTx`

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/kernel/tests.rs
+++ b/crates/agglayer-node/src/kernel/tests.rs
@@ -207,6 +207,60 @@ async fn interop_executor_verify_signature() {
         .unwrap();
 }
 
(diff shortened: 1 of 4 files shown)


crates/agglayer-node/src/kernel/tests.rs
```diff
@@ -207,6 +207,60 @@ async fn interop_executor_verify_signature() {
         .unwrap();
 }
 
+/// Test that check if the verify_signature method works with proof signer
```
muursh, line 210:
```suggestion
/// Test that check if the verify_signature method works with proof signer.
```

muursh (approved):

PR #168, opened 2024-07-17
feat: Local pessimistic proof with SMTs

### LINKED ISSUE #109: Handle balance on origin network correctly
`TokenInfo` identifies a token by origin network and the bridge correctly tracks this info. This means we can easily identify when an asset is bridged back to it's origin network and this case needs specific handling. I haven't thought it through completely but I *believe* one way to handle this is to effectively not manage a 'balance' entry for assets that are native to a network. This is because from the standpoint of the proof, the native network will look like it is minting assets that weren't bridged to it - but this is (obviously?) correct since it is managed on the network.

```
pub struct TokenInfo {
    /// Network which the token originates from
    /// The address of the token on the origin network
    pub origin_token_address: Address,
}
```

### LINKED ISSUE #110: Leaf proof: use precomputed "empty roots" for each depth
Here, "empty root" denotes the root of an empty Merkle tree. We currently [recompute these roots](https://github.com/AggLayer/invariant-proof/blob/0e85096db86fe222b800b61735d142534ab0c280/invariant_proof/src/local_exit_tree/mod.rs#L87) every time we compute a tree's root in `LocalExitTree::get_root()`.

We should instead pre-compute 32 empty roots, roughly as:

```
precomputed = [
  hash([0; 32], [0; 32]),                // root of tree of depth 1
  hash(precomputed[0], precomputed[0]),  // root of tree of depth 2
  hash(precomputed[1], precomputed[1]),  // root of tree of depth 3
  ...
]
```

wborgeaud:
Same changes as #150 with some gaps filled using the SMT and local exit tree implementations introduced in #164 and #163.

~Testing still TODO~
Added some E2E tests including one generating an SP1 proof (uses 2'405'226 cycles).

Resolves #109, resolves #110, closes #150

## PR Checklist:


crates/pessimistic-proof/src/nullifier_tree/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+            empty_hash_at_height,
+        }
+    }
+    pub fn new_with_root(root: H::Digest) -> Self {
+        let mut res = Self::new();
+        res.root = root;
+        res
+    }
+
+    // TODO: Consider batching the updates per network for efficiency
```
bpfarmer, line 93: Yeah - this is why I had the awkward/lazy NetworkNullifierSet struct instead of a single tree. Depending on how claims are distributed, I think it could be a nice future optimization to batch nullifier set updates for a single network to save on hashing. Especially in the synchronous case, we might expect to see many messages claimed from a particular chain in a given batch.

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -16,38 +17,126 @@ pub struct LocalNetworkState {
...
+            self.nullifier_set.verify_and_update(nullifier_key, nullifier_path)?;
+
+            if batch_header.origin_network
+                == imported_bridge_exit.bridge_exit.token_info.origin_network
+            {
+                // When the token is native to the chain, we don't care about the local balance
+                continue;
+            }
+            // Check the balance tree inclusion path and update the balance tree
+            let old_balance = imported_bridge_exit.current_balance;
```
bpfarmer, line 90: Totally valid to do it this way, but I wonder if in the future it's simpler to omit current_balance from ImportedBridgeExit, sum over BridgeExit and ImportedBridgeExits to get net flows for each token address, and then load the current balances for each affected token address.

Would save the AL service from needing to calculate and insert current balances for each ImportedBridgeExit, though this isn't a big deal.

crates/pessimistic-proof/src/nullifier_tree/mod.rs
wborgeaud: Yep for sure. I think it's cleaner that way, but definitely more efficient to batch by network. So my idea was to implement the batching in a separate PR to keep this one light.

crates/pessimistic-proof/src/local_state.rs
wborgeaud: Yes good idea! Would save a lot of hashing when a lot of events are on the same token. Will implement it 👍

crates/pessimistic-proof/src/nullifier_tree/mod.rs
bpfarmer: Makes total sense - definitely a better way to do it and batching is probably premature at this stage.

crates/pessimistic-proof/src/local_state.rs
wborgeaud: Implemented in 892714b. It reduced the number of SP1 cycles by 50% in the simple E2E test.
It still isn't totally optimal, as I verify and update each token Merkle path individually instead of using a batch method. Added a TODO for that and will tackle it in a future PR.
Also, using the full 192-bit token info (160 for the address + 32 for the network id) is probably not optimal too. Using some sort of indexed Merkle tree would probably be more efficient. Same for the nullifier tree. Will also look into that in a future PR.

bpfarmer: Niiiiice!

crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -16,7 +16,8 @@ use jsonrpsee::{
         ErrorObject, ErrorObjectOwned,
     },
 };
-use pessimistic_proof::certificate::Certificate;
+use pessimistic_proof::batch_header::BatchHeader;
```
muursh, line 19: I think we need to rethink the name `BatchHeader` it's already caused some confusion with the CDK team

wborgeaud: I don't have a strong opinion on this. What name should we change it to?

crates/pessimistic-proof/src/batch_header.rs
```diff
@@ -0,0 +1,133 @@
...
+where
+    H: Hasher,
+    H::Digest: Eq + Hash + Copy + Serialize + DeserializeOwned,
+{
+    /// The origin network which emitted this BatchHeader.
+    /// TODO: we should clarify naming. We use origin to refer to the issuing network of a token. Could consider using "sending" or "local" to refer to the network that created a batch.
+    pub origin_network: NetworkId,
+
+    // /// The start slot height and end block height for the batch. Unsure if required
+    // pub start_block_height: u32,
```
hadjiszs, line 38: yeah not required I think

crates/pessimistic-proof/src/batch_header.rs
```diff
@@ -0,0 +1,133 @@
...
+    // pub end_block_height u32,
+
+    // /// The state root produced by the current block
+    // TODO: Determine if this is necessary
```
hadjiszs, line 42: not necessary I think

crates/pessimistic-proof/src/batch_header.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    /// A commitment to the set of imported bridge exits for which the origin network is the target.
+    #[serde_as(as = "Option<_>")]
+    pub imported_exits_root: Option<H::Digest>,
+
+    /// The set of imported local exit roots
+    // TODO: benchmark if BTreeMap is the best choice in terms of SP1 cycles
+    pub imported_local_exit_roots: BTreeMap<NetworkId, H::Digest>,
+
+    /// A map from token info to the token balance of the receiving network before any bridge event is processed,
```
hadjiszs, line 68: you meant the emitting/sending network of this `BatchHeader` right?

```suggestion
    /// A map from token info to the token balance of the origin network before any bridge event is processed,
```

crates/pessimistic-proof/src/imported_bridge_exit.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ImportedBridgeExit {
+    /// The bridge exit initiated on another network, called the "foreign" network.
+    /// Need to verify that the destination network matches the current network, and that
+    /// the bridge exit is included in an imported LER
+    pub bridge_exit: BridgeExit,
+
+    /// The Network ID for the foreign network. May not be strictly
+    /// necessary depending on the final structure of the Imported Local Exit Root
+    pub sending_network: NetworkId,
```
hadjiszs, line 19: nit: referred as `foreign network` in the rustdoc also above for `bridge_exit`, but the variable is named as `sending_network`, any of the two names is ok with me

crates/pessimistic-proof/tests/leaf_proof.rs
```diff
@@ -2,11 +2,12 @@ use std::collections::BTreeMap;
```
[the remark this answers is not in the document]
wborgeaud: Sounds good, will delete this file 👍

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -16,38 +19,141 @@ pub struct LocalNetworkState {
...
+        // TODO: benchmark if BTreeMap is the best choice in terms of SP1 cycles
+        let mut new_balances: BTreeMap<_, _> =
+            batch_header.balances_proofs.iter().map(|(k, v)| (*k, v.0)).collect();
+
+        // TODO: check batch_header.imported_exits_root
+        // Apply the imported bridge exits
+        for (imported_bridge_exit, nullifier_path) in &batch_header.imported_bridge_exits {
+            if matches!(imported_bridge_exit.bridge_exit.leaf_type, LeafType::Message) {
+                // TODO: handle bridge messages
+                panic!()
```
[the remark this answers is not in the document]
wborgeaud: Can you expand on that? I didn't really find docs on bridge messages. How do they differ from bridge transfers?

crates/pessimistic-proof/src/batch_header.rs
wborgeaud: That's right, I meant "origin".

crates/pessimistic-proof/src/imported_bridge_exit.rs
wborgeaud: I find "sending" more explicit. Will change in the doc comments 👍

crates/pessimistic-proof/src/local_state.rs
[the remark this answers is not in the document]
hadjiszs: In terms of doc, maybe these slides from slide 5: https://github.com/0xPolygonHermez/zkevm-techdocs/blob/main/slides/zkevm-architecture-part3-layer-communication.pdf

When `LeafType` is `Message`, the `metadata` field carries the payload for the function call to the contract `bridge_exit.destination_address`. My understanding is that we do not need to care about the semantic of this contract call in the pessimistic proof, so in that sense I believe that it could take the same execution path as for the native token asset transfers as in skipping the insertion in the balance tree (but still being included in the local exit tree)

That said I'm ok having that in another PR to check all that properly :ok_hand:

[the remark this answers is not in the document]
bpfarmer: Paging @invocamanman - I **think** messages can have native ETH transfers so it could be a bit more complicated to parse that part of the message and update balances. 🙃

crates/pessimistic-proof/src/batch_header.rs
bpfarmer: Definitely not necessary for PP generation, but if the same header is used for execution and consensus proofs (which it should be) then it could be useful to include. Happy to omit for now though.

crates/pessimistic-proof/src/local_state.rs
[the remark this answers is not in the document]
wborgeaud: Thanks for the details. I think it's best to leave this to a separate PR. 
One thought though, maybe it the API would be a bit cleaner if `BridgeExit` was an enum with two variants (`Transfer` and `Message`) instead of the current version where `BridgeExit` holds an enum `leaf_type: LeafType`.

crates/agglayer-node/src/rpc/mod.rs
hadjiszs: cc @krlosMata @zuiris
Is `MultiBatchHeader` ok with you to replace `Certificate`? Or do you have any other name idea? It refers to the data structure submitted to the AggLayer

muursh: Why don't we go with `MultiBatchHeader` for now to get this merged? We have a while where we could change it if someone has a "better" idea anyway.

wborgeaud: Renamed to `MultiBatchHeader` in 2730558.

hadjiszs (approved):

crates/agglayer-node/src/rpc/mod.rs
zuiris: I don't find this discussion particularly productive tbh, I'm pretty sure there is a better application for everyone's brain capacity than having a discussion about how to name a data structure. If we want to optimize for "less confusing" name, `certificate` is less confusing IMO because it is at least signals of a new entity in a protocol, which can't be said about any name, which includes "batch" in it.

Freyskeyd: I totally agree on what @zuiris are saying. We already had discussions about that during meetings when discussing the specification and we end up agreeing on `Certificate`, as it was less confusing with other structs and concept used by other components in relation with the `agglayer`. I don't have a strong opinion on what we should use, but I would like to have something on which we can rely on without have to add extra context.
I may not have the full picture, but does `MultiBatchHeader` mean that we can somehow get a `BatchHeader` or even a `Batch` using this struct ? Does it mean that we can receive multiple instances of what we were calling `Certificate` in one RPC call ?

bpfarmer: @zuiris agree that there are more productive discussions than naming. But actually thinking about the future direction of the protocol is probably a very good use of brainpower. 

If we project a tiny bit into the future, what will we call the data structure that we use to authenticate a signing key for a chain, or a relay, or some other service? This is what the term "Certificate" actually refers to. 

>  it is at least signals of a new entity in a protocol

This is not correct and (imo) signals a misunderstanding of the protocol. I suggested BatchHeader[1] initially because it would be better, imo, for the verifier to use the same header as a public input for the pessimistic proof and (aggregated) execution and consensus proofs, because it's simpler and we're less likely to mess up by not checking that two inputs are consistent. The point is that we should be committing to the same public inputs across proofs to avoid errors.

Like I said in my above comment, if we're carrying technical debt that doesn't allow us to call any sequence of blocks a "Batch" then we can call it something else. I want to be clear though, the point here is not to be pedantic, it's to ensure that we share a consistent understanding of the protocol and how it's supposed to work.

[1] You might notice that the data structure in the PP has other fields, like ImportedBridgeExits, that are included in plaintext but don't belong in a "Header". I had left comments that these should eventually be moved to separate structs for auxiliary data that we provide as private inputs.

PR #173, opened 2024-07-18
fix: convert the remaining config keys to `kebab-case`

iljakuklic:
# Description

Some were previously missed.

Related:
* #151 
* #147 
* #83

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/src/auth.rs
+++ b/crates/agglayer-config/src/auth.rs
@@ -19,6 +19,7 @@ use serde_with::{serde_as, NoneAsEmptyString};
 /// - If the `GOOGLE_APPLICATION_CREDENTIALS` environment is set, attempt to
 ///   load a service account JSON from this path.
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
 pub enum AuthConfig {
     Local(LocalConfig),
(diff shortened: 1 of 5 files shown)


crates/agglayer-config/src/auth.rs
```diff
@@ -19,6 +19,7 @@ use serde_with::{serde_as, NoneAsEmptyString};
 /// - If the `GOOGLE_APPLICATION_CREDENTIALS` environment is set, attempt to
 ///   load a service account JSON from this path.
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
 pub enum AuthConfig {
```
iljakuklic, line 23: I am not completely convinced by this change. Converting it to kebab makes the configuration look more consistent. This is, however, an enum, so maybe there is some merit in keeping a syntactic differentiation between structs and enums (when only one of the keys is expected to be present).

Thoughts?

Freyskeyd: mmh I think we should stick to having the same format.

And after removing all the workaround to make it backward compatible, you will not be able to have multiple `auth` config. Like for `[certificate-orchestrator.prover]`

Freyskeyd (approved):

muursh (approved):

PR #182, opened 2024-07-23
feat: adding nonce-manager

Freyskeyd:
# Description

To prevent `nonce` concurrency error, we introduce a new middleware on the `Provider`, the `NonceManagerMiddleware`.

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/kernel/mod.rs
+++ b/crates/agglayer-node/src/kernel/mod.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use agglayer_config::Config;
 use ethers::prelude::*;
 use thiserror::Error;
-use tracing::{info, instrument};
+use tracing::{info, instrument, warn};
 
 use crate::{
     contracts::{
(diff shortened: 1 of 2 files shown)


crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -329,12 +329,23 @@ where
...
         let f = self
             .build_verify_batches_trusted_aggregator_call(signed_tx)
             .await
             .map_err(SettlementError::ContractError)?;
 
+        if let Ok(Some(tx)) = self.check_tx_status(hex_hash).await {
+            warn!(hash, "Transaction already settled: {:?}", tx);
+        }
+
+        info!(
```
hadjiszs, line 344: maybe debug?

hadjiszs (approved):

crates/agglayer-node/src/kernel/mod.rs
Freyskeyd: For now I let it as `info` in order to have it on the logs in datadog, but after the debugging phase we can reduce its level.

PR #218, opened 2024-08-13
feat: handle the native eth balance for the bridge messages

hadjiszs:
# Description

Remove the panic for the bridge messages.
If the leaf type is a message, then the non-zero `amount` corresponds to a native eth transfer

## PR Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -250,9 +250,9 @@ checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
 name = "alloy-chains"
-version = "0.1.25"
+version = "0.1.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


muursh (approved):

bpfarmer:
Perfect - was worried about this case. Does it make sense to include some test data from production to double-check that we're parsing the message correctly?

hadjiszs:
> Does it make sense to include some test data from production to double-check that we're parsing the message correctly?

Yeah, there is one already now reintroduced, `test_local_exit_root` with the 10k events for the LER computation (the `BridgeEvent` structure that is parsed is the same for exits and messages)

bpfarmer:
> Yeah, there is one already now reintroduced, test_local_exit_root with the 10k events for the LER computation (the BridgeEvent structure that is parsed is the same for exits and messages)

Ah ok nice, just wasn't sure if there were messages as well as exits. Thanks!!

Issue #219, opened 2024-08-15
Rate-limiting AggLayer transaction settlement


iljakuklic:
The frequency of L1 transaction settlement `AggLayer` performs on behalf of participating chains needs to be limited. This is needed to avoid excessive L1 fees.

## Requirements

* Networks have to be prevented from settling more often than the configured rate.
* The limit is applied to the two RPC calls that eventually result in posting a transaction on L1:
  * `sendTx` - limit the number of transactions submitted within given time interval.
  * `sendCertificate` - limit the number of transactions submitted within a specified number of epochs.
* The settlement rate limits are configurable using a configuration file.
  * There is a configurable default settlement rate.
  * The default rate can be overridden on per-network basis.
* Participating chains are notified in case they have been rate-limited.
  * The response is informative and easily differentiated from other error states.
  * The error information is machine readable to allow for automated handling.

Non-goals:

* Not a general DoS prevention mechanism.
  * Other mechanisms are responsible for this.
* We do not consider the option for participating chains to pay for a more frequent settlement at this stage.
  * This is a longer-term goal for the decentralized version of AggLayer.

## Related

* #213

Freyskeyd:
Superseded by agglayer/protocol-research#147

PR #222, opened 2024-08-16
feat: return proper error codes from RPC calls

iljakuklic:
Introduce a mechanism to return the proper error codes from failing RPC calls and use it in the existing RPC methods. This also lays the groundwork for adding arbitrary structured data to RPC errors.

This is to make the future rate limiting errors easier to distinguish from other kinds of errors and to prepare for passing additional machine readable data in the RPC error response.

Related to:
* #219
* #213 

## Breaking changes

* Error codes returned from RPC calls have been changed from generic ones, like -20602 (invalid params) to newly introduced ones that specify the issue more precisely. Anyone relying on these error codes will have to update how they are handled. This is unlikely since the original codes were not particularly informative.
* The error message string has moved from the `data` field of the error object to the `message` field which is a better match. Anyone parsing (not the brightest idea) or reading the error message will have to update which field is read.
* The `data` field has been cleared to make space for error-specific structured data to be returned together with the error code and message.

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/rpc/error.rs
+++ b/crates/agglayer-node/src/rpc/error.rs
@@ -1,0 +1,164 @@
+//! Support for structured errors in RPC.
+
+use ethers::{middleware::Middleware, types::H256};
+use jsonrpsee::types::error::ErrorObjectOwned;
+use serde::Serialize;
+
+use crate::kernel::{CheckTxStatusError, SignatureVerificationError, ZkevmNodeVerificationError};
+
+/// RPC error codes.
+pub mod code {
+    /// Rollup is not registered.
+    pub const ROLLUP_NOT_REGISTERED: i32 = -10001;
+
+    /// Rollup signature verification failure.
+    pub const SIGNATURE_MISMATCH: i32 = -10002;
+
+    /// Proof or state validation failed.
+    pub const VALIDATION_FAILURE: i32 = -10003;
+
+    /// L1 settlement failure.
+    pub const SETTLEMENT_ERROR: i32 = -10004;
+
+    /// Transaction status retrieval error.
+    pub const STATUS_ERROR: i32 = -10005;
+}
+
+#[derive(PartialEq, Eq, Serialize, Debug, Clone, thiserror::Error)]
+#[serde(rename_all = "kebab-case")]
+pub enum ValidationError {
+    #[error("Dry run failed")]
+    DryRun { detail: String },
+
+    #[error("State root verification failed")]
+    RootVerification { detail: String },
+}
+
+#[derive(PartialEq, Eq, Serialize, Debug, Clone, thiserror::Error)]
+#[serde(rename_all = "kebab-case")]
+pub enum SettlementError {
+    #[error("No receipt")]
+    NoReceipt,
+
+    #[error("IO error")]
+    IoError(String),
+
+    #[error("Contract error")]
+    Contract { detail: String },
+}
+
+impl<R: Middleware> From<crate::kernel::SettlementError<R>> for SettlementError {
+    fn from(err: crate::kernel::SettlementError<R>) -> Self {
+        match err {
+            crate::kernel::SettlementError::NoReceipt => Self::NoReceipt,
+            crate::kernel::SettlementError::ProviderError(e) => Self::IoError(e.to_string()),
+            crate::kernel::SettlementError::ContractError(e) => {
+                let detail = e.to_string();
+                Self::Contract { detail }
+            }
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, Serialize, Debug, Clone, thiserror::Error)]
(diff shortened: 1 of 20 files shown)


crates/agglayer-node/src/rpc/error.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct RpcError(ErrorObjectOwned);
+
+impl RpcError {
+    /// Create an [RpcError] from a jsonrpsee error object.
+    pub fn form_error_object(err_obj: ErrorObjectOwned) -> Self {
+        Self(err_obj)
+    }
+
+    /// Create a basic [RpcError] with just an error code and a message.
+    pub fn basic(code: i32, message: impl ToString) -> Self {
```
Freyskeyd, line 69:
```suggestion
    pub fn basic<C: Into<i32>>(code: C, message: impl ToString) -> Self {
```

iljakuklic: I think it's better to keep the interfaces tight, only allowing things that make sense. There is, for example, a `From<bool>` for `i32`. The suggested change would allow `RpcError::basic(false, "foo")`. It's not clear at all that the first argument is some sort of error code.

I'm actually thinking of going in the opposite direction: Introduce a new type for agglayer-specific error codes and accept it here rather than plain `i32`, so that using the named constants is encouraged.

Freyskeyd: Yes, we can introduce an enum `RpcCode` that resolve to the constant and either impl `From` to use it in the various needed method or have a custom method for it:

```rust
pub const ROLLUP_NOT_REGISTERED: i32 = -10101;

pub enum RpcCode {
  RollupNotRegistered
}

impl From<RpcCode> for i32 {
  fn from(value: RpcCode) -> i32 {
    match value {
      RpcCode::RollupNotRegistered => ROLLUP_NOT_REGISTERED
    }
  }
}

```

iljakuklic: So I did something like this but with a struct which is just a wrapper around the code, with a bunch of associated constants. The struct is a bit more open-ended in that it still allows being constructed form arbitrary integers. Over all it probably does not make a massive amount of difference.

crates/agglayer-node/src/rpc/error.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+#[derive(PartialEq, Eq, Serialize, Debug, thiserror::Error)]
+#[serde(rename_all = "kebab-case")]
+pub enum StatusError {
+    #[error("Transaction not found")]
+    TxNotFound { hash: H256 },
+
+    #[error("Failed get the current L1 block")]
```
Freyskeyd, line 69:
```suggestion
    #[error("Failed to get the current L1 block")]
```

crates/agglayer-node/src/rpc/tests.rs
```diff
@@ -106,6 +107,52 @@ async fn check_tx_status() {
     assert_eq!(res, "done");
 }
 
+#[tokio::test]
+async fn check_tx_status_fail() {
```
Freyskeyd, line 111: We could add some tests with `insta` to acknowledge the shape of the errors.

crates/agglayer-node/src/rpc/error.rs
```diff
@@ -0,0 +1,157 @@
...
+    #[error("Dry run failed")]
+    DryRun { detail: String },
+
+    #[error("State root verification failed")]
+    RootVerification { detail: String },
+}
+
+#[derive(PartialEq, Eq, Serialize, Debug, thiserror::Error)]
+#[serde(rename_all = "kebab-case")]
+pub enum SettlementError {
```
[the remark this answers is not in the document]
iljakuklic: This is mostly due to the `Serialize` instance which is used to populate the `"data"` field of the json error. Putting `Serialize` on `kernel::SettlementError` would probably require a custom impl or heavy annotation since most of the fields are not `Serialize`. Another way would be to come up with a different way of creating the structured data but I find the `Serialize`-based approach rather neat.

Also the error messages provided by thiserror here are a bit more brief. They end up in the `"message"` field, leaving most of the detail to the `"data"` field.

In summary this approach gives us control over how the json error is rendered without compromising on information stored in `kernel::SettlementError`.

crates/agglayer-node/src/rpc/tests.rs
iljakuklic: Added a bunch of them covering all the error cases defined in the `rpc::error` module.

crates/agglayer-node/src/rpc/error.rs
iljakuklic: Resolving this thread as the discussion does not apply anymore

PR #237, opened 2024-08-23
feat: aligns the pp inputs with the contracts

hadjiszs:
# Description

Aligns the PP inputs with the smart contracts

## PR Checklist:


crates/pessimistic-proof/src/proof.rs
```diff
@@ -53,20 +57,59 @@ pub enum ProofError {
 pub type ExitRoot = Digest;
 pub type BalanceRoot = Digest;
 pub type NullifierRoot = Digest;
-pub type LeafProofOutput = (ExitRoot, BalanceRoot, NullifierRoot);
+
+/// Outputs of the leaf proof.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LeafProofOutput {
```
wborgeaud, line 63: Just for my own understanding, is there a reason to use "leaf proof" and not "pessimistic proof" here?

wborgeaud (approved):

crates/pessimistic-proof/src/proof.rs
hadjiszs: That's a term introduced some time ago to distinguish the proofs and their aggregation.
This term isnt used that much anymore though, I'll switch to "pessimistic proof" and then for the aggregation that would just be "aggregation proof"

PR #246, opened 2024-08-26
fix: handle network vs. rollup id

### LINKED ISSUE #240: Issue: Rollup index does not match the networkID, therefore some checks must be reviewed.
# Context

Just to try to avoid confusions, that might be tricky:

The first networkID is 0 and it's mainnet
The first rollupID is 1, and the first networkID assigned to a rollup is as well 1.
In that sense RollupID == networkID in all cases, except for networkID = 0 which is reserved for mainnet.

Then we have the global Index, which is a unique index for all the leafs in the GER ( global exit root) which is defined here:
https://github.com/0xPolygonHermez/zkevm-contracts/blob/a5eacc6e51d7456c12efcabdfc1c37457f2219b2/contracts/v2/PolygonZkEVMBridgeV2.sol#L430-L434

# Issue
Note that the first rollup ( RollupID 1 == networkID 1) will be on the first leaf of the Rollups Exit tree and therefore will be the index 0, so that means that: 

- RollupID == networkID 
- LeafIndex ==  RollupID - 1
   - Note that RollupID is always >=1 ^^

So we have to tweak theses checks:
https://github.com/agglayer/agglayer/blob/fee10896ad8906a0897f18188016a174dc06fab4/crates/pessimistic-proof/src/local_state.rs#L88

https://github.com/agglayer/agglayer/blob/fee10896ad8906a0897f18188016a174dc06fab4/crates/pessimistic-proof/src/local_state.rs#L106

We could  for example add a function/parameter tot he `global_index.rs` file to calculate the networkID based on the global index, and use that instead of the rollup index, which will be something like this:

```
let network_id;
if global_index.mainnet_flag {
    network_id = 0;
} else {
    network_id = rollup_index + 1;
}
```

hadjiszs:
# Description

Remove discrepancies between the network id and the rollup id. 

Fixes #240 

## PR Checklist:


wborgeaud (approved):
👍

crates/pessimistic-proof/src/global_index.rs
```diff
@@ -15,6 +17,10 @@ pub struct GlobalIndex {
 
 impl GlobalIndex {
     const MAINNET_FLAG_OFFSET: usize = 2 * 32;
+
+    pub fn network_id(&self) -> NetworkId {
```
Freyskeyd, line 21: Are we sure we want to allow `NetworkId(0)` ?
Does `NetworkId(0)` useful for any action?

hadjiszs: Yeah `NetworkId(0)` corresponds to the L1 network

PR #249, opened 2024-08-27
feat: adding storage crate

Freyskeyd:
# Description

This PR adds the storage crate to the repository.

It is a barebone crate which contains nothing more that the structure of the different component of the storage.

It is divided into:

- `columns` which contains the column definition and the different `codec` to serialize the data
- `storage` which contains the `rocksdb` abstraction (physical storage)
- `stores` which contains the store implementation to interact with the physical storage (logical stores)

## PR Checklist:

### DIFF
--- a/crates/agglayer-storage/src/lib.rs
+++ b/crates/agglayer-storage/src/lib.rs
@@ -1,0 +1,18 @@
+// Physical storage
+#[rustfmt::skip]
+pub mod storage;
+// Logical store
+#[rustfmt::skip]
+pub mod stores;
+
+#[rustfmt::skip]
+pub mod columns;
+#[rustfmt::skip]
+pub mod error;
+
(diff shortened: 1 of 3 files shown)


crates/agglayer-storage/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+// Physical storage
+#[rustfmt::skip]
+pub mod storage;
```
iljakuklic, line 3: Does the `rustfmt::skip` attribute apply only here, or to the whole module, including the contents in `storage.rs`?

Freyskeyd: This is temporary to not fail the CI with empty files, until the implementation PR are set

iljakuklic (approved):

PR #250, opened 2024-08-27
feat: adding columns to storage

Freyskeyd:
# Description

This PR adds the different columns to the storage crate.

## PR Checklist:

### DIFF
--- a/crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
+++ b/crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
@@ -7,14 +7,15 @@ use crate::types::Hash;
 mod tests;
 
 /// Column family for the balance tree per network.
+///
 /// ## Column definition
-/// ```
-/// |-key-----------------------------------|    |-value-----------------------------|
-/// | (NetworkId, KeyType::Root)              =>  (hash(root.left) hash(root.right)) |
-/// | (NetworkId, hash(node))                 =>  (hash(node.left) hash(node.right)) |
-/// | (NetworkId, hash(node))                 =>  (hash(leaf))                       |
-/// | (NetworkId, KeyType::Leaf(hash(leaf)))  =>  (U256)                             |
-/// ```
+///
+/// | key                            | value                                   |
+/// | ---                            | --                                      |
+/// | (`NetworkId`, `KeyType::Root`) | (`hash(root.left)`, `hash(root.right)`) |
+/// | (`NetworkId`, `hash(node)`)    | (`hash(node.left)`, `hash(node.right)`) |
+/// | (`NetworkId`, `hash(node)`)    | (`hash(leaf)`)                          |
+/// | (`NetworkId`, `KeyType::Leaf`) | [`U256`](type@reth_primitives::U256)    |
 pub struct BalanceTreePerNetworkColumn;
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
--- a/crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
+++ b/crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
@@ -1,19 +1,20 @@
-use bincode::Options;
 use serde::{Deserialize, Serialize};
 
-use super::{default_bincode_options, Codec, ColumnSchema, BALANCE_TREE_PER_NETWORK_CF};
-use crate::{error::Error, types::Hash};
+use super::{Codec, ColumnSchema, BALANCE_TREE_PER_NETWORK_CF};
+use crate::types::Hash;
 
 #[cfg(test)]
 mod tests;
 
 /// Column family for the balance tree per network.
-///
-/// | ------------ key ------------ |    | ------------- value -------------- |
-/// | NetworkId "root"              | => | hash(node.left) hash(node.right)   |
-/// | NetworkId hash(node)          | => | hash(node.left) hash(node.right)   |
-/// | NetworkId hash(node)          | => | hash(leaf)                         |
-/// | NetworkId "leaves" hash(leaf) | => | leaf bytes array                   |
+/// ## Column definition
+/// ```
+/// |-key-----------------------------------|    |-value-----------------------------|
+/// | (NetworkId, KeyType::Root)              =>  (hash(root.left) hash(root.right)) |
+/// | (NetworkId, hash(node))                 =>  (hash(node.left) hash(node.right)) |
+/// | (NetworkId, hash(node))                 =>  (hash(leaf))                       |
+/// | (NetworkId, KeyType::Leaf(hash(leaf)))  =>  (U256)                             |
+/// ```
 pub struct BalanceTreePerNetworkColumn;
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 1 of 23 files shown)


crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+use super::{default_bincode_options, Codec, ColumnSchema, BALANCE_TREE_PER_NETWORK_CF};
+use crate::{error::Error, types::Hash};
+
+#[cfg(test)]
+mod tests;
+
+/// Column family for the balance tree per network.
+///
+/// | ------------ key ------------ |    | ------------- value -------------- |
+/// | NetworkId "root"              | => | hash(node.left) hash(node.right)   |
```
wborgeaud, line 13: nit: `hash(node.left) hash(node.right)` -> `hash(root.left) hash(root.right)`

crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+/// | NetworkId hash(node)          | => | hash(node.left) hash(node.right)   |
+/// | NetworkId hash(node)          | => | hash(leaf)                         |
+/// | NetworkId "leaves" hash(leaf) | => | leaf bytes array                   |
```
wborgeaud, line 16: The balances in the balance tree are `U256`, so this map is not really required. We could just store the balances directly.

crates/agglayer-storage/src/columns/nullifier_tree_per_network/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+#[cfg(test)]
+mod tests;
+
+/// Column family for the nullifier tree per network.
+///
+/// | ------------ key ------------ |    | ------------- value -------------- |
+/// | NetworkId "root"              | => | hash(node.left) hash(node.right)   |
+/// | NetworkId hash(node)          | => | hash(node.left) hash(node.right)   |
+/// | NetworkId hash(node)          | => | hash(leaf)                         |
+/// | NetworkId "leaves" hash(leaf) | => | leaf bytes array                   |
```
wborgeaud, line 16: Same as for the balances, the nullifier tree leaves are boolean, so we could just store them directly.

crates/agglayer-storage/src/columns/proof_per_certificate/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+/// Column family that returns the generated proof for one certificate.
+///
+/// | --- key ------  |    | --- value ---- |
+/// | CertificateId   | => | Proof bytes    |
+pub struct ProofPerCertificateColumn;
+
+impl Codec for Vec<u8> {
+    fn encode(&self) -> Result<Vec<u8>, crate::error::Error> {
+        Ok(default_bincode_options().serialize(&self)?)
```
iljakuklic, line 17: Nit: this pattern involving `Ok(default_bincode_options().serialize(&self)?)` could be a helper function in the module where `Codec` is defined.

crates/agglayer-storage/src/error.rs
```diff
@@ -0,0 +1,11 @@
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Trying to access an unknown ColumnFamily")]
+    ColumnFamilyNotFound,
+
+    #[error("Serialization error: {0}")]
+    Serialization(#[from] bincode::Error),
+    #[error(r#"An unexpected error occurred: {0}
+        This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
+    Unexpected(String),
```
iljakuklic, line 10: Would a `Serialization` issue also qualify as a critical bug that needs to be reported? I'd expect the data structures that are stored in the permanent storage to be defined in a way to make them always serializable without an issue.

crates/agglayer-storage/src/columns/certificate_per_network/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn decode(buf: &[u8]) -> Result<Self, Error> {
+        Ok(default_bincode_options().deserialize(buf)?)
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Value {
+    certificate_id: [u8; 32],
+    epoch_number: u64,
+    certificate_index: u64,
```
iljakuklic, line 36: Why not use `types::CertificateId`, `types::EpochNumber`, `types::CertificateIndex`, respectively?

crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
Freyskeyd: I wasn't sure about the payload of the leaves. Going to switch this to `U256`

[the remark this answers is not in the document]
Freyskeyd: Yes, I wasn't sure about the syntax here. I did some tests, but I'm totally fine changing it.
Will try to push a commit with a new syntax version, let me know what you think about it.

crates/agglayer-storage/src/columns/latest_certificate_per_network/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+/// Column family for the latest settled certificate per network.
+/// The key is the network_id and the value is the certificateID,
+/// the height and the epoch_number.
+///
+/// | --- key ---- |    | ---------- value ----------  |
+/// | NetworkID    | => | Proof CertificateID Height   |
+pub struct LatestSettledCertificatePerNetworkColumn;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ProvenCertificate(pub [u8; 32], pub u64, pub u64);
```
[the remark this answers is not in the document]
Freyskeyd: Yes I agree, this structure changes in a next PR tho

crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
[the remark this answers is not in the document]
Freyskeyd: Here's the possible format that I think about:

Screenshot 2024-09-03 at 11 05 13

Screenshot 2024-09-03 at 11 05 22

I think I prefer the one in codeblock at the end, or the second one (without the fat arrow), what do you think @iljakuklic ?

crates/agglayer-storage/src/columns/proof_per_certificate/mod.rs
Freyskeyd: Indeed, I will put a default trait function

crates/agglayer-storage/src/error.rs
Freyskeyd: Yes we can do that

crates/agglayer-storage/src/columns/balance_tree_per_network/mod.rs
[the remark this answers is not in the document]
iljakuklic: > I think I prefer the one in codeblock at the end, or the second one (without the fat arrow), what do you think @iljakuklic ?

I agree with this assesment. I'd say proper table is preferable from the point of view of semantic encoding into html, so I am slightly leaning towards that one. Codeblock is also fine if you like it more.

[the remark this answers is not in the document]
Freyskeyd: Still trying to find the best solution here:

This version use rustdoc to generate the table

```rust
/// ## Column definition
///
/// | key                            | value                                   |
/// | ---                            | --                                      |
/// | ([NetworkId], [KeyType::Root]) | ([hash(root.left)], [hash(root.right)]) |
/// | ([NetworkId], [hash(node)])    | ([hash(node.left)], [hash(node.right)]) |
/// | ([NetworkId], [hash(node)])    | ([hash(leaf)])                          |
/// | ([NetworkId], [KeyType::Leaf]) | ([`U256`](type@reth_primitives::U256))  |
///
```

which render to:

Screenshot 2024-09-04 at 15 18 31

iljakuklic (approved):

crates/agglayer-storage/src/columns/latest_certificate_per_network/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Column family for the latest settled certificate per network.
+/// The key is the network_id and the value is the certificateID,
+/// the height and the epoch_number.
+///
+/// ## Column definition
+///
+/// | key         | value                                |
+/// | --          | --                                   |
+/// | `NetworkId` | (`Proof`, `CertificateId`, `Height`) |
```
hadjiszs, line 17: (saw this while reviewing 252, inconsistency between the comment and the table)

PR #253, opened 2024-08-27
feat: adding config for the storage

Freyskeyd:
# Description

This PR adds the capability to configure the storage in the config file.

Parent: #252 

## PR Checklist:


crates/agglayer-config/src/storage.rs
```diff
@@ -0,0 +1,63 @@
...
+            pending_path: default_pending_path(),
+            state_path: default_state_path(),
+            epochs_path: default_epochs_path(),
+        }
+    }
+}
+
+/// Returns the default path to the metadata storage directory.
+fn default_metadata_path() -> PathBuf {
+    match dirs::config_dir() {
```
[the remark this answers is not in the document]
Freyskeyd: I was planning to use `config_dir` to store the database close to the config itself, so that we have one folder which contains everything related to the AggLayer.

crates/agglayer/tests/config.rs
```diff
@@ -9,7 +9,18 @@ fn config_display() -> Result<(), Box<dyn std::error::Error>> {
...
+
+#[cfg(test)]
+#[allow(dead_code)]
+pub fn sanitize_config_folder_path(cmd_out: &str) -> String {
+    let dir = dirs::config_dir().unwrap().join("agglayer");
+    let pattern =
+        regex::Regex::new(dir.to_str().expect("Unable to create regex dir exclusion")).unwrap();
+    pattern
+        .replace_all(cmd_out, "/tmp/agglayer-test")
+        .to_string()
```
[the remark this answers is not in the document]
Freyskeyd: Yeah, I planned to make this function more configurable but end up not having the need for, will remove that !

iljakuklic (approved):

crates/agglayer-config/src/storage.rs
```diff
@@ -0,0 +1,145 @@
...
+pub struct StorageConfig {
+    /// Base path of the storage.
+    pub base_path: PathBuf,
+    /// Custom metadata storage path or infered from the base path.
+    pub metadata_path: PathBuf,
+    /// Custom pending storage path or infered from the base path.
+    pub pending_path: PathBuf,
+    /// Custom state storage path or infered from the base path.
+    pub state_path: PathBuf,
+    /// Custom epochs storage path or infered from the base path.
```
hadjiszs, line 20:
```suggestion
    /// Custom metadata storage path or inferred from the base path.
    pub metadata_path: PathBuf,
    /// Custom pending storage path or inferred from the base path.
    pub pending_path: PathBuf,
    /// Custom state storage path or inferred from the base path.
    pub state_path: PathBuf,
    /// Custom epochs storage path or inferred from the base path.
```

hadjiszs (approved):

crates/agglayer-config/src/storage.rs
```diff
@@ -0,0 +1,153 @@
...
+pub struct StorageConfig {
+    /// Custom metadata storage path or inferred from the db path.
+    pub metadata_db_path: PathBuf,
+    /// Custom pending storage path or inferred from the db path.
+    pub pending_db_path: PathBuf,
+    /// Custom state storage path or inferred from the db path.
+    pub state_db_path: PathBuf,
+    /// Custom epochs storage path or inferred from the db path.
+    pub epochs_db_path: PathBuf,
+}
```
iljakuklic, line 24: At risk of overcomplicating things... Could there be an intermediate `storage_dir` config, defaulting to `"storage"`?

Then, for example, the path to the epochs db would be:
```rust
base_dir.join(storage_dir).join(epochs_db_path)
```

This should work as expected even in presence of absolute paths as noted in the other comment.

It allows the user to change the location of all the databases without changing `base_dir` or individual database paths. Having the databases in an extra directory may be useful if the user wants to nuke the database without deleting other stuff that might live in the base directory (configs, logs, whatever).

crates/agglayer-config/src/storage.rs
```diff
@@ -0,0 +1,153 @@
...
+        let base = value.join(STORAGE_DIR);
+
+        Self {
+            metadata_db_path: base.join(METADATA_DB_NAME),
+            pending_db_path: base.join(PENDING_DB_NAME),
+            state_db_path: base.join(STATE_DB_NAME),
+            epochs_db_path: base.join(EPOCHS_DB_PATH),
+        }
+    }
+}
```
iljakuklic, line 71: Having this as a `From` impl seems a bit too magical. Would prefer an explicit constructor, like `StorageConfig::from_base_path` for this.

crates/agglayer-config/src/storage.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// Returns the default path to the state storage directory.
+fn default_state_path(path: &Path) -> PathBuf {
+    path.join("state")
+}
+
+/// Returns the default path to the epochs storage directory.
+fn default_epochs_path(path: &Path) -> PathBuf {
+    path.join("epochs")
```
iljakuklic, line 130: There are constants for the directory names at the top of the file.

crates/agglayer-config/src/storage.rs
Freyskeyd: That's exactly what the `Deserialize` implementation do. Currently you just need to define the `db-path` and all the other path are inferred from it. 
I removed the fact that one can define a custom path for one database only, if we need it we'll add it later.

The configuration looks like this now:

```toml

[storage]
db-path = "./database"
```

iljakuklic (approved):

PR #254, opened 2024-08-27
fix: use U512 in `apply_batch_header` for balances

### LINKED ISSUE #242: Issue: imported_bridge_exit and bridge_exit can easily overflow the amount. An attacker could DoS easily the system.
# Issue 

In those lines the balances are udpated:
https://github.com/agglayer/agglayer/blob/fee10896ad8906a0897f18188016a174dc06fab4/crates/pessimistic-proof/src/local_state.rs#L125-L133

https://github.com/agglayer/agglayer/blob/fee10896ad8906a0897f18188016a174dc06fab4/crates/pessimistic-proof/src/local_state.rs#L170-L182

The problem comes where an overflow could be easily performed by an attacker, wasting lots of resources to the system or even stopping it:

An attacker creates a new token on network 1, and transfer N times the max amount ( Max_UINT_256) to network 2.

Now the attacker just bridge back his token to his network N times with the same amount.  In order to be able to proof that interaction in the pessimistic, due the overflow issues, we must generate a pessimistic proof N times ( the same N as the bridges) in order to be able to proof this, since we cannot "accumulate" the bridges amount due to overflow issues.

Other issues even bigger could be that a transaction could not be proven never, even if the "execution" seemed fine, example:

Attacker bridge 2 times (Max_UINT_256) ATTACK tokens to network 2 .
Now bridges back 1 Attack Token, and then Max_UINT_256 ATTACK tokens.
The pessimistic won't be able to proof this, since, in order to proof that has enough ends in an overflow!

As a mitigation for this issue i propose to expand the balance tree uint256 to uint512

wborgeaud:
Use `U512` instead of `U256` in the temporary balances used in `apply_batch_header`.
Partially fixes #242.

## PR Checklist:


crates/pessimistic-proof/src/local_state.rs
```diff
@@ -197,6 +200,10 @@ impl LocalNetworkState {
         // TODO: implement batch `verify_and_update` for the LBT
         for (token, (old_balance, balance_path)) in &multi_batch_header.balances_proofs {
             let new_balance = new_balances[token];
+            if new_balance > U512::from(U256::MAX) {
+                return Err(ProofError::BalanceOverflowInBridgeExit);
+            }
+            let new_balance = U256::from(new_balance);
```
iljakuklic, line 206: Nitpick: is there something `U256::try_from` to roll the check and conversion into one call? I see `U256::uint_try_from` exists but not sure how that works.

iljakuklic (commented):
+1. Would be interesting to see how this impacts the cycle counts.

wborgeaud:
> +1. Would be interesting to see how this impacts the cycle counts.

For `test_sp1_simple` the cycle count goes from 2495476 to 2495795, so 319 additional cycles with this change. For real-world PPs it will obviously be more, but still fairly negligible I think.

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -197,6 +200,11 @@ impl LocalNetworkState {
         // TODO: implement batch `verify_and_update` for the LBT
         for (token, (old_balance, balance_path)) in &multi_batch_header.balances_proofs {
             let new_balance = new_balances[token];
+            let new_balance = if let Ok(balance) = U256::uint_try_from(new_balance) {
+                balance
+            } else {
+                return Err(ProofError::BalanceOverflowInBridgeExit);
+            };
```
iljakuklic, line 207: Nitpick:
```suggestion
            let new_balance = U256::uint_try_from(new_balance).map_err(|_| ProofError::BalanceOverflowInBridgeExit)?;
```

iljakuklic (approved):

Issue #234, opened 2024-08-23
[audit] keccak256_combine is not a safe generic hashing function


tob-joe:
Severity: Informational
Difficulty: Not Applicable
Type: Cryptography
Target: crates/pessimistic-proof/src/keccak.rs

## Description
The function `keccak256_combine`, shown below, returns a hash of an array of values.
https://github.com/agglayer/agglayer/blob/d9d33885b6a396a69f57054d7b549edc6ca1ef51/crates/pessimistic-proof/src/keccak.rs#L40-L54
Although it is generic over the type parameter `T`, `keccak256_combine` is not guaranteed to be collision-resistant for sequences when `T` has values with different byte-lengths, due to ambiguous encodings. For example, the following test, where `T` is `Vec<u8>`, fails:
```rs
#[test]
fn test_keccak256_combine_collision() {
    assert_ne!(keccak256_combine(vec![vec![0,1],vec![2]]),
               keccak256_combine(vec![vec![0],vec![1,2]]))
}
```

This construction is an example of [the YoloMultiHash ad-hoc hash construction](https://blog.trailofbits.com/2024/08/21/yolo-is-not-a-valid-hash-construction/), which is unsafe when fields of the data structure can overlap, thus leading to hash values for semantically different inputs.

We have not discovered any current uses of `keccak256_combine` in the `pessimistic-proof` crate which are vulnerable due to this issue. Most inputs are implicitly protected because they have fixed lengths. For example, the `BridgeExit` hash has a fixed input length of 85 bytes, even though it uses `keccak256_combine` where `T` is the variable-length type `[u8]` https://github.com/agglayer/agglayer/blob/d9d33885b6a396a69f57054d7b549edc6ca1ef51/crates/pessimistic-proof/src/bridge_exit.rs#L101-L112

## Recommendations
Short term, review all use of `keccak256_combine` to ensure unambiguous encodings. Consider modifying `keccak256_combine` to generically ensure unambiguous encoding. For example, by restricting `T` to fixed-size types or prefixing each entry with its length.

Long term, consider adopting standard schemes for hashing structured data, e.g. TupleHash.

## References
[“YOLO” is not a valid hash construction](https://blog.trailofbits.com/2024/08/21/yolo-is-not-a-valid-hash-construction/)

wborgeaud:
Added a comment regarding the safety of `keccak256_combine` in #256. 
[...]

PR #256, opened 2024-08-28
chore: add safety comment to `keccak256_combine`

wborgeaud:
As noted in #234, `keccak256_combine` should only be called with fixed-size inputs (which is currently the case in the code). This PR adds a comment mentioning that in the function's doc comment so that implementers are aware.

## PR Checklist:

### DIFF
--- a/crates/pessimistic-proof/src/keccak.rs
+++ b/crates/pessimistic-proof/src/keccak.rs
@@ -38,6 +38,7 @@ pub fn keccak256(data: &[u8]) -> Digest {
 }
 
 /// Hashes the input items using a Keccak hasher with a 256-bit security level.
+/// Safety: This function should only be called with fixed-size items to avoid collisions.
 pub fn keccak256_combine<I, T>(items: I) -> Digest
 where
     I: IntoIterator<Item = T>,


crates/pessimistic-proof/src/keccak.rs
```diff
@@ -38,6 +38,7 @@ pub fn keccak256(data: &[u8]) -> Digest {
 }
 
 /// Hashes the input items using a Keccak hasher with a 256-bit security level.
+/// Safety: This function should only be called with fixed-size items to avoid collisions.
```
iljakuklic, line 41: Maybe we could try to impose the constraint by types. Now we have `T: AsRef<[u8]>`. Maybe changing it to something like `T: AsRef<[u8; N]>` or `T: Borrow<[u8; N]>` or `T: Into<[u8; N]>` (not sure which one to pick here, maybe just try which one, if any, already compiles) would work. There would have to be a new generic param `const N: usize`.

iljakuklic (approved):

crates/pessimistic-proof/src/keccak.rs
wborgeaud: I think this wouldn't allow hashing items of different size, which we do for example here https://github.com/agglayer/agglayer/blob/e462db6b30a76807f2763a320f7a552c5181b1dd/crates/pessimistic-proof/src/bridge_exit.rs#L103

iljakuklic: > I think this wouldn't allow hashing items of different size, which we do for example here
> 
> https://github.com/agglayer/agglayer/blob/e462db6b30a76807f2763a320f7a552c5181b1dd/crates/pessimistic-proof/src/bridge_exit.rs#L103

For hashing fixed number of heterogeneous data item, we could just put all the items into a tuple and hash that. I think it gives the same result, does it not?

wborgeaud: Yes, as mentioned #234 there are secure ways to serialize data for hashing, but they are only required when hashing variable length data, which we don't do in our crates. So I think it's fine to keep using `keccak256_combine` with this safety note so that everyone is aware of the potential vulnerability.

Issue #231, opened 2024-08-21
Investigate how to log RPC call issues upon client disconnect


iljakuklic:
The RPC methods, such as `sendTx`, take some care to log any issues encountered. However, some corner cases are missed. When a client waiting for a result of an RPC call disconnects, the `jsonrpsee` library drops the future representing the RPC call execution. This is an error state that bypasses the logging present in the `sendTx` implementation.

Ways to log disconnection should be explored. This should not be done naively, since logging every time the future is dropped would result in the log being easily filled up with spurious data, i.e. the log should only happen after disconnection of requests where some basic checks have already been performed.

Related:
* #230

Issue #230, opened 2024-08-21
Review RPC timeouts


iljakuklic:
The Agglayer node connects to various other services via RPC, most notably the L1 node and L2 nodes. The L2 queries are used to obtain the current state roots. There should be a sensible timeout if an RPC call to any of these gets stuck.

This applies to:
* Agglayer <-> L1 node communication
* Agglayer <-> L2 nodes communication
* Agglayer <-> proof generation service communication

As noted by @praetoriansentry:

> Agglayer needs time outs - is it just waiting forever right now? This should have been picked up in reliability / chaos testing... which we never did.

This may result in the `sendTx` call being stuck rather than returning an error. If the client just disconnects, it is not obvious how to diagnose the issue.

iljakuklic:
I took a look at how the Agglayer <-> L2 nodes RPC communication is done. Initial investigation suggests there is the default `jsonrpsee` timeout of 60 seconds. This is not currently configurable.

The reason we have not seen (and logged) the timeout error is (most likely) that the request was dropped before the timeout had the chance to occur as described in #231.

Freyskeyd:
Yes, the default value is 60sec for every RPCs that use `Http` client of `jsonrpsee`. We can configure this with this [method](https://docs.rs/jsonrpsee-http-client/0.24.3/jsonrpsee_http_client/struct.HttpClientBuilder.html#method.request_timeout).

iljakuklic:
Continuing investigation... for contacting the L1 node, the `Http` provider from the `ethers` library is used with default settings. Timeouts are handled by the underlying library which in this case is `reqwest`. By default, the `reqwest::Client` imposes **no timeout** on the duration of a request. This could be changed by creating the `reqwest::Client` object manually with appropriate options and passing it to the ethers provider upon construction.

iljakuklic:
As for the proof generation service, it's not quite clear to me what will be used in the end so there's not much to say about it at the moment.

iljakuklic:
For completeness. The RPC server implementation that handles the RPC calls to the Agglayer node is `jsonrpsee`. It is possible to impose timeout for the duration of any request served by the agglayer node. It can be done by adding the `timeout` layer to the http middleware when building the RPC server object. It is demonstrated [in the docs for the `ServerBuilder::set_middleware` function](https://docs.rs/jsonrpsee-server/0.24.3/jsonrpsee_server/struct.ServerBuilder.html#method.set_http_middleware).

[...]

PR #259, opened 2024-08-28
fix: adjust RPC timeouts

iljakuklic:
# Description

Addresses
* #230 
* #231 

For timeouts to the L1 node requests, there was previously none, so introduce one of 45s. Contract execution should fit within 45s easily.

For timeouts to the L2 node requests, there was one of 60s. However, if the request to the agglayer node is dropped before that timeout, we do not get any useful diagnostic. So lower it to 45s as well, so we increase chances that our request to the L2 node times out before the client request to agglayer, which gives us more room to log the error. The requests to L2 nodes are rather trivial so 45s should be more than enough. This is only a partial fix / workaround, ideally cancelled requests are logged regardless of timeout.

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/kernel/mod.rs
+++ b/crates/agglayer-node/src/kernel/mod.rs
@@ -75,9 +75,11 @@ impl<RpcProvider> Kernel<RpcProvider> {
             .get(&rollup_id)
             .ok_or(ZkevmNodeVerificationError::InvalidRollupId(rollup_id))?;
 
-        Ok(ZkevmNodeClient::new(
-            jsonrpsee::http_client::HttpClientBuilder::new().build(url.as_str())?,
-        ))
+        let client = jsonrpsee::http_client::HttpClientBuilder::new()
+            .request_timeout(std::time::Duration::from_secs(45))
+            .build(url.as_str())?;
+
+        Ok(ZkevmNodeClient::new(client))
     }
 
     /// Verify that the given [`SignedTx`] is valid according to the ZkEVM node.
--- a/crates/agglayer-node/src/kernel/mod.rs
+++ b/crates/agglayer-node/src/kernel/mod.rs
@@ -76,7 +76,7 @@ impl<RpcProvider> Kernel<RpcProvider> {
             .ok_or(ZkevmNodeVerificationError::InvalidRollupId(rollup_id))?;
 
         let client = jsonrpsee::http_client::HttpClientBuilder::new()
-            .request_timeout(std::time::Duration::from_secs(45))
+            .request_timeout(self.config.l2.rpc_timeout)
             .build(url.as_str())?;
 
         Ok(ZkevmNodeClient::new(client))
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -70,7 +70,7 @@ impl Node {
         let address = signer.address();
         // Create a new L1 RPC provider with the configured signer.
         let client = reqwest::Client::builder()
-            .timeout(std::time::Duration::from_secs(45))
+            .timeout(config.l1.rpc_timeout)
             .build()?;
         let rpc = Provider::new(Http::new_with_client(config.l1.node_url.clone(), client))
             .with_signer(signer)
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -69,7 +69,10 @@ impl Node {
         let signer = ConfiguredSigner::new(config.clone()).await?;
         let address = signer.address();
         // Create a new L1 RPC provider with the configured signer.
-        let rpc = Provider::<Http>::try_from(config.l1.node_url.as_str())?
+        let client = reqwest::Client::builder()
+            .timeout(std::time::Duration::from_secs(45))
+            .build()?;
+        let rpc = Provider::new(Http::new_with_client(config.l1.node_url.clone(), client))
             .with_signer(signer)
             .nonce_manager(address);
 
--- a/crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
+++ b/crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
@@ -1,0 +1,83 @@
+//! RPC middleware for recording cancelled requests.
+
+use std::{borrow::Cow, future::Future};
+
+use jsonrpsee::{
+    server::middleware::rpc::RpcServiceT,
+    types::{Id, Request},
+    MethodResponse,
+};
+
+/// An RPC layer that logs request cancellations.
+#[derive(Clone, Debug)]
+pub struct CancelLoggerLayer {}
+
+impl CancelLoggerLayer {
+    pub fn new() -> Self {
+        CancelLoggerLayer {}
+    }
+}
+
+impl<S> tower::Layer<S> for CancelLoggerLayer {
+    type Service = CancelLoggerService<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        CancelLoggerService(inner)
+    }
+}
+
+pub struct CancelLoggerService<S>(S);
+
+impl<'a, S: RpcServiceT<'a>> RpcServiceT<'a> for CancelLoggerService<S> {
+    type Future = CancelLoggerFuture<'a, S::Future>;
+
+    fn call(&self, request: Request<'a>) -> Self::Future {
+        CancelLoggerFuture {
+            completed: false,
+            method: request.method.clone(),
+            request_id: request.id.clone(),
+            inner: self.0.call(request),
+        }
+    }
+}
+
+#[pin_project::pin_project(PinnedDrop)]
+pub struct CancelLoggerFuture<'a, F> {
+    // Future state.
+    completed: bool,
+
+    // Method information.
+    method: Cow<'a, str>,
+    request_id: Id<'a>,
+
+    // The future to execute under a timeout.
+    #[pin]
+    inner: F,
+}
+
+impl<F: Future<Output = MethodResponse>> Future for CancelLoggerFuture<'_, F> {
+    type Output = MethodResponse;
+
+    fn poll(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        let this = self.project();
+        let poll_result = this.inner.poll(cx);
+        if poll_result.is_ready() {
+            *this.completed = true;
+        }
+        poll_result
+    }
+}
+
+#[pin_project::pinned_drop]
+impl<F> PinnedDrop for CancelLoggerFuture<'_, F> {
+    fn drop(self: std::pin::Pin<&mut Self>) {
+        if !self.completed {
+            let method = &*self.method;
+            let id = &self.request_id;
+            tracing::warn!("Request ID `{id}` to `{method}` was cancelled");
+        }
+    }
+}
--- a/crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
+++ b/crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
@@ -73,7 +73,7 @@ impl<F> PinnedDrop for CancelLoggerFuture<'_, F> {
         if !self.completed {
             let method = &*self.request_info.method;
             let id = &self.request_info.request_id;
-            tracing::warn!("Request ID `{id}` to `{method}` was cancelled");
+            tracing::info!("Request ID `{id}` to `{method}` was cancelled");
         }
     }
 }
--- a/crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
+++ b/crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
@@ -1,12 +1,10 @@
 //! RPC middleware for recording cancelled requests.
 
-use std::{borrow::Cow, future::Future};
+use std::future::Future;
 
-use jsonrpsee::{
-    server::middleware::rpc::RpcServiceT,
-    types::{Id, Request},
-    MethodResponse,
-};
+use jsonrpsee::{server::middleware::rpc::RpcServiceT, types::Request, MethodResponse};
+
+use super::RequestInfo;
 
 /// An RPC layer that logs request cancellations.
 #[derive(Clone, Debug)]
@@ -34,8 +32,7 @@ impl<'a, S: RpcServiceT<'a>> RpcServiceT<'a> for CancelLoggerService<S> {
     fn call(&self, request: Request<'a>) -> Self::Future {
         CancelLoggerFuture {
             completed: false,
-            method: request.method.clone(),
-            request_id: request.id.clone(),
+            request_info: RequestInfo::from_request(&request),
             inner: self.0.call(request),
         }
     }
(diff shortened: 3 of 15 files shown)


crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -75,9 +75,11 @@ impl<RpcProvider> Kernel<RpcProvider> {
             .get(&rollup_id)
             .ok_or(ZkevmNodeVerificationError::InvalidRollupId(rollup_id))?;
 
-        Ok(ZkevmNodeClient::new(
-            jsonrpsee::http_client::HttpClientBuilder::new().build(url.as_str())?,
-        ))
+        let client = jsonrpsee::http_client::HttpClientBuilder::new()
+            .request_timeout(std::time::Duration::from_secs(45))
```
Freyskeyd, line 79: How did you end up choosing 45 secs?
This shouldn't be a fixed value but a configurable one using the config file

crates/agglayer-node/src/node.rs
```diff
@@ -69,7 +69,10 @@ impl Node {
         let signer = ConfiguredSigner::new(config.clone()).await?;
         let address = signer.address();
         // Create a new L1 RPC provider with the configured signer.
-        let rpc = Provider::<Http>::try_from(config.l1.node_url.as_str())?
+        let client = reqwest::Client::builder()
+            .timeout(std::time::Duration::from_secs(45))
```
Freyskeyd, line 73: The settlement on L1 is waiting for `N` confirmations, maybe 45sec can be too short. The average block time on ETH is 12.06, so I think we should be good.
This also needs to be configurable to be able to update this value without rebuilding the agglayer

Freyskeyd (changes requested):
I'm wondering if we could add some tests for checking that the timeout is working as expected, I'm afraid that the query (for jsonrpsee) is dropped without any logs or infos on our side.

crates/agglayer-node/src/node.rs
iljakuklic: > The settlement on L1 is waiting for `N` confirmations, maybe 45sec can be too short. The average block time on ETH is 12.06, so I think we should be good.

I think the checking whether we have the required number of confirmations does not happen in single RPC request, or does it? Will double check once again.

> This also needs to be configurable to be able to update this value without rebuilding the agglayer

crates/agglayer-node/src/kernel/mod.rs
iljakuklic: > How did you end up choosing 45 secs?

As written in the PR description:

> If the request to the agglayer node is dropped before that timeout, we do not get any useful diagnostic. So lower it to 45s as well, so we increase chances that our request to the L2 node times out before the client request to agglayer, which gives us more room to log the error. The requests to L2 nodes are rather trivial so 45s should be more than enough. This is only a partial fix / workaround, ideally cancelled requests are logged regardless of timeout.

So it's basically the common default of 60s minus some slack for various latencies.

crates/agglayer-node/src/node.rs
iljakuklic: > > The settlement on L1 is waiting for `N` confirmations, maybe 45sec can be too short. The average block time on ETH is 12.06, so I think we should be good.
> 
> I think the checking whether we have the required number of confirmations does not happen in single RPC request, or does it? Will double check once again.

Yea, basically  `await`ing on `ethers::PendingTransaction` periodically polls `get_transaction` until it shows up in the chain. When that is successful, it periodically polls `get_transaction_receipt` until the sufficient number of confirmations is obtained. The timeout introduced above applies to the single `get_transaction` or `get_transaction_receipt` request.

However, there is a different issue. If I'm reading the code correctly, there is no limit on the number of retries of these requests. So for example, if the L1 node is down, it keeps retrying indefinitely. As far as I can see, there is a mechanism to flag it up to the agglayer node operators (other than perhaps some very low-level logging).

Freyskeyd: Ah sorry, I though the PR was ready to review 😅 

The `retries` defined in the `PendingTransaction` is for unconfirmed transaction, I do not see any limit to the number of request too

iljakuklic: > Ah sorry, I though the PR was ready to review 😅

It was at the time you reviewed, so it's a mistake on my part.

iljakuklic:
Made the timeouts configurable.

However, there is still the issue with the settlement transaction submission taking a potentially unbounded number of requests, so it might take an unbounded amount of time.

Freyskeyd:
> Made the timeouts configurable.
> 
> However, there is still the issue with the settlement transaction submission taking a potentially unbounded number of requests, so it might take an unbounded amount of time.

I think that for that we need to switch the whole `jsonrpsee` to a `tower` service with a timeout layer

iljakuklic:
> > However, there is still the issue with the settlement transaction submission taking a potentially unbounded number of requests, so it might take an unbounded amount of time.
> 
> I think that for that we need to switch the whole `jsonrpsee` to a `tower` service with a timeout layer

I am not quite sure how that would work. The L1 client uses `reqwest` for issuing jsonrpc calls, and I think this is pretty much hardcoded in the `ethers` library if I'm not mistaken.

Are you referring to the jsonrpc server in the agglayer node? In that case, `jsonrpsee` allows you to inject a custom `tower` layers into the stack. Adding the `Timeout` layer here would limit the total running time of any given request to agglayer. There are some drawbacks. As far as I can see, the `Timeout` layer does not log the fact that the request timed out, so maybe we'd prefer a custom one. Also, it is not possible to set different timeouts for different kinds of requests with it (e.g. for `sendTx` vs `getTxStatus`).

It seems to me that similar effect could be achieved by just a strategically placed `tokio::timeout` around the `PendingTransaction` future. This approach is pretty flexible but compared to a global tower layer, it is more easily forgotten when developing new code which might re-introduce the issue under different conditions.

Freyskeyd:
> > > However, there is still the issue with the settlement transaction submission taking a potentially unbounded number of requests, so it might take an unbounded amount of time.
> > 
> > 
> > I think that for that we need to switch the whole `jsonrpsee` to a `tower` service with a timeout layer
> 
> I am not quite sure how that would work. The L1 client uses `reqwest` for issuing jsonrpc calls, and I think this is pretty much hardcoded in the `ethers` library if I'm not mistaken.
> 
> Are you referring to the jsonrpc server in the agglayer node? In that case, `jsonrpsee` allows you to inject a custom `tower` layers into the stack. Adding the `Timeout` layer here would limit the total running time of any given request to agglayer. There are some drawbacks. As far as I can see, the `Timeout` layer does not log the fact that the request timed out, so maybe we'd prefer a custom one. Also, it is not possible to set different timeouts for different kinds of requests with it (e.g. for `sendTx` vs `getTxStatus`).
> 
> It seems to me that similar effect could be achieved by just a strategically placed `tokio::timeout` around the `PendingTransaction` future. This approach is pretty flexible but compared to a global tower layer, it is more easily forgotten when developing new code which might re-introduce the issue under different conditions.

Yes, the goal is to spawn `jsonrpsee` service into tokio and handle the requests from `hyper` in a tower service that call the underlying `jsonrpsee` service. That way, we can detect when a connection is dropped.

We can also have a custom RpcMiddleware that handles the timeout per RPC.

iljakuklic:
> We can also have a custom RpcMiddleware that handles the timeout per RPC.

Added a custom middleware that, in addition to limiting the total amount of time taken up by request processing, also issues a log entry if the timeout is hit. There is a caveat. The implementation assumes there is some sort of RPC limiting in place lower in the stack. If it's not in place, an adversary or a misbehaving client could fill the logs with spam entries just by establishing RPC requests and letting them time out.

iljakuklic:
Added a middleware for logging cancelled RPC requests. Similar caveat as for the timeout logging applies here. In the absence of an external rate limiter, and adversary or a misbehaving client will be able to spam the logs.

Still needs some tests.

crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
```diff
@@ -0,0 +1,83 @@
...
+#[pin_project::pin_project(PinnedDrop)]
+pub struct CancelLoggerFuture<'a, F> {
+    // Future state.
+    completed: bool,
+
+    // Method information.
+    method: Cow<'a, str>,
+    request_id: Id<'a>,
+
+    // The future to execute under a timeout.
```
Freyskeyd, line 53:
```suggestion
    /// The future to execute under a timeout.
```

crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+}
+
+#[pin_project::pin_project(PinnedDrop)]
+pub struct CancelLoggerFuture<'a, F> {
+    // Future state.
+    completed: bool,
+
+    // Method information.
```
Freyskeyd, line 49:
```suggestion
    /// Method information.
```

crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
```diff
@@ -0,0 +1,83 @@
...
+    method: Cow<'a, str>,
+    request_id: Id<'a>,
```
Freyskeyd, line 51:
```suggestion
    /// Unique ID for the request.
    request_id: Id<'a>,
```

crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
```diff
@@ -0,0 +1,83 @@
...
+            method: request.method.clone(),
+            request_id: request.id.clone(),
+            inner: self.0.call(request),
+        }
+    }
+}
+
+#[pin_project::pin_project(PinnedDrop)]
+pub struct CancelLoggerFuture<'a, F> {
+    // Future state.
```
Freyskeyd, line 46:
```suggestion
    /// Future completion state.
```

crates/agglayer-node/src/rpc/rpc_middleware/cancel_logger.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+#[pin_project::pinned_drop]
+impl<F> PinnedDrop for CancelLoggerFuture<'_, F> {
+    fn drop(self: std::pin::Pin<&mut Self>) {
+        if !self.completed {
+            let method = &*self.method;
+            let id = &self.request_id;
+            tracing::warn!("Request ID `{id}` to `{method}` was cancelled");
```
Freyskeyd, line 80: Maybe we can just put that in `info` for now, until we see how it behaves in production

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+            method: request.method.clone(),
+            request_id: request.id.clone(),
+            inner: self.inner.call(request),
+        }
+    }
+}
+
+#[pin_project::pin_project]
+pub struct LoggingTimeoutFuture<'a, F> {
+    // Timeout duration.
```
Freyskeyd, line 63:
```suggestion
    /// Timeout duration.
```

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+    timeout: Duration,
+
+    // Method information.
```
Freyskeyd, line 66:
```suggestion
    /// Method information.
```

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+use tracing::warn;
+
+/// Error code to return when the response time is too long.
+pub const TIMEOUT_ERROR_CODE: i32 = -32001;
+
+/// A layer that applies a timeout on a request and issues a log entry if the
+/// timeout expires before the request is completed.
+#[derive(Clone, Debug)]
+pub struct LoggingTimeoutLayer {
+    timeout: Duration,
```
Freyskeyd, line 20:
```suggestion
    /// Maxmum duration for the request to complete.
    timeout: Duration,
```

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+impl<S> tower::Layer<S> for LoggingTimeoutLayer {
+    type Service = LoggingTimeoutService<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        LoggingTimeoutService::new(self.timeout, inner)
+    }
+}
+
+pub struct LoggingTimeoutService<S> {
+    inner: S,
```
Freyskeyd, line 38:
```suggestion
    /// The underlying service.
    inner: S,
```

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+    timeout: Duration,
```
Freyskeyd, line 39:
```suggestion
    /// Maxmum duration for the request to complete.
    timeout: Duration,
```

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[pin_project::pin_project]
+pub struct LoggingTimeoutFuture<'a, F> {
+    // Timeout duration.
+    timeout: Duration,
+
+    // Method information.
+    method: Cow<'a, str>,
+    request_id: Id<'a>,
```
Freyskeyd, line 68:
```suggestion
     /// Unique ID for the request.
    request_id: Id<'a>,
```

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    // The future to execute under a timeout.
```
Freyskeyd, line 70:
```suggestion
    /// The future to execute under a timeout.
```

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
```diff
@@ -0,0 +1,97 @@
...
+    type Output = MethodResponse;
+
+    fn poll(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        let this = self.project();
+        let timeout = *this.timeout;
+
+        let fut = tokio::time::timeout(timeout, this.inner).unwrap_or_else(move |e| {
```
Freyskeyd, line 85: Does the future for the timeout isn't recreated at each poll?
Shouldn't we store the `timeout_fut` as a `[pin]` future inside `LoggingTimeoutFuture` ?

iljakuklic: > Does the future for the timeout isn't recreated at each poll? Shouldn't we store the `timeout_fut` as a `[pin]` future inside `LoggingTimeoutFuture` ?

Good call, I think you're correct.

crates/agglayer-node/src/rpc/rpc_middleware/logging_timeout.rs
iljakuklic: Will package the two fields into a helper struct. It's the same between the two middlewares.

iljakuklic:
Finally managed to find a sensible way to test these middlewares.

Freyskeyd (approved):

PR #284, opened 2024-09-06
feat: create pp inputs from certificate and state

### LINKED ISSUE #265: Extend the API endpoint `send_certificate` for CDK
- Extend the API endpoint `send_certificate` for the CDK with the latest structure
- Create one multi batch header out of one certificate

hadjiszs:
# Description

- Adapt the `send_certificate` endpoint to take one `Certificate` instead of a `MultiBatchHeader`
- Complete the multi batch header creation from certificate and current state
- Add structures for distinction full state <> roots
- Adapt the native/prover flow within the orchestrator

Fixes #265 

## PR Checklist:


crates/pessimistic-proof/src/certificate.rs
```diff
@@ -1,41 +1,149 @@
...
     }
+
+    pub fn hash(&self) -> Digest {
+        keccak256_combine([
+            self.network_id.to_be_bytes().as_slice(),
+            self.height.to_be_bytes().as_slice(),
+            self.prev_local_exit_root.as_slice(),
+            self.new_local_exit_root.as_slice(),
+        ])
+    }
```
iljakuklic, line 73: It may be useful to add a comment up where the `Certificate` struct is defined that anyone who adds or modifies the fields there should think whether `Certificate::hash` also needs updating. Seems it could be easily forgotten.

Even better, factor out the parts that are committed to into a separate struct and define the hash as just the hash of a serialization of the inner struct. That way, adding a field to the new struct is automatically picked up when hashing. I am not sure about possible drawbacks of this approach in terms of code clarity and so on but may be worth considering.

crates/pessimistic-proof/src/certificate.rs
```diff
@@ -1,41 +1,149 @@
...
+            signer,
+            signature: self.signature,
+            imported_rollup_exit_root: [0; 32],
+            imported_mainnet_exit_root: [0; 32],
+            imported_exits_root: None,
+            target: StateCommitment {
+                exit_root: self.new_local_exit_root,
+                balance_root: Default::default(), // computed by agglayer
+                nullifier_root: Default::default(), // computed by agglayer
+            },
```
iljakuklic, line 146: Not necessarily for this PR, just a general comment. I think it would be cleaner to introduce a separate type for PP input rather than recycling `MultiBatchHeader` and filling it with dummy fields. Common parts could be factored out.

crates/pessimistic-proof/src/certificate.rs
```diff
@@ -1,41 +1,149 @@
...
+                        .clone()
+                        .get_inclusion_proof_zero(token)
+                        .unwrap()
+                } else {
+                    state.balance_tree.get_inclusion_proof(token).unwrap()
+                };
+                res.insert(token, (initial_balance, path));
+            });
+            res
+        };
```
iljakuklic, line 111: Minor: It seems to me it should be possible to use `mutated_tokens.map(|token| ...).collect()` here in place of `for_each`/`insert`.

crates/pessimistic-proof/src/certificate.rs
```diff
@@ -1,41 +1,149 @@
...
+                    // TODO: dont clone once get_inclusion_proof_zero doesnt mutate anymore
+                    state
+                        .balance_tree
+                        .clone()
+                        .get_inclusion_proof_zero(token)
+                        .unwrap()
+                } else {
+                    state.balance_tree.get_inclusion_proof(token).unwrap()
+                };
+                res.insert(token, (initial_balance, path));
```
iljakuklic, line 108: What if a token is present in both `imported_tokens` and `exported_tokens`? Seems the `exported_tokens` path would overwrite whatever was initially inserted by the `imported_tokens` entry. Is that the intended behaviour?

hadjiszs: Yeah indeed :+1: fixed by accumulating in a set

crates/pessimistic-proof/src/certificate.rs
```diff
@@ -1,41 +1,149 @@
...
 }
 
 impl Certificate {
-    /// Creates a new [`Certificate`].
-    pub fn new(
-        origin_network: NetworkId,
-        prev_local_exit_root: Digest,
-        bridge_exits: Vec<BridgeExit>,
-    ) -> Self {
+    pub fn new_for_test(network_id: NetworkId, height: Height) -> Self {
```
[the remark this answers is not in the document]
hadjiszs: In the end I just removed it, given that it is not used in this PR

crates/pessimistic-proof/src/certificate.rs
hadjiszs: > I am not sure about possible drawbacks of this approach in terms of code clarity and so on but may be worth considering.

It touches the most commonly accessed fields of this struct, so having some kind of `certificate.inner.foo` whenever we access them might not be so great no?

The access could be behind methods like `network_id()` `prev_local_exit_root()` etc but not sure either how it would impact the code clarity on some area

In any case, I added a note in the `Certificate` rustdoc

PR #297, opened 2024-09-09
fix: proper encoding for the pp public inputs

hadjiszs:
# Description

- Adapt the public inputs encoding for the pp proof
- Misc changes on `ppgen` such as outputting the signer and log e2e latency

## PR Checklist:


crates/pessimistic-proof-program/src/main.rs
```diff
@@ -1,16 +1,23 @@
...
     let initial_state = sp1_zkvm::io::read::<LocalNetworkState>();
     let batch_header = sp1_zkvm::io::read::<MultiBatchHeader<Keccak256Hasher>>();
 
     let outputs = generate_pessimistic_proof(initial_state, &batch_header).unwrap();
 
-    sp1_zkvm::io::commit(&outputs);
+    let pp_inputs = bincode::DefaultOptions::new()
+        .with_big_endian()
+        .with_fixint_encoding()
+        .allow_trailing_bytes()
```
Freyskeyd, line 18: Why do we need this one?

Freyskeyd (approved):

crates/pessimistic-proof-test-suite/src/runner.rs
```diff
@@ -41,8 +42,13 @@ impl Runner {
     }
 
     /// Extract outputs from the committed public values.
-    pub fn extract_output(mut public_vals: SP1PublicValues) -> PessimisticProofOutput {
-        public_vals.read::<PessimisticProofOutput>()
+    pub fn extract_output(public_vals: SP1PublicValues) -> PessimisticProofOutput {
+        bincode::DefaultOptions::new()
+            .with_big_endian()
+            .with_fixint_encoding()
+            .allow_trailing_bytes()
```
iljakuklic, line 49: Ideally these options should be taken out into a helper function so they can be easily shared between the encoding and decoding side for consistency.

crates/pessimistic-proof-program/src/main.rs
hadjiszs: The motivation is to avoid trimmed prefixed zeroes, but that's already guaranteed by the `with_fixint_encoding` so I removed the `allow_trailing_bytes`

PR #300, opened 2024-09-11
feat: hash global index for imported bridge exits

hadjiszs:
# Description

Simple change for the hash of the imported bridge exit

- before: hash of the underlying bridge exit
- after: hash of the global index

Rationale being to have less fields to hash

## PR Checklist:


crates/pessimistic-proof/src/global_index.rs
```diff
@@ -46,8 +46,7 @@ impl From<U256> for GlobalIndex {
 }
 
 impl GlobalIndex {
-    #[allow(unused)]
-    fn to_u256(self) -> U256 {
+    pub fn to_u256(self) -> U256 {
```
Freyskeyd, line 49: It wasn't possible to have an `Into` implementation?

Freyskeyd (approved):

crates/pessimistic-proof/src/global_index.rs
hadjiszs: yeah I moved it to From/Into implementation

PR #303, opened 2024-09-11
feat: rate limiting for `sendTx`

### LINKED ISSUE #291: Implement rate limiting for `sendTx`
The `sendTx` transaction settlement should be rate limited according to configuration.

Related:
* General rate limiting
  * #219 
  * #213 
* Progress PRs / partial tasks
  * #222 
* Other possibly relevant but not directly related to `sendTx`
  * #271

iljakuklic:
# Description

Initial rate limiting implementation. This is a draft because tests are missing. Feel free to comment.

## Related issues
* fixes #291 
* spec #213, the implementation from this PR differs slightly.

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/src/rate_limiting.rs
+++ b/crates/agglayer-config/src/rate_limiting.rs
@@ -8,8 +8,11 @@ pub type RollupId = u32;
 #[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(rename_all = "kebab-case")]
 pub enum TimeRateLimit {
+    /// Don't apply any rate limiting, allowing the client to make requests as
+    /// often as desired.
     Unlimited,
 
+    /// Apply rate limit of `max_per_interval` events in given `time_interval`.
     #[serde(untagged, rename_all = "kebab-case")]
     Limited {
         max_per_interval: u32,
@@ -53,12 +56,11 @@ impl PerNetworkRateLimitOverride {
 impl TryFrom<BTreeMap<String, RateLimitOverride>> for PerNetworkRateLimitOverride {
     type Error = <RollupId as std::str::FromStr>::Err;
 
-    fn try_from(value: BTreeMap<String, RateLimitOverride>) -> Result<Self, Self::Error> {
-        let overrides = value
+    fn try_from(overrides: BTreeMap<String, RateLimitOverride>) -> Result<Self, Self::Error> {
+        overrides
             .into_iter()
             .map(|(k, v)| Ok((k.parse::<RollupId>()?, v)))
-            .collect::<Result<Self, Self::Error>>()?;
-        Ok(overrides)
+            .collect::<Result<Self, Self::Error>>()
     }
 }
 
--- a/crates/agglayer-config/src/rate_limiting.rs
+++ b/crates/agglayer-config/src/rate_limiting.rs
@@ -94,6 +94,15 @@ impl RateLimitingConfig {
         }
     }
 
+    /// Override `sendTx`setting for given network
+    pub fn with_send_tx_override(mut self, nid: RollupId, limit: TimeRateLimit) -> Self {
+        let limit_override = RateLimitOverride {
+            send_tx: Some(limit),
+        };
+        let _ = self.network.0.insert(nid, limit_override);
+        self
+    }
+
     /// Get rate limiting for the `sendTx` call for given network.
     pub fn send_tx_limit(&self, nid: RollupId) -> &TimeRateLimit {
         self.override_for(nid)
--- a/crates/agglayer-config/src/rate_limiting.rs
+++ b/crates/agglayer-config/src/rate_limiting.rs
@@ -1,0 +1,187 @@
+use std::{collections::BTreeMap, time::Duration};
+
+use serde::Deserialize;
+
+pub type RollupId = u32;
+
+/// Time-based rate limiting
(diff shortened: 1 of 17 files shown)


crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -139,6 +139,12 @@ where
             return Err(Error::rollup_not_registered(tx.tx.rollup_id));
         }
 
+        // Check rate limiting early so that we can report it before entering
+        // the subsequent expensive checks.
+        self.kernel
+            .rate_limiter()
+            .check_send_tx_now(tx.tx.rollup_id)?;
```
Freyskeyd, line 146: Does it mean that if 10 tx are proceed at the same time, all of them are proceed for the sanity check but only one are settled and the 9 other got an error ?
Shouldn't it be better to discard the call directly?

iljakuklic: The question is whether the participating network should be rate limited even if the checks error out and no settlement actually happens. I did not want to apply the settlement rate limit unless there are good chances settlement happens. If that's not the desired behaviour, let me know.

As a side note, I also experimented with a more accurate approach where a rate limiting slot would be "reserved" when the call is first made and then recorded on settlement or released on error. However, it ended up being quite complex so I opted for two checks even though it may lead to spurious errors as you describe.

crates/agglayer-node/src/rate_limiting/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// Check rate limiting for `sendTx` and record the request with given
+    /// request timestamp.
+    pub fn limit_send_tx(&self, nid: RollupId, now: Instant) -> Result<(), Error> {
+        self.lock().limit_send_tx(nid, now)
+    }
+
+    fn lock(&self) -> MutexGuard<RateLimiterImpl> {
+        self.0.lock().expect("mutex poisoned")
```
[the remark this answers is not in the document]
iljakuklic: > Can we use `parking_lot` here?

I guess. Do you suggest it for performance or are there other reasons as well?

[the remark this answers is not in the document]
Freyskeyd: Better performance and cannot poison :)

crates/agglayer-node/src/rpc/mod.rs
iljakuklic: > As we scope the rate limiting per network, we could move the `check_signature` out from the `try_join`, in order to execute it first and then, if it passes, we're sure that that's the right network that submitted this `SignedTx`, thus we can directly apply the rate limiting, as having multiple submission isn't allowed.

I moved the rate limiting check after signature check. However, there is still the question whether a transaction submission that fails the checks should still be counted towards the rate limit even though no settlement actually occurs in that case. There are two checks exactly to avoid that scenario. The first check only checks whether the rate limiter is "full", the other one just before settlement actually records the event in the rate limiter.

Freyskeyd: Indeed, maybe we need to think about that as slot per network?

crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -343,6 +352,9 @@ where
             warn!(hash, "Transaction already settled: {:?}", tx);
         }
 
+        self.rate_limiter
+            .limit_send_tx_now(signed_tx.tx.rollup_id)?;
```
Freyskeyd, line 356: If the transaction fails, the rate limit is still applied right?

iljakuklic: > If the transaction fails, the rate limit is still applied right?

Yea, if the tx submission to L1 fails, it still contributes to the rate limit in this case. It would be more accurate to use the actual L1 state to determine how many transaction settlements actually happened. That would probably be a fair bit more complicated. Especially considering the various delays and weird "in-between" states involved. For example, should a settlement transaction in mempool contribute towards rate limit? At that point, whether it is going to be confirmed or not is outside of our control to a large extent.

crates/agglayer-node/src/rpc/mod.rs
iljakuklic: > Indeed, maybe we need to think about that as slot per network?

I am not sure what you mean? The settlement limit is applied on the per-network basis.

crates/agglayer-config/src/rate_limiting.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /// Override `sendTx`setting for given network
+    pub fn with_send_tx_override(mut self, nid: RollupId, limit: TimeRateLimit) -> Self {
+        let limit_override = RateLimitOverride {
+            send_tx: Some(limit),
+        };
+        let _ = self.network.0.insert(nid, limit_override);
+        self
+    }
```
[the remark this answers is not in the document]
Freyskeyd: That's somehow similar but represent different things. I would say that we can use `network_id` type everywhere

crates/agglayer-node/src/rpc/mod.rs
Freyskeyd: Rethinking about this, another solution would be to "queue" the queries by fetching an async `lockguard` per network, it would still allow the network to send queries, but only one would be settled.

iljakuklic:
Closing in favour of #316

PR #308, opened 2024-09-18
refactor: generalize rate limiter

iljakuklic:
This is to prepare the rate limiter to be further extended to be applied to `sendCertificate` settlement and other RPC calls / event types.

## PR Checklist:


crates/agglayer-node/src/rate_limiting/local/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+
+pub use interface::RateLimiter;
+
+pub struct LocalRateLimiter {
+    /// Rate limiter for `sendTx` settlement.
+    send_tx: trivial::RateLimiter<wall_clock::RateLimiter>,
+}
+
+impl LocalRateLimiter {
+    pub fn from_config(send_tx_config: &TimeRateLimit) -> Self {
```
Freyskeyd, line 19: If we have more RPC added to this struct, it means that the arguments of this method will change right?
Could we use the base `config` and build the proper argument inside the function?
If we need to, we can have this method returning a `Result`.

crates/agglayer-node/src/rate_limiting/local/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+// TODO: make private if possible
+mod interface;
+pub mod trivial;
+pub mod wall_clock;
+
+pub use interface::RateLimiter;
+
+pub struct LocalRateLimiter {
+    /// Rate limiter for `sendTx` settlement.
+    send_tx: trivial::RateLimiter<wall_clock::RateLimiter>,
```
Freyskeyd, line 15: I'm wondering if we could use a custom struct such as `SendTx {}` and `SendCertificate`, and have a dynamic hashmap where we use a `TypeId` on that struct. So that we can call `rate_limiter.check::<SendTx>(Instant::now())`. Maybe it's a bad idea, or not even possible, I didn't tested it.

crates/agglayer-node/src/rate_limiting/local/trivial.rs
```diff
@@ -0,0 +1,50 @@
+/// Rate limiter handling edge cases.
+///
+/// This rate limiter handles edge cases, namely if no rate limit is imposed or
+/// if rate limit is zero. Otherwise, it delegates to the inner limiter.
+pub enum RateLimiter<L> {
+    Disabled,
+    Limited(L),
+    Unlimited,
```
Freyskeyd, line 8: Could we have some explanation of the different variants ?
Such as `Disabled` mean that its an error and that the RPC is disabled?

crates/agglayer-node/src/rate_limiting/local/trivial.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    fn limit(&mut self, time: Self::Instant) -> Result<(), Self::RateLimited> {
+        self.with_inner(|inner| inner.limit(time))
+    }
+
+    fn is_clear(&mut self, time: Self::Instant) -> bool {
+        match self {
+            Self::Disabled => true,
+            Self::Limited(inner) => inner.is_clear(time),
+            Self::Unlimited => true,
```
Freyskeyd, line 41:
```suggestion
            Self::Limited(inner) => inner.is_clear(time),
            Self::Disabled | Self::Unlimited => true,
```

Freyskeyd (commented):
Nice, just dropping some comments, but overall it is good for me !

crates/agglayer-node/src/rate_limiting/local/mod.rs
iljakuklic: > I'm wondering if we could use a custom struct such as `SendTx {}` and `SendCertificate`, and have a dynamic hashmap where we use a `TypeId` on that struct. So that we can call `rate_limiter.check::<SendTx>(Instant::now())`. Maybe it's a bad idea, or not even possible, I didn't tested it.

I think it is possible but in my opinion not worth all the trait magic needed to accomplish this. Also the implementation for the epoch-based rate limiting will have a different (more efficient) internal representation since it does not need to track the list of events in given window, just the number of events in the current epoch.

Freyskeyd: Yeah, I was wondering how the api of the LocalRateLimiter would look like if we had 10 or 20 RPC, but we have time to think about that

iljakuklic:
Closing in favour of #316

PR #316, opened 2024-09-24
feat: Rate limiting for `sendTx` with slot reservation

### LINKED ISSUE #291: Implement rate limiting for `sendTx`

iljakuklic:
# Description

A new PR for `sendTx` transaction limiting with some parts substantially rewritten to address various annoying corner cases. When a request is issued, a slot is reserved in the rate limiter. If settlement ends up being successful, the event is recorded into the rate limiter. If settlement does not happen in the end for some reason, the slot is freed so the client can attempt again.

The [`rate_limiting/local/mod.rs`](https://github.com/agglayer/agglayer/blob/feat/rate-limit-reserve/crates/agglayer-node/src/rate_limiting/local/mod.rs) module contains a top-level comment with the description of layers the implementation consists of. It may serve as a useful guide to suggest a sensible review order.

* fixes #291
* discussion #213
* obsoletes #303 
* obsoletes #308
* `main` port #318

## PR Checklist:


crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -136,22 +136,26 @@ where
...
+            Error::signature_mismatch(e)
+        })?;
+
+        agglayer_telemetry::VERIFY_SIGNATURE.add(1, metrics_attrs);
+
+        // Reserve a rate limiting slot.
+        let guard = self
+            .kernel
+            .rate_limiter()
+            .reserve_send_tx(tx.tx.rollup_id, tokio::time::Instant::now())?;
```
iljakuklic, line 153: Just to make sure, is the `tx.tx.rollup_id` here the same ID we want to be specified in the config file? I vaguely remember there being distinction between `network_id` and `rollup_id`.

Freyskeyd: `network_id` and `rollup_id` are the same things unless `network_id = 0` which is L1
And yes this is the same ID as in the config, but I would like to not rely on the config for defining the "participant" on the agglayer ^^

iljakuklic:
This is on top of 0.1.7 branch. I guess we can just close this now.

PR #318, opened 2024-09-25
chore: port `sendTx` rate limiting to `main`

iljakuklic:
# Description

Same as #316 but ported from `0.1.7` to `main` with some very minor modifications.

* port of #316

## PR Checklist:


crates/agglayer-node/src/rate_limiting/local/inner/component.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// Get the rate limiter for this component.
+    fn component(limiter: &mut LocalRateLimiter) -> &mut RateLimiter<Self::State>;
+
+    /// Convert the component-specific error into the rate limited error.
+    fn error(info: limiter::RateLimited<Self::LimitedInfo>) -> RateLimited;
+}
+
+/// The rate limiter component for `sendTx` settlement.
+pub enum SendTx {}
```
Freyskeyd, line 32: Why is it an `enum`?

crates/agglayer-node/src/rate_limiting/local/limiter/core.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn new(raw_state: S) -> Self {
+        let state = State::new(raw_state);
+        let reserved = 0;
+        Self { state, reserved }
+    }
+
+    /// Query the total occupancy, including the reserved slots.
+    pub fn query(&mut self, time: S::Instant) -> usize {
+        let num = self.state.query(time) + self.reserved;
+        assert!(num <= self.state.max_events());
```
Freyskeyd, line 31: Should'nt it be `debug_assert` here?
What's the rational behind allowing panics here?

Freyskeyd (approved):
Nice PR ! Just some question related to the `assert` usage and the different `panic`, what is your idea behind that?

crates/agglayer-node/src/rate_limiting/local/inner/component.rs
iljakuklic: > Why is it an `enum`?

This is used purely as a phantom type tag. Since it makes no sense to construct an object of this type, I opted for something that cannot be constructed to begin with (there are no enum arms).

crates/agglayer-node/src/rate_limiting/local/limiter/core.rs
iljakuklic: > Should'nt it be `debug_assert` here? What's the rational behind allowing panics here?

I don't have a strong opinion on the use of `debug_assert` vs. plain `assert`. Here, it checks an invariant maintained by a higher-level interface, so it will not happen if the external interface is used (unless the implementation is broken).

iljakuklic: Changed to `log_assert`, which asserts in debug and logs an error in release.

PR #320, opened 2024-09-26
build: updating github build

Freyskeyd:
# Description

This PR adds `cargo-make` and update some github workflows.

## PR Checklist:

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -12,7 +12,7 @@ on:
       - "**"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 env:
@@ -52,6 +52,8 @@ jobs:
 
   lint:
     name: Lint
+    # Disabled due to: https://github.com/rust-lang/rust-clippy/issues/13458
+    if: false
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
@@ -66,12 +68,15 @@ jobs:
       - name: Set up rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          cache-on-failure: true
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - uses: taiki-e/install-action@protoc
 
-      - name: Run cargo clippy
-        run: cargo clippy --all-targets -- -D warnings
+      - name: Install cargo-make
+        run: cargo install --debug --locked cargo-make
+
+      - name: Run clippy
+        run: cargo make ci-clippy
 
   fmt:
     name: Format
@@ -89,7 +94,10 @@ jobs:
       - name: Set up rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          cache-on-failure: true
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
-      - name: Run cargo fmt
-        run: cargo fmt --all --check
+      - name: Install cargo-make
+        run: cargo install --debug --locked cargo-make
+
+      - name: Run format
+        run: cargo make ci-format
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Test
+name: Test - units
 
 on:
   push:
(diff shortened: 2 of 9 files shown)


.github/workflows/quality.yml
```diff
@@ -66,12 +66,15 @@ jobs:
       - name: Set up rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          cache-on-failure: true
+          save-if: ${{ github.ref == 'refs/heads/main' }}
```
iljakuklic, line 69: Would it make sense to also include the `release/*` branches?

Makefile.ci.toml
```diff
@@ -0,0 +1,27 @@
...
+    "--all-targets",
+    "--all-features",
+    "--tests",
+    "--benches",
+    "--examples",
+    "--bins",
+    "--",
+    "-D",
+    "warnings",
+]
```
iljakuklic, line 27: It may be worth (not necessarily now) to go through [clippy checks that are disabled by default](https://rust-lang.github.io/rust-clippy/master/index.html#?levels=allow,none,warn) and enable some of them.

This may introduce a higher rate of false positives. However, it may still be useful to have them flagged up and have the dev `#[allow()]` them in places selectively to acknowledge the issue has been considered and it's deemed to break the lint rule in this particular instance.

It is also possible to run clippy twice. Once with a more permissive setting and apply it to all the code base, and once with stricter settings and apply it only to the main code base (excluding tests, benches, etc. where a more lax approach is often acceptable).

Makefile.ci.toml
```diff
@@ -0,0 +1,27 @@
...
+command = "cargo"
+args = [
+    "clippy",
+    "--all-targets",
+    "--all-features",
+    "--tests",
+    "--benches",
+    "--examples",
+    "--bins",
+    "--",
```
iljakuklic, line 24: I don't see the `--workspace` switch. Which crates does the command apply to in this case? Wondering if this covers everything we want.

Even with the `--workspace` switch, I don't think it would run on the pessimistic proof crates as these technically are under a separate workspace. Or am I mistaken here and it processes all the local dependencies anyway?

leovct (approved):
lgtm

.github/workflows/test.yml
```diff
@@ -32,201 +30,23 @@ jobs:
...
-          docker image ls -a
-
-      - name: Publishing image
-        env:
-          DOCKER_IMAGE: ${{ needs.pre-e2e.outputs.tags }}
-        run: docker push ${{ env.DOCKER_IMAGE }}
-
-
+      - name: Test
+        run: cargo nextest run --workspace --nocapture && cargo test --doc --workspace
```
iljakuklic, line 52: The `--nocapture` (which is an alias to `--no-capture`) forces the tests to run serially. Is that the intention?

I'd say using Nextest's `--success-output=immediate` and `--failure-output=immediate-final` is preferable here.

[Relevant docs here](https://nexte.st/docs/reporting/#standard-output-and-standard-error)

iljakuklic (commented):
Adding some comments. I only skimmed through the kurtosis stuff as I don't understand it well enough.

.github/workflows/quality.yml
Freyskeyd: The goal is to only cache what is linked to the `main` lock file hash. The release branches aim at producing a docker image which doesn't use the `rust-cache` tho.
They can be some improvements here but I didn't pushed too far.

Makefile.ci.toml
Freyskeyd: Yes, I agree that we should customize the clippy config. I didn't want to have a huge diff on this PR, but I agree that we can introduce more linter if they're acknowledged and relevant.

.github/workflows/test:e2e.yml
```diff
@@ -0,0 +1,125 @@
...
+
+      - name: Check that batches are progressing
+        continue-on-error: true
+        env:
+          ETH_RPC_URL: ${{ env.ETH_RPC_URL }}
+          TIMEOUT_SECONDS: 900
+        run: ./.github/scripts/monitor-verified-batches.sh --rpc-url "$ETH_RPC_URL"
+
+      - name: debug
+        run: kurtosis service logs cdk-v1 zkevm-agglayer-001
```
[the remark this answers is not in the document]
leovct: It won't work because we're still using kurtosis-cdk `v0.2.3` in CI.

.github/workflows/test:e2e.yml
```diff
@@ -0,0 +1,125 @@
...
+          ref: ${{ env.KURTOSIS_CDK_GIT_REF }}
+
+      - name: Replace param with local agglayer-rs docker img
+        run: |
+          new_value="${{ needs.pre-e2e.outputs.tags }}"
+
+          # Escape special characters in the new value
+          escaped_new_value=$(sed 's/[\/&]/\\&/g' <<< "$new_value")
+
+          sed -i "s/zkevm_agglayer_image: .*/zkevm_agglayer_image: $escaped_new_value/" "params.yml"
```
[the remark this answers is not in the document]
Freyskeyd: The CI is using `0.2.3` currently, so it is `zkevm_agglayer_image`

hadjiszs (approved):

PR #329, opened 2024-10-01
feat: adding getCertificateHeader RPC

Freyskeyd:
# Description

This PR adds a new RPC that provide access to the `CertificateHeader`.

The goal is to provide a way to monitor and check the status of a particular `CertificateId`.

The definition is:

```
path: getCertificateHeader (interop_getCertificateHeader)
params: [ CertificateId ]
response: CertificateHeader
```

The `CertificateHeader` contains multiple things:

- `epoch_number` which is null until the certificate is settled in an epoch
- `certificate_index` which is null until the certificate is settled in an epoch, it represents the index of the certificate in the epoch
- `status` which represent the status of the certificate, it can be `Pending`, `InError` or `Settled`

An example of a `CertificateHeader` of a failed certificate with the reason:

```json
{
  "network_id": 1,
  "height": 0,
  "epoch_number": null,
  "certificate_index": null,
  "certificate_id": "0x548faaa7bd84fdd200e56791f0d88f97bc11f897b1bdc1f75704d9b99610b9e8",
  "new_local_exit_root": "0x030b33b7ce8e655ee40bf325c31d7f0ec29854f73c73f9e054bd820c79ea726e",
  "status": {
    "InError": {
      "error": {
        "ProofGenerationError": {
          "generation_type": "Native",
          "source": {
            "InvalidInitialLocalExitRoot": {
              "got": "0x27ae5ba08d7291c96c8cbddcc148bf48a6d68c7974b94356f53754ef6171d757",
              "expected": "0xf99fbc86af88be1a031b1d3aa12352bbc35c660f84f127100d98c722980dd5d7"
            }
          }
        }
      }
    }
  }
}
```

## Additions and Changes

- `sendCertificate` RPC now returns a `CertificateId`

## PR Checklist:

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/sp1/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/sp1/mod.rs
@@ -60,6 +60,8 @@ where
                     )
                     .map_err(|error| {
                         let error_description = error.to_string();
+                        // TODO: Find a better solution than downcasting the error into a
+                        // ProofError.
                         if let Ok(error) = error.downcast::<ProofError>() {
                             error
                         } else {
--- a/crates/agglayer-aggregator-notifier/src/sp1/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/sp1/mod.rs
@@ -1,13 +1,15 @@
 use std::sync::Arc;
 
+use agglayer_types::ProofVerificationError;
 use futures::{future::BoxFuture, FutureExt as _};
-use pessimistic_proof::LocalNetworkState;
+use pessimistic_proof::{LocalNetworkState, ProofError};
 use serde::Serialize;
 use sp1_prover::components::DefaultProverComponents;
 use sp1_sdk::{
     CpuProver, MockProver, NetworkProver, Prover as _, SP1ProvingKey, SP1Stdin, SP1VerifyingKey,
 };
 use tokio::task::spawn_blocking;
+use tracing::error;
 
 use super::Proof;
 
@@ -38,7 +40,7 @@ where
         &self,
         initial_state: LocalNetworkState,
         certificate: I,
-    ) -> BoxFuture<'_, Result<Proof, anyhow::Error>> {
+    ) -> BoxFuture<'_, Result<Proof, ProofError>> {
         let mut stdin = SP1Stdin::new();
         stdin.write(&initial_state);
         stdin.write(&certificate);
@@ -48,21 +50,33 @@ where
 
         async move {
             spawn_blocking(move || {
-                prover.prove(
-                    &proving_key,
-                    stdin,
-                    Default::default(),
-                    Default::default(),
-                    Default::default(),
-                )
+                prover
+                    .prove(
+                        &proving_key,
+                        stdin,
+                        Default::default(),
+                        Default::default(),
+                        Default::default(),
+                    )
+                    .map_err(|error| {
+                        let error_description = error.to_string();
+                        if let Ok(error) = error.downcast::<ProofError>() {
+                            error
+                        } else {
+                            error!("Error while proving: {}", error_description);
+
+                            ProofError::Unknown(error_description)
+                        }
+                    })
             })
-            .await?
+            .await
+            .map_err(|_| ProofError::Unknown("Unable to properly execute Proof task".to_string()))?
             .map(Proof::SP1)
         }
         .boxed()
     }
 
-    fn verify(&self, proof: &Proof) -> Result<(), anyhow::Error> {
+    fn verify(&self, proof: &Proof) -> Result<(), ProofVerificationError> {
         let Proof::SP1(proof) = proof;
 
         Ok(self.prover.verify(proof, &self.verifying_key)?)
@@ -77,7 +91,7 @@ where
         &self,
         initial_state: LocalNetworkState,
         certificate: I,
-    ) -> BoxFuture<'_, Result<Proof, anyhow::Error>> {
+    ) -> BoxFuture<'_, Result<Proof, ProofError>> {
         let mut stdin = SP1Stdin::new();
         stdin.write(&initial_state);
         stdin.write(&certificate);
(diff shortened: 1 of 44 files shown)


crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -258,12 +271,26 @@ where
...
-        _ = self.pending_store.insert_pending_certificate(
-            certificate.network_id,
-            certificate.height,
-            &certificate,
-        );
+    async fn send_certificate(&self, certificate: Certificate) -> RpcResult<CertificateId> {
+        let id = certificate.hash();
+        // TODO: Batch the different queries.
+        // Insert the certificate header into the state store.
+        _ = self
```
iljakuklic, line 278: Have not seen the `_ = foo();` construct. How does it work? I always used `let _ = foo();`.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -30,21 +33,54 @@ impl StateStore {
...
                 epoch_number: None,
                 certificate_index: None,
-                new_local_exit_root: certificate.new_local_exit_root,
+                new_local_exit_root: certificate.new_local_exit_root.into(),
+                status: status.clone(),
             },
-        )
+        )?;
+
+        if let CertificateStatus::Settled = status {
```
hadjiszs, line 55: in this case the fields `epoch_number` and `certificate_index` are not written? or will it be updated afterwards?

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -573,7 +586,68 @@ where
...
+                        .update_certificate_header_status(
+                            &certificate_id,
+                            &CertificateStatus::InError { error },
+                        )
+                        .is_err()
+                    {
+                        error!(
+                            "Certificate in error and failed to update the certificate header \
+                             status"
+                        );
```
hadjiszs, line 647: maybe also logging both errors: the `CertificateStatusError` and the error raised while attempting to update the certificate status

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -573,7 +586,68 @@ where
...
+                };
+
+                if let Some((certificate_id, error)) = certificate_error {
+                    if self
+                        .state_store
+                        .update_certificate_header_status(
+                            &certificate_id,
+                            &CertificateStatus::InError { error },
+                        )
+                        .is_err()
```
iljakuklic, line 642: Are the error-ed out certificate headers cleaned up from storage at some point or do we keep them forever? Seems like a possible DoS vector, at least until some sort of rate limiting is in place.

crates/agglayer-node/src/rpc/tests/mod.rs
```diff
@@ -259,6 +285,73 @@ async fn check_tx_status_fail() {
     }
 }
 
+#[tokio::test]
+async fn get_certificate_header_after_sending_the_certif() {
```
hadjiszs, line 289:
```suggestion
async fn get_certificate_header_after_sending_certificate() {
```

crates/agglayer-aggregator-notifier/src/sp1/mod.rs
```diff
@@ -48,21 +50,33 @@ where
...
+                    .prove(
+                        &proving_key,
+                        stdin,
+                        Default::default(),
+                        Default::default(),
+                        Default::default(),
+                    )
+                    .map_err(|error| {
+                        let error_description = error.to_string();
+                        if let Ok(error) = error.downcast::<ProofError>() {
```
iljakuklic, line 63: Maybe the `AggregatorProver` should specify the error state as an associated type or something like that to avoid `downcast` here. Seems we are trying to recover some type information at runtime even though it could be propagated at compilation time.

crates/agglayer-node/src/rpc/error.rs
```diff
@@ -25,6 +25,9 @@ pub mod code {
 
     /// Error submitting a ceritficate.
     pub const SEND_CERTIFICATE: i32 = -10006;
+
+    /// Resource not found.
+    pub const RESOURCE_NOT_FOUND: i32 = -10007;
```
hadjiszs, line 30:
```suggestion
    /// Error submitting a certificate.
    pub const SEND_CERTIFICATE: i32 = -10006;

    /// Resource not found.
    pub const RESOURCE_NOT_FOUND: i32 = -10007;
```

hadjiszs (commented):
:+1: I'll review again after rebase

crates/agglayer-storage/src/stores/state/mod.rs
Freyskeyd: Will be done when we tackle the settlement on L1

crates/agglayer-certificate-orchestrator/src/lib.rs
Freyskeyd: The error is already logged above the match, currently in `warn` as I don't really know if all the errors are critical. We may move some to `error` if needed.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -74,6 +119,9 @@ impl StateReader for StateStore {
             .get::<CertificatePerNetworkColumn>(&certificate_per_network::Key {
                 network_id: *network_id,
                 height,
+            })?
+            .map_or(Ok(None), |certificate_id| {
+                self.get_certificate_header(&certificate_id)
```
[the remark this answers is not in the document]
Freyskeyd: Technically, we shouldn't have the second case if the first is `Ok` as the relationship between the two is 1:1.
I will put a TODO and a warning to track if it happens

crates/agglayer-node/src/rpc/mod.rs
Freyskeyd: That's [destructuring-assignment](https://rust-lang.github.io/rfcs/2909-destructuring-assignment.html) RFC, basically it is the same things as `let _` but in a more concise and clear way.

crates/agglayer-aggregator-notifier/src/sp1/mod.rs
Freyskeyd: The downcast is due to the fact that the `SP1Prover` returns an `anyhow::Error` which can be a `ProofError` from the `pessimistic-proof` program. We have no choice than `downcasting` the error to try getting more context.
I will put a TODO when we have time to check for a better solution.

hadjiszs (approved):

PR #338, opened 2024-10-04
feat: use `humantime` for `Duration` in config

iljakuklic:
# Description

There is an inconsistency in time format used in the config. The rate limiter uses human readable strings like "1h" or "15min" while all other components use a plain number interpreted as seconds.

Let's address the inconsistency while improving the UX and maintaining backwards compatibility. Make the config parser to accept both human strings and number of seconds, at user discretion.

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/src/epoch.rs
+++ b/crates/agglayer-config/src/epoch.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
 
 /// The Epoch configuration.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
@@ -20,8 +20,7 @@ impl Default for Epoch {
 pub struct TimeClockConfig {
     #[serde(
         default = "default_epoch_duration",
-        serialize_with = "serialize_duration",
-        deserialize_with = "deserialize_duration",
+        with = "crate::with::HumanDuration",
         alias = "EpochDuration"
     )]
     pub epoch_duration: Duration,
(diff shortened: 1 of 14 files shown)


hadjiszs (approved):

crates/agglayer-config/Cargo.toml
```diff
@@ -21,6 +21,8 @@ rstest.workspace = true
 serde_json = { workspace = true }
 toml.workspace = true
 
+agglayer-config = { path = ".", features = ["testutils"] }
```
Freyskeyd, line 24: why do we need that?

crates/agglayer-config/src/epoch.rs
```diff
@@ -65,7 +48,7 @@ mod tests {
         let epoch = Epoch::TimeClock(TimeClockConfig::default());
         let serialized = serde_json::to_string(&epoch).unwrap();
 
-        assert_eq!(serialized, r#"{"time-clock":{"epoch-duration":5}}"#);
+        assert_eq!(serialized, r#"{"time-clock":{"epoch-duration":"5s"}}"#);
```
Freyskeyd, line 51: How does it deal with not having the `s` ?
Do we have a fallback for config that would use the previous format?

Freyskeyd (approved):

crates/agglayer-config/src/epoch.rs
iljakuklic: > How does it deal with not having the `s` ? Do we have a fallback for config that would use the previous format?

Yes the old config still loads fine. You can choose either to use string with units, like `"5s"` or `"15h"`, or just a json number which is treated as number of seconds, same as with `DurationSeconds`.

crates/agglayer-config/Cargo.toml
iljakuklic: > why do we need that?

It failed to build if only building `agglayer-config` crate but it looks like it's no longer needed, so I removed it.

PR #351, opened 2024-10-11
feat: retrieve the l1 info tree leaf count from certificate

hadjiszs:
# Description

Add one function to retrieve the L1 Info Tree leaf count considered for a given Certificate.

It corresponds to the highest leaf index considered among the imported bridge exits of a given Certificate.

## PR Checklist:

### DIFF
--- a/crates/agglayer-types/src/lib.rs
+++ b/crates/agglayer-types/src/lib.rs
@@ -197,6 +197,15 @@ pub struct Certificate {
 }
 
 impl Certificate {
+    /// Default L1 Info Tree leaf count used for state transitions without
+    /// imported bridge exits.
+    const DEFAULT_L1_INFO_LEAF_COUNT: u32 = 0;
+    /// Default L1 Info Tree root used for state transitions without imported
+    /// bridge exits.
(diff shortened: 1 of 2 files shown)


crates/agglayer-types/src/lib.rs
```diff
@@ -219,6 +228,40 @@ impl Certificate {
...
+            .unwrap_or(Self::DEFAULT_L1_INFO_LEAF_COUNT)
+    }
+
+    /// Returns the L1 Info Root considered for this [`Certificate`].
+    /// Fails if multiple L1 Info Root are considered among the inclusion proofs
+    /// of the imported bridge exits.
+    pub fn l1_info_root(&self) -> Result<Digest, Error> {
+        let Some(l1_info_root) = self
+            .imported_bridge_exits
+            .first()
```
Freyskeyd, line 249: Why do we take the first one?

Freyskeyd (approved):

crates/agglayer-types/src/lib.rs
hadjiszs: just ensuring that they are all equals, so just picking the first one and checking that all others are equal. If there is none, then returning the default hash defined by the constant

PR #359, opened 2024-10-17
chore: adding prover config generator and remove useless doctests

Freyskeyd:
# Description

This PR adds a new command to the CLI to generate the default `agglayer-prover` config

## PR Checklist:


crates/agglayer/src/main.rs
```diff
@@ -11,7 +11,15 @@ fn main() -> anyhow::Result<()> {
     match cli.cmd {
         cli::Commands::Run { cfg } => agglayer_node::main(cfg)?,
         cli::Commands::Prover { cfg } => agglayer_prover::main(cfg)?,
-        cli::Commands::Config { base_dir } => println!(
+        cli::Commands::Config { prover: true, .. } => println!(
```
iljakuklic, line 14: Since this ignores the `base_dir` value, it would probably be cleaner if it was a separate subcommand, i.e. we'd have something like `generate-node-config` and `generate-prover-config` each accepting switches relevant to given mode.

iljakuklic (approved):

crates/agglayer/src/main.rs
iljakuklic: Also, a snapshot test for default prover config would be nice, similar to what we have for node config.

Freyskeyd: Yeah I didn't wanted to have too much changes on that but it makes sense 👌🏼

PR #362, opened 2024-10-18
feat: adding block clock by default

Freyskeyd:
# Description

This PR adds and refactor how the clock behave when starting.
Basically, when started with a genesis block in the future, the clock will wait for that block number to start.

## PR Checklist:

### DIFF
--- a/crates/agglayer-clock/src/block.rs
+++ b/crates/agglayer-clock/src/block.rs
@@ -13,7 +13,7 @@ use ethers::{
 use futures::StreamExt as _;
 use tokio::sync::broadcast;
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error};
+use tracing::{debug, error, info};
 
 use crate::{Clock, ClockRef, Error, Event, BROADCAST_CHANNEL_SIZE};
 
@@ -70,6 +70,12 @@ impl<P> BlockClock<P> {
             epoch_duration,
             current_epoch: Arc::new(AtomicU64::new(0)),
         }
+    }
+
+    /// Reinitialize the current Epoch number based on the current Block height.
+    fn reinitialize_epoch_number(&mut self, current_block: u64) {
+        let current_epoch = self.calculate_epoch_number(current_block);
+        self.current_epoch.store(current_epoch, Ordering::SeqCst);
     }
 
     /// Updates the current Epoch of this [`BlockClock`].
@@ -135,15 +141,17 @@ where
         sender: broadcast::Sender<Event>,
         cancellation_token: CancellationToken,
     ) -> Result<(), BlockClockError> {
+        info!("Starting the BlockClock task");
         // Start by setting the current Block height based on the current L1 Block
         // number. If the current L1 Block number is less than the genesis block
-        // number, we exit the process with an println message.
+        // number, we walk the L1 block stream until reaching the genesis block.
         let current_l1_block = self
             .provider
             .get_block_number()
             .await
             .map_err(|_| BlockClockError::GetBlockNumber)?;
 
+        debug!("Current L1 Block number: {}", current_l1_block);
         let provider = self.provider.clone();
 
         // Subscribe to the L1 Block stream.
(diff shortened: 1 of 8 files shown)


crates/agglayer-clock/src/block.rs
```diff
@@ -151,13 +152,16 @@ where
...
 
-        let current_l1_block = current_l1_block.as_u64();
-        if current_l1_block < self.genesis_block {
-            return Err(BlockClockError::GenesisBlockNumber(
-                current_l1_block,
-                self.genesis_block,
-            ));
-        };
+        let mut current_l1_block = current_l1_block.as_u64();
+        while !current_l1_block < self.genesis_block {
```
hadjiszs, line 156: not sure of why the negation?

there is also the first comment at the beginning of this function to adapt

```diff
        // Start by setting the current Block height based on the current L1 Block
        // number. If the current L1 Block number is less than the genesis block
-       // number, we exit the process with an println message.
+       // number, we walk the L1 block stream until reaching the genesis block.
```

crates/agglayer-clock/src/block.rs
```diff
@@ -302,26 +306,37 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_block_clock_starting_with_wrong_genesis() {
+    async fn test_block_clock_with_genesis_in_future() {
+        let anvil = Anvil::new().block_time(1u64).spawn();
+        let client = Provider::<Ws>::connect(anvil.ws_endpoint()).await.unwrap();
+
+        let clock = BlockClock::new(client, 2, NonZeroU64::new(3).unwrap());
```
hadjiszs, line 313: I'd either increase the block time or move the genesis slightly more in the future than block 2, otherwise with a block time of 1s for anvil, it feels that the genesis block might not be in the future once spawning the block clock no?

crates/agglayer/tests/snapshots/config__config_display.snap
```diff
@@ -45,8 +45,9 @@ private-keys = []
 [telemetry]
 prometheus-addr = "0.0.0.0:3000"
 
-[epoch.time-clock]
-epoch-duration = 60
+[epoch.block-clock]
+epoch-duration = 8
```
iljakuklic, line 49: May be my Rust biases speaking but duration invokes real time rather than blocks. How about something like `blocks-per-epoch`? Hmm, maybe that's not better... idk.

crates/agglayer-clock/src/block.rs
```diff
@@ -69,6 +72,12 @@ impl<P> BlockClock<P> {
         }
     }
 
+    /// Reinitialize the current Epoch number based on the current Block height.
+    fn reinitialize_epoch_number(&mut self, current_block: u64) {
+        let current_epoch = self.calculate_epoch_number(current_block);
+        self.current_epoch.store(current_epoch, Ordering::SeqCst);
```
iljakuklic, line 78: Seems error prone, and possibly racy, to keep both the current block and current epoch in sync. Could we just keep track of the current block and calculate the epoch from that on demand?

iljakuklic (approved):

crates/agglayer-clock/src/block.rs
Freyskeyd: This method is only used when starting the clock, I don't expect much risk on this. The `SeqCst` provides some guarantee too.

After that, only the `clock` that is spawn is updating the different counter (both `block` and `epoch`). The `ClockRef` which can be cloned and move from thread to thread has a ref to those counter but only read them to provide method such as `current_epoch` and `current_block_height`.

Those counter are not exposed outside the scope of the `clock` module, thus can't be read nor update by any other process. 

I agree that it could be better to have the epoch calculated on the flight, I'll create a ticket to address that.
One question that I have is, how could we properly deal with a change of epoch duration ?

Freyskeyd: Here it is: #375

crates/agglayer/tests/snapshots/config__config_display.snap
Freyskeyd: Yeah I too had some back and forth between both options. I end up letting it with `duration` as we refer to the `duration` of an epoch when discussing. And the `blocks` are here to have a global clock when we'll have multiple nodes. The goal was to define that if one block is 15sec, then if we want to have epoch that are 1h duration, we need 250 block duration.

iljakuklic: Ok. No strong opinion here, we can leave it as is.

crates/agglayer-clock/src/block.rs
```diff
@@ -151,22 +160,31 @@ where
...
-        if current_l1_block < self.genesis_block {
-            return Err(BlockClockError::GenesisBlockNumber(
-                current_l1_block,
-                self.genesis_block,
-            ));
-        };
+        debug!("Successfully subscribed to the L1 Block stream");
+
+        let mut current_l1_block = current_l1_block.as_u64();
+        while !current_l1_block < self.genesis_block {
```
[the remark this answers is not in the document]
Freyskeyd: I update the condition and added one more test ! thankt !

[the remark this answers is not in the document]
iljakuklic: Hmm I think it's possibly even worse. If it parses as `(!current_l1_block) < self.genesis_block`, we are comparing with a bitwise complement of the current block number.

hadjiszs (approved):
:+1:

PR #364, opened 2024-10-21
feat: inclusion proof of sub-local exit tree

wborgeaud:
Implement a function `LocalExitTree::is_subtree()` to check if a local exit tree is a sub-tree of another one.
See #119 for discussions.

## PR Checklist:

### DIFF
--- a/crates/pessimistic-proof/src/local_exit_tree/mod.rs
+++ b/crates/pessimistic-proof/src/local_exit_tree/mod.rs
@@ -1,3 +1,4 @@
+use data::LETMerkleProof;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_with::serde_as;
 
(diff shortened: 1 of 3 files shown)


crates/pessimistic-proof/src/local_exit_tree/tests.rs
```diff
@@ -25,6 +26,32 @@ fn test_local_exit_tree_basic() {
...
+
+#[test]
+fn test_local_exit_tree_is_subtree_different_sizes() {
+    local_exit_tree_is_subtree(100, 200);
+}
+
+#[test]
+fn test_local_exit_tree_is_subtree_empty() {
+    local_exit_tree_is_subtree(0, 100);
+}
```
hadjiszs, line 53: Maybe adding at least one failing case? Also the case with two trees that are the same?

wborgeaud: Done 👍

crates/pessimistic-proof/src/local_exit_tree/mod.rs
```diff
@@ -106,6 +107,41 @@ where
 
         root
     }
+
+    /// Returns true if `other_root` is the root of a LET whose first
+    /// `self.leaf_count` leaves are the same as the leaves of `self`.
+    /// `next_leaf` is the leaf at index `self.leaf_count` in the other tree and
+    /// `proof` is the Merkle proof of `next_leaf`.
+    pub fn is_subtree(
```
iljakuklic, line 115: If I read the description correctly, this could be called `is_subtree_of` to make it a bit clearer which way around the subtree relation goes.

iljakuklic: Also, the presence of `next_leaf` suggests this checks strict subset/subtree. May be worth mentioning in the comment

crates/pessimistic-proof/src/local_exit_tree/tests.rs
```diff
@@ -25,6 +26,69 @@ fn test_local_exit_tree_basic() {
...
+
+#[test]
+fn test_local_exit_tree_is_subtree_same_tree() {
+    let leaf_count_a = 100;
+    let mut let_a = LocalExitTreeData::<Keccak256Hasher>::new();
+    for _ in 0..leaf_count_a {
+        let_a.add_leaf(random());
+    }
+    let mut let_b = let_a.clone();
+    let_b.add_leaf([0; 32]); // Hack so that `let_b.get_proof(leaf_count_a)` doesn't panic
```
iljakuklic, line 63: Aside: Would it make sense to change `get_proof` so that instead of panicking, it treats out of range leaves as being 0?

iljakuklic (dismissed):

crates/pessimistic-proof/src/local_exit_tree/tests.rs
wborgeaud: Hm I think it's a nice fail-safe to panic on out-of-range leaves. In practice we won't need this hack because (as mentioned in the other comment) we can just compare Merkle roots when the trees are equal.

hadjiszs:
gonna propagate it in `interop` repo when needed 👍

PR #371, opened 2024-10-21
chore: refactor config to make it mandatory

Freyskeyd:
# Description

This PR refactors how we handle the configuration.
No more backward compatibility, but a new `validate-config` command and also more flexibility on how to hanle the serialization/deserialization.
## PR Checklist:


crates/agglayer/tests/snapshots/config__config_display.snap
```diff
@@ -35,6 +35,7 @@ settlement-timeout = 1200
 [l1]
 chain-id = 1337
 node-url = "http://zkevm-mock-l1-network:8545/"
+ws-node-url = "ws://zkevm-mock-l1-network:8546/"
```
iljakuklic, line 38: Would it make sense to only have one L1 node url, the protocol to be determined by the scheme (`http://` or `ws://`)? The `ws://` variant required for block based epoch clock.

crates/agglayer-types/Cargo.toml
```diff
@@ -13,9 +13,9 @@ serde_with.workspace = true
 sp1-sdk.workspace = true
 sp1-core-machine.workspace = true
 thiserror.workspace = true
+bincode.workspace = true
 
 [dev-dependencies]
-bincode.workspace = true
```
iljakuklic, line 18: Incredibly nitpicky nitpick: maybe remove the whole `[dev-dependencies]` section?

crates/agglayer-config/tests/snapshots/validate_deserialize__empty_rpcs.snap
```diff
@@ -0,0 +1,60 @@
...
+node-url = "http://zkevm-mock-l1-network:8545/"
+ws-node-url = "ws://zkevm-mock-l1-network:8546/"
+rollup-manager-contract = "0xb7f8bc63bbcad18155201308c8f3540b07f84f5e"
+rpc-timeout = 45
+
+[auth.local]
+private-keys = []
+
+[telemetry]
+prometheus-addr = "0.0.0.0:3000"
```
iljakuklic, line 46: Should probably remove this default too, it does not make much sense.

crates/agglayer-config/src/storage.rs
```diff
@@ -104,6 +118,36 @@ impl From<StorageConfig> for StorageConfigHelper {
...
 }
 
+fn default_db_path() -> PathBuf {
+    PathBuf::new().join(STORAGE_DIR)
+}
+
+pub(crate) fn normalize_path(path: &Path) -> PathBuf {
+    let mut components = path.components().peekable();
+    let mut ret = if let Some(c @ Component::Prefix(..)) = components.peek().cloned() {
+        components.next();
```
iljakuklic, line 128: Why is the prefix component removed? A comment explaining it would be useful.

crates/agglayer-config/src/storage.rs
```diff
@@ -104,6 +118,36 @@ impl From<StorageConfig> for StorageConfigHelper {
...
+    for component in components {
+        match component {
+            Component::Prefix(..) => unreachable!(),
+            Component::RootDir => {
+                ret.push(component.as_os_str());
+            }
+            Component::CurDir => {}
+            Component::ParentDir => {
+                ret.pop();
+            }
```
iljakuklic, line 143: What's the intended behaviour in presence of symlinks? Should `foo/..` resolve to `.` even if `foo` is a symlink pointing somewhere else? There's the `Path::canonicalize` method which follows symlinks but it requires the path to physically exist.

iljakuklic (commented):
Some comments below. The PR changed during the review so apologies if some of the comments are outdated at this point.

crates/agglayer/tests/snapshots/config__config_display.snap
Freyskeyd: I would prefer to have different one as `http` and `ws` are sometimes on different ports

Cargo.toml
```diff
@@ -25,7 +25,11 @@ ethers-signers = "2.0.14"
 fail = "0.5.1"
 futures = "0.3.31"
 hex = "0.4.3"
-jsonrpsee = { version = "0.24.7", features = ["full"] }
+insta = { git = "https://github.com/freyskeyd/insta", branch = "chore/updating-deps-to-avoid-serialize-error", features = [
```
hadjiszs, line 28: Temporary right?

crates/agglayer-config/src/storage.rs
Freyskeyd: This part of the method inherited from `cargo`

crates/agglayer-config/tests/snapshots/validate_deserialize__empty_rpcs.snap
Freyskeyd: it does, it is the metrics entrypoint

iljakuklic: What does the `0.0.0.0` mean? Does it disable prometheus? If so, I'd probably prefer it to be a bit more explicit about that, using e.g. `null` or `"disabled"` or something.

crates/agglayer-config/src/storage.rs
```diff
@@ -104,6 +118,36 @@ impl From<StorageConfig> for StorageConfigHelper {
     }
 }
 
+fn default_db_path() -> PathBuf {
+    PathBuf::new().join(STORAGE_DIR)
+}
+
+pub(crate) fn normalize_path(path: &Path) -> PathBuf {
```
hadjiszs, line 125: I'd propagate the rustdoc from the cargo [code](https://github.com/rust-lang/cargo/blob/edd36eba5e0d6e0cfcb84bd0cc651ba8bf5e7f83/crates/cargo-util/src/paths.rs#L76-L83) that you shared maybe, at least the first line `Normalize a path, removing things like . and ..`

hadjiszs (approved):

crates/agglayer-config/src/lib.rs
```diff
@@ -48,24 +41,30 @@ pub use rate_limiting::RateLimitingConfig;
...
+#[serde(deny_unknown_fields)]
+#[serde(default)]
 pub struct Config {
     /// A map of Zkevm node RPC endpoints for each rollup.
     ///
     /// The key is the rollup ID, and the value is the URL of the associated RPC
     /// endpoint.
-    #[serde(alias = "FullNodeRPCs", deserialize_with = "deserialize_rpc_map")]
+    #[serde_as(as = "HashMap<DisplayFromStr, _>")]
     pub full_node_rpcs: HashMap<u32, Url>,
```
iljakuklic, line 55: Note: There's also similar thing in rate limiter config for per-network rate limiting configuration but uses an ad-hoc implementation. This approach with `serde_as` is better.

iljakuklic (approved):

crates/agglayer-config/src/lib.rs
Freyskeyd: I'll not touch more than the scope of this PR, I created a ticket: #374 for that

PR #378, opened 2024-10-25
feat: add metadata field in the certificate

### LINKED ISSUE #377: Add one `metadata: U256` field in the Certificate
# Rationale

Upon failure and restart, the aggsender needs to know from which block it should continue the certificate production. To do so, the latest LER is not enough given that the imported bridge exits does not impact the LER.

The chosen solution is to add one metadata field `U256` in the Certificate in order to let the aggsender populate this field with the block number and extend it as needed in the future.

# Acceptance Criteria

- [ ] The field is added in the `Certificate`
- [ ] The field is exposed in the `CertificateHeader`

hadjiszs:
# Description

Adding one metadata field in the certificate for chain purposes

Fixes #377 

## PR Checklist:

### DIFF
--- a/crates/agglayer-types/src/lib.rs
+++ b/crates/agglayer-types/src/lib.rs
@@ -39,8 +39,8 @@ pub struct CertificateHeader {
     pub certificate_index: Option<CertificateIndex>,
     pub certificate_id: CertificateId,
     pub new_local_exit_root: Hash,
-    pub status: CertificateStatus,
     pub metadata: Metadata,
+    pub status: CertificateStatus,
 }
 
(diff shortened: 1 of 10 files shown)


crates/agglayer-types/src/lib.rs
```diff
@@ -39,6 +40,7 @@ pub struct CertificateHeader {
     pub certificate_id: CertificateId,
     pub new_local_exit_root: Hash,
     pub status: CertificateStatus,
+    pub metadata: Metadata,
```
Freyskeyd, line 43: Can we place the `metadata` before `status` so that we have consistent size until the `status` which can contains different length?

Freyskeyd (approved):

PR #388, opened 2024-10-29
feat: adding epoch config rpc

Freyskeyd:
# Description

This PR adds a new RPC `interop_getClockConfiguration` to retrieve the information related to how epochs are calculated.

Currently, only the `Block` clock is returned (as it doesn't really make sense for other type to be exposed).

The RPC is returning a `ClockConfiguration` struct which defines `genesis_block` and `epoch_duration`.

Snapshot of both success and error are available in the repo and tested, they look like that:

### Success with block config
```json
{
  "id": 0,
  "jsonrpc": "2.0",
  "result": {
    "epoch_duration": 6,
    "genesis_block": 0
  }
}
```

### Error with time config
```json
{
  "error": {
    "code": -32603,
    "data": {
      "internal": "AggLayer isn't configured with a BlockClock configuration, thus no ClockConfiguration is available"
    },
    "message": "Internal error: AggLayer isn't configured with a BlockClock configuration, thus no ClockConfiguration is available"
  },
  "id": 0,
  "jsonrpc": "2.0"
}
```
## PR Checklist:


crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -50,6 +53,9 @@ trait Agglayer {
         &self,
         certificate_id: CertificateId,
     ) -> RpcResult<CertificateHeader>;
+
+    #[method(name = "getClockConfiguration")]
+    async fn get_clock_configuration(&self) -> RpcResult<ClockConfiguration>;
```
hadjiszs, line 58: nit: maybe "epoch configuration" fits better than "clock configuration"

hadjiszs (approved):

iljakuklic (approved):

PR #402, opened 2024-11-05
fix: different size for usize between native and prover

hadjiszs:
# Description

`usize` became 4bytes in the prover while being 8bytes in 64bit native machines leading to failures for some comparisons

## PR Checklist:


crates/agglayer-types/src/lib.rs
```diff
@@ -325,10 +328,9 @@ impl LocalNetworkStateData {
 
         let l1_info_root = certificate.l1_info_root()?;
 
-        certificate
-            .bridge_exits
-            .iter()
-            .for_each(|e| self.exit_tree.add_leaf(e.hash()));
+        for e in certificate.bridge_exits.iter() {
+            self.exit_tree.add_leaf(e.hash())?;
```
Freyskeyd, line 332: No need to revert the changes made before the failure?

crates/pessimistic-proof-test-suite/src/bin/ppgen.rs
```diff
@@ -89,13 +89,14 @@ pub fn main() {
...
         .expect("proving failed");
     let duration = start.elapsed();
-
     info!(
         "Successfully generated the plonk proof with a latency of {:?}",
         duration
     );
 
     let vkey = vk.bytes32().to_string();
+    info!("vkey: {}", vkey);
```
Freyskeyd, line 98: left over debug?

crates/pessimistic-proof/src/local_exit_tree/mod.rs
```diff
@@ -69,12 +71,10 @@ where
...
-    pub fn add_leaf(&mut self, leaf: H::Digest) {
-        assert!(
-            (self.leaf_count as usize) < (1usize << TREE_DEPTH) - 1,
-            "Can have at most {} leaves.",
-            (1usize << TREE_DEPTH) - 1
-        );
+    pub fn add_leaf(&mut self, leaf: H::Digest) -> Result<u32, LocalExitTreeError> {
+        if self.leaf_count as u64 >> TREE_DEPTH != 0 {
+            return Err(LocalExitTreeError::LeafIndexOverflow);
+        }
```
wborgeaud, line 77: nit: Not very important but this checks that `self.leaf_count < 2^TREE_DEPTH`, when it should really check that `self.leaf_count < 2^TREE_DEPTH - 1` (the LET implementation doesn't support having `2^TREE_DEPTH - 1` leaves).

wborgeaud (approved):
LGTM

crates/agglayer-types/src/lib.rs
hadjiszs: yeah for the rollback in case of failure one should use `make_multi_batch_header` which applies the certificate on a temporary state

iljakuklic (approved):

crates/pessimistic-proof-test-suite/src/bin/ppgen.rs
Freyskeyd: Okey, at the end it would be better to have a dedicated CLI command on `agglayer` to retrieve it directly

PR #417, opened 2024-11-15
fix: switching to alloy for blocks

Freyskeyd:
# Description

This PR switches from ethers to alloy for listening on Blocks.

It uses a backoff mechanisms in case of network failure

## PR Checklist:


crates/agglayer-clock/src/block.rs
```diff
@@ -4,19 +4,36 @@ use std::{
...
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, trace, warn};
 
 use crate::{Clock, ClockRef, Error, Event, BROADCAST_CHANNEL_SIZE};
 
+type BlockProvider = FillProvider<
+    JoinFill<
+        Identity,
+        JoinFill<GasFiller, JoinFill<BlobGasFiller, JoinFill<NonceFiller, ChainIdFiller>>>,
+    >,
```
iljakuklic, line 31: Nitpick:

```suggestion
    JoinFill<
        Identity,
        Ethereum::RecommendedFillers,
    >,
```

(probably needs to import / qualify with the `RecommendedFillers` trait, will not work as suggested)

crates/agglayer-clock/src/block.rs
```diff
@@ -260,21 +315,59 @@ where
...
+                Ok(self.0.try_reconnect().await?)
+            },
+        )
+        .await
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use std::{num::NonZeroU64, time::Duration};
```
iljakuklic, line 265: Nitpick: Imho there's now enough tests to deserve a separate file.

iljakuklic (approved):
+1, no important comments, just some nits.

crates/agglayer-clock/src/block.rs
Freyskeyd: I tried to use that format earlier, but it doesn't reduce the boilerplate as I needed to define `Ethereum as ...`

PR #432, opened 2024-11-25
feat: adding more context and error management for settlement

Freyskeyd:
# Description

Adding more context when Settlement contract call fails and catch it on the network task.

## PR Checklist:

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/packer/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/packer/mod.rs
@@ -13,7 +13,8 @@ use agglayer_types::{
 use bincode::Options;
 use futures::future::BoxFuture;
 use pessimistic_proof::PessimisticProofOutput;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
+use tracing::{Instrument, Span};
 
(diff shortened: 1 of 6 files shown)


crates/agglayer-aggregator-notifier/src/packer/mod.rs
```diff
@@ -113,78 +114,90 @@ where
                 proof.into(),
             );
 
+        let span = Span::current();
+        span.record("certificate_id", hash.clone());
+        span.record("network_id", *network_id);
+        if let Ok(data) = serde_json::to_string(&contract_call.function.inputs) {
+            span.record("call_data", data);
+        } else {
```
hadjiszs, line 122: I think `contract_call.function.inputs` seems to be the list of the input ABI variables but without their actual values, we may want what is under `contract_call.tx.data` instead if possible, they seem to provide one method `contract_call.call_data()`

hadjiszs: discussed offline: will get back to that if we fall back again to one InvalidProof

hadjiszs (approved):

PR #433, opened 2024-11-25
fix: dont fail the sync when epoch already packed

Freyskeyd:
# Description

This PR prevents the agglayer epoch sync to fail when an epoch is already packed.

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/epoch_synchronizer.rs
+++ b/crates/agglayer-node/src/epoch_synchronizer.rs
@@ -39,7 +39,7 @@ impl EpochSynchronizer {
                         error
                     );
 
-                    return Err(error)?;
+                    return Err(error.into());
                 }
                 Ok(_) => {}
             }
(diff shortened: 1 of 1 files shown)


hadjiszs (approved):

crates/agglayer-node/src/epoch_synchronizer.rs
```diff
@@ -21,7 +25,24 @@ impl EpochSynchronizer {
...
+                    );
+                }
+                Err(error) => {
+                    error!(
+                        "Error starting packing for epoch {}: {:?}",
+                        opened_epoch.get_epoch_number(),
+                        error
+                    );
+
+                    return Err(error)?;
```
iljakuklic, line 42: The question mark operator is a bit wierd.

Would one of the following work?
```suggestion
                    return Err(error);
```
```suggestion
                    return Err(error.into());
```

iljakuklic (approved):

PR #440, opened 2024-11-27
fix: bats test

atanmarko:
# Description

Add bats dependencies as submodules of the project.
Updated `_common_setup`.

## PR Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 tests/kurtosis
+.idea
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "tests/e2e/helpers/bats-assert"]
+	path = tests/e2e/helpers/bats-assert
+	url = https://github.com/ztombol/bats-assert
+[submodule "tests/e2e/helpers/bats-support"]
+	path = tests/e2e/helpers/bats-support
+	url = https://github.com/ztombol/bats-support
--- a/Makefile.e2e.toml
+++ b/Makefile.e2e.toml
@@ -23,5 +23,5 @@ category = "E2E - PREPARE"
 cwd = "tests/kurtosis"
 script = '''
 kurtosis clean --all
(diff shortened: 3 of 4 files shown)


Makefile.e2e.toml
```diff
@@ -23,5 +23,5 @@ category = "E2E - PREPARE"
 cwd = "tests/kurtosis"
 script = '''
 kurtosis clean --all
-kurtosis run --enclave cdk --args-file .github/tests/fork12-pessimistic.yml --image-download always . '{"args": {"agglayer_prover_sp1_key": "\'$AGGLAYER_PROVER_SP1_KEY'"}}'
+kurtosis run --enclave cdk --args-file .github/tests/fork12-pessimistic.yml --image-download always . '{"args": {"agglayer_prover_sp1_key": "'$AGGLAYER_PROVER_SP1_KEY'"}}'
```
0xaatif, line 26: very nice :)

0xaatif (commented):
Love this, but I think submodules are a bit of a pain - can you think of another way? Is there a reason `homebrew` isn't the best solution?

atanmarko:
> Is there a reason homebrew isn't the best solution?  

https://github.com/ztombol/bats-docs?tab=readme-ov-file#installation

Documentation says it is only for `macOS` users. Many people have Linux. Even if you can setup `brew` on your system, I am not happy to introduce another package manager to my system just to be able to install two supporting libraries.

>  Do they not ship bats-assert and bats-core?

Homebrew, npm and source code.

.gitmodules
Freyskeyd, line 1: I would prefer relying on npm version of those libraries: 

- https://www.npmjs.com/package/bats-assert
- https://www.npmjs.com/package/bats-support

.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 tests/kurtosis
+.idea
```
Freyskeyd, line 3: I think that could be place in a global `.gitignore` on your machine ^^

PR #442, opened 2024-11-27
feat: add a debug-mode to the node

Freyskeyd:
# Description

This PR is adding a new field in the config to activate a `debug-mode`.
The Debug mode will save all certificates received in another DB.
A new API is also exposed to retrieve them.

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/src/lib.rs
+++ b/crates/agglayer-config/src/lib.rs
@@ -104,6 +104,10 @@ pub struct Config {
     #[serde(default = "default_prover_entrypoint")]
     #[serde(skip_serializing_if = "String::is_empty")]
     pub prover_entrypoint: String,
+
+    #[serde(default)]
+    #[serde(skip_serializing_if = "<&bool as std::ops::Not>::not")]
+    pub debug_mode: bool,
 }
 
 impl Config {
@@ -153,6 +157,7 @@ impl Config {
             shutdown: Default::default(),
             certificate_orchestrator: Default::default(),
             prover_entrypoint: default_prover_entrypoint(),
+            debug_mode: false,
         }
     }
 
--- a/crates/agglayer-config/src/lib.rs
+++ b/crates/agglayer-config/src/lib.rs
@@ -106,7 +106,7 @@ pub struct Config {
     pub prover_entrypoint: String,
 
     #[serde(default)]
-    #[serde(skip_serializing_if = "<&bool as std::ops::Not>::not")]
+    #[serde(skip_serializing_if = "is_false")]
     pub debug_mode: bool,
 }
 
@@ -224,4 +224,8 @@ impl<'de> DeserializeSeed<'de> for ConfigDeserializer<'_> {
             .validate()
             .map_err(|e| serde::de::Error::custom(e.to_string()))
     }
+}
+
+fn is_false(b: &bool) -> bool {
+    !*b
 }
--- a/crates/agglayer-node/src/rpc/mod.rs
+++ b/crates/agglayer-node/src/rpc/mod.rs
@@ -4,6 +4,8 @@ use agglayer_config::epoch::BlockClockConfig;
 use agglayer_config::Config;
 use agglayer_config::Epoch;
 use agglayer_storage::columns::latest_settled_certificate_per_network::SettledCertificate;
+use agglayer_storage::stores::DebugReader;
+use agglayer_storage::stores::DebugWriter;
 use agglayer_storage::stores::PendingCertificateReader;
 use agglayer_storage::stores::PendingCertificateWriter;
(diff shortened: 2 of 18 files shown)


hadjiszs (approved):

iljakuklic (approved):

crates/agglayer-config/src/lib.rs
```diff
@@ -104,6 +104,10 @@ pub struct Config {
     #[serde(default = "default_prover_entrypoint")]
     #[serde(skip_serializing_if = "String::is_empty")]
     pub prover_entrypoint: String,
+
+    #[serde(default)]
+    #[serde(skip_serializing_if = "<&bool as std::ops::Not>::not")]
```
0xaatif, line 109: style: I tend to write a `is_false`, which reads a bit easier ;)

crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -452,6 +476,35 @@ where
             Ok(None)
         }
     }
+
+    async fn debug_get_certificate(
+        &self,
+        certificate_id: CertificateId,
+    ) -> RpcResult<(Certificate, Option<CertificateHeader>)> {
+        trace!("Received request to get certificate for id {certificate_id}");
```
0xaatif, line 484: smell: We're `trace!`-ing entry to this function, and `error!`-ing appropriately, which feels similar to the `tracing::instrument` macro :D

crates/agglayer-config/src/lib.rs
atanmarko: We probably don't need trace there at all.

crates/agglayer-node/src/rpc/mod.rs
atanmarko: We probably don't need trace there at all.

Freyskeyd: > We probably don't need trace there at all.

It is useful when locally debugging access, as the `Ok` doesn't log anything. If it becomes too annoying, we'll remove it.

> smell: We're `trace!`-ing entry to this function, and `error!`-ing appropriately, which feels similar to the `tracing::instrument` macro :D

The `instrument::instrument` macro is just here to define the `Span` in which the function is part of, with pre-initialized fields and stuff. It doesn't provide any logs and in fact won't do anything if there is no event produced in it.

0xaatif: > The instrument::instrument macro is just here to define the Span in which the function is part of, with pre-initialized fields and stuff. It doesn't provide any logs and in fact won't do anything if there is no event produced in it.

I think `tracing` is more poweful than you give it credit for :)
```rust
use std::{env, fs::File, io::{self, Write as _}, path::Path};

#[tracing::instrument(skip(bytes), err, level = "info")]
fn write_to_file(path: &Path, bytes: &[u8]) -> io::Result<()> {
    File::create(path)?.write_all(bytes)
}

fn main() {
    tracing_subscriber::fmt().init();
    for path in env::args_os().skip(1) {
        let _ = write_to_file(path.as_ref(), b"foo");
    }
}
```

```console
$ cargo run --quiet --example demo -- /dev/null /does/not/exist
2024-11-29T18:04:26.472443Z ERROR write_to_file{path="/does/not/exist"}: demo: error=No such file or directory (os error 2)
```

0xaatif: - `skip(bytes)`: don't include `bytes` in the span.
- `err`: emit an event for `Result::Err`
- `level = "info"` level for the span events when doing this:
```rust
    tracing_subscriber::fmt()
        .with_span_events(FmtSpan::FULL)
        .init();
```

Freyskeyd: I know what `tracing` can do, but in the scope of this PR, introducing changes related to tracing and how we track and expose the instrumented function doesn't seem to be a good idea. It should be a dedicated PR that asserts how logs are displayed for both devs and tools like Datadog.

I really think that we have room for improvement on this part, both in terms of logs but also in terms of general observability. It is just a matter of defining how we want to do it and doing it at the right time and the right place. For that, I created a [ticket](https://github.com/orgs/agglayer/projects/4?query=sort%3Aupdated-desc+is%3Aopen&pane=issue&itemId=89193327) that you can edit or comment with suggestions if you have some 🙂 

As this PR is part of an ongoing effort to production, I'll remove this log as it is not really important, but if I should have to rely on `instrument` for replacing this single `trace` I would have used something like `#[instrument(level = "trace", skip(self), err(level = Level::TRACE))]` with only `FmtSpan::NEW` (or maybe `FmtSpan::NEW | FmtSpan::CLOSE`) to not be too verbose, I think.

PR #444, opened 2024-11-28
fix: get rid of stale nodes before going through disk

hadjiszs:
# Description

- the read smt db operation was not working for smt containing stale nodes
- fixes:
  - make the read smt db operation robust against smt with stale nodes
  - prune the stale nodes before each and every pass through the DB

## PR Checklist:


crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -508,14 +513,21 @@ where
...
+                    if let Some(mut new) = self.pending_state.take() {
                         debug!(
                             "Updated the state for network {} with the new state {} > {}",
                             self.network_id,
                             self.local_state.get_roots().display_to_hex(),
                             new.get_roots().display_to_hex()
                         );
 
+                        // Prune the SMTs of the state
+                        new.prune_stale_nodes()
```
Freyskeyd, line 525: Can we do that in the `cerifier` instead?

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -396,9 +398,16 @@ impl StateStore {
                         left: *left.as_bytes(),
                         right: *right.as_bytes(),
                     });
-
-                    keys.push_back(SmtKeyType::Node(left));
-                    keys.push_back(SmtKeyType::Node(right));
+                    // update here
```
Freyskeyd, line 401:
```suggestion

```

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -396,9 +398,16 @@ impl StateStore {
...
+                    if !already_read.contains(&left) {
+                        already_read.insert(left);
+                        keys.push_back(SmtKeyType::Node(left));
+                    }
+                    if !already_read.contains(&right) {
+                        already_read.insert(right);
+                        keys.push_back(SmtKeyType::Node(right));
+                    }
+                    // update end
```
Freyskeyd, line 410:
```suggestion
```

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -381,6 +382,7 @@ impl StateStore {
         let mut nodes: Vec<Node<Keccak256Hasher>> = Vec::new();
         nodes.push(root_node);
 
+        let mut already_read = BTreeSet::new();
```
Freyskeyd, line 385: Maybe we can find a better name, not a blocker tho

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -485,11 +494,21 @@ impl StateReader for StateStore {
         &self,
         network_id: NetworkId,
     ) -> Result<Option<LocalNetworkStateData>, Error> {
-        match (
-            self.read_local_exit_tree(network_id)?,
-            self.read_smt::<BalanceTreePerNetworkColumn, LOCAL_BALANCE_TREE_DEPTH>(network_id)?,
-            self.read_smt::<NullifierTreePerNetworkColumn, NULLIFIER_TREE_DEPTH>(network_id)?,
-        ) {
+        debug!("Reading local network state for network_id: {}", network_id);
```
Freyskeyd, line 497: I don't think we need to keep all of those debug right?

crates/agglayer-storage/src/stores/state/tests.rs
```diff
@@ -135,3 +137,95 @@ fn can_detect_inconsistent_state(network_id: NetworkId, store: StateStore) {
...
+
+use pessimistic_proof_test_suite::sample_data::{self as data};
+
+#[rstest]
+fn can_read(network_id: NetworkId, store: StateStore) {
+    let certificates: Vec<Certificate> = [
+        "n15-cert_h0.json",
+        "n15-cert_h1.json",
+        "n15-cert_h2-v2.json",
+        //"n15-cert_h3-v2.json",
```
Freyskeyd, line 149: Should it be uncommented or is it a leftover?

Freyskeyd (approved):

crates/agglayer-storage/src/stores/state/mod.rs
hadjiszs: I renamed it to `queued`, happy to rename it to something else

crates/agglayer-storage/src/stores/state/mod.rs
hadjiszs: Removed, I kept only the 1st log

crates/agglayer-storage/src/stores/state/tests.rs
hadjiszs: Leftover that I removed, this certificate was an attempt to replicate the h3 issue, I also renamed `n15-cert_h2-v2.json` to `n15-cert_h2.json` and deleted the unused ones

crates/agglayer-storage/src/stores/state/mod.rs
Freyskeyd: I think that `queue` isn't the right name either, we're storing the `hash` of every node that the SMT contains, shouldn't it be something like `smt_hashes` or `node_hashes` ?
I also think that we could do the `if insert..` before even being in the match, but I'm fine letting it this way for now.

PR #448, opened 2024-12-02
feat: general-purpose SP1 Authenticator

### LINKED ISSUE #542: Update `generate_pessimistic_proof` method to support the new `aggchain_type`
## Description

The `generate_pessimistic_proof` method needs to be updated to support the new `aggchain_type` which will validate the `aggchain_proof` using `aggchain_params` and `vkey` as input.

wborgeaud:
Replace the current ECDSA authenticator with a general purpose SP1 authenticator.
See https://github.com/agglayer/protocol-research/discussions/53 for discussion.

# TODO

- [x] Update the PP program to verify either the ECDSA signature or the sp1 stark proof
- [x] Update the Certificate to contain either the signature or the aggchain proof 

Fixes #542 #541 #464

## PR Checklist:


pgebheim:
Should this be called "Authentication via SP1 proof" or something to link the topics?

wborgeaud:
Quick update on this PR:

Ideally, we would have an `SP1Proof` consensus proof in the multi-batch header/certificate to replace the signature we have currently. However, I've had issues making SP1 compile when importing `sp1-sdk` (to reproduce, uncomment this line https://github.com/agglayer/agglayer/blob/8f12583bc2cc2674e2bba3be33ad5d65f81cca14/crates/pessimistic-proof/Cargo.toml#L19 and run `cargo prove build` in `pessimistic-proof-program`). 

I've pushed a solution to this issue in [65bffe6](https://github.com/agglayer/agglayer/pull/448/commits/65bffe68ee7d866ee9060243ec871474ccce97d3), where I add a `#[cfg(not(target_os = "zkvm"))]` on all `sp1-sdk` related code in `pessimistic-proof`. That way, when compiling the PP with SP1, we do not have to explicitly handle the consensus proof which is provided in a separate buffer during witness generation.

One minor issue is that I haven't found a way to verify an `SP1Proof` with just a `SP1VerifyingKey` and the public inputs (I've only found APIs using an `SP1Prover`). So currently, the consensus proof is not verified when running `apply_batch_header`  natively. See this TODO https://github.com/agglayer/agglayer/blob/65bffe68ee7d866ee9060243ec871474ccce97d3/crates/pessimistic-proof/src/local_state.rs#L266

crates/pessimistic-proof-core/src/aggchain_proof.rs
```diff
@@ -0,0 +1,103 @@
...
+            ]),
+        }
+    }
+}
+
+/// ECDSA variant of the aggchain proof.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct AggchainProofECDSAData {
+    /// Signer committing to the state transition.
+    pub signer: Address,
```
Freyskeyd, line 56: This address shouldn't be `recovered` from the signature and the `combined_hash`?

crates/pessimistic-proof-core/src/aggchain_proof.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+pub type Vkey = [u32; 8];
+pub type StarkVkey = Vec<u8>;
+pub type StarkProof = Vec<u8>;
+
+/// Aggchain Proof which is either one ECDSA signature, or one stark proof.
+/// Contains all the necessary data for verification.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum AggchainProofData {
```
Freyskeyd, line 28: Shouldn't it be just `AggchainProof` here?

crates/pessimistic-proof/src/aggchain_proof.rs
```diff
@@ -0,0 +1,36 @@
...
+pub enum AggchainProof {
+    ECDSA { signature: Signature },
+    SP1 { aggchain_proof: AggchainProofSP1 },
+}
+
+impl AggchainProof {
+    pub fn aggchain_type(&self) -> AggchainType {
+        match self {
+            AggchainProof::ECDSA { signature: _ } => AggchainType::ECDSA,
+            AggchainProof::SP1 { aggchain_proof: _ } => AggchainType::SP1,
```
Freyskeyd, line 21: This could be a `From` impl, but you can also update that like that:

```suggestion
            AggchainProof::ECDSA { .. } => AggchainType::ECDSA,
            AggchainProof::SP1 { .. } => AggchainType::SP1,
```

Freyskeyd (approved):

crates/pessimistic-proof-core/src/aggchain_proof.rs
```diff
@@ -0,0 +1,103 @@
...
+            self.prev_local_exit_root.as_slice(),
+            self.new_local_exit_root.as_slice(),
+            self.l1_info_root.as_slice(),
+            &self.origin_network.to_be_bytes(),
+            self.commit_imported_bridge_exits.as_slice(),
+            self.aggchain_params.as_slice(),
+        ]
+        .concat();
+
+        Sha256::digest(&public_values).into()
```
Freyskeyd, line 101: Just for my understanding, why is it `sha256` and not `keccak` here?

crates/pessimistic-proof-core/src/aggchain_proof.rs
hadjiszs: I wanted to have the `AggchainProof` name for the structure in the `Certificate` which includes only 
- if SP1 proof: `proof` + `aggchain_params`
- if ECDSA: `signature`


While `AggchainProofData` would be the entire aggchain proof information given to the prover, so:

- if SP1 proof: includes also the `vkey` retrieved from L1
- if ECDSA: includes also the `signer` retrieved from L1

crates/pessimistic-proof-core/src/aggchain_proof.rs
hadjiszs: Here it is the targeted `signer` retrieved from L1 during the witness generation. If the `recovered` signer from `combined_hash` isn't equal to this `signer`, then we want the prover to fail with the error `InvalidSigner { declared: Address, recovered: Address }`

crates/pessimistic-proof/src/aggchain_proof.rs
hadjiszs: I ended up deleted it in the end as it was not used

crates/pessimistic-proof-core/src/aggchain_proof.rs
hadjiszs: keccak is mainly used for the hashes which need to be recomputed on the solidity side, but here this hash is solely for the stark verification

crates/pessimistic-proof-core/src/aggchain_proof.rs
Freyskeyd: I would have preferred to have `AggchainProof` and `AggchainProofVerifier` that clearly defined what is part of the `aggchain-proof` and what is needed to verify it.
At some point if we're moving the `pessimistic-proof-core` and `program` to another repo, we'll maybe loose this context.

Another solution could be to have something like:

```rs
enum AggchainProof {
	ECDSA(..),
	SP1(..)
}

enum AggchainProofVerifier {
	ECDSA(..),
	SP1(..)
}

enum AggchainProofWithVerifier {
	proof: AggchainProof,
	verifier: AggchainProofVerifier
}
```

crates/pessimistic-proof/src/aggchain_proof.rs
```diff
@@ -0,0 +1,27 @@
...
+
+// Aggchain proof values submitted via the [`Certificate`].
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(untagged)]
+pub enum AggchainProof {
+    ECDSA { signature: Signature },
+    SP1 { aggchain_proof: AggchainProofSP1 },
+}
+
+// TODO: Replace with the proper format (fixed size buffer of ~7kb)
```
Ekleog-Polygon, line 17: Do we have a ticket open for this TODO (and the ones above)?

Ekleog-Polygon (commented):
Overall LGTM! Got a few comments below, pointing out things that look weird to me, but the base looks great to me, thank you!

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/program/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    assert_eq!(recovered_signer.as_slice(), aggchain_ecdsa.signer);
+
+    let aggchain_proof_inputs = bincode::DefaultOptions::new()
+        .with_big_endian()
+        .with_fixint_encoding()
+        .serialize(&aggchain_ecdsa.public_values())
+        .unwrap();
+
+    sp1_zkvm::io::commit_slice(&aggchain_proof_inputs);
```
iljakuklic, line 47: Why the custom serialization and `commit_slice` instead of plain `commit` which should use the default serialization format? I see `Serialize` is implemented for `AggchainProofPublicValues`.

crates/pessimistic-proof-core/src/aggchain_proof.rs
iljakuklic: > Another solution could be to have something like:
> 
> ```rust
> enum AggchainProof {
> 	ECDSA(..),
> 	SP1(..)
> }
> 
> enum AggchainProofVerifier {
> 	ECDSA(..),
> 	SP1(..)
> }
> 
> enum AggchainProofWithVerifier {
> 	proof: AggchainProof,
> 	verifier: AggchainProofVerifier
> }
> ```

Ideally, the types make nonsensical states impossible. The `AggchainProofWithVerifier` suggested above makes it possible to mix together ECDSA "challenge" and SP1 "witness data" and vice versa, whereas the current representation in the PR does not.

crates/pessimistic-proof-core/src/aggchain_proof.rs
hadjiszs: >Unresolving the conversation: I’d argue for using keccak still here

ah I don't think we can really choose here, the public values are expected to be hashed with sha256 I believe https://github.com/succinctlabs/sp1/blob/dev/crates/primitives/src/io.rs#L55-L79
https://github.com/succinctlabs/sp1/blob/dev/examples/aggregation/program/src/main.rs#L54-L55

>Wondering why is sha256 required / preferable in this case?

I'm not expert tbh, @Nashtare would know more

crates/pessimistic-proof-core/src/aggchain_proof.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub aggchain_params: Digest,
+}
+
+impl AggchainProofPublicValues {
+    pub fn hash(&self) -> [u8; 32] {
+        let public_values = [
+            self.prev_local_exit_root.as_slice(),
+            self.new_local_exit_root.as_slice(),
+            self.l1_info_root.as_slice(),
+            &self.origin_network.to_be_bytes(),
```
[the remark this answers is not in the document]
hadjiszs: Here there is no limitation given that this proof won't need to be verified on-chain, so we can be `be` or `le` as we wish, I certainly put `be` by simple copypaste from the PP public inputs (where it needs to be `be` because verified on-chain)

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s00_be000.snap
```diff
@@ -1,42 +1,41 @@
 ---
```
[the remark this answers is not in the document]
hadjiszs: @Nashtare might know more on that

crates/pessimistic-proof-test-suite/tests/e2e_local_pp.rs
```diff
@@ -105,22 +107,31 @@ fn test_sp1_simple() {
...
         .make_multi_batch_header(&certificate, forest.get_signer(), l1_info_root)
         .unwrap();
 
-    let initial_state = LocalNetworkState::from(initial_state);
+    // Set the aggchain proof to the sp1 variant
+    multi_batch_header.aggchain_proof = AggchainProofData::SP1(AggchainProofSP1Data {
+        aggchain_params: aggchain_params.into(),
+        aggchain_vkey: aggchain_vkey.hash_u32(),
+        stark_proof: Default::default(),
+    });
```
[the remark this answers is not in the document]
hadjiszs: I believe that chains with the simple ecdsa signature will continue to exist for some time if @Freyskeyd confirms, so we probably want to keep those tests for some time

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/script/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "sp1-ecdsa-verification-script"
+edition = "2021"
+
+[dependencies]
+sp1-sdk = "=4.0.0"
+alloy = { version = "0.6.4", features = ["signers", "signer-local"] }
+rand = "0.8.5"
+
+[build-dependencies]
+sp1-helper = { git = "https://github.com/succinctlabs/sp1.git", rev = "v1.0.0-testnet" }
```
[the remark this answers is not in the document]
hadjiszs: I removed the `script/` folder in the end, redundant with the test `test_sp1_simple`

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/script/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    let client = ProverClient::new();
+    let (pk, vk) = client.setup(ELF);
+    let proof = client
+        .prove(&pk, stdin)
+        .compressed()
+        .run()
+        .expect("proving failed");
+
+    dbg!("done proof");
```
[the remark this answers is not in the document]
hadjiszs: I removed the `script/` folder in the end, redundant with the test `test_sp1_simple`

crates/pessimistic-proof-test-suite/src/forest.rs
```diff
@@ -186,6 +206,57 @@ impl Forest {
...
+        let certificate = Certificate {
+            network_id: NETWORK_B,
+            height: 0,
+            prev_local_exit_root,
+            new_local_exit_root,
+            bridge_exits,
+            imported_bridge_exits: imported_bridge_exits.clone(),
+            signature,
+            metadata: Default::default(),
+        };
```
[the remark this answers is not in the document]
hadjiszs: Indeed, done 👍

crates/pessimistic-proof-core/src/local_state.rs
```diff
@@ -223,26 +226,58 @@ impl NetworkState {
...
+                    return Err(ProofError::InvalidSigner {
+                        declared: aggchain_proof_ecdsa.signer,
+                        recovered: signer,
+                    });
+                }
+            }
+            #[cfg(not(target_os = "zkvm"))]
+            AggchainProofData::SP1(_) => {
+                // NOTE: No stark verification in the native rust code due to
+                // the sp1_zkvm::lib::verify::verify_sp1_proof syscall
```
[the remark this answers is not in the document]
hadjiszs: added 👍

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/program/elf/riscv32im-succinct-zkvm-elf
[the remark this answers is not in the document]
hadjiszs: Hm I don't think that's necessary, I'd add a ci check if I see that I have to modify it again

crates/pessimistic-proof/src/aggchain_proof.rs
hadjiszs: In the scope of #540 I'd say

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/program/src/main.rs
iljakuklic: > That's because the public inputs are re-constructed on the smart contracts to feed the on-chain verifier, and solidity manipulates big-endian with this encoding

Makes sense. A code comment pointing that out would be useful.

crates/pessimistic-proof/src/aggchain_proof.rs
iljakuklic: > In the scope of #540 I'd say

Do you want me to include it in #597 or will you do it in your upcoming PR on top of it?

hadjiszs: @iljakuklic happy if you can add it to ensure proper storage compatibility

crates/pessimistic-proof-core/src/aggchain_proof.rs
```diff
@@ -92,7 +92,7 @@ impl AggchainProofPublicValues {
             self.prev_local_exit_root.as_slice(),
             self.new_local_exit_root.as_slice(),
             self.l1_info_root.as_slice(),
-            &self.origin_network.to_be_bytes(),
+            &self.origin_network.to_le_bytes(),
```
Ekleog-Polygon, line 95: Considering your comment on https://github.com/agglayer/agglayer/pull/448#discussion_r1950275561, are we sure we'll never want to verify the proof on-chain? If we ever want to, it might actually be better to keep big endian?

(My comment from before was mostly about "why do we use both BE and LE encodings, is there any reason?" — it wasn't requesting actual changes in this case)

Ekleog-Polygon (commented):
LGTM! Not approving yet, until we get @Nashtare's feedback on the few questions we don't know the answer of ourselves :)

crates/pessimistic-proof-core/src/aggchain_proof.rs
hadjiszs: > are we sure we'll never want to verify the proof on-chain?

yeah never especially given that this aggchain proof is a stark and not a snark, it won't be verified on-chain, only the pp (which verifies this aggchain proof) get verified on-chain

>(My comment from before was mostly about "why do we use both BE and LE encodings, is there any reason?" — it wasn't requesting actual changes in this case)

yeah I still changed it for a bit more consistency

crates/pessimistic-proof-core/src/aggchain_proof.rs
Nashtare: From a security standpoint, SHA3 (keccak) is technically subtly superior in some regards to SHA2 (SHA256), but that doesn't matter here.
I agree the point of @Ekleog-Polygon of unifying where possible / meaningful the hash function used (unless there are counter-reasons) makes sense, however from a performance standpoint, I'd expect SHA2 circuit to be more efficient than KECCAK, which may impair performance of the consensus proof. Maybe worth benchmarking the cycle count.

atanmarko (approved):

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s00_be000.snap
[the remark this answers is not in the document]
Ekleog-Polygon: @Nashtare Do you have any feedback on this too? The perf win here seems too good to be true considering we semantically just switched from verifying an ECDSA signature to verifying a proof, so I'm wondering whether there's something wrong here :/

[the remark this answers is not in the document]
Nashtare: Ah sorry I had missed this ping.
Against which version has it been tested to obtain this 10% perf win? A lot of things can explain the difference, SP1 recursion VM is pretty fast with SP1 Turbo, so that may come from this. I don't know the `sp1_verify` syscall though, can you point me to it?

Ekleog-Polygon (commented):
About the sha2/sha3 question, upon more thought we can likely land as-is for now, and keep investigating. We'll still be able to change that until we release it anyway :)

However, I do think that answering the question of the perf win before landing would be worth it: it might mean that we're not actually running some code that we should be running, which would be very wrong. My first intuition would be, we're not actually verifying the aggchain proof, because we're not seeing the sp1_verify syscall in the syscall list.

Maybe it's related to the way the cycle tracker works, as I think it might be executing outside of a proving system, though inside the SP1 VM?

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s00_be000.snap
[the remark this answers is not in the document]
iljakuklic: The certificates fed into the cycle tracker still use the ECDSA signature if I'm not mistaken. Or have I missed something?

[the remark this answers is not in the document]
hadjiszs: Regarding the `verify_sp1_proof` I dont know the performance of it, here is the syscall we're referring to @Nashtare : https://github.com/succinctlabs/sp1/blob/6a921fb256ce949c57bae7c83c49f8aaa6c35f3c/crates/zkvm/lib/src/lib.rs#L94

Actually the cycle tracker track solely the path where the `aggchain_proof` is the simple ecdsa signature. Then, one difference introduced is that to compute the `combined_hash` which is the hash which is supposed to be signed, we now take directly the `multi_batch_header.target.new_local_exit_root` instead of re-calling `self.exit_tree.get_root()`, so basically we save a root computation of the LET (so less of keccak), we keep the same guarantees given that the `target.new_local_exit_root` is already verified right after. That's where the -10% are probably coming from.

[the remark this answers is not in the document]
Ekleog-Polygon: Makes sense to me, thank you for the research! Approving :)

Ekleog-Polygon (approved):

crates/pessimistic-proof-core/src/aggchain_proof.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// SP1 variant of the aggchain proof.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct AggchainProofSP1Data {
+    /// Chain-specific commitment forwarded by the PP.
+    pub aggchain_params: Digest,
+    /// SP1 verifying key for the SP1 aggchain proof program.
+    pub aggchain_vkey: Vkey,
+    /// STARK proof.
+    pub stark_proof: StarkProof,
```
iljakuklic, line 69: It seems to me that this field does could be deleted. The proof is fed in separately using `stdin.write_proof(..)` and read implicitly when `verify_sp1_proof` is called. So it does not need to be included in the multi batch header.

crates/pessimistic-proof-test-suite/tests/e2e_local_pp.rs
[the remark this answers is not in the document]
Freyskeyd: yes chains with only ecdsa (signature) will still exists

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s00_be000.snap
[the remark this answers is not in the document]
Nashtare: Indeed makes sense @hadjiszs, Keccak is heavy so any removal can lead to nice savings like the observed one.
And thanks for the pointer.

crates/pessimistic-proof-core/src/aggchain_proof.rs
hadjiszs: yeah true 👍  only needed for the structure contained in the certificate

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/program/src/main.rs
hadjiszs: my bad I mixed up the program we were talking about, what I said is true for the pp program only, that's why we have this custom serialization in the pp program. However for the aggchain program, the public inputs are solely reconstructed within the pp-prover, so here I removed this custom serialization which isn't needed for the aggchain proof example

crates/pessimistic-proof-core/src/local_state.rs
```diff
@@ -223,26 +228,59 @@ impl NetworkState {
...
+                warn!("verify_sp1_proof is not callable outside of SP1");
```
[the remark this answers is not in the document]
hadjiszs: yeah here: #611

iljakuklic (approved):

PR #451, opened 2024-12-03
fix: backport discrepancy fix on bridge messages

hadjiszs:
# Description

the witness generation was creating the balance proof against the token info contained in the bridge messages (which is wrong), as opposed to within the PP where L1 ETH was properly considered for all bridge messages.

## PR Checklist:


iljakuklic (approved):
Makes me wonder if `BridgeExit` should be an enum with `Message` and `Transfer` arms, each containing only the relevant fields for given case. Would make these sort of issues less likely imho.

hadjiszs:
> Makes me wonder if `BridgeExit` should be an enum with `Message` and `Transfer` arms, each containing only the relevant fields for given case. Would make these sort of issues less likely imho.

propagated to #454 for a new rc given that this PR is associated to rc18 which is already under testing

PR #463, opened 2024-12-09
feat: include the LER leaf count in the pessimistic root

### LINKED ISSUE #248: [audit] Exit tree hash does not commit to the tree size
Severity: Medium
Difficulty: High
Type: Cryptography
Target: crates/pessimistic-proof/src/local_exit_tree/mod.rs, crates/pessimistic-proof/src/local_exit_tree/mod.rs

## Description
The local exit tree is represented by its root hash in the leaf proof output, shown below
https://github.com/agglayer/agglayer/blob/820180bb3f4a77369e72472b8e2847485b9551e2/crates/pessimistic-proof/src/proof.rs#L59-L72

While two trees with the same root must have the same sequence of leaves, a tree with the same leaves but a higher value of `leaf_count` will not have a different root. This allows an attacker with control of the signing key to choose the index at which the exits are added. By doing this, an attacker can fill an exit tree and create a denial of service that makes further transfers out of the exploited chain impossible.

This issue is Medium severity because it can only directly impact the value in a chain which already has its bridge key compromised, and cannot violate the overall `pessimistic-proof` invariants. However, by halting the bridge, it could lead to significant loss of funds.

## Exploit Scenario
Alice compromises the BobChain signing key and submits an SP1 proof of a multi-batch with a single exit starting from a `leaf_count` value of `2^32 - 2`. After this multi-batch, the BobChain local exit tree can no longer have new leaves added, and BobChain is unable to bridge any additional assets through AggLayer.

## Recommendations
Short term, ensure that the leaf count is included in the commitment to the local exit tree — e.g., include the leaf count in the root hash, or include the leaf count as an additional field of the leaf proof output.

Long term, ensure that all state variables of an authenticated data structure — i.e., any variable that can cause a different result of an update function — is included in the overall data structure commitment.

wborgeaud:
Include the LER leaf count in the pessimistic root to make sure it is committed. Fixes #248.

## PR Checklist:


hadjiszs:
I think we also need to update the value of `EMPTY_PP_ROOT` used [here](https://github.com/agglayer/agglayer/blob/wb-pessimistic-root-leaf-count/crates/pessimistic-proof/src/proof.rs#L219-L236)

wborgeaud:
> I think we also need to update the value of `EMPTY_PP_ROOT` used [here](https://github.com/agglayer/agglayer/blob/wb-pessimistic-root-leaf-count/crates/pessimistic-proof/src/proof.rs#L219-L236)

Right, good catch. It's fixed now.

hadjiszs (approved):

PR #491, opened 2024-12-19
chore: updating vkey

Freyskeyd:
# Description

This PR aims at changing the `vKey`.

Old one: `0x00e360ff2c18d456d6ee96c17f4e746216b475b62174bd2e9e8775ed99368ca3`
New one: `0x002d324867c82b5b5e0c66a822b004694cb73166d74c0713add59a282b8ce3e0`

## PR Checklist:


Ekleog-Polygon (commented):
AFAIU this is committing the result of [this xtask command](https://github.com/agglayer/agglayer/blob/a4566a7f74ce5fdbb24e6bc63538285b3f33484d/xtask/src/commands.rs#L9).

Can we, instead of committing it, add it to the build.rs so that it's automatically rebuilt upon each `cargo build`? Or is there a reason why we wouldn't want to auto-rebuild? The three I could think of are:
- it way too slow to actually rebuild,
- we want people without the succinct toolchain to be able to run our builds, or
- the build is non-deterministic and we otherwise depend on the hash of the verifying key being stable.

Freyskeyd:
> AFAIU this is committing the result of [this xtask command](https://github.com/agglayer/agglayer/blob/a4566a7f74ce5fdbb24e6bc63538285b3f33484d/xtask/src/commands.rs#L9).
> Can we, instead of committing it, add it to the build.rs so that it's automatically rebuilt upon each `cargo build`? Or is there a reason why we wouldn't want to auto-rebuild? The three I could think of are:

The goal is to not have the pessimistic-proof-program as part of the `crates` to avoid having to use the `succinct` toolchain everywhere.

> * we want people without the succinct toolchain to be able to run our builds, or

Yes, it is part of why we want to have this committed

> * the build is non-deterministic and we otherwise depend on the hash of the verifying key being stable.

The build is currently altered by PATH, we have a task to make the ELF generated from docker which will solve a lot of this.
There is already a ticket to address that, and @iljakuklic is working on it.

If you feel that there is some improvement to be made on this, please, feel free to raise any concern in a ticket 🙏🏼

Ekleog-Polygon (dismissed):
Makes sense! And considering the build is not deterministic yet, we cannot yet add a test in CI that'd enforce we remember to update the file when modifying the pessimistic proof 

Hopefully once builds are deterministic we can add this check and remove this easy-to-forget chore!

r+ing as AFAIU there's not much else I can do to validate further, due to nondeterministic builds

Freyskeyd:
> And considering the build is not deterministic yet, we cannot yet add a test in CI that'd enforce we remember to update the file when modifying the pessimistic proof

We had that a while ago, and that's how we discover that the build wasn't deterministic, ahah
 

> r+ing as AFAIU there's not much else I can do to validate further, due to nondeterministic builds

This PR will certainly not be merged as is, it is mainly here to trigger a `vKey` change so that we can test things :)

PR #493, opened 2024-12-20
feat: add version info

atanmarko:
# Description

Adds detailed version info cli command and prints version info on startup of node and prover.

It looks like this:

```
$ cargo run -- --version  
$ cargo run -- -V

agglayer agglayer (v0.2.0-rc.20-20-g7c3e48d-dirty) [git commit timestamp: 2025-01-08T18:13:56.000000000Z]

```

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/lib.rs
+++ b/crates/agglayer-node/src/lib.rs
@@ -27,7 +27,7 @@ use agglayer_telemetry::ServerBuilder as MetricsBuilder;
 ///
 /// This function returns on fatal error or after graceful shutdown has
 /// completed.
-pub fn main(cfg: PathBuf) -> Result<()> {
+pub fn main(cfg: PathBuf, version: &str) -> Result<()> {
     let cfg = cfg.canonicalize().map_err(|_| {
         anyhow::Error::msg(format!(
             "Configuration file path must be absolute, given: {}",
@@ -49,6 +49,8 @@ pub fn main(cfg: PathBuf) -> Result<()> {
 
     // Initialize the logger
     logging::tracing(&config.log);
+
+    info!("Starting agglayer node version info: {}", version);
 
     let node_runtime = tokio::runtime::Builder::new_multi_thread()
         .thread_name("agglayer-node-runtime")
--- a/crates/agglayer/src/cli.rs
+++ b/crates/agglayer/src/cli.rs
@@ -3,8 +3,12 @@ use std::path::PathBuf;
 
 use clap::{Parser, Subcommand, ValueHint};
 
+use crate::version;
+
 /// Agglayer command line interface.
 #[derive(Parser)]
+#[command(version = version())]
+#[command(propagate_version = true)]
 pub(crate) struct Cli {
     #[command(subcommand)]
     pub(crate) cmd: Commands,
@@ -48,6 +52,4 @@ pub(crate) enum Commands {
     },
 
     Vkey,
-
-    Version,
 }
--- a/crates/agglayer/src/cli.rs
+++ b/crates/agglayer/src/cli.rs
@@ -48,4 +48,6 @@ pub(crate) enum Commands {
     },
 
     Vkey,
+
+    Version,
 }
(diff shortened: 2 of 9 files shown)


crates/agglayer-node/src/lib.rs
```diff
@@ -27,7 +27,7 @@ use agglayer_telemetry::ServerBuilder as MetricsBuilder;
 ///
 /// This function returns on fatal error or after graceful shutdown has
 /// completed.
-pub fn main(cfg: PathBuf) -> Result<()> {
+pub fn main(cfg: PathBuf, version: &str) -> Result<()> {
```
atanmarko, line 30: Note: to read cargo env variables `build.rs` has to be executed for the particular crate. Felt like overkill to make separate `build.rs` for `agglayer`,  `agglayer-prover` and `agglayer-node`, so I chose to pass the global `agglayer` package version info instead.

crates/agglayer/src/main.rs
```diff
@@ -31,7 +31,18 @@ fn main() -> anyhow::Result<()> {
...
     }
 
     Ok(())
 }
+
+/// Common version information about the executed agglayer binary.
+pub fn version() -> String {
+    let pkg_name = env!("CARGO_PKG_NAME");
+    let git_describe = env!("VERGEN_GIT_DESCRIBE");
+    let timestamp = env!("VERGEN_BUILD_TIMESTAMP");
```
iljakuklic, line 46: How is the timestamp obtained? It seems a bit odd if just running `cargo build` some time later with no other changes results in a different binary. Imho that sort of non-determinism should be limited. If it actually takes the current time, i's advise against it. On the other hand, taking the timestamp from the latest commit is fine.

atanmarko: It takes the timestamp from the local system in the moment of the build. You can uniquely identify the commit with the commit hash available with the `VERGEN_GIT_DESCRIBE` and that does not change in the other build of the same commit.
I don't think the time of the build hurts, it is clearly marked as the build time information - nothing more, nothing less.

atanmarko: Not to digress too far, even `Git Release process` has its own date/time when the binaries/artifacts are created/packed and uploaded on the release page, not the time of the latest commit that was released.

iljakuklic: > It takes the timestamp from the local system in the moment of the build. You can uniquely identify the commit with the commit hash available with the `VERGEN_GIT_DESCRIBE` and that does not change in the other build of the same commit. I don't think the time of the build hurts, it is clearly marked as the build time information - nothing more, nothing less.

There are some drawbacks in builds not being deterministic. For example, it makes it more difficult for third parties to check the result of the build by reproducing it with the same build inputs. Consider building a Docker image with agglayer node in it. It can uniquely identified by a hash. However, we will get different resulting image hash even if the exact same sources are used to build the node.

More info [here](https://reproducible-builds.org/). Vergen has a section on deterministic outputs [here](https://docs.rs/vergen/9.0.2/vergen/#support-deterministic-output).

iljakuklic: > Having the very same agglayer source commit built on the two different machines - you will certainly have different agglayer file hash - and docker image hash built at different time with the same commit will certainly be different. It will depend on many different parameters - file system creation times, compiler version etc. A little experiment - take the current `origin/main`, build the `agglayer` binary and check the `md5sum` of it. Then `cargo clean`, change the compiler version and repeat the same. They will have different md5 hash.

Yea, the output depends on many variables, including the directory path where the build happens etc, so it's not very practical at the moment. However, I'd say it's preferable if we directionally move towards more determinism, or at least not introduce much more of it.

iljakuklic (approved):

iljakuklic (dismissed):

crates/agglayer/src/cli.rs
```diff
@@ -48,4 +48,6 @@ pub(crate) enum Commands {
     },
 
     Vkey,
+
+    Version,
```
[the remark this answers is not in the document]
atanmarko: Sure, why not.

iljakuklic (approved):

Freyskeyd (approved):

PR #496, opened 2024-12-22
fix: ppgen PPGenArgs default value failure fix

BrianSeong99:
# Description

The initial value set for `n_imported_exits` in `PPGenArgs` is 0, which the chain doesnt have tokens at the beginning to bridge out which will cause failure when building `multi_batch_header`.

### Bug fix (non-breaking change which fixes an issue)

so we need to change the `n_imported_exits` default value to some other numbers (10 worked in this case) so that n_exits can be executed.

## PR Checklist:

### DIFF
--- a/crates/pessimistic-proof-test-suite/src/bin/ppgen.rs
+++ b/crates/pessimistic-proof-test-suite/src/bin/ppgen.rs
@@ -25,7 +25,7 @@ struct PPGenArgs {
     n_exits: usize,
 
     /// The number of imported bridge exits.
-    #[clap(long, default_value = "0")]
+    #[clap(long, default_value = "10")]
     n_imported_exits: usize,
 
     /// The optional output directory to write the proofs in JSON. If not set,


hadjiszs (approved):
Ideally we have `n_imported_exits` with the default value equal to the `n_exits` value (we just need `n_imported_exits >= n_exits`)

Now if `n_exits` is set to 11 and `n_imported_exits` omitted (and so defaults to `10`), then it will fail

Freyskeyd:
@BrianSeong99 your commit need to be signed to contribute.

crates/pessimistic-proof-test-suite/src/bin/ppgen.rs
```diff
@@ -25,7 +25,7 @@ struct PPGenArgs {
     n_exits: usize,
 
     /// The number of imported bridge exits.
-    #[clap(long, default_value = "0")]
+    #[clap(long, default_value = "10")]
```
iljakuklic, line 28: I'm fine with this. An alternative would be to remove the default so the user has to specify the exit count explicitly. While it's not much of an issue for this tool, the default of 10 feels a bit arbitrary.

iljakuklic (approved):

Ekleog-Polygon (commented):
I just approved the workflow runs based on the two previous accepting reviews. Not reviewing myself as review seems already complete :)

iljakuklic (approved):

PR #499, opened 2024-12-23
feat: setup otel tracing

### LINKED ISSUE #498: Setup agglayer open telemetry
Add optional open telemetry initialization to existing log setup.

atanmarko:
# Description

Consolidate log initialization.
Implement optional `otlp` tracing setup.

Fixes #498 

## PR Checklist:


crates/agglayer-telemetry/Cargo.toml
```diff
@@ -7,15 +7,21 @@ edition.workspace = true
...
-opentelemetry_sdk = { version = "0.27.1", features = ["metrics"] }
+opentelemetry_sdk = { version = "0.27.1", features = ["metrics", "rt-tokio"] }
 prometheus = "0.13.3"
 thiserror.workspace = true
 tokio = { workspace = true, features = ["full"] }
 tokio-util = { workspace = true }
 tracing.workspace = true
+tracing-opentelemetry = "0.28"
+tracing-subscriber = { workspace = true, features = ["env-filter", "json"] }
+
```
Freyskeyd:
```suggestion
```

crates/agglayer-config/src/tracing.rs
```diff
@@ -85,22 +94,25 @@ impl<'de> Deserialize<'de> for LogOutput {
         // If the string is not recognized to be either `stdout` or `stderr`,
         // it is assumed to be a file path.
         match s.as_str() {
-            "stdout" => Ok(LogOutput::Stdout),
-            "stderr" => Ok(LogOutput::Stderr),
-            _ => Ok(LogOutput::File(PathBuf::from(s))),
+            "stdout" => Ok(TracingOutput::Stdout),
+            "stderr" => Ok(TracingOutput::Stderr),
+            "otlp" => Ok(TracingOutput::Otlp),
+            _ => Ok(TracingOutput::File(PathBuf::from(s))),
```
Ekleog-Polygon, line 100: This change means that if we previously had people logging to an `otlp`-named file, they'll be silently moved to an otlp tracing output.

Could we maybe force `File` to eg. end in `.log`, so that we can add other variants in the future without any such fear?

crates/agglayer-config/src/tracing.rs
```diff
@@ -85,22 +94,25 @@ impl<'de> Deserialize<'de> for LogOutput {
...
+    /// This can be used to plug the tracing output into the tracing subscriber.
     pub fn as_make_writer(&self) -> BoxMakeWriter {
         match self {
-            LogOutput::Stdout => BoxMakeWriter::new(std::io::stdout),
-            LogOutput::Stderr => BoxMakeWriter::new(std::io::stderr),
-            LogOutput::File(path) => {
+            TracingOutput::Stdout => BoxMakeWriter::new(std::io::stdout),
+            TracingOutput::Stderr => BoxMakeWriter::new(std::io::stderr),
+            // For OTLP, output traces also to stdout
+            TracingOutput::Otlp => BoxMakeWriter::new(std::io::stdout),
```
Ekleog-Polygon: @atanmarko you reacted to the bot with a thumbs up, but then marked the conversation as resolved without actually making the changes, did you forget to push?

crates/agglayer-node/src/lib.rs
```diff
@@ -48,8 +47,8 @@ pub fn main(
         bail!("Received cancellation signal before starting the node.");
     }
 
-    // Initialize the logger
-    logging::tracing(&config.log);
+    // // Initialize the logger
+    // logging::tracing(&config.log);
```
Ekleog-Polygon: I'm not sure I get why this got commented out, if it's replaced by the metrics_runtime.block_on call below it should just be deleted?

crates/agglayer-telemetry/Cargo.toml
```diff
@@ -7,15 +7,20 @@ edition.workspace = true
...
 lazy_static.workspace = true
-opentelemetry = { version = "0.27.1", features = ["metrics"] }
-opentelemetry-prometheus = "0.27.0"
-opentelemetry_sdk = { version = "0.27.1", features = ["metrics"] }
-prometheus = "0.13.3"
+opentelemetry = { version = "0.29.1", features = ["metrics"] }
+opentelemetry-otlp = { version = "0.29.0", features = ["trace", "grpc-tonic"]}
+opentelemetry-prometheus = "0.29.1"
+opentelemetry_sdk = { version = "0.29.0", features = ["metrics", "rt-tokio", "trace"] }
+prometheus = "0.14.0"
```
Ekleog-Polygon: Can we add all these dependencies to the workspace `Cargo.toml`? We have different versions in a lot of places, and even if we don't think we'll ever have these deps outside of agglayer-telemetry, having them in the workspace doesn't cost us anything and allows to have a single file to view at a glance which dependency versions we use throughout the whole codebase

crates/agglayer-node/src/lib.rs
```diff
@@ -5,7 +5,6 @@ use anyhow::{bail, Result};
 use node::Node;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info};
-mod logging;
```
Ekleog-Polygon, line 8: Delete the associated module file

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_sdk::{
+    propagation::TraceContextPropagator,
+    trace::{BatchConfigBuilder, BatchSpanProcessor, Sampler, SpanLimits},
+    Resource,
+};
+use tracing_subscriber::{prelude::*, util::SubscriberInitExt, EnvFilter};
+
+pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) -> anyhow::Result<()> {
+    let writer = config.outputs.first().cloned().unwrap_or_default();
```
Ekleog-Polygon: I'm not sure I understand, we seem to be configuring many outputs but only actually using the first one and checking if otlp is in? If my understanding is correct, can we fail this if we have unexpected outputs, eg. two file outputs?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let mut layers = Vec::new();
+
+    // Setup instrumentation if both otlp agent url and
+    // otlp service name are provided as arguments
+    if config.outputs.contains(&TracingOutput::Otlp) {
+        if let (Some(otlp_agent), Some(otlp_service_name)) =
```
Ekleog-Polygon: Nit: as you're having a huge if just for a very small else, the code would be easier to read like:
```rust
let (Some(a), Some(b)) = (aa, bb) else {
    anyhow::bail!("foo");
}
```

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+            let resources = build_resources(otlp_service_name, version);
+            let otlp_exporter = opentelemetry_otlp::SpanExporter::builder()
+                .with_tonic()
+                .with_endpoint(otlp_agent)
+                .build()?;
+
+            let batch_processor_config = BatchConfigBuilder::default()
+                .with_scheduled_delay(match std::env::var("OTLP_BATCH_SCHEDULED_DELAY") {
+                    Ok(v) => Duration::from_millis(v.parse::<u64>().unwrap_or(5_000)),
+                    _ => Duration::from_millis(5_000),
```
Ekleog-Polygon: Nit: move these values to constants at the top of the file. Also, are the default values from the opentelemetry crates not good enough, for us to provide our own additional defaults?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+                    _ => 2048,
+                })
+                .with_max_export_batch_size(
+                    match std::env::var("OTLP_BATCH_MAX_EXPORTER_BATCH_SIZE") {
+                        Ok(v) => v.parse::<usize>().unwrap_or(512),
+                        _ => 512,
+                    },
+                );
+
+            let span_limits_default = SpanLimits::default();
```
Ekleog-Polygon: IIUC this is used to set the defaults to the same value as they would have been by default, if there's no env var. Can we just not call `.with_max_events_per_span` and co if there's no env var, and avoid that complexity? Something like this:
```rust
if let Ok(Ok(foo)) = env::var("FOO").map(u64::from_str) {
    builder.with_foo(foo);
}
```

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+                        .unwrap_or(span_limits_default.max_events_per_span),
+                    _ => span_limits_default.max_events_per_span,
+                })
+                .with_max_attributes_per_span(match std::env::var("OTLP_MAX_ATTRIBUTES_PER_SPAN") {
+                    Ok(v) => v
+                        .parse::<u32>()
+                        .unwrap_or(span_limits_default.max_attributes_per_span),
+                    _ => span_limits_default.max_attributes_per_span,
+                })
+                .with_max_links_per_span(match std::env::var("OTLP_MAX_LINK_PER_SPAN") {
```
Ekleog-Polygon:
```suggestion
                .with_max_links_per_span(match std::env::var("OTLP_MAX_LINKS_PER_SPAN") {
```

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+
+            let tracer = trace_provider.tracer("agglayer-otlp");
+
+            let _ = global::set_tracer_provider(trace_provider);
+
+            layers.push(
+                tracing_opentelemetry::layer()
+                    .with_tracer(tracer)
+                    .with_filter(
+                        EnvFilter::try_from_default_env().unwrap_or_else(|_| config.level.into()),
```
Ekleog-Polygon, line 131: Hmmm I'm not sure we actually want the same filter as for log metrics? For OTLP traces I'd expect most of the time we'd want to always trace everything, and just not keep a big retention; whereas logs would usually be much less enabled

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+                    )
+                    .boxed(),
+            );
+
+            global::set_text_map_propagator(TraceContextPropagator::new());
```
Ekleog-Polygon, line 136: Do we really need this? It should be the default, is it not?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        TracingFormat::Json => tracing_subscriber::fmt::layer()
+            .json()
+            .with_writer(writer.as_make_writer())
+            .with_filter(EnvFilter::try_from_default_env().unwrap_or_else(|_| config.level.into()))
+            .boxed(),
+    });
+
+    // We are using try_init because integration test may try
+    // to initialize this multiple times.
```
Ekleog-Polygon, line 159: Can we at least log a warning message here with `eprintln`, so that if we do break tracing we still have some feedback?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+fn build_resources(otlp_service_name: &str, version: &str) -> Vec<KeyValue> {
+    let mut resources = Vec::new();
+
+    resources.push(KeyValue::new("service.name", otlp_service_name.to_string()));
+    resources.push(KeyValue::new("service.version", version.to_string()));
+
+    let custom_resources: Vec<_> = std::env::var("AGGLAYER_OTLP_TAGS")
+        .unwrap_or_default()
+        .split(',')
+        // NOTE: limit to 10 tags to avoid exploit
```
Ekleog-Polygon: What exploit are you speaking about? This is an env var, so it should be admin-controlled anyway

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,162 @@
...
+        .filter_map(|tag_raw| {
+            let mut v = tag_raw.splitn(2, '=');
+            match (v.next(), v.next()) {
+                (Some(key), Some(value)) if !key.trim().is_empty() && !value.trim().is_empty() => {
+                    Some(KeyValue::new(
+                        key.trim().to_string(),
+                        value.trim().to_string(),
+                    ))
+                }
+                _ => None,
```
Ekleog-Polygon: Can we log some message if it's ill-configured? Same, can be with eprintln, we just need some feedback that config is wrong

Ekleog-Polygon (commented):
Very happy to see we're hopefully soon getting traces out of our system! I still have a few comments on implementation below

crates/agglayer-config/src/tracing.rs
atanmarko: No, I have added the comment, as suggested by the bot :)

crates/agglayer-config/src/tracing.rs
atanmarko: @Ekleog-Polygon I think probability for this issue is extremely small. I would say forcing somebody to have particular extensions for log files is bigger hurdle than the possibility that he would use `otlp` for the file name. I don't think we would have another output type for the traces beside the existing ones.

crates/agglayer-node/src/lib.rs
atanmarko: Will remove it, it is replaced by `agglayer_telemetry::traces::setup_tracing`. My idea in the near future is to move tracing/telemetry crates to `interop`, so there will be more refactors.

crates/agglayer-telemetry/src/traces.rs
atanmarko: I guess the initial idea is to have for example both file and stdout outputs. As the config option for this parameter is already an array, I will refactor this a bit to fix dillema.

atanmarko: I have refactored this function. Now you would explicitly set  `[stdout, otlp]` or even `[stdout, otlp, logfile.log]` as the output types in the config

crates/agglayer-telemetry/src/traces.rs
atanmarko: I think this is debatable. I'll leave it like here, we can change if there is a need (or even introduce separate filter for otlp).

crates/agglayer-telemetry/src/traces.rs
atanmarko: I think it is [useful](https://docs.rs/opentelemetry/latest/opentelemetry/propagation/text_map_propagator/trait.TextMapPropagator.html) if we combine for example agglayer, aggkit and aggkit-prover traces to be shown in the context of the same request. Even if we don't use this possibility, I don't think it hurts.

crates/agglayer-telemetry/src/traces.rs
atanmarko: Some pervert attack with zillion tags from environment :)

atanmarko: I'll remove it :)

crates/agglayer-config/src/tracing.rs
Ekleog-Polygon: My thinking was it's better to fail loud 99% of the time than silently fail 1% of the time, but it's not a strong comment — as you say it's unlikely enough, and requiring a `.log` extension would only be useful for future-proofing

Cargo.toml
```diff
@@ -96,6 +96,11 @@ insta = { git = "https://github.com/freyskeyd/insta", branch = "chore/updating-d
 jsonrpsee = { version = "0.24.7", features = ["full"] }
 lazy_static = "1.5"
 mockall = "0.13.1"
+opentelemetry = { version = "0.29.1", features = ["metrics"] }
+opentelemetry-otlp = { version = "0.29.0", features = ["trace", "grpc-tonic"]}
+opentelemetry-prometheus = "0.29.1"
+opentelemetry_sdk = { version = "0.29.0", features = ["metrics", "rt-tokio", "trace"] }
+prometheus = "0.14.0"
```
Ekleog-Polygon: Nit: this is not in alphabetical order

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -10,17 +11,25 @@ use opentelemetry_sdk::{
...
+    for writer in &config.outputs {
+        // Setup instrumentation if both otlp agent url and
+        // otlp service name are provided as arguments
+        if writer == &TracingOutput::Otlp {
+            let (Some(otlp_agent), Some(otlp_service_name)) =
+                (&config.otlp_agent, &config.otlp_service_name)
+            else {
+                anyhow::bail!(
+                    "Otlp tracing requires both otlp agent url and otlp service provided"
+                );
```
Ekleog-Polygon, line 26: Actually, looking at it again it could also be done like — not sure which one would look best, that said, as it'd depend on the error message length I think:
```
let otlp_agent = config.otlp_agent.ok_or_else(|| anyhow!("blah"))?;
let otlp_service_name = [same];
```

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -29,56 +38,67 @@ pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) ->
...
-                    Ok(v) => Duration::from_millis(v.parse::<u64>().unwrap_or(5_000)),
-                    _ => Duration::from_millis(5_000),
+                    Ok(v) => {
+                        if let Ok(millis) = v.parse::<u64>() {
+                            Duration::from_millis(millis)
+                        } else {
+                            OTLP_BATCH_SCHEDULED_DELAY
+                        }
+                    }
+                    _ => OTLP_BATCH_SCHEDULED_DELAY,
```
Ekleog-Polygon: Is `OTLP_BATCH_SCHEDULED_DELAY` different from the default? If it's the same as the default, can we do like other settings and just only set it if it's properly set?

Also, we should probably log an error if the env var is set but doesn't parse correctly, eg. if people try to set `OTLP_BATCH_SCHEDULED_DELAY=10min` or similar

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -29,56 +38,67 @@ pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) ->
...
             let batch_processor_config = BatchConfigBuilder::default()
                 .with_scheduled_delay(match std::env::var("OTLP_BATCH_SCHEDULED_DELAY") {
-                    Ok(v) => Duration::from_millis(v.parse::<u64>().unwrap_or(5_000)),
-                    _ => Duration::from_millis(5_000),
+                    Ok(v) => {
+                        if let Ok(millis) = v.parse::<u64>() {
+                            Duration::from_millis(millis)
+                        } else {
+                            OTLP_BATCH_SCHEDULED_DELAY
+                        }
```
Ekleog-Polygon: Why change the previous `.unwrap_or`?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -29,56 +38,67 @@ pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) ->
...
+                    }
+                    _ => OTLP_BATCH_SCHEDULED_DELAY,
                 })
                 .with_max_queue_size(match std::env::var("OTLP_BATCH_MAX_QUEUE_SIZE") {
-                    Ok(v) => v.parse::<usize>().unwrap_or(2048),
-                    _ => 2048,
+                    Ok(v) => v.parse::<usize>().unwrap_or(OTLP_BATCH_MAX_QUEUE_SIZE),
+                    _ => OTLP_BATCH_MAX_QUEUE_SIZE,
```
Ekleog-Polygon: Same here, can we just use upstream's default values, or would 2048 be significantly better?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -29,56 +38,67 @@ pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) ->
...
                 })
                 .with_max_export_batch_size(
                     match std::env::var("OTLP_BATCH_MAX_EXPORTER_BATCH_SIZE") {
-                        Ok(v) => v.parse::<usize>().unwrap_or(512),
-                        _ => 512,
+                        Ok(v) => v
+                            .parse::<usize>()
+                            .unwrap_or(OTLP_BATCH_MAX_EXPORTER_BATCH_SIZE),
+                        _ => OTLP_BATCH_MAX_EXPORTER_BATCH_SIZE,
```
Ekleog-Polygon: And same here

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -29,56 +38,67 @@ pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) ->
...
+                    if let Ok(value) = max_attributes_per_event.parse::<u32>() {
+                        span_limits.max_attributes_per_event = value;
+                    }
+                }
+
+                if let Ok(max_attributes_per_link) = std::env::var("OTLP_MAX_ATTRIBUTES_PER_LINK") {
+                    if let Ok(value) = max_attributes_per_link.parse::<u32>() {
+                        span_limits.max_attributes_per_link = value;
+                    }
+                }
```
Ekleog-Polygon, line 108: In all these checks, log an error if the env var exists but doesn't parse correctly

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -29,56 +38,67 @@ pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) ->
...
+                span_limits
+            };
+
+            // Ensure that the span limits are not too low
```
Ekleog-Polygon: I'm not sure I get this comment, the code above is just making the default span limits configurable, not actually ensuring that the span limits are not too low?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -151,7 +174,12 @@ fn build_resources(otlp_service_name: &str, version: &str) -> Vec<KeyValue> {
                         value.trim().to_string(),
                     ))
                 }
-                _ => None,
+                _ => {
+                    eprint!(
```
Ekleog-Polygon: Nit: `eprintln` for readability

Ekleog-Polygon (commented):
Thank you for the changes! I just have a few last comments 😄

crates/agglayer-telemetry/src/traces.rs
atanmarko: I'll keep the current one.

crates/agglayer-telemetry/src/traces.rs
atanmarko: I'll remove comment.

Ekleog-Polygon (approved):
LGTM! I'm just thinking the code might look slightly better with eg. a tiny macro to do the `if env::var("") { if foo.parse { ... } else { eprintln } }`code that is repeated 8 times; but it's definitely not a blocker 😄 

Something that'd look like this:
```rust
if_env_var!(
    OTLP_BATCH_SCHEDULED_DELAY,
    |m: u64| batch_processor_config = batch_processor_config.with_scheduled_delay(Duration::from_millis(m)),
);
```

crates/agglayer-telemetry/src/traces.rs
Freyskeyd: I think that if we're not filtering properly, we could end up flooding the collector with a lot of traces related to lib (such as h2 or tonic), but that's something that we try

crates/agglayer-config/src/tracing.rs
```diff
@@ -3,24 +3,32 @@ use std::{fmt::Display, path::PathBuf};
 use serde::{Deserialize, Deserializer, Serialize};
 use tracing_subscriber::{fmt::writer::BoxMakeWriter, EnvFilter};
 
-/// The log configuration.
+/// The tracing configuration.
 #[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq)]
 #[serde(rename_all = "kebab-case")]
-pub struct Log {
+pub struct TracingConfig {
```
Freyskeyd, line 9: Are you renaming this because we're using `tracing` lib?

crates/agglayer-config/src/tracing.rs
```diff
@@ -3,24 +3,32 @@ use std::{fmt::Display, path::PathBuf};
...
+    /// configuration tracing level.
     #[serde(default)]
-    pub level: LogLevel,
+    pub level: TracingLevel,
     #[serde(default)]
-    pub outputs: Vec<LogOutput>,
+    pub outputs: Vec<TracingOutput>,
     #[serde(default)]
-    pub format: LogFormat,
+    pub format: TracingFormat,
```
Freyskeyd, line 17: This format represents the format of the `logs` exposed by the agglayer-node, I don't understand why we rename this to `TracingFormat`

crates/agglayer-config/src/tracing.rs
```diff
@@ -69,14 +77,15 @@ impl From<LogLevel> for EnvFilter {
 /// appropriate enum variant. If the string is not recognized to be either
 /// `stdout` or `stderr`, it is assumed to be a file path.
 #[derive(Serialize, Debug, Clone, Default, PartialEq, Eq)]
-pub enum LogOutput {
+pub enum TracingOutput {
     #[default]
     Stdout,
     Stderr,
     File(PathBuf),
+    Otlp,
```
Freyskeyd, line 85: I'm not sure `Otlp` is an output format.
It means that if we're defining `Otlp` here, we do not have a way to define the `log` output?

crates/agglayer-config/src/tracing.rs
```diff
@@ -3,24 +3,32 @@ use std::{fmt::Display, path::PathBuf};
...
-    pub format: LogFormat,
+    pub format: TracingFormat,
+    /// Socket of the open telemetry agent endpoint.
+    /// If not provided open telemetry will not be used.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub otlp_agent: Option<String>,
+    /// Otlp service name.
+    /// If not provided open telemetry will not be used.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub otlp_service_name: Option<String>,
```
Freyskeyd, line 25: This should maybe be placed in a `TraceConfig` struct, avoiding modifying everything for the logs and scoping everything there

crates/agglayer-node/src/lib.rs
```diff
@@ -64,6 +60,10 @@ pub fn main(
         .enable_all()
         .build()?;
 
+    // Initialize the tracing
+    metrics_runtime
+        .block_on(async { agglayer_telemetry::traces::setup_tracing(&config.log, version) })?;
```
Freyskeyd, line 65: This isn't async, thus you can execute the `setup_tracing` at the same place as `logging::tracing`

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,201 @@
...
+};
+use tracing_subscriber::{prelude::*, util::SubscriberInitExt, EnvFilter};
+
+pub fn setup_tracing(config: &agglayer_config::TracingConfig, version: &str) -> anyhow::Result<()> {
+    let mut layers = Vec::new();
+
+    for writer in &config.outputs {
+        // Setup instrumentation if both otlp agent url and
+        // otlp service name are provided as arguments
+        if writer == &TracingOutput::Otlp {
```
Freyskeyd, line 20: Maybe we should move the `otlp` configuration out of this loop, or even have a dedicated `TracesConfig` struct that have a method that build and export the layer directly

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,201 @@
...
+                .with_span_limits(span_limits)
+                .with_resource(Resource::builder().with_attributes(resources).build())
+                .with_span_processor(
+                    BatchSpanProcessor::builder(otlp_exporter)
+                        .with_batch_config(batch_processor_config.build())
+                        .build(),
+                )
+                .build();
+
+            let tracer = trace_provider.tracer("agglayer-otlp");
```
Freyskeyd, line 123: Could this be a `const` ?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,201 @@
...
+
+            let _ = global::set_tracer_provider(trace_provider);
+
+            layers.push(
+                tracing_opentelemetry::layer()
+                    .with_tracer(tracer)
+                    .with_filter(
+                        EnvFilter::try_from_default_env().unwrap_or_else(|_| config.level.into()),
```
Freyskeyd, line 131: Maybe we can have a method on `LogLevel` crate for this ?

crates/agglayer-telemetry/src/traces.rs
```diff
@@ -0,0 +1,201 @@
...
+    Ok(())
+}
+
+fn build_resources(otlp_service_name: &str, version: &str) -> Vec<KeyValue> {
+    let mut resources = Vec::new();
+
+    resources.push(KeyValue::new("service.name", otlp_service_name.to_string()));
+    resources.push(KeyValue::new("service.version", version.to_string()));
+
+    let custom_resources: Vec<_> = std::env::var("AGGLAYER_OTLP_TAGS")
```
Freyskeyd, line 176: Should it be scoped as `AGGLAYER_NODE_OTLP_TAGS` ?
Do we want those tags to be shared among other software?

Cargo.toml
```diff
@@ -96,8 +96,13 @@ insta = { git = "https://github.com/freyskeyd/insta", branch = "chore/updating-d
 jsonrpsee = { version = "0.24.7", features = ["full"] }
 lazy_static = "1.5"
 mockall = "0.13.1"
+opentelemetry = { version = "0.29.1", features = ["metrics"] }
+opentelemetry-otlp = { version = "0.29.0", features = ["trace", "grpc-tonic"]}
+opentelemetry-prometheus = "0.29.1"
```
iljakuklic, line 101: The `opentelemetry-prometheus` crate has not updated for quite some time. The [crate description](https://crates.io/crates/opentelemetry-prometheus) states:

> Warning: This crate is no longer recommended for use.
>
> Development of the Prometheus exporter has been discontinued. See the related [issue](https://github.com/open-telemetry/opentelemetry-rust/issues/2451). This crate depends on the unmaintained protobuf crate and has unresolved security vulnerabilities. Version 0.29 will be the final release.
>
> For Prometheus integration, we strongly recommend using the [OTLP](https://docs.rs/opentelemetry-otlp/latest/opentelemetry_otlp/) exporter instead. Prometheus [natively supports OTLP](https://prometheus.io/docs/guides/opentelemetry/#enable-the-otlp-receiver), providing a more stable and actively maintained solution.

I see we pull in `otlp` too. What do we actually rely on `opentelemetry-prometheus` for? How difficult would it be to remove the dependency altogether?

Freyskeyd:
closing, will reopen if needed

PR #501, opened 2024-12-27
feat: await for the l1 info tree entry to be finalized

### LINKED ISSUE #479: Switch checks from `latest` (default) to await for `finalized` for the L1 checks
# Description

Behavior as of today:
- Check that the L1 Info Root / leaf count referred by the Certificate exists on the L1 (default: check only on the `latest` block so far)

What we want:
- Check that the L1 Info Root / leaf count referred by the Certificate exists on the `latest` L1 block, reject otherwise
- Await the certifier process until that the referred L1 Info Root is referred in a `finalized` block

hadjiszs:
# Description

- Check that the L1 Info Root / leaf count referred by the Certificate exists on the latest L1 block, reject otherwise
- Await the certifier process until that the referred L1 Info Root is referred in a finalized block

If the L1 ends up doing a reorg leading to the referred L1 info tree entry to be non-existent, then the Certificate will become `InError` with the error `CertificationError::L1InfoRootNotFound`.

Fixes #479 

## PR Checklist:

### DIFF
--- a/crates/agglayer-contracts/src/lib.rs
+++ b/crates/agglayer-contracts/src/lib.rs
@@ -35,10 +35,14 @@ pub trait RollupContract {
         proof_signers: HashMap<u32, Address>,
     ) -> Result<Address, ()>;
 
-    async fn get_l1_info_root(&self, l1_leaf_count: u32) -> Result<[u8; 32], ()>;
+    async fn get_l1_info_root(&self, l1_leaf_count: u32) -> Result<[u8; 32], L1RpcError>;
 
     fn default_l1_info_tree_entry(&self) -> (u32, [u8; 32]);
 }
+
+/// Polling tick interval used to check for one block to be finalized.
+const CHECK_BLOCK_FINALIZED_TICK_INTERVAL: tokio::time::Duration =
+    tokio::time::Duration::from_secs(10);
 
 pub struct L1RpcClient<RpcProvider> {
     rpc: Arc<RpcProvider>,
@@ -54,6 +58,16 @@ pub enum L1RpcInitializationError {
     InitL1InfoRootMapEventNotFound(String),
     #[error("Event InitL1InfoRootMap returned null value for L1 info root, leaf count: {0}")]
     InvalidL1InfoRootFromEvent(u32),
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum L1RpcError {
+    #[error("Failed to get the `UpdateL1InfoTreeV2` events: {0}")]
+    UpdateL1InfoTreeV2EventFailure(String),
+    #[error("Unable to find `UpdateL1InfoTreeV2` events")]
+    UpdateL1InfoTreeV2EventNotFound,
+    #[error("Unable to fetch the latest finalized block")]
+    LastFinalizedBlockNotFound,
 }
 
 impl<RpcProvider> L1RpcClient<RpcProvider>
@@ -130,7 +144,7 @@ where
         self.default_l1_info_tree_entry
     }
 
-    async fn get_l1_info_root(&self, l1_leaf_count: u32) -> Result<[u8; 32], ()> {
+    async fn get_l1_info_root(&self, l1_leaf_count: u32) -> Result<[u8; 32], L1RpcError> {
         // Get `UpdateL1InfoTreeV2` event for the given leaf count from the latest block
         let filter = Filter::new()
             .address(self.l1_info_tree.address())
@@ -143,7 +157,7 @@ where
             .client()
             .get_logs(&filter)
             .await
-            .map_err(|_| ())?;
+            .map_err(|e| L1RpcError::UpdateL1InfoTreeV2EventFailure(e.to_string()))?;
 
         // Extract event details
         let (l1_info_root, event_block_number) = events
@@ -156,7 +170,7 @@ where
                     _ => None,
                 }
             })
-            .ok_or(())?;
+            .ok_or(L1RpcError::UpdateL1InfoTreeV2EventNotFound)?;
 
         debug!(
             "Retrieved UpdateL1InfoTreeV2 event from block {}. L1 info tree leaf count: {}, root: \
@@ -167,7 +181,7 @@ where
         // Await for the related block to be finalized
         // NOTE: Cannot use block subscription because the provider is not websocket
         {
-            let mut tick = tokio::time::interval(tokio::time::Duration::from_secs(10));
+            let mut tick = tokio::time::interval(CHECK_BLOCK_FINALIZED_TICK_INTERVAL);
             let mut finalized_block_number: U64 = 0.into();
             while finalized_block_number < event_block_number {
                 tick.tick().await;
@@ -179,7 +193,7 @@ where
                     .ok()
                     .flatten()
                     .and_then(|block| block.number)
-                    .ok_or(())?;
+                    .ok_or(L1RpcError::LastFinalizedBlockNotFound)?;
 
                 debug!(
                     "Awaiting L1 info tree leaf count ({}) set at block {} to be finalized. \
--- a/crates/agglayer-contracts/src/lib.rs
+++ b/crates/agglayer-contracts/src/lib.rs
@@ -124,8 +124,9 @@ where
             }
 
             debug!(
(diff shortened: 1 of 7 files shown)


Freyskeyd (dismissed):

Freyskeyd (approved):

crates/agglayer-contracts/src/lib.rs
```diff
@@ -126,10 +131,65 @@ where
...
+            .topic1(U256::from_big_endian(&l1_leaf_count.to_be_bytes()))
+            .from_block(BlockNumber::Earliest);
+
+        let events = self
+            .l1_info_tree
+            .client()
+            .get_logs(&filter)
             .await
-            .map_err(|_| ())
+            .map_err(|_| ())?;
```
iljakuklic, line 146: Should a more descriptive error type be introduced in place of `()`?

crates/agglayer-contracts/src/lib.rs
```diff
@@ -126,10 +131,65 @@ where
...
+            .ok_or(())?;
+
+        debug!(
+            "Retrieved UpdateL1InfoTreeV2 event from block {}. L1 info tree leaf count: {}, root: \
+             {:?}",
+            event_block_number, l1_leaf_count, l1_info_root
+        );
+
+        // Await for the related block to be finalized
+        // NOTE: Cannot use block subscription because the provider is not websocket
```
iljakuklic, line 168: It may be wort considering switching to websocket for this. I believe it's already used for the block-based epoch clock. (Not for this PR, but filing an issue would be appreciated if none exists yet.)

crates/agglayer-contracts/src/lib.rs
```diff
@@ -126,10 +131,65 @@ where
...
+        {
+            let mut tick = tokio::time::interval(tokio::time::Duration::from_secs(10));
```
iljakuklic, line 170: The tick duration should be a constant (or possibly even a configuration value).

crates/agglayer-contracts/src/lib.rs
```diff
@@ -126,10 +131,65 @@ where
...
+            let mut finalized_block_number: U64 = 0.into();
+            while finalized_block_number < event_block_number {
```
iljakuklic, line 172: Should there be a maximum number of attempts too?

crates/agglayer-contracts/src/lib.rs
```diff
@@ -126,10 +131,65 @@ where
...
+                tick.tick().await;
+
+                finalized_block_number = self
+                    .rpc
+                    .get_block(BlockNumber::Finalized)
+                    .await
+                    .ok()
+                    .flatten()
+                    .and_then(|block| block.number)
+                    .ok_or(())?;
```
iljakuklic, line 182: If `get_block` fails, it will break out of the loop instead of waiting for the next attempt. Is that the intended behaviour?

crates/agglayer-contracts/src/lib.rs
hadjiszs: done :+1:

crates/agglayer-contracts/src/lib.rs
hadjiszs: Moved to as constant

crates/agglayer-contracts/src/lib.rs
hadjiszs: I added a timeout based on the time for finality on ethereum

crates/agglayer-contracts/src/lib.rs
hadjiszs: #530

crates/agglayer-contracts/src/lib.rs
hadjiszs: Hm I believe that yes if `get_block` fails we want to get out of the loop and return an error

iljakuklic (approved):

crates/agglayer-contracts/src/lib.rs
```diff
@@ -125,11 +149,82 @@ where
...
+                        Some((event.current_l1_info_root, log.block_number?))
+                    }
+                    _ => None,
+                }
+            })
+            .ok_or(L1RpcError::UpdateL1InfoTreeV2EventNotFound)?;
+
+        debug!(
+            "Retrieved UpdateL1InfoTreeV2 event from block {}. L1 info tree leaf count: {}, root: \
+             {:?}",
```
Freyskeyd, line 182: The `l1_info_root` will not be a hex encoded value right? just a bytearray?

Freyskeyd (approved):

Ekleog-Polygon (dismissed):
LGTM, thank you for having applied most of the comments! :)

crates/agglayer-contracts/src/lib.rs
```diff
@@ -125,11 +149,82 @@ where
...
+                            .and_then(|block| block.number)
+                            .ok_or(L1RpcError::LastFinalizedBlockNotFound)?;
+
+                        debug!(
+                            "Awaiting L1 info tree leaf count ({}) set at block {} to be finalized. \
+                             Latest finalized block: {}",
+                            l1_leaf_count, event_block_number, finalized_block_number,
+                        );
+
+                        if finalized_block_number >= event_block_number {
```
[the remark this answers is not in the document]
hadjiszs: yes great catch thank you :ok_hand: fixed

crates/agglayer-contracts/src/lib.rs
hadjiszs: Indeed, converted it to H256 for display

crates/agglayer-contracts/src/lib.rs
```diff
@@ -125,11 +149,82 @@ where
...
+            .topic1(U256::from_big_endian(&l1_leaf_count.to_be_bytes()))
+            .from_block(BlockNumber::Earliest);
+
+        let events = self
+            .l1_info_tree
+            .client()
+            .get_logs(&filter)
             .await
-            .map_err(|_| ())
+            .map_err(|e| L1RpcError::UpdateL1InfoTreeV2EventFailure(e.to_string()))?;
```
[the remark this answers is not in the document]
hadjiszs: Hm I did not figure out easy way to do so given that the involved Error type is under some generics, happy to change it if I missed something obvious

hadjiszs:
> LGTM, thank you for having applied most of the comments! :)

thanks sorry pushed again to fix @Freyskeyd 's comment which I missed

crates/agglayer-contracts/src/lib.rs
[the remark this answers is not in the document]
Ekleog-Polygon: Makes sense, it's probably not worth spending too much time on it :)

Ekleog-Polygon (approved):

crates/agglayer-contracts/src/lib.rs
```diff
@@ -125,11 +154,111 @@ where
...
+        // Await for the related block to be finalized
+        // NOTE: Cannot use block subscription because the provider is not websocket
+        {
+            let mut tick = tokio::time::interval(CHECK_BLOCK_FINALIZED_TICK_INTERVAL);
+            let mut timeout_tick = tokio::time::interval(TIME_TO_FINALITY_ETHEREUM);
+            let mut finalized_block_number: U64 = 0.into();
+
+            // The first tick completes immediately
+            tick.tick().await;
+            timeout_tick.tick().await;
```
iljakuklic, line 204: Wondering if `sleep` could be used instead of these intervals, seems a better match at a first glance. That would make dropping the initial `tick`s unnecessary.

iljakuklic (approved):

crates/agglayer-contracts/src/lib.rs
iljakuklic: > Wondering if `sleep` could be used instead of these intervals, seems a better match at a first glance. That would make dropping the initial `tick`s unnecessary.

Note: might only apply to the total timeout.

hadjiszs: >I guess wrapping the whole thing in tokio::time::timeout would be even more descriptive.

done

Ekleog-Polygon (approved):

Freyskeyd (approved):

iljakuklic (approved):

PR #505, opened 2025-01-02
feat: enable backup for storage DBs with purge mechanism

Freyskeyd:
# Description

This PR is adding a `BackupEngine` and its configuration to the `agglayer` node.

The goal is to allow `backups` of `state`, `pending` and `epoch` databases when triggered by the application.
The backups per database are cleaned up, keeping only a certain amount of backups.

## Config changes

```toml
[storage.backup.enabled]
path = "PATH TO BACKUP"
state_max_backup_number = 100 # default: 100
pending_max_backup_number = 100 # default: 100
```

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/src/storage/backup.rs
+++ b/crates/agglayer-config/src/storage/backup.rs
@@ -1,0 +1,45 @@
+use std::path::PathBuf;
+
+use serde::{Deserialize, Serialize};
+
+/// Configuration for Storage backups.
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub enum BackupConfig {
+    /// Backups are disabled.
+    #[default]
+    Disabled,
+
+    /// Backups are enabled.
+    Enabled {
+        /// Path to the directory where backups are stored.
+        path: PathBuf,
+        /// Maximum number of backups to keep for the state storage.
+        #[serde(default = "default_max_backup_number")]
+        state_max_backup_number: usize,
+        /// Maximum number of backups to keep for the pending storage.
+        #[serde(default = "default_max_backup_number")]
+        pending_max_backup_number: usize,
+    },
+}
+
+impl BackupConfig {
+    /// Default maximum number of backups to keep.
+    const DEFAULT_MAX_BACKUP_NUMBER: usize = 100;
+
+    pub fn with_path<P: Into<PathBuf>>(path: P) -> Self {
+        BackupConfig::Enabled {
+            path: path.into(),
+            state_max_backup_number: default_max_backup_number(),
+            pending_max_backup_number: default_max_backup_number(),
+        }
+    }
+
+    pub fn is_disabled(&self) -> bool {
+        *self == BackupConfig::Disabled
+    }
+}
+
+const fn default_max_backup_number() -> usize {
+    BackupConfig::DEFAULT_MAX_BACKUP_NUMBER
+}
--- a/crates/agglayer-config/src/storage/backup.rs
+++ b/crates/agglayer-config/src/storage/backup.rs
@@ -11,15 +11,16 @@ pub enum BackupConfig {
     Disabled,
 
     /// Backups are enabled.
+    #[serde(untagged)]
     Enabled {
         /// Path to the directory where backups are stored.
         path: PathBuf,
         /// Maximum number of backups to keep for the state storage.
         #[serde(default = "default_max_backup_number")]
-        state_max_backup_number: usize,
+        state_max_backup_count: usize,
         /// Maximum number of backups to keep for the pending storage.
         #[serde(default = "default_max_backup_number")]
-        pending_max_backup_number: usize,
+        pending_max_backup_count: usize,
     },
 }
 
@@ -30,8 +31,8 @@ impl BackupConfig {
     pub fn with_path<P: Into<PathBuf>>(path: P) -> Self {
         BackupConfig::Enabled {
             path: path.into(),
-            state_max_backup_number: default_max_backup_number(),
-            pending_max_backup_number: default_max_backup_number(),
+            state_max_backup_count: default_max_backup_number(),
+            pending_max_backup_count: default_max_backup_number(),
         }
     }
 
--- a/crates/agglayer-config/tests/snapshots/backup__backup_enabled.snap
+++ b/crates/agglayer-config/tests/snapshots/backup__backup_enabled.snap
@@ -62,7 +62,7 @@ input-backpressure-buffer-size = 1000
 [storage]
 db-path = "/tmp/agglayer/tests/fixtures/valide_config/storage"
 
-[storage.backup.enabled]
+[storage.backup]
 path = "./test/"
-state_max_backup_number = 100
-pending_max_backup_number = 100
+state_max_backup_count = 100
+pending_max_backup_count = 100
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -3,14 +3,17 @@ use std::{num::NonZeroU64, sync::Arc};
 use agglayer_aggregator_notifier::{CertifierClient, EpochPackerClient};
 use agglayer_certificate_orchestrator::CertificateOrchestrator;
 use agglayer_clock::{BlockClock, Clock, TimeClock};
-use agglayer_config::{Config, Epoch};
+use agglayer_config::{storage::backup::BackupConfig, Config, Epoch};
 use agglayer_contracts::{
     polygon_rollup_manager::PolygonRollupManager,
     polygon_zkevm_global_exit_root_v2::PolygonZkEVMGlobalExitRootV2, L1RpcClient,
 };
 use agglayer_signer::ConfiguredSigner;
 use agglayer_storage::{
-    storage::DB,
+    storage::{
+        backup::{BackupClient, BackupEngine},
+        DB,
+    },
     stores::{
         debug::DebugStore, epochs::EpochsStore, pending::PendingStore, state::StateStore,
         PerEpochReader as _,
@@ -87,7 +90,27 @@ impl Node {
             agglayer_storage::storage::state_db_cf_definitions(),
         )?);
 
-        let state_store = Arc::new(StateStore::new(state_db.clone()));
+        // Initialize backup engine
+        let backup_client = if let BackupConfig::Enabled {
+            path,
+            state_max_backup_number,
+            pending_max_backup_number,
+        } = &config.storage.backup
+        {
+            let (backup_engine, client) = BackupEngine::new(
+                path,
+                state_db.clone(),
+                pending_db.clone(),
+                *state_max_backup_number,
+                *pending_max_backup_number,
+            )?;
+            tokio::spawn(async move { backup_engine.run().await });
+
+            client
+        } else {
+            BackupClient::noop()
+        };
+        let state_store = Arc::new(StateStore::new(state_db.clone(), backup_client));
         let pending_store = Arc::new(PendingStore::new(pending_db.clone()));
         let debug_store = if config.debug_mode {
             Arc::new(DebugStore::new_with_path(&config.storage.debug_db_path)?)
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -93,19 +93,19 @@ impl Node {
         // Initialize backup engine
         let backup_client = if let BackupConfig::Enabled {
             path,
-            state_max_backup_number,
-            pending_max_backup_number,
+            state_max_backup_count,
+            pending_max_backup_count,
         } = &config.storage.backup
         {
             let (backup_engine, client) = BackupEngine::new(
                 path,
                 state_db.clone(),
                 pending_db.clone(),
-                *state_max_backup_number,
-                *pending_max_backup_number,
+                *state_max_backup_count,
+                *pending_max_backup_count,
+                cancellation_token.clone(),
             )?;
-            let backup_engine_cancellation_token = cancellation_token.child_token();
-            tokio::spawn(async move { backup_engine.run(backup_engine_cancellation_token).await });
+            tokio::spawn(backup_engine.run());
 
             client
         } else {
@@ -162,7 +162,7 @@ impl Node {
             current_epoch,
             pending_store.clone(),
             state_store.clone(),
-            backup_client.clone(),
+            backup_client,
         )?);
 
         info!("Epoch synchronization started.");
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -104,13 +104,14 @@ impl Node {
                 *state_max_backup_number,
                 *pending_max_backup_number,
             )?;
-            tokio::spawn(async move { backup_engine.run().await });
+            let backup_engine_cancellation_token = cancellation_token.child_token();
+            tokio::spawn(async move { backup_engine.run(backup_engine_cancellation_token).await });
 
             client
         } else {
             BackupClient::noop()
         };
-        let state_store = Arc::new(StateStore::new(state_db.clone(), backup_client));
+        let state_store = Arc::new(StateStore::new(state_db.clone(), backup_client.clone()));
         let pending_store = Arc::new(PendingStore::new(pending_db.clone()));
         let debug_store = if config.debug_mode {
             Arc::new(DebugStore::new_with_path(&config.storage.debug_db_path)?)
@@ -161,6 +162,7 @@ impl Node {
             current_epoch,
             pending_store.clone(),
             state_store.clone(),
+            backup_client.clone(),
         )?);
 
         info!("Epoch synchronization started.");
--- a/crates/agglayer-storage/src/storage/backup/mod.rs
+++ b/crates/agglayer-storage/src/storage/backup/mod.rs
@@ -1,10 +1,17 @@
 use std::{
+    collections::BTreeMap,
+    fs::read_dir,
     path::{Path, PathBuf},
     sync::Arc,
 };
 
-use rocksdb::backup::{BackupEngine as RocksBackupEngine, RestoreOptions};
+use rocksdb::backup::{
+    BackupEngine as RocksBackupEngine, BackupEngineInfo as RocksBackupEngineInfo,
+    BackupEngineOptions, RestoreOptions,
+};
+use serde::{Deserialize, Serialize};
 use tokio::sync;
+use tokio_util::sync::CancellationToken;
 use tracing::{error, info};
 
 use super::{BackupError, DB};
@@ -16,6 +23,7 @@ pub struct BackupRequest {
 }
 
 /// Client used to request a backup.
+#[derive(Clone)]
 pub struct BackupClient {
     sender: Option<sync::mpsc::Sender<BackupRequest>>,
 }
@@ -79,6 +87,9 @@ impl BackupEngine {
 
         let (sender, backup_request) = sync::mpsc::channel(10);
 
+        let epochs_path = path.join("epochs");
+        std::fs::create_dir_all(&epochs_path)?;
+
         Ok((
             Self {
                 state_engine: RocksBackupEngine::open(&state_opts, &env)?,
@@ -86,7 +97,7 @@ impl BackupEngine {
                 env,
                 state_db,
                 pending_db,
-                epochs_path: path.join("epochs"),
+                epochs_path,
                 backup_request,
                 state_max_backup_number,
                 pending_max_backup_number,
@@ -101,34 +112,6 @@ impl BackupEngine {
     /// This function will also purge old backups as configured.
     pub fn create_new_backup(&mut self, request: &BackupRequest) -> Result<(), BackupError> {
         info!("Creating new backup");
-
-        if let Err(error) = self
-            .state_engine
-            .create_new_backup_flush(&self.state_db.rocksdb, true)
-        {
-            error!("Failed to create backup for state db: {:?}", error);
-        }
-
-        if let Err(error) = self
-            .state_engine
-            .purge_old_backups(self.state_max_backup_number)
-        {
-            error!("Failed to purge old backup for state db: {:?}", error);
-        }
-
-        if let Err(error) = self
-            .pending_engine
-            .create_new_backup_flush(&self.pending_db.rocksdb, true)
-        {
-            error!("Failed to create backup for pending db: {:?}", error);
-        }
-
-        if let Err(error) = self
-            .pending_engine
-            .purge_old_backups(self.pending_max_backup_number)
-        {
-            error!("Failed to purge old backup for pending db: {:?}", error);
-        }
 
         if let Some((db, epoch_number)) = request.epoch_db.as_ref() {
             let epochs_opts = rocksdb::backup::BackupEngineOptions::new(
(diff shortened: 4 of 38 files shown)


hadjiszs (approved):

crates/agglayer-node/src/node.rs
```diff
@@ -138,6 +162,7 @@ impl Node {
             current_epoch,
             pending_store.clone(),
             state_store.clone(),
+            backup_client.clone(),
```
atanmarko, line 165: You don't need to clone here

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    /// Restore the state database from the latest backup.
+    pub fn restore(path: &Path, db_path: &Path) -> Result<(), BackupError> {
+        let env = rocksdb::Env::new()?;
+        let opts = rocksdb::backup::BackupEngineOptions::new(path)?;
+
+        let mut engine = RocksBackupEngine::open(&opts, &env)?;
+
+        std::fs::create_dir_all(db_path).unwrap();
```
atanmarko, line 190: Should we panic here or bubble error up?

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    /// Restore the state database from the defined backup version.
+    pub fn restore_at(path: &Path, db_path: &Path, version: u32) -> Result<(), BackupError> {
+        let env = rocksdb::Env::new()?;
+        let opts = rocksdb::backup::BackupEngineOptions::new(path)?;
+
+        let mut engine = RocksBackupEngine::open(&opts, &env)?;
+
+        std::fs::create_dir_all(db_path).unwrap();
```
atanmarko, line 202: Again, we should not panic without explanation

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+            backup_id: info.backup_id,
+            timestamp: info.timestamp,
+            size: info.size,
+            num_files: info.num_files,
+        }
+    }
+}
+
+impl BackupReport {
+    pub fn state<V: IntoIterator<Item = T>, T: Into<BackupEngineInfo>>(&mut self, value: V) {
```
atanmarko, line 295: Nit: I think `set_state` (and `set_pending`, `set_epochs`) would be more appropriate. This way it feels like a getter.

crates/agglayer/src/main.rs
```diff
@@ -31,6 +34,48 @@ fn main() -> anyhow::Result<()> {
             let vkey = agglayer_prover::get_vkey();
             println!("{}", vkey);
         }
+
+        cli::Commands::Backups(cli::Backups::List { cfg }) => {
+            let cfg = agglayer_config::Config::try_load(&cfg)?;
+
+            if let BackupConfig::Enabled { path, .. } = cfg.storage.backup {
+                let result =
+                    agglayer_storage::storage::backup::BackupEngine::list_backups(&path).unwrap();
```
atanmarko, line 43: We should exit program with error

tests/integrations/src/agglayer_setup.rs
```diff
@@ -119,12 +125,13 @@ pub async fn start_agglayer(tmp_dir: &Path, l1: &L1Docker) -> (oneshot::Receiver
...
 
-    let config_file = tmp_dir.join("config.toml");
+    let config_file = config_path.join("config.toml");
     let toml = toml::to_string_pretty(&config).unwrap();
     std::fs::write(&config_file, toml).unwrap();
 
+    let gracefull_shutdown_token = shutdown_token.clone();
     let handle = std::thread::spawn(move || {
-        _ = agglayer_node::main(config_file);
+        agglayer_node::main(config_file, Some(gracefull_shutdown_token)).unwrap();
```
atanmarko, line 134: Could we print here some useful error info before panicking? Maybe we should refactor `start_agglayer` to return error and then bubble up

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        Ok(engine.restore_from_backup(db_path, db_path, &RestoreOptions::default(), version)?)
+    }
+
+    pub fn list_backups(path: &Path) -> Result<BackupReport, BackupError> {
+        let env = rocksdb::Env::new()?;
+
+        let mut report = BackupReport::default();
+
+        let opts = BackupEngineOptions::new(path.join("state"))?;
```
iljakuklic, line 212: Minor: Would probably be better to have some sort of `BackupConfig` object instead of bare path. The object can have methods like `state_backup_path`, `epoch_backup_path`, etc. That would avoid the repeated use of string literals which are somewhat error prone and harder to maintain.

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+
+impl Drop for BackupEngine {
+    fn drop(&mut self) {
+        self.env.set_background_threads(0);
+        self.env.set_low_priority_background_threads(0);
+        self.env.set_high_priority_background_threads(0);
+        self.env.set_bottom_priority_background_threads(0);
+
+        self.env.join_all_threads();
+    }
```
iljakuklic, line 254: It's not quite clear to me what's happening here, particularly why the `set_*_threads` calls are required. A comment would be appreciated.

crates/agglayer/src/cli.rs
```diff
@@ -52,4 +52,72 @@ pub(crate) enum Commands {
     },
 
     Vkey,
+
+    #[clap(subcommand)]
+    Backups(Backups),
```
Ekleog-Polygon, line 57: Nit: rename `Backup`? `backup list / backup restore` sounds better than `backups list / backups restore` (especially the `backups restore` one)

crates/agglayer/src/cli.rs
```diff
@@ -52,4 +52,72 @@ pub(crate) enum Commands {
...
+    List {
+        #[arg(long, short, value_hint = ValueHint::FilePath, default_value = "agglayer.toml", env = "CONFIG_PATH")]
+        cfg: PathBuf,
+    },
+
+    /// Restore from a backup.
+    Restore {
+        #[arg(long, short, value_hint = ValueHint::FilePath, default_value = "agglayer.toml", env = "CONFIG_PATH")]
+        cfg: PathBuf,
+        #[arg( value_parser = parse_db_kind_version)]
```
Ekleog-Polygon, line 72:
```suggestion
        #[arg(value_parser = parse_db_kind_version)]
```

tests/integrations/Cargo.toml
```diff
@@ -19,7 +19,7 @@ hex.workspace = true
 http-body-util = "0.1.2"
 hyper.workspace = true
 hyper-util = { version = "0.1.6", features = ["client"] }
-insta.workspace = true
+insta = { workspace = true, features = ["json"] }
```
Ekleog-Polygon, line 22: Nit: add this right in the workspace `Cargo.toml`? It sounds like it’d be useful everywhere, it’s used only for tests, and it’ll make for fewer rebuilds when testing individual crates (because builds will be cached better)

tests/integrations/src/agglayer_setup.rs
```diff
@@ -119,12 +125,13 @@ pub async fn start_agglayer(tmp_dir: &Path, l1: &L1Docker) -> (oneshot::Receiver
         }],
     });
 
-    let config_file = tmp_dir.join("config.toml");
+    let config_file = config_path.join("config.toml");
     let toml = toml::to_string_pretty(&config).unwrap();
     std::fs::write(&config_file, toml).unwrap();
 
+    let gracefull_shutdown_token = shutdown_token.clone();
```
Ekleog-Polygon, line 132:
```suggestion
    let graceful_shutdown_token = shutdown_token.clone();
```

Ekleog-Polygon (commented):
I still need to review the `crates/agglayer-storage/src/storage/backup/mod.rs` and `tests/integrations/tests/storage_backups/main.rs`, but here are a few comments already.

I’ll resume the review soon!

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+            .filter_map(|d| {
+                d.file_name()
+                    .to_string_lossy()
+                    .parse::<u64>()
+                    .map(|r| (r, d.path()))
+                    .ok()
+            })
+            .collect::<Vec<_>>();
+
+        epochs.sort_by(|(p, _), (n, _)| p.cmp(n));
```
Ekleog-Polygon, line 234: I think this is basically the default sorting method, considering that all epoch numbers are different? AFAIU the default sort would do lexicographical ordering.

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+            report.epochs(epoch_number, engine.get_backup_info());
+        }
+
+        Ok(report)
+    }
+}
+
+impl Drop for BackupEngine {
+    fn drop(&mut self) {
+        self.env.set_background_threads(0);
```
Ekleog-Polygon, line 248:
```suggestion
        info!("Waiting for all requested backups to complete");

        self.env.set_background_threads(0);
```

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+pub struct BackupReport {
+    epochs: BTreeMap<u64, Vec<BackupEngineInfo>>,
+    state: Vec<BackupEngineInfo>,
+    pending: Vec<BackupEngineInfo>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct BackupEngineInfo {
+    pub backup_id: u32,
+    #[serde(serialize_with = "timestamp_to_readable")]
```
Ekleog-Polygon, line 267: AFAIU with this Serialize and Deserialize do not actually round-trip correctly. Maybe just remove the Deserialize bound, that is likely not of much use?

Also, can we make `timestamp` just be `: chrono::DateTime<chrono::Utc>`, which should both be more user-friendly and automatically take care of this?

Ekleog-Polygon (commented):
Here is, as promised, the end of my review! :)

crates/agglayer/src/cli.rs
```diff
@@ -48,4 +48,72 @@ pub(crate) enum Commands {
...
+
+impl std::str::FromStr for DbKind {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "state" => Ok(DbKind::State),
+            "pending" => Ok(DbKind::Pending),
+            s if s.starts_with("epoch_") => {
+                let parts: Vec<&str> = s.split("_").collect();
```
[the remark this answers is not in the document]
Freyskeyd: I think this is an experimental format, I'll try it

crates/agglayer/src/cli.rs
```diff
@@ -48,4 +48,72 @@ pub(crate) enum Commands {
...
+
+    #[clap(subcommand)]
+    Backups(Backups),
+}
+
+#[derive(Subcommand)]
+pub(crate) enum Backups {
+    /// List all backups.
+    List {
+        #[arg(long, short, value_hint = ValueHint::FilePath, default_value = "agglayer.toml", env = "CONFIG_PATH")]
```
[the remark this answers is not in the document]
Freyskeyd: Yes, or even remove the env variable if not needed

crates/agglayer-storage/src/storage/backup/mod.rs
Freyskeyd: It prevents `rocksdb` to spawn any more threads and wait for completion of the current ones.

crates/agglayer/src/cli.rs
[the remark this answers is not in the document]
Freyskeyd: @iljakuklic the suggestion is experimental, I refactored a bit this part. Let me know if that's ok for you !

crates/agglayer-node/src/node.rs
```diff
@@ -87,7 +90,28 @@ impl Node {
...
+        {
+            let (backup_engine, client) = BackupEngine::new(
+                path,
+                state_db.clone(),
+                pending_db.clone(),
+                *state_max_backup_number,
+                *pending_max_backup_number,
+            )?;
+            let backup_engine_cancellation_token = cancellation_token.child_token();
+            tokio::spawn(async move { backup_engine.run(backup_engine_cancellation_token).await });
```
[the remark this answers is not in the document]
Freyskeyd: @iljakuklic @Ekleog-Polygon 
I refactor a bit the `Drop` of the `BackupEngine` so that it cancel the token when dropped.

crates/agglayer-storage/src/storage/mod.rs
```diff
@@ -27,6 +29,18 @@ pub enum DBError {
...
+#[derive(Debug, thiserror::Error)]
+pub enum BackupError {
+    #[error("Unable to send backup request")]
+    UnableToSendBackupRequest,
+
+    #[error("RocksDB error: {0}")]
+    RocksDB(#[from] rocksdb::Error),
+
+    #[error("IO Error: {0}")]
+    IO(#[from] std::io::Error),
```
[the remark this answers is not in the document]
Freyskeyd: I think it is quite interesting, but I don't want to introduce too many changes in this PR.
I created a [ticket](https://github.com/agglayer/protocol-research/issues/115) related to that, if you want to add your ideas here 🙏🏼

tests/integrations/tests/storage_backups/main.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    let config = agglayer_config::Config::new(&tmp_dir.path);
+    std::fs::remove_dir_all(&config.storage.pending_db_path).unwrap();
+    std::fs::remove_dir_all(&config.storage.epochs_db_path).unwrap();
+    std::fs::remove_dir_all(&config.storage.state_db_path).unwrap();
+
+    let backup_report = BackupEngine::list_backups(&backup_dir.path).unwrap();
+
+    assert_eq!(backup_report.get_state().len(), 4);
+    assert_eq!(backup_report.get_pending().len(), 4);
```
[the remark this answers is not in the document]
Freyskeyd: There are 4 backups because 2 actions triggers a backup per certs:

- One when the L1 `tx_hash` is known
- One when the `Certificate` is settled and the network state is updated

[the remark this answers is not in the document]
Ekleog-Polygon: Got it, thank you!

crates/agglayer-storage/src/storage/backup/mod.rs
Freyskeyd: As the initial vector is coming from a `read_dir` which list the folder in the system order (treated as str), the order isn't guaranteed.
We are parsing the epoch numbers as u64, in order to have them properly sorted:

```
# Example of the output of `read_dir`

>> println!("{:#?}", std::fs::read_dir("./")?.flatten().collect::<Vec<_>>())
[
    DirEntry(
        "./20",
    ),
    DirEntry(
        "./11",
    ),
    DirEntry(
        "./1",
    ),
    DirEntry(
        "./10",
    ),
    DirEntry(
        "./2",
    ),
]
```

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    pub fn list_backups(path: &Path) -> Result<BackupReport, BackupError> {
+        let env = rocksdb::Env::new()?;
+
+        let mut report = BackupReport::default();
+
+        let opts = BackupEngineOptions::new(path.join("state"))?;
+        let engine = RocksBackupEngine::open(&opts, &env)?;
+
+        report.state(engine.get_backup_info());
```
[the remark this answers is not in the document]
Freyskeyd: @Ekleog-Polygon I made some changes, let me know if its ok for you

crates/agglayer-storage/src/storage/backup/mod.rs
Ekleog-Polygon: Right! What I mean is that you don’t actually need `.sort_by()`, but `.sort()` would be enough, because it sorts in lexicographical order by default :)

crates/agglayer-storage/src/storage/backup/mod.rs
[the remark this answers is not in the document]
Ekleog-Polygon: LGTM, thanks!

crates/agglayer-storage/src/storage/backup/mod.rs
Freyskeyd: Oh, right ! I'll fix that !

Ekleog-Polygon (approved):
LGTM! Just a very minor code formatting nit remaining, but I think everyone’s review comments have been properly addressed :)

crates/agglayer/src/cli.rs
```diff
@@ -52,4 +52,71 @@ pub(crate) enum Commands {
...
+
+impl std::str::FromStr for DbKind {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str().trim() {
+            "state" => Ok(DbKind::State),
+            "pending" => Ok(DbKind::Pending),
+            s if s.starts_with("epoch_") => {
+                if let Some(s) = s.strip_prefix("epoch_") {
```
iljakuklic, line 92: Nit: The `strip_prefix` and `starts_with` perform redundant checks.
```suggestion
            s => {
                if let Some(s) = s.strip_prefix("epoch_") {
```

(May need to adjust the error messages and the wildcard branch can be removed after this.)

crates/agglayer-config/src/storage/backup.rs
```diff
@@ -0,0 +1,45 @@
...
+/// Configuration for Storage backups.
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub enum BackupConfig {
+    /// Backups are disabled.
+    #[default]
+    Disabled,
+
+    /// Backups are enabled.
+    Enabled {
```
iljakuklic, line 14: Would it make sense to add `#[serde(untagged)]` here?
```suggestion
    /// Backups are enabled.
    #[serde(untagged)]
    Enabled {
```

Not sure if that would play nicely with toml serializer but it would avoid specifying the annoying `enabled` key in the config file.

crates/agglayer-config/tests/snapshots/backup__backup_enabled.snap
```diff
@@ -0,0 +1,68 @@
...
+[certificate-orchestrator]
+input-backpressure-buffer-size = 1000
+
+[certificate-orchestrator.prover.sp1-local]
+
+[storage]
+db-path = "/tmp/agglayer/tests/fixtures/valide_config/storage"
+
+[storage.backup.enabled]
+path = "./test/"
```
iljakuklic, line 66: Here is what I'm talking about in the other comment. It would become:
```toml
[storage.backup]
path = "./test/"
```

or just:
```toml
storage.backup = "disabled"
```
to disable backups explicitly

crates/agglayer-storage/src/storage/backup/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+    state_max_backup_number: usize,
+    pending_max_backup_number: usize,
+    cancellation_token: CancellationToken,
+}
+
+// # Safety
+//
+// RocksBackupEngine is a simple pointer wrapper, so it's safe to send to
+// another thread since the underlying RocksDB backup engine is thread-safe.
+unsafe impl Send for BackupEngine {}
```
iljakuklic, line 93: What is preventing this from being auto-inferred? Is it because `RocksBackupEngine` is `!Sync`?

crates/agglayer/src/main.rs
```diff
@@ -31,6 +34,51 @@ fn main() -> anyhow::Result<()> {
...
+                        }
+                        cli::DbKind::Pending => (
+                            cfg.storage.pending_db_path.join("pending"),
+                            path.join("pending"),
+                        ),
+                        cli::DbKind::Epoch(epoch_number) => (
+                            cfg.storage.epochs_db_path.join(format!("{}", epoch_number)),
+                            path.join(format!("epochs/{}", epoch_number)),
+                        ),
+                    };
```
iljakuklic, line 69: Nit: The path calculation could be a separate function. I suspect it's also useful where the backups are generated?

crates/agglayer-storage/src/storage/backup/mod.rs
Freyskeyd: Yes, it's because of [this](https://docs.rs/rocksdb/latest/src/rocksdb/backup.rs.html#39)

Ekleog-Polygon: Uhh I had missed this change. Do you know why the developers of the `rocksdb` crate did not already do this? It seems weird that they'd not implement it if they could

Ekleog-Polygon (commented):
Dismissed my r+ due to the unsafe keyword @iljakuklic noticed. I don't know how I missed it, maybe it was part of the review doing a branch change and I missed it within all the rebase noise? Anyway, thank you @iljakuklic for pointing it out!

crates/agglayer-storage/src/storage/backup/mod.rs
Ekleog-Polygon: Actually, looking more into it, the rocksdb developers seem to already have done exactly this [here](https://docs.rs/rocksdb/latest/src/rocksdb/backup.rs.html#53). So you theoretically shouldn't need to do it yourself, or there's something weirder going on?

iljakuklic: > Actually, looking more into it, the rocksdb developers seem to already have done exactly this [here](https://docs.rs/rocksdb/latest/src/rocksdb/backup.rs.html#53). So you theoretically shouldn't need to do it yourself, or there's something weirder going on?

I did a little digging and the `Send` impl for `rocksdb::backup::BackupEngine` was only added in version 0.23, whereas we pull in version 0.22.

Ekleog-Polygon: Then we should probably land my deps upgrade PR, that bumps to 0.23, rather than adding our own unsafe impl :smile:

iljakuklic: > Then we should probably land my deps upgrade PR, that bumps to 0.23, rather than adding our own unsafe impl 😄

Alternatively, if that would create a chain of PR dependencies, I'd be fine with just bumping that one dep as a part of this PR.

Ekleog-Polygon: This version bump would actually be the only part of #547 that could be controversial, because it bumps rocksdb from 8 to 9, so I’d think it’s better to keep them separate, if only for the git log to include the full rationale of each change :)

This being said, I rebased to 0.2.2 and signed the commit, so #547 is ready to land as soon as it gets the green checkmarks!

Ekleog-Polygon (dismissed):
Considering that:
- The rocksdb upgrade from 8 to 9 is relatively risky
- We’re targeting 0.2.2 for this change
- 0.2.2 will have relatively little time for testing
- 0.3.0 should come soon, and can have all the missing updates
- `BackupEngine` has been thread-safe [since v6.20](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/utilities/backup_engine.h#L654) and we’re currently on v8

@Freyskeyd and I both estimate that it should be fine to land the `unsafe impl` as-is.

As a consequence, I’m approving this PR and adjusting my deps-bumping PRs :)

iljakuklic (approved):

PR #521, opened 2025-01-08
refactor: minimize pp program dependencies

### LINKED ISSUE #450: Refactor pessimistic-proof crate to include only struct/trait used by the program and needed outside of the program
## Description

The goal of this ticket is to refactor the `pessimistic-proof` crate in order to make it lighter to avoid having the need to update the `VerificationKey` (vkey) every time we change something in it.

## Context

The diagram below shows the dependencies between the different crates and the PP library. As you can see, it is not optimized as a lot of high level crates are depending on the PP lib.

<div align="center">

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/23bfdb11-7d52-48aa-934c-559305312f19">
  
</picture>

</div>

The goal would be to either:

- Move a lot of structs, trait and tests in `agglayer-types` so that `pessimistic-proof` library one contains shared struct with the `program`.
- Create a new `pessimistic-proof-core` crate that contains only shared struct with the `program` such as:

<div align="center">

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/ea5031fb-88f4-4e89-bb17-0ee404560a14">
  
</picture>

</div>

## Technical details

- A lot of `structs` in `pessimistic-proof` are only used by `agglayer-*` crates
- Changing the `program` (and its associated `lib`) is a breaking

hadjiszs:
# Description

- [x] introduce `pessimistic-proof-core` crate by pruning what is not necessary for the program
- [x] plug `pessimistic-proof-program` to `pessimistic-proof-core`
- [ ] adapt `pessimistic-proof` to leverage `pessimitic-proof-core` structures

Fixes #450 

## PR Checklist:


iljakuklic (commented):
Only skimmed through the contents of `pessimisitic-proof-core`, assuming it's mostly a copy-paste of the previous code from the `pessimistic-proof` crate.

iljakuklic (approved):

crates/pessimistic-proof/src/lib.rs
```diff
@@ -12,8 +10,19 @@ pub mod bridge_exit;
...
+    pub use pessimistic_proof_core::keccak::*;
+}
+
+pub use pessimistic_proof_core::local_state::NetworkState;
+pub use pessimistic_proof_core::multi_batch_header;
+pub use pessimistic_proof_core::proof::ProofError;
+
+pub mod core {
+    pub use pessimistic_proof_core::generate_pessimistic_proof;
+}
```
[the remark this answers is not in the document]
Freyskeyd: At first, that what I did, but as the goal is really to decouple the pp lib from the program, it didn't make sense to offer a way for `agglayer-*` crates to access all the struct. I picked only the relevant one and re-export them in dedicated module, so that it is transparent if we want to migrate some in further iteration.

Freyskeyd (approved):

iljakuklic (dismissed):

crates/pessimistic-proof-test-suite/src/forest.rs
```diff
@@ -35,7 +39,7 @@ impl Default for Forest {
             local_exit_tree_data_a: LocalExitTreeData::new(),
             l1_info_tree: Default::default(),
             state_b: LocalNetworkStateData {
-                exit_tree: LocalExitTree::new(),
+                exit_tree: pessimistic_proof::local_exit_tree::LocalExitTree::new(),
```
Ekleog-Polygon, line 42: Nit: import `LocalExitTree` to avoid these changes

crates/pessimistic-proof-test-suite/src/sample_data.rs
```diff
@@ -3,48 +3,45 @@
...
-type TreeHasher = local_exit_tree::hasher::Keccak256Hasher;
+type TreeHasher = pessimistic_proof::local_exit_tree::hasher::Keccak256Hasher;
 type LocalExitTree = local_exit_tree::LocalExitTree<TreeHasher>;
-type LocalBalanceTree = local_balance_tree::LocalBalanceTree<TreeHasher>;
 
 lazy_static::lazy_static! {
-    pub static ref NETWORK_A: NetworkId = 0.into();
-    pub static ref NETWORK_B: NetworkId = 1.into();
+    pub static ref NETWORK_A: NetworkId = 0u32.into();
+    pub static ref NETWORK_B: NetworkId = 1u32.into();
```
Ekleog-Polygon, line 23:
```suggestion
    pub static ref NETWORK_A: NetworkId = 0;
    pub static ref NETWORK_B: NetworkId = 1;
```

crates/pessimistic-proof-test-suite/src/sample_data.rs
```diff
@@ -3,48 +3,45 @@
...
     };
 }
 
 pub fn empty_state() -> LocalNetworkState {
     LocalNetworkState {
-        balance_tree: LocalBalanceTree::new(),
         exit_tree: LocalExitTree::new(),
-        nullifier_tree: NullifierTree::new(),
+        balance_tree: pessimistic_proof::local_balance_tree::LocalBalanceTree::new(),
+        nullifier_tree: pessimistic_proof::nullifier_tree::NullifierTree::new(),
```
Ekleog-Polygon, line 38: Similarly, import `LocalBalanceTree` and `NullifierTree` to simplify here

crates/pessimistic-proof-test-suite/src/sample_data.rs
```diff
@@ -3,48 +3,45 @@
...
     }
 }
 
-fn sample_exit_tree_01() -> LocalExitTree {
+fn sample_exit_tree_01() -> pessimistic_proof::local_exit_tree::LocalExitTree<TreeHasher> {
```
Ekleog-Polygon, line 42: And `LocalExitTree`

crates/pessimistic-proof/Cargo.toml
```diff
@@ -1,23 +1,27 @@
...
-tracing.workspace = true
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["arbitrary_precision"] }
 serde_with = { version = "3" }
 thiserror.workspace = true
-tiny-keccak = { git = "https://github.com/sp1-patches/tiny-keccak", branch = "patch-v2.0.2", features = [
-    "keccak",
-] }
+hex.workspace = true
+hex-literal = "0.4"
```
Ekleog-Polygon, line 16: Nit: the deps were alphabetically sorted before

crates/pessimistic-proof/src/bridge_exit.rs
```diff
@@ -104,23 +76,26 @@ impl BridgeExit {
...
-    }
-
     pub fn is_transfer(&self) -> bool {
         self.leaf_type == LeafType::Transfer
     }
+}
+
+impl Hashable for BridgeExit {
+    fn hash(&self) -> Digest {
+        pessimistic_proof_core::bridge_exit::bridge_exit_hasher(
```
Ekleog-Polygon, line 86: This could just call `BridgeExit::hash(self)` and we can remove `bridge_exit_hasher` as suggested above

crates/pessimistic-proof/src/local_exit_tree/mod.rs
```diff
@@ -1,55 +1,93 @@
...
+pub use pessimistic_proof_core::local_exit_tree::LocalExitTreeError;
+
 #[cfg(test)]
 mod tests;
 
+impl<H, const TREE_DEPTH: usize> TryFrom<&LocalExitTreeData<H, TREE_DEPTH>>
+    for LocalExitTree<H, TREE_DEPTH>
+where
+    H: Hasher,
+    H::Digest: std::fmt::Debug + Copy + Default + Serialize + DeserializeOwned,
```
Ekleog-Polygon, line 18: Is the new `Debug` bound really required, or was it just used while developing this PR?

crates/pessimistic-proof/src/local_exit_tree/mod.rs
```diff
@@ -1,55 +1,93 @@
...
+}
+
 /// Represents a local exit tree as defined by the LxLy bridge.
 #[serde_as]
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct LocalExitTree<H, const TREE_DEPTH: usize = 32>
 where
     H: Hasher,
-    H::Digest: Serialize + for<'a> Deserialize<'a>,
+    H::Digest: Copy + std::fmt::Debug + Serialize + for<'a> Deserialize<'a>,
```
Ekleog-Polygon, line 53: Same here

crates/pessimistic-proof/src/local_exit_tree/mod.rs
```diff
@@ -1,55 +1,93 @@
...
-    #[error("Leaf index overflow")]
-    LeafIndexOverflow,
-    #[error("Index out of bounds")]
-    IndexOutOfBounds,
-    #[error("Frontier index out of bounds")]
-    FrontierIndexOutOfBounds,
+impl<H, const TREE_DEPTH: usize> Default for LocalExitTree<H, TREE_DEPTH>
+where
+    H: Hasher,
+    H::Digest: std::fmt::Debug + Copy + Default + Serialize + for<'a> Deserialize<'a>,
```
Ekleog-Polygon, line 61: And here (I’ll not copy that again in this file, there’s probably more down below)

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -27,6 +19,22 @@ pub struct LocalNetworkState {
     pub nullifier_tree: NullifierTree<Keccak256Hasher>,
 }
 
+impl LocalNetworkState {
+    pub fn into_core(self) -> pessimistic_proof_core::NetworkState {
```
Ekleog-Polygon, line 23: Not sure it’s worth adding a function just to do `into()`, but maybe it helps type inference sometimes?

crates/pessimistic-proof/src/utils/smt.rs
```diff
@@ -331,7 +308,8 @@ where
         let mut siblings = vec![];
         let mut hash = self.root;
         let bits = key.to_bits();
-        for i in 0..DEPTH {
+
+        for item in bits.iter().take(DEPTH) {
```
Ekleog-Polygon, line 312:
```suggestion
        for bit in bits.iter().take(DEPTH) {
```

crates/pessimistic-proof-core/src/local_balance_tree.rs
```diff
@@ -0,0 +1,60 @@
...
+        key: TokenInfo,
+        path_to_update: &LocalBalancePath<H>,
+        old_balance: U256,
+        new_balance: U256,
+    ) -> Result<(), ProofError> {
+        self.root = path_to_update
+            .verify_and_update(
+                key,
+                H::Digest::from_u256(old_balance),
+                H::Digest::from_u256(new_balance),
```
Ekleog-Polygon, line 53: This feels like it could deserve a more semantic trait method, like `FromBalance::from_balance`?

Ekleog-Polygon (dismissed):
I mostly reviewed the code more than the concepts, as I’m not yet very familiar with PP, but please find a few nits below.

This being said, there’s no blocker, so I’m r+ing. Thank you for the split, that should make future changes simpler! :)

crates/agglayer-prover/src/executor/tests.rs
```diff
@@ -248,7 +254,8 @@ async fn executor_fails_because_of_concurrency_cpu() {
         1,
         service_fn(|r: Request| async move {
             tokio::time::sleep(Duration::from_secs(20)).await;
-            let Proof::SP1(mut proof) = Proof::new_for_test(&r.initial_state, &r.batch_header);
+            let Proof::SP1(mut proof) =
+                Proof::new_for_test(&r.initial_state.into(), &r.batch_header);
```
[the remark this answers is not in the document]
Ekleog-Polygon: Hmm... in this case I think they could be meaningful? This is test code, so it should definitely be covered by some tests, or it'd mean that these tests are not running (or this is dead code). Or did I miss something?

Anyway this is unrelated to this PR, but it looks weird to me :sweat_smile:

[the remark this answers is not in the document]
Freyskeyd: You can read [this](https://docs.codecov.com/docs/commit-status#patch-status) if you want to understand the difference between `project` coverage and `patch` coverage.

crates/pessimistic-proof-core/src/bridge_exit.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub amount: U256,
+
+    pub metadata: Option<Digest>,
+}
+
+const EMPTY_METADATA_HASH: Digest = Digest(hex!(
+    "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"
+));
+
+pub fn bridge_exit_hasher<T: Into<[u8; 20]>>(
```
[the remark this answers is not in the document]
Freyskeyd: This function is used by both `pp-lib::BridgeExit` and `pp-core::BridgeExit`, we want the `hash` function to be similar for those structs.
The generic can be removed, indeed, as we have `agglayer-primitives` types now.

crates/pessimistic-proof-core/src/nullifier_tree/mod.rs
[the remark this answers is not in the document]
Freyskeyd: It was defined like that in the original crate version, will try to `git mv` to reduce diff size

crates/pessimistic-proof-core/src/local_state.rs
```diff
@@ -0,0 +1,250 @@
...
+        // TODO: benchmark if BTreeMap is the best choice in terms of SP1 cycles
+        let mut new_balances = BTreeMap::new();
+        for (k, v) in &multi_batch_header.balances_proofs {
+            if new_balances.insert(*k, U512::from(v.0)).is_some() {
+                return Err(ProofError::DuplicateTokenBalanceProof(*k));
+            }
+        }
+
+        // Check batch_header.imported_exits_root
+        let imported_exits_root = commit_imported_bridge_exits(
```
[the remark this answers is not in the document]
Freyskeyd: The content didn't change. It is still a `root`

crates/pessimistic-proof-core/src/local_state.rs
```diff
@@ -0,0 +1,250 @@
...
+            // For message exits, the origin network in token info should be the origin
+            // network of the batch header.
+            if bridge_exit.is_message()
+                && bridge_exit.token_info.origin_network != multi_batch_header.origin_network
+            {
+                return Err(ProofError::InvalidMessageOriginNetwork);
+            }
+
+            // For ETH transfers, we need to check that the origin network is the L1 network
+            if bridge_exit.token_info.origin_token_address.is_zero()
```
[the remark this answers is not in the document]
Freyskeyd: The origin token address 0x00 is reserved for Ether, as described in the [protocol](https://docs.polygon.technology/agglayer/unified-bridge/).

crates/pessimistic-proof-test-suite/src/sample_data.rs
Freyskeyd: `NetworkId` is not a `u32` for `agglayer-types` only in `pp-core`

crates/pessimistic-proof/src/bridge_exit.rs
Freyskeyd: We would need to convert the `pp-lib::BridgeExit` to `pp-core::BridgeExit` to use what you suggest, which isn't what we want for every hash generation.

crates/pessimistic-proof-core/src/nullifier_tree/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    // TODO: Consider batching the updates per network for efficiency
+    pub fn verify_and_update(
+        &mut self,
+        key: NullifierKey,
+        path_to_update: &NullifierPath<H>,
+    ) -> Result<(), ProofError> {
+        self.root = path_to_update
+            .verify_and_update(
+                key,
+                H::Digest::from_bool(true),
```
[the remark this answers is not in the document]
Ekleog-Polygon: Then, would it make sense to instead have something like `MakeFromWasClaimed::from_was_claimed(bool)`?

crates/pessimistic-proof-test-suite/src/sample_data.rs
Ekleog-Polygon: Oh right! Sorry, I had misread the use statement

crates/pessimistic-proof/src/bridge_exit.rs
Ekleog-Polygon: I see, I missed the fact that the types were different

crates/pessimistic-proof-core/src/nullifier_tree/mod.rs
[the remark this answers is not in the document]
Freyskeyd: I don't have a strong opinion on that, but this kind of refactor isn't for this PR.
I also want to point out that this kind of structure needs to be as simple as possible, that why we're trying to not put too much product oriented things to make it more reusable.

crates/pessimistic-proof/src/local_exit_tree/mod.rs
Freyskeyd: The `Debug` was needed as the `pp-core::LocalExitTree` derive `Debug`. I didn't search too much on resolving this, I must say

iljakuklic: This is a consequence of the auto-derived `Debug` impl for `pessimistic_proof::local_exit_tree::LocalExitTree`. It does not place the correct trait bounds (which is my long-standing frustration with built-in auto-derived traits).

Here's a snippet of `cargo expand` output:
```rust
impl<H: ::core::fmt::Debug, const TREE_DEPTH: usize> ::core::fmt::Debug for LocalExitTree
```

Observe that it generates the `H: Debug` bound which is not correct. Should be `pp_core::LocalExitTree<H, TREE_DEPTH>: Debug`.

I do not see how to fix this other than implementing `Debug` on `pp::LocalExitTree` manually. I looked over crates.io if there are crates with a fixed derive macro with no luck but that was a couple of years back.

crates/pessimistic-proof-core/src/local_balance_tree.rs
Freyskeyd: Again, no strong opinion on this, but not the place for such update in this PR

Ekleog-Polygon (approved):
LGTM, thank you for having gone through the nits!

crates/pessimistic-proof/src/local_exit_tree/mod.rs
iljakuklic: For context:
* Rust issue: https://github.com/rust-lang/rust/issues/26925
* A crate mentioned there (not an endorsement, more research required): https://crates.io/crates/derive-where

Freyskeyd: I think we'll leave the `Debug` trait currently. It is helpful for debugging and until we have a better solution, I'm fine with it.

iljakuklic (approved):

PR #528, opened 2025-01-13
build(deps): bump deps

Ekleog-Polygon:
# Description

Take over https://github.com/agglayer/agglayer/pull/527. Includes bumping axum to 0.8.

Changelog review:

- All changelog entries look reasonably harmless
- RocksDB 0.22 -> 0.23 will lead to the rocksdb 8 -> 9 major version bump. This [should be harmless](https://github.com/facebook/rocksdb/wiki/RocksDB-Compatibility-Between-Different-Releases), but it may mean that we could have difficulty rolling back this upgrade in production (though in theory we shouldn’t). However, we definitely don’t want to stay on rocksdb 8 forever, so let’s just upgrade.

Fixes #527

## Breaking changes

In theory (according to its doc), bumping rocksdb from 8 to 9 should not be a breaking change. However it’s still a major upgrade, so I’m writing it out here for full clarity.

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/Cargo.toml
+++ b/crates/agglayer-node/Cargo.toml
@@ -18,8 +18,8 @@ hex.workspace = true
 hyper.workspace = true
 jsonrpsee = { workspace = true, features = ["full"] }
 parking_lot.workspace = true
-pin-project = "1.1.7"
-reqwest = "0.12.9"
+pin-project.workspace = true
+reqwest = "0.12.12"
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 serde_with.workspace = true
--- a/tests/integrations/Cargo.toml
+++ b/tests/integrations/Cargo.toml
@@ -8,7 +8,7 @@ license.workspace = true
 workspace = true
 
 [dependencies]
-alloy = { version = "0.6.4", features = ["full"] }
+alloy.workspace = true
 anyhow.workspace = true
 arc-swap.workspace = true
 buildstructor.workspace = true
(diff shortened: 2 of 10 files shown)


tests/integrations/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license.workspace = true
 workspace = true
 
 [dependencies]
-alloy = { version = "0.6.4", features = ["full"] }
+alloy = { version = "0.8.1", features = ["full"] }
```
iljakuklic, line 11:
```suggestion
alloy.workspace = true
```

crates/agglayer-node/Cargo.toml
```diff
@@ -17,8 +17,8 @@ hex.workspace = true
 hyper.workspace = true
 jsonrpsee = { workspace = true, features = ["full"] }
 parking_lot.workspace = true
-pin-project = "1.1.7"
-reqwest = "0.12.9"
+pin-project = "1.1.8"
```
iljakuklic, line 20: I'd say `pin-project` is a fairly generic dependency, could be moved to workspace deps.

iljakuklic (dismissed):
Looks good. Adding some comments, mostly to do with structuring `Cargo.toml` files rather than updating dependencies per se.

Ekleog-Polygon:
Done, thank you!

iljakuklic (dismissed):

atanmarko (commented):
Just maybe wait with the merge until `agglayer`/`provers` dependency is resolved

atanmarko (dismissed):

Ekleog-Polygon:
Makes sense, I’ll rebase and re-request review once #543 has landed :)

Ekleog-Polygon:
As per https://github.com/agglayer/agglayer/pull/505#pullrequestreview-2573086271 I’m reviving this PR, which should now be ready to land! 🎉  It’s the same changes as #547, ie. the changes originally in this PR plus splitting pessimistic-proof to not use the workspace (because we’ll often want to bump workspace deps without affecting PP)

iljakuklic (approved):

Issue #525, opened 2025-01-10
Add the MockProver to the `agglayer-prover`


atanmarko:
## Description

This ticket covers how to add the `MockProver` to the existing agglayer-prover. 

MockProver should be of the same type and treated equally as the `NetworkProver` or `CpuProver`.

## Acceptance

Implement the test running and verifying elf of the pessimistic proof program with the `MockProver`.

Run the existing tests/scenarios with the `MockProver`.

Those acceptances also need to be covered by insta snapshots.

### Configuration with MockProver

- Given a configuration with a MockProver
- When I parse it and get a ProvenConfig
- Then the parsed configuration should define that the primary prover is Mock and fallback prover is None

PR #537, opened 2025-01-20
feat: refactor prover config

### LINKED ISSUE #514: Refactor the `ProverConfig` and `Executor` to handle the different prover types in a more flexible way
## Description

This ticket covers how to update the existing `agglayer-prover` to be more flexible with handling the prover types.

The `ProverConfig` should have obligatory `primary` and optional `fallback` provers, and their type should be easily determined/configurable from the config file.

## Acceptance

Run the existing tests/kurtosis scenarios with the new configuration/executor.

Those acceptances need to be covered by `insta` snapshots.

### Configuration with NetworkProver

- Given a configuration with a NetworkProver
- When I parse it and get a `ProvenConfig`
- Then the parsed configuration should define that the primary prover is `NetworkProver` and fallback prover is `None`

### Configuration with CpuProver

- Given a configuration with a CpuProver
- When I parse it and get a `ProvenConfig`
- Then the parsed configuration should define that the primary prover is `CpuProver` and fallback prover is `None`

### Configuration with primary NetworkProver and fallback CpuProver

-  Given a configuration with NetworkProver and  CpuProver
- When I parse it and get a `ProvenConfig`
- Then the parsed configuration should define that the primary prover is `NetworkProver` and fallback prover is `CpuProver`

atanmarko:
# Description

The ProverConfig is changed to have obligatory primary and optional fallback prover, and their type should be easily determined/configurable from the config file.

Fixes #514 

## Breaking changes

Configuration file has been changed. Primary and fallback provers are now set in the following way:

```
[primary-prover.network-prover]
proving-timeout = 900

[fallback-prover.cpu-prover]
max-concurrency-limit = 100
proving-timeout = 900
```

## PR Checklist:


crates/agglayer-config/src/prover.rs
```diff
@@ -67,21 +64,47 @@ impl Default for ProverConfig {
...
+
+        let deserializer = toml::de::Deserializer::new(&reader);
+        serde::Deserialize::deserialize(deserializer)
+            .map_err(ConfigurationError::DeserializationError)
+    }
+}
+
+/// Type of the prover to be used for generation of the pessimistic proof
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub enum AgglayerProverType {
```
iljakuklic, line 91: Other config `enum`s use `#[serde(rename_all = "kebab-case")]`. See for example `agglayer_config::epoch::Epoch`. The config key names will then be `network-prover` instead of `NetworkProver` and such.

crates/agglayer-prover/src/executor.rs
```diff
@@ -33,8 +33,8 @@ pub(crate) const ELF: &[u8] =
 
 #[derive(Clone)]
 pub struct Executor {
-    network: Option<BoxCloneService<Request, Response, Error>>,
-    local: Option<BoxCloneService<Request, Response, Error>>,
+    primary: BoxCloneService<Request, Response, Error>,
+    fallback: Option<BoxCloneService<Request, Response, Error>>,
```
iljakuklic, line 37: Would it make sense for this to have an arbitrary number of fallbacks (including zero)? I.e. replace the `Option` with a sequence of some sort?

iljakuklic (commented):
Not a full review, just some things I noticed so far.

crates/agglayer-prover/src/executor.rs
atanmarko: We could. Do we need to at this time?

iljakuklic: > We could. Do we need to at this time? I mean would we realistically use that sequencer any time?

No, just a random suggestion. Just `Option` as a parameter is often a code smell that suggests a generalization is possible.

atanmarko: In this case, feels cleaner/simpler to ask the user to set a simple fallback prover that exists/doesn't exist in the toml file than to create an array to say I want one fallback prover. Happy to add this sequence if we meet the case where we need it.

iljakuklic (dismissed):

crates/agglayer-config/src/prover.rs
```diff
@@ -67,21 +64,48 @@ impl Default for ProverConfig {
...
-            gpu_prover: GpuProverConfig::default(),
+            primary_prover: AgglayerProverType::NetworkProver(NetworkProverConfig::default()),
+            fallback_prover: None,
             grpc: Default::default(),
         }
     }
 }
 
+impl ProverConfig {
+    pub fn try_load(path: &Path) -> Result<Self, ConfigurationError> {
```
Freyskeyd, line 75: Is it only for tests purpose? If so, can we flag it as is?

crates/agglayer-prover/src/executor.rs
```diff
@@ -87,59 +87,70 @@ impl Executor {
...
-                config.cpu_prover.max_concurrency_limit,
-                LocalExecutor {
-                    prover: Arc::new(prover),
-                    proving_key,
-                    verification_key,
-                },
-            ))
-        } else {
-            None
+        let primary = match &config.primary_prover {
```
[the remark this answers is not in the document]
atanmarko: I'll refactor it

crates/agglayer-config/src/prover.rs
atanmarko: At the moment it is for the test purpose. But it could be used in the code if needed.
Alternatively I should make a 
```
pub mod test {
     pub fn try_load_prover_config(path: &Path) -> Result<ProverConfig, ConfigurationError> {
}
```
as the `#[cfg(test)]` function could not be exported, even if used only in tests.
I think it would be better to stay as it is -  we have the same function for the `agglayer_config::Config`.

Freyskeyd: If a function needs to be exported to other crate, while aiming at being use for tests, we have a dedicated `feature` that cover that in other crate under: `testutils`.

I'm fine letting this function public if you think it can be valuable soon.

crates/agglayer-config/src/prover.rs
```diff
@@ -67,21 +64,48 @@ impl Default for ProverConfig {
...
+        let deserializer = toml::de::Deserializer::new(&reader);
+        serde::Deserialize::deserialize(deserializer)
+            .map_err(ConfigurationError::DeserializationError)
+    }
+}
+
+/// Type of the prover to be used for generation of the pessimistic proof
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub enum AgglayerProverType {
```
Ekleog-Polygon, line 92: I'd have likes adding a `FallbackProver` variant over setting default/fallback provers right in the config more. This being said, it should be enough for a first itération, considering AFAIU we don't care at all about breaking changes in configuration for now, as we're the only ones with AggLayer deployed.

crates/agglayer-config/src/prover.rs
```diff
@@ -180,7 +196,41 @@ impl GpuProverConfig {
...
+            proving_request_timeout: None,
+            proving_timeout: default_cpu_proving_timeout(),
+        }
+    }
+}
+
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct MockProverConfig {
```
Ekleog-Polygon, line 209: This is the one that "proves" anything without any actual verification, right? If so I'd say it needs big warnings to not be used in prod, like having the config option be like `unsafe-mock` or similar

crates/agglayer-config/src/prover.rs
```diff
@@ -180,7 +196,41 @@ impl GpuProverConfig {
 impl Default for GpuProverConfig {
     fn default() -> Self {
         Self {
-            enabled: default_activation_gpu_prover(),
+            max_concurrency_limit: default_max_concurrency_limit(),
+            proving_request_timeout: None,
+            proving_timeout: default_cpu_proving_timeout(),
```
Ekleog-Polygon, line 201: Nit: rename to not be "cpu", if it's used for all local timeouts.

crates/agglayer-config/tests/validate_deserialize.rs
```diff
@@ -88,3 +89,59 @@ fn prover_grpc_max_decoding_message_size() {
 
     assert_eq!(config.grpc.max_decoding_message_size, 100 * 1024 * 1024);
 }
+
+#[test]
+fn network_prover() {
+    let input = "./tests/fixtures/valide_config/prover_config_network_prover.toml";
```
Ekleog-Polygon, line 95:
```suggestion
    let input = "./tests/fixtures/validate_config/prover_config_network_prover.toml";
```

crates/agglayer-config/tests/validate_deserialize.rs
```diff
@@ -88,3 +89,59 @@ fn prover_grpc_max_decoding_message_size() {
...
+            proving_timeout: std::time::Duration::from_secs(600),
+        })
+    );
+
+    assert_eq!(config.fallback_prover, None);
+}
+
+#[test]
+fn cpu_prover() {
+    let input = "./tests/fixtures/valide_config/prover_config_cpu_prover.toml";
```
Ekleog-Polygon, line 111:
```suggestion
    let input = "./tests/fixtures/validate_config/prover_config_cpu_prover.toml";
```

And similar for the whole file, won't copy further

crates/agglayer-prover/src/executor.rs
```diff
@@ -87,59 +87,80 @@ impl Executor {
...
         } else {
-            None
-        };
-
-        if network.is_none() && local.is_none() {
-            panic!("No prover enabled");
+            Self {
+                primary,
+                fallback: None,
+            }
```
Ekleog-Polygon, line 162: This could be more concisely expressed as

```rust
Self {
    primary: Self::create_prover(&config.primary_prover),
    fallback: config.fallback_prover.as_ref().map(Self::create_prover),
}
```

crates/agglayer-prover/src/executor.rs
```diff
@@ -87,59 +87,80 @@ impl Executor {
...
+            AgglayerProverType::NetworkProver(network_prover_config) => {
+                debug!("Creating network prover executor...");
+                let network_prover = ProverClient::builder().network().build();
+                let (proving_key, verification_key) = network_prover.setup(ELF);
+                Self::build_network_service(
+                    network_prover_config
+                        .proving_request_timeout
+                        .unwrap_or_else(|| {
+                            network_prover_config.proving_timeout
+                                + NetworkProverConfig::DEFAULT_PROVING_TIMEOUT_PADDING
```
Ekleog-Polygon, line 109: Can we move that unwrap and addition to a method on network_prover_config, like you did for MockProver?

Ekleog-Polygon (approved):
LGTM, thank you! Here are just a few nits on format rather than contents :)

Freyskeyd (approved):

crates/agglayer-config/src/prover.rs
atanmarko: Actually it does produce some kind of proof and has verification - more details [here](https://github.com/agglayer/protocol-research/issues/58). If used in production verification would fail on the smart contract level, but I agree we could put this behind some mock feature.

crates/agglayer-config/src/prover.rs
atanmarko: I am not sure this is the approach I would take. `FallbackProver` is not type of the prover.

crates/agglayer-config/tests/validate_deserialize.rs
atanmarko: @Ekleog-Polygon Was already set by Simon before, didn't want to ruin the French touch :)

crates/agglayer-prover/src/executor.rs
atanmarko: Nice

crates/agglayer-prover/src/executor.rs
atanmarko: Will do it in the next PR on this topic

crates/agglayer-config/src/prover.rs
atanmarko: I'll add mock feature to the agglayer prover in the PR covering https://github.com/agglayer/agglayer/issues/525

PR #547, opened 2025-01-24
build(deps): bump all dependencies

Ekleog-Polygon:
Take over #527. Rebase of #528 onto 0.2.2. Includes bumping axum to 0.8.
    
## Changelog review:
    
All changelog entries look reasonably harmless
RocksDB 0.22 -> 0.23 will lead to the rocksdb 8 -> 9 major version bump. This should be harmless, but it may mean that we could have difficulty rolling back this upgrade in production (though in theory we shouldn’t). However, we definitely don’t want to stay on rocksdb 8 forever, so let’s just upgrade.
Fixes #527

## Breaking changes

In theory (according to its doc), bumping rocksdb from 8 to 9 should not be a breaking change. However it’s still a major upgrade, so I’m writing it out here for full clarity.

## PR Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas (N/A)
- [ ] I have made corresponding changes to the documentation (N/A)
- [ ] I have added or updated tests that comprehensively prove my change is effective or that my feature works (N/A)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,7 +51,7 @@ dependencies = [
 
 [[package]]
 name = "agglayer"
-version = "0.2.1"
+version = "0.1.0"
 dependencies = [
  "agglayer-config",
  "agglayer-node",
@@ -68,7 +68,7 @@ dependencies = [
 
 [[package]]
 name = "agglayer-aggregator-notifier"
-version = "0.2.1"
+version = "0.1.0"
 dependencies = [
  "agglayer-certificate-orchestrator",
  "agglayer-config",
@@ -94,7 +94,7 @@ dependencies = [
  "sp1-prover",
  "sp1-sdk",
  "test-log",
- "thiserror 2.0.8",
+ "thiserror 2.0.11",
  "tokio",
  "tokio-util",
  "tonic 0.12.3",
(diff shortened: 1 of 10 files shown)


iljakuklic (approved):
Looks good. Apparently, `Cargo.lock` has conflicts. Will re-approve if necessary after resolving the conflicts.

Ekleog-Polygon:
That’s because I forgot to set the right base branch: this targets 0.2.2. I have set the right branch now!

atanmarko (approved):

iljakuklic (dismissed):

crates/pessimistic-proof-program/elf/riscv32im-succinct-zkvm-elf
Freyskeyd, line 1: We don't want this to be updated on 0.2.2

Freyskeyd (changes requested):

crates/pessimistic-proof-program/elf/riscv32im-succinct-zkvm-elf
Ekleog-Polygon: Great point! It naturally stemmed from having `.workspace = true` dependencies from `pessimistic-proof-program`.

I just updated the PR to remove all these dependencies, and restore them to the previous, locked crate version that we already had.

atanmarko (approved):

Ekleog-Polygon:
Closing as per https://github.com/agglayer/agglayer/pull/505#pullrequestreview-2573086271 ; I’ll revive #528 with the new changes brought here instead

PR #554, opened 2025-01-27
refactor: move logic from rpc to a separate component

### LINKED ISSUE #544: Refactor the JSON-RPC api to extract the logic into another crate/service
## Description

This ticket defines the effort to move the logic of the `JSON-RPC` api into a dedicated service in order to share this between JSON-RPC and gRPC.

iljakuklic:
# Description

Move the logic from rpc implementation into a separate component, Agglayer service. This will allow it to be reused when introducing new external interfaces, initially gRPC.

* fixes #544

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/service/error.rs
+++ b/crates/agglayer-node/src/service/error.rs
@@ -1,0 +1,94 @@
+//! Error types for the top-level Agglayer service.
+
+use agglayer_contracts::{
+    polygon_rollup_manager::PolygonRollupManagerErrors, polygon_zk_evm::PolygonZkEvmErrors,
+};
+pub use agglayer_storage::error::Error as StorageError;
+pub use agglayer_types::Digest;
+use ethers::{contract::ContractError, providers::Middleware, types::H256};
+
+pub use crate::{
+    kernel::{
+        CheckTxStatusError, SettlementError, SignatureVerificationError, ZkevmNodeVerificationError,
+    },
+    rate_limiting::RateLimited as RateLimitedError,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum CertificateRetrievalError {
+    #[error(transparent)]
+    Storage(#[from] StorageError),
+
+    #[error("Data for certificate {certificate_id} not found")]
+    NotFound { certificate_id: Digest },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CertificateSubmissionError {
+    #[error(transparent)]
+    Storage(#[from] StorageError),
+
+    #[error("Failed to send the certificate to the orchestrator")]
+    OrchestratorNotResponsive,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum TxStatusError<Rpc: 'static + Middleware> {
+    #[error("Status retrieval error: {0}")]
+    StatusCheck(CheckTxStatusError<Rpc>),
+
+    #[error("Failed to get L1 block: {0}")]
+    L1BlockRetrieval(CheckTxStatusError<Rpc>),
+
+    #[error("Transaction {hash} not found")]
+    TxNotFound { hash: H256 },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum SendTxError<Rpc: 'static + Middleware> {
+    #[error("Rate limited: {0}")]
+    RateLimited(#[from] RateLimitedError),
+
+    #[error(transparent)]
+    SignatureError(#[from] SignatureVerificationError<Rpc>),
+
+    #[error("Rollup {rollup_id} not registered")]
+    RollupNotRegistered { rollup_id: u32 },
+
+    #[error("Zkevm error during dry run: {0:?}")]
+    DryRunZkEvm(PolygonZkEvmErrors),
+
+    #[error("Rollup manager error during dry run: {0:?}")]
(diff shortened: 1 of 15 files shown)


crates/agglayer-node/src/service/mod.rs
```diff
@@ -0,0 +1,390 @@
+use std::sync::Arc;
```
Ekleog-Polygon, line 1: I mostly did not review this file, assuming it is basically copy-paste from the rpc/mod.rs file. Is there anything else of note I should look at?

crates/agglayer-node/src/service/error.rs
```diff
@@ -0,0 +1,94 @@
...
+    NotFound { certificate_id: Digest },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CertificateSubmissionError {
+    #[error(transparent)]
+    Storage(#[from] StorageError),
+
+    #[error("Failed to send the certificate to the orchestrator")]
+    OrchestratorNotResposnsive,
```
Ekleog-Polygon, line 32:
```suggestion
    OrchestratorNotResponsive,
```

Ekleog-Polygon (approved):
LGTM! I assumed lots of it were copy-paste so I didn’t review in-depth

crates/agglayer-node/src/service/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+use crate::{kernel::Kernel, signed_tx::SignedTx};
+
+pub mod error;
+
+/// The RPC agglayer service implementation.
+pub(crate) struct AgglayerService<Rpc, PendingStore, StateStore, DebugStore> {
+    kernel: Kernel<Rpc>,
+    certificate_sender: mpsc::Sender<(NetworkId, Height, CertificateId)>,
+    pending_store: Arc<PendingStore>,
+    state: Arc<StateStore>,
```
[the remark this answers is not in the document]
iljakuklic: It's mostly a copy-paste from the service. However, at least `pending_store` and `state` are shared with the orchestrator which should not have access to `kernel`. While the use of `Arc`s could be optimized, relying on the outer `Arc` alone does not give us sufficient granularity.

crates/agglayer-node/src/service/mod.rs
iljakuklic: Yea mostly copy-paste but also some light code improvements which hopefully do not change behaviour.

crates/agglayer-node/src/service/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+    PendingStore: PendingCertificateWriter + PendingCertificateReader + 'static,
+    StateStore: StateReader + StateWriter + 'static,
+    DebugStore: DebugReader + DebugWriter + 'static,
+{
+    #[instrument(skip(self, tx), fields(hash, rollup_id = tx.tx.rollup_id), level = "info")]
+    pub async fn send_tx(&self, tx: SignedTx) -> Result<H256, SendTxError<Rpc>> {
+        let hash = format!("{:?}", tx.hash());
+        tracing::Span::current().record("hash", &hash);
+
+        info!(
```
atanmarko, line 87: Nit: we should probably use consistently structured logging format for all new log lines - discussed [here](https://github.com/agglayer/protocol-research/issues/59).

crates/agglayer-node/src/service/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+            error!(
+                error = %e,
+                hash,
+                "Failed to settle transaction {hash} on L1: {e}"
+            )
+        })?;
+
+        agglayer_telemetry::SETTLE.add(1, metrics_attrs);
+
+        info!(hash, "Successfully settled transaction {hash}");
```
atanmarko, line 164: Nit - should be enough. Also on other places.
```
 info!(hash, "Successfully settled transaction");
```

crates/agglayer-node/src/service/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+        self.certificate_sender
+            .send((
+                certificate.network_id,
+                certificate.height,
+                certificate.hash(),
+            ))
+            .await
+            .map_err(|error| {
+                error!("Failed to send certificate: {error}");
+                CertificateSubmissionError::OrchestratorNotResponsive
```
atanmarko, line 230: I would argue that in general better approach is to keep original errors, just wrap them and bubble up.

iljakuklic: The `SendError` returned here contains the original sent value which I wanted to get rid of.

crates/agglayer-node/src/service/mod.rs
iljakuklic: This is taken from the original code in the `rpc` module. I did want to minimize behaviour changes in this PR (although some log lines may have changed), so would prefer this as a follow up.

atanmarko: Agreed, we should do this consistently as a separate task.

Ekleog-Polygon (approved):

crates/agglayer-node/src/service/mod.rs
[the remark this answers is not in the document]
Ekleog-Polygon: Got it, makes sense!

PR #556, opened 2025-01-27

### LINKED ISSUE #553: Add a signature check in the `send_certificate` RPC to prevent DDOS
## Description

In order to prevent exploit, we should check the signature of the certificate before inserting it into the storage.

### Technical details

- [ ] Check the certificate's signature in `sendCertificate` RPC, using the same mechanism as the `FEP` (send_tx)
- [ ] Switch the storage calls from `insert header > insert pending` to `insert pending > insert header` to avoid having dangling headers

Ekleog-Polygon:
# Description

Fixes #553

## PR Checklist:


Ekleog-Polygon:
Closing in favor of #557, as this PR could not have passed CI due to being from an external repo

PR #557, opened 2025-01-28

Ekleog-Polygon:
# Description

Fixes https://github.com/agglayer/agglayer/issues/553

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/kernel/mod.rs
+++ b/crates/agglayer-node/src/kernel/mod.rs
@@ -6,7 +6,12 @@ use agglayer_contracts::{
     polygon_rollup_manager::{PolygonRollupManager, RollupIDToRollupDataReturn},
     polygon_zk_evm::PolygonZkEvm,
 };
-use ethers::prelude::*;
+use agglayer_types::Certificate;
+use ethers::{
+    contract::{ContractCall, ContractError},
+    providers::{Middleware, ProviderError},
+    types::{Address, TransactionReceipt, H160, H256, U64},
+};
 use thiserror::Error;
 use tracing::{info, instrument, warn};
 
@@ -158,9 +163,14 @@ pub(crate) enum SignatureVerificationError<RpcProvider>
 where
     RpcProvider: Middleware,
 {
-    /// The signer could not be recovered from the signature.
-    #[error("could not recover signer: {0}")]
-    CouldNotRecoverSigner(SignatureError),
+    /// The signer could not be recovered from the transaction signature.
+    #[error("could not recover transaction signer: {0}")]
+    CouldNotRecoverTxSigner(#[source] ethers::types::SignatureError),
+
+    /// The signer could not be recovered from the certificate signature.
+    #[error("could not recover certificate signer: {0}")]
+    CouldNotRecoverCertSigner(#[source] alloy::primitives::SignatureError),
+
     /// The signer of the proof is not the trusted sequencer for the given
     /// rollup id.
     #[error("invalid signer: expected {trusted_sequencer}, got {signer}")]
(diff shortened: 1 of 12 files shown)


crates/agglayer-telemetry/src/lib.rs
```diff
@@ -41,6 +41,9 @@ lazy_static! {
     pub static ref VERIFY_SIGNATURE: opentelemetry::metrics::Counter<u64> = global::meter(AGGLAYER_KERNEL_OTEL_SCOPE_NAME)
         .u64_counter("verify_signature")
         .with_description("Number of signature verifications")
+    // TODO: THIS IS WRONG. This is the number of times we checked the signer is the expected signer,
```
Freyskeyd, line 44: I'm not sure to follow, the `verify_signature` method is [recovering](https://github.com/agglayer/agglayer/blob/main/crates/agglayer-node/src/signed_tx.rs#L126) the `signer` address by using the `recover` method with the structured expected hash [here](https://github.com/agglayer/agglayer/blob/main/crates/agglayer-node/src/signed_tx.rs#L107).

If the recovered address matches the expected one (from L1 contract), it means that the signature is valid.
That's exactly what the [verify](https://docs.rs/ethers-core/2.0.14/src/ethers_core/types/signature.rs.html#77-84) method of `ethers` do, we just do it in a single method than two (calling `hash` and then `verify` on the signature directly)

crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -158,9 +163,14 @@ pub(crate) enum SignatureVerificationError<RpcProvider>
 where
     RpcProvider: Middleware,
 {
-    /// The signer could not be recovered from the signature.
-    #[error("could not recover signer: {0}")]
-    CouldNotRecoverSigner(SignatureError),
+    /// The signer could not be recovered from the transaction signature.
```
Freyskeyd, line 166:
```suggestion
    /// FEP (0.1): The signer could not be recovered from the [`SignedTx`].
```

Freyskeyd (commented):
This PR needs to be based on `release/0.2.2` branch

crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -304,7 +315,35 @@ where
...
+            });
+        }
+
+        Ok(())
+    }
+
+    /// Verify that the signer of the given [`Certificate`] is the trusted
+    /// sequencer for the rollup id it specified.
+    #[instrument(skip(self), level = "debug")]
+    pub(crate) async fn verify_certificate_signer(
```
iljakuklic, line 333: I'm not sure it's a good idea to put this into the `kernel` component. It is (if I understand it correctly) basically a backwards compatibility layer for the previous version of agglayer. Now placing it here allows you to reuse some code but it would better be factored out imho. Not sure how pressing this is though. Or am I mistaken here @Freyskeyd?

iljakuklic (approved):

crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -229,7 +229,7 @@ where
             return Err(Error::rollup_not_registered(tx.tx.rollup_id));
         }
 
-        self.kernel.verify_signature(&tx).await.map_err(|e| {
+        self.kernel.verify_tx_signer(&tx).await.map_err(|e| {
```
iljakuklic, line 232: The naming changes should probably be reverted. Verifying tx signer implies signature validity.

crates/agglayer-telemetry/src/lib.rs
Ekleog-Polygon: Oh yes sorry, I was planning to re-check this before submitting the PR and I completely forgot about it. Thank you for the clarification! I’ll remove this comment, adjust the naming and rebase to 0.2.2

Ekleog-Polygon:
Closing in favor of #558

PR #558, opened 2025-01-28

Ekleog-Polygon:
# Description

Fixes https://github.com/agglayer/agglayer/issues/553
Replaces https://github.com/agglayer/agglayer/pull/557 ; handles review comments and adds metrics for certificate signature verifications (with a separate label)

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/kernel/mod.rs
+++ b/crates/agglayer-node/src/kernel/mod.rs
@@ -6,7 +6,12 @@ use agglayer_contracts::{
     polygon_rollup_manager::{PolygonRollupManager, RollupIDToRollupDataReturn},
     polygon_zk_evm::PolygonZkEvm,
 };
-use ethers::prelude::*;
+use agglayer_types::Certificate;
+use ethers::{
+    contract::{ContractCall, ContractError},
+    providers::{Middleware, ProviderError},
+    types::{Address, TransactionReceipt, H160, H256, U64},
+};
 use thiserror::Error;
 use tracing::{info, instrument, warn};
 
@@ -158,9 +163,14 @@ pub(crate) enum SignatureVerificationError<RpcProvider>
 where
     RpcProvider: Middleware,
 {
-    /// The signer could not be recovered from the signature.
-    #[error("could not recover signer: {0}")]
-    CouldNotRecoverSigner(SignatureError),
+    /// FEP (0.1): The signer could not be recovered from the [`SignedTx`].
+    #[error("could not recover transaction signer: {0}")]
+    CouldNotRecoverTxSigner(#[source] ethers::types::SignatureError),
+
+    /// The signer could not be recovered from the certificate signature.
+    #[error("could not recover certificate signer: {0}")]
+    CouldNotRecoverCertSigner(#[source] alloy::primitives::SignatureError),
+
     /// The signer of the proof is not the trusted sequencer for the given
     /// rollup id.
     #[error("invalid signer: expected {trusted_sequencer}, got {signer}")]
@@ -170,6 +180,7 @@ where
         /// The trusted sequencer address.
         trusted_sequencer: Address,
     },
+
     /// Generic network error when attempting to retrieve the trusted sequencer
     /// address from the rollup contract.
     #[error("contract error: {0}")]
@@ -294,7 +305,7 @@ where
     /// Verify that the signer of the given [`SignedTx`] is the trusted
     /// sequencer for the rollup id specified in the proof.
     #[instrument(skip(self), level = "debug")]
-    pub(crate) async fn verify_signature(
+    pub(crate) async fn verify_tx_signature(
         &self,
         signed_tx: &SignedTx,
     ) -> Result<(), SignatureVerificationError<RpcProvider>> {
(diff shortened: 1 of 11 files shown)


crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -304,8 +315,40 @@ where
...
+        let sequencer_address = self
+            .get_trusted_sequencer_address(u32::from(cert.network_id))
+            .await?;
+
+        let signer = cert
+            .signer()
+            .map_err(SignatureVerificationError::CouldNotRecoverCertSigner)?;
+
+        // Convert signer from alloy to ethers
+        let signer = H160::from(signer.0 .0);
```
iljakuklic, line 348: Weird spacing in the parens. Irritatingly enough, it passes the format checks.

Ekleog-Polygon: This is mandatory: `signer.0.0` does not parse as valid Rust code, though I don’t remember the rationale it’s always been this way

crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -304,8 +315,40 @@ where
...
+        Ok(())
+    }
+
+    /// Verify that the signer of the given [`Certificate`] is the trusted
+    /// sequencer for the rollup id it specified.
+    #[instrument(skip(self), level = "debug")]
+    pub(crate) async fn verify_cert_signature(
+        &self,
+        cert: &Certificate,
+    ) -> Result<(), SignatureVerificationError<RpcProvider>> {
```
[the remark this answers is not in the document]
Ekleog-Polygon: I looked and I couldn’t find any pre-existing `verify_tx_signature` unit test that I could use as an example. Considering this must land for 0.2.2, I thought it’d be better to postpone these for when we rebase to main. Does that make sense?

crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -304,8 +315,40 @@ where
 
         let signer = signed_tx
             .signer()
-            .map_err(|e| SignatureVerificationError::CouldNotRecoverSigner(e))?;
+            .map_err(SignatureVerificationError::CouldNotRecoverTxSigner)?;
+
+        // k256 signature verification works by recovering the public key from the
+        // signature, and then checking that it is the expected one.
```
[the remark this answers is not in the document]
Ekleog-Polygon: Fixed :)

crates/agglayer-node/src/kernel/mod.rs
iljakuklic: > This is mandatory: `signer.0.0` does not parse as valid Rust code, though I don’t remember the rationale it’s always been this way

Ah, I see. Maybe it thinks `0.0` is a floar constant?

Ekleog-Polygon: Yup, that’s exactly the reason! By rationale, I mean I don’t really get why it cannot detect that it’s in a `.something` to infer-out float literals. My guess would be because float literals get recognized in the tokenizer before the parser, but I’m not 100% sure about that, as I think Rust left the LR category for GLR (I think?) a while ago. Or maybe I misremember and that was all discussions about the future of turbofish that’s still here, so maybe it hasn’t actually changed yet. Anyway, it would be invalid syntax ^^

crates/agglayer-node/src/kernel/mod.rs
```diff
@@ -304,8 +315,40 @@ where
...
+        let sequencer_address = self
+            .get_trusted_sequencer_address(u32::from(cert.network_id))
+            .await?;
+
+        let signer = cert
+            .signer()
+            .map_err(SignatureVerificationError::CouldNotRecoverCertSigner)?;
+
+        // Convert signer from alloy to ethers
+        let signer = H160::from(signer.0 .0);
```
Freyskeyd, line 348: Maybe you can do something like that?

```suggestion
        let signer: H160 = cert
            .signer()
            .map_err(SignatureVerificationError::CouldNotRecoverCertSigner)
            .map(|signer| signer.into_array().into())?;
```

crates/agglayer-node/src/rpc/mod.rs
```diff
@@ -336,29 +340,41 @@ where
...
+        self.kernel.verify_cert_signature(&certificate).await.map_err(|e| {
+            error!(error = %e, hash = hash_string, "Failed to verify the signature of certificate {hash}: {e}");
+            Error::signature_mismatch(e)
+        })?;
+
+        let metrics_attrs = &[
+            KeyValue::new("rollup_id", certificate.network_id.to_string()),
+            KeyValue::new("type", "cert"),
+        ];
+        agglayer_telemetry::VERIFY_SIGNATURE.add(1, metrics_attrs);
```
Freyskeyd, line 360: I don't think we should mix the two flow together on metrics.
And I'm not sure if it is really valuable to track this one for PP path. Maybe a better one would be to have the failure one

crates/agglayer-node/src/kernel/mod.rs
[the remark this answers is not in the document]
Ekleog-Polygon: Done :)

crates/agglayer-node/src/rpc/mod.rs
Ekleog-Polygon: Here they’re not actually mixed: we’ll have two series, one with `type=tx` and one with `type=cert`. Without it, the "verify signature" name would become wrong.

Do you have any other idea for how to present this to the UI? I would expect renaming VERIFY_SIGNATURE to VERIFY_TX_SIGNATURE to likely be problematic from a backward-compatibility point of view :/ (especially, more so than just adding a `type=tx` filter that downstream users could want to add if they wanted to get the previous meaning of verify-signature).

And actually, AFAIU SignedTx is hopefully going away soon. So hopefully, verify-signature will soon only have type=cert records?

Freyskeyd: Those metrics are used by others teams and dashboard, I would prefer not to touch them to not break anything

Ekleog-Polygon: Hmm ok. I’m adjusting to still add the `type=tx` label, as AFAIK it cannot break anything. This way at least if we add a new type later we won’t have to think about this. But I’m now removing the new `type=cert`label for now and we can reconsider later :)

Does that make sense?

Freyskeyd (approved):

PR #565, opened 2025-01-29

### LINKED ISSUE #553: Add a signature check in the `send_certificate` RPC to prevent DDOS

Ekleog-Polygon:
# Description

Forward-port #558 to main

Closes #553 

## PR Checklist:

### DIFF
--- a/crates/agglayer-node/src/kernel/tests.rs
+++ b/crates/agglayer-node/src/kernel/tests.rs
@@ -9,6 +9,7 @@ use agglayer_contracts::polygon_rollup_manager::{
     RollupIDToRollupDataCall, RollupIDToRollupDataReturn, VerifyBatchesTrustedAggregatorCall,
 };
 use agglayer_contracts::polygon_zk_evm::{TrustedSequencerCall, TrustedSequencerReturn};
+use agglayer_types::Certificate;
 use ethers::core::utils;
 use ethers::prelude::*;
 use ethers::signers::LocalWallet;
@@ -214,9 +215,9 @@ async fn interop_executor_verify_zkp_failure() {
     mock.assert_request("eth_call", [tx_verify_batch, block])
         .unwrap();
 }
-/// Test that check if the verify_signature method
+/// Test that checks if the verify_tx_signature method
 #[tokio::test]
-async fn interop_executor_verify_signature() {
+async fn interop_executor_verify_tx_signature() {
     let config = Arc::new(Config::new_for_test());
 
     let (provider, mock) = providers::Provider::mocked();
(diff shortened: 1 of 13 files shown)


iljakuklic (approved):

crates/agglayer-node/src/kernel/tests.rs
```diff
@@ -214,9 +215,9 @@ async fn interop_executor_verify_zkp_failure() {
     mock.assert_request("eth_call", [tx_verify_batch, block])
         .unwrap();
 }
-/// Test that check if the verify_signature method
+/// Test that checks if the verify_tx_signature method
```
atanmarko, line 218:
```suggestion
/// Test that checks if verify_tx_signature method is working
```

crates/agglayer-node/src/kernel/tests.rs
```diff
@@ -336,6 +337,62 @@ async fn interop_executor_verify_signature_proof_signer() {
     ));
 }
 
+/// Test that checks if the verify_cert_signature method
```
atanmarko, line 340: Update comment

crates/agglayer-node/src/rpc/tests/send_certificate.rs
```diff
@@ -98,3 +101,50 @@ async fn send_certificate_method_can_be_called_and_fail() {
 
     assert!(res.is_err());
 }
+
+#[test_log::test(tokio::test)]
+async fn send_certificate_method_requires_known_signer() {
+    let _ = tracing_subscriber::FmtSubscriber::builder()
```
atanmarko, line 107: Do you need to explicitly initialize this when you use `test_log`?

atanmarko (approved):

crates/agglayer-node/src/rpc/tests/send_certificate.rs
Ekleog-Polygon: It seems to change the behavior: I tried removing it both here and in the `send_certificate_method_can_be_called` test I copy-pasted it from, and in both cases it made `INFO`-level logs appear.

This being said, even without it, setting `RUST_LOG=trace` also logged more stuff, and the `INFO` logs are not so verbose that we’d benefit that much from not having them by default.

Hence, I’ll remove them. I’ll do so in a follow-up PR, just like your other two comments, because this PR seems to have matched all merge conditions and to be otherwise ready to merge :)

PR #572, opened 2025-01-31
feat: prepare ci for enabling merge queues

Ekleog-Polygon:
# Description

Enable all required tests for merge queues.

Once this lands, we’ll still need to configure merge queues in the github admin interface, as per [the documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue).

Incidentally, removed syntax errors from `workflow_dispatch` triggers.

## PR Checklist:

### DIFF
--- a/.github/workflows/pr-checking.yml
+++ b/.github/workflows/pr-checking.yml
@@ -1,4 +1,4 @@
-name: Checking PR semantic
+name: Checking PR title
 
 on:
   pull_request_target:
@@ -23,12 +23,3 @@ jobs:
         with:
           ignoreLabels: |
             release
-
-  commits:
-    name: Validate PR commits
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - uses: wagoid/commitlint-github-action@v6
--- a/.github/workflows/commit-checking.yml
+++ b/.github/workflows/commit-checking.yml
@@ -1,0 +1,22 @@
+name: Checking commit message
+
+on:
+  push:
+    branches:
+      - main
+  merge_group:
+
+permissions:
+  contents: read
+  pull-requests: read
+  statuses: write
+
+jobs:
+  commits:
+    name: Validate PR commits
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: wagoid/commitlint-github-action@v6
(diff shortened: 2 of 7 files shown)


.github/workflows/pr-checking.yml
```diff
@@ -23,12 +23,3 @@ jobs:
...
             release
-
-  commits:
-    name: Validate PR commits
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - uses: wagoid/commitlint-github-action@v6
```
Freyskeyd, line 34: I would prefer keeping that as the `squash and merge` will combine every commits by default.
Meaning that if a commit is malformed, we may have rejected PR because of that

Ekleog-Polygon: Hmm… So I was actually thinking of that as a feature. Basically:
- Currently, if any commit is malformed in the PR, we’ll fail CI
- With my proposed changes, if the result of the squash-and-merge is malformed, we’ll fail CI

So it all boils down to what we want to ensure, and what the likelihood of rejection is.

For what we want to ensure, I’d argue that ensuring properties on the result of the squash-and-merge would be the best option, because it’s what actually enters the git history. We don’t really care about the contents of individual commit messages themselves, only about the contents of the resulting squashed commit.

As for the rejection likelihood:
- The result of the squash-and-merge is basically the PR title (that we already check) and then, in the commit body, the concatenation of commit bodies
- So, AFAIU, any squash-and-merge of valid commits would be a valid commit
- OTOH, if any commit is invalid, CI currently fails
- But with the suggested changes, CI would not fail, and we would just need to edit the message appropriately when clicking the "add to merge queue for squashing" button
- In particular, we could merge dependabot PRs without any special case in the git commit checks, by just deleting the (pretty useless) long commit bodies upon clicking the merge button
- I’d even argue that most of the time, people clicking "squash and merge" would want to delete the default commit messages — and even if they don’t, if the commits were initially valid, they’ll stay valid

What’s particularly annoying to me with CI failing for wrong commit messages is, that it forces us to force-push to the branch to "fix" the commit message. Yet, force-pushing is pretty bad for various reasons, not least the review time that gets demultiplied. I’m actually trying to prepare an explanation for the next meeting, to push for easier-to-review git handling — I’ll add it to the agenda if I have enough time to get my demonstration ready by then.

All that to say, I quite strongly think that we should not have this check for the PR CI, and only for the merge queue.

Does that make sense?

Ekleog-Polygon:
@Freyskeyd I think this matches more or less what you’ve presented in the meeting today, do you have the slides available somewhere so I can adjust the last remaining bits?

Ekleog-Polygon:
Closing, as it’s being replaced with https://github.com/agglayer/agglayer/pull/585

PR #574, opened 2025-01-31
build: update commitlint for dependabot

Freyskeyd:
# Description

This PR is preventing dependabot to fails on the commit message.
Until they provide a way to customize this, we ignore the checks.

## PR Checklist:

### DIFF
--- a/.github/commitlint.config.mjs
+++ b/.github/commitlint.config.mjs
@@ -1,0 +1,4 @@
+export default {
+    extends: ["@commitlint/config-conventional"],
+    ignores: [(message) => /^build\(deps\): bump .*$/m.test(message)],
+};
--- a/.github/workflows/pr-checking.yml
+++ b/.github/workflows/pr-checking.yml
@@ -26,6 +26,7 @@ jobs:
 
   commits:
     name: Validate PR commits
+    if: (github.actor != 'dependabot[bot]') && (contains(github.head_ref, 'dependabot/') == false)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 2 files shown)


.github/commitlint.config.mjs
```diff
@@ -0,0 +1,4 @@
+export default {
+    extends: ["@commitlint/config-conventional"],
+    ignores: [(message) => /^build\(deps\): bump .*$/m.test(message)],
```
iljakuklic, line 3: From now on, I will start all my PRs with `build(deps): bump <whatever>` :smiling_imp:.

On a bit more serious note, is it not possible to filter by author?

Freyskeyd: Ahah let me add another rule :D 

I didn't find anyway to filter per author in commitlint directly

iljakuklic (approved):

.github/workflows/pr-checking.yml
```diff
@@ -26,9 +26,12 @@ jobs:
 
   commits:
     name: Validate PR commits
+    if: (github.actor != 'dependabot[bot]') && (contains(github.head_ref, 'dependabot/') == false)
```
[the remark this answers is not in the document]
Ekleog-Polygon: With this, dependabot will not have commitlints, nor any PR starting with `build(deps)`.

I’m fine with it, but TBH would rather have merge queues, both for convenience (we’ll stop having to re-approve the same PR so many times) and for efficiency (because we will then be able to remove this dependabot-specific rule).

So I’d say if we expect to have merge queues within one month, we can probably just live with things as-is, and keep ensuring clean history. But if we don’t expect to have them, then it might be better to land this indeed, to allow us to merge dependabot PRs.

WDYT?

Ekleog-Polygon (approved):

0xaatif:
@Freyskeyd is this still relevant?

PR #585, opened 2025-02-03
build: update ci jobs to match new organization

Freyskeyd:
# Description

This PR is refactoring the different jobs and introduce a new set of workflows to cover the future merge queue.

## PR Checklist:


.github/workflows/coverage.yml
```diff
@@ -4,11 +4,10 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
```
Ekleog-Polygon, line 8: Hmm… I seemed to remember that code coverage CI would stay on PRs (even if non-blocking), so that we at least keep informing developers and reviewers about which lines of code are (un)covered. Here, moving them to merge queue would make these checks mostly useless I think?

Ekleog-Polygon (commented):
Looks like this deletes e2e tests and docker image publishing (`test:e2e.yml` and `util:docker.yml` that I’m not seeing moved to merge queue either). I didn’t see that in the PR description, is it on purpose?

.github/workflows/coverage.yml
Freyskeyd: mmh indeed, I'll put them back!

Ekleog-Polygon (approved):
LGTM! Just one thing: with these changes, `docker-build-local` and `e2e-kurtosis-cdk` will no longer run before merge groups. This being said, I’m not sure what tests are exactly there, nor whether it’s actually useful to have them run before? If they fail very rarely (and I’ve never seen them fail yet), we can probably keep them on merge groups :)

Ekleog-Polygon:
Also, 🎉 🎉 🎉  merge queues! 🎉 🎉 🎉

Freyskeyd:
> LGTM! Just one thing: with these changes, `docker-build-local` and `e2e-kurtosis-cdk` will no longer run before merge groups. This being said, I’m not sure what tests are exactly there, nor whether it’s actually useful to have them run before? If they fail very rarely (and I’ve never seen them fail yet), we can probably keep them on merge groups :)

Those tests are e2e tests that can be really long to run, they don't expect to fail unless there is an integration issue.

Ekleog-Polygon:
👍 SGTM!

iljakuklic (approved):
Looks good to me. I am, however, not an expert on github job definition files, so take this with a grain of salt.

Issue #587, opened 2025-02-04
Batch Merkle proofs for better PP cycle count


Nashtare:
## Description

The pessimistic proof program main overhead in cycle count at the moment is the import of bridge exits, mainly due to the various tree proof verifications (balance and nullifiers).
An example profiling trace (PP for 10 bridge exits & 20 imported bridge exits) available [here](https://share.firefox.dev/3EiTnWk) shows that `SmtNonInclusionProof::verify_and_update()` takes about 30% of the total cycles count, while `ImportedBridgeExit::verify_path()` takes about 20%, both accounting for half of the total PP program cycle count.

## Proposal

Instead of iteratively verifying each imported bridge exit independently, the Merkle proofs of (non)-inclusion could be batched and verified together to save cycles. Depending on the number of imported bridge exits to verify, this could remove from 10% to 50% of the number of nodes stored and needed to verify.

Issue #586, opened 2025-02-04
Use `rkyv` zero-copy serialization for better PP cycle count


Nashtare:
## Description

The pessimistic proof program is currently heavily impacted by SP1 default deserialization (bincode and serde under the hood of `sp1_io`.
An example profiling trace (PP for 10 bridge exits & 20 imported bridge exits) available [here](https://share.firefox.dev/3EiTnWk) shows it takes about 37% of the total cycles count.

## Proposal

Similarly to [op-succinct](https://github.com/succinctlabs/op-succinct/tree/main) state initialization, we should benefit from zero-copy deserialization with a library like `rkyv`, where the start of the PP program would go from:

```rust
    let initial_state = sp1_zkvm::io::read::<NetworkState>();
    let batch_header = sp1_zkvm::io::read::<MultiBatchHeader<Keccak256Hasher>>();
```

to 

```rust
    let initial_state_bytes = sp1_zkvm::io::read_vec();
    let initial_state = NetworkState::from_raw_bytes(initial_state_bytes);
    let batch_header_bytes = sp1_zkvm::io::read_vec();
    let batch_header = MultiBatchHeader<Keccak256Hasher>::from_raw_bytes(batch_header_bytes);
```

The benefits lies in getting rid of most of the overhead induced by `serde` complex deserialization.

### Note

`rkyv` zero-copy deserialization requires security guarantees in the shape of the data it is being provided. While fast, unchecked deserialization is acceptable if the data being fed to the deserializer is "controlled" (i.e. checked by some known entity) but may yield undefined behavior on the prover side otherwise. In the event the PP generation was delegated to chains themselves, it may be preferable to use the [validation API](https://docs.rs/rkyv/latest/rkyv/fn.access.html) to prevent any inconsistency.

iljakuklic:
Just for reference, this is also mentioned in succinct docs: https://docs.succinct.xyz/docs/sp1/optimizing-programs/basics#zero-copy-deserialization

PR #594, opened 2025-02-05
perf: Improve pessimistic-proof program to reduce # of cycles

Nashtare:
# Description

Introduces several improvements to the PP program in order to reduce the number of cycle counts. Below is a snapshot of the # of cycles for main VS this branch, for different number of (I) imported bridge exits and (E) bridge exits (3 runs each):

| Branch  | 10 E / 20 I | 10 E / 50 I | 30 E / 50 I | 100 E / 200 I |
| -------- | ------- | ------- | ------- | ------- |
| Old  | 8701768 | 19894737 | 20019004 | 77819172 |
| New | 7324101 | 16677887 | 16801915 | 65342788 |
| Improvement | 15.83% | 16.17% | 16.07% | 16.03% |

## Additions and Changes

The list of changes is the following:

- removed recomputation of `to_bits`: this was being called twice in `verify_and_update()`, it's now being computed once and reused
- added some `inlining` to reduce function call overhead for tiny methods
- tweaked very lightly some functions to remove branching (minor speedup) 
- precomputed `empty_hash_at_height` instead of recomputing them at every call of `get_root()`
- changed `NetworkId` to be 16-bit long instead of 32 bits: there was a comment already mentioning this. Bringing it down to 16 bits (i.e. 65536 possible chains) does remove a consequent number of `keccak` permutations when hashing the tries due to their smaller depth, yielding important savings

The last change (`NetworkId` implicit inner type as `u16`) hasn't been fully propagated to the rest of the codebase, given this has a _major_ impact on the API. Happy to pursue this / tackle it differently if needed.

Unrelated to other improvement suggestions (#586 and #587)

## PR Checklist:

### DIFF
--- a/crates/pessimistic-proof-core/src/bridge_exit.rs
+++ b/crates/pessimistic-proof-core/src/bridge_exit.rs
@@ -70,18 +70,18 @@ const EMPTY_METADATA_HASH: Digest = Digest(hex!(
 
 pub fn bridge_exit_hasher(
     leaf_type: u8,
-    origin_network: u32,
+    origin_network: u16,
     origin_token_address: Address,
-    dest_network: u32,
+    dest_network: u16,
     dest_address: Address,
     amount: U256,
     metadata: Option<Digest>,
 ) -> Digest {
     keccak256_combine([
         [leaf_type].as_slice(),
-        &u32::to_be_bytes(origin_network),
+        &u16::to_be_bytes(origin_network),
         origin_token_address.as_slice(),
-        &u32::to_be_bytes(dest_network),
+        &u16::to_be_bytes(dest_network),
         dest_address.as_slice(),
         &amount.to_be_bytes::<32>(),
         &metadata.unwrap_or(EMPTY_METADATA_HASH).0,
@@ -94,9 +94,9 @@ impl BridgeExit {
     pub fn hash(&self) -> Digest {
         bridge_exit_hasher(
             self.leaf_type as u8,
-            self.token_info.origin_network,
+            self.token_info.origin_network as u16,
             self.token_info.origin_token_address,
-            self.dest_network,
+            self.dest_network as u16,
             self.dest_address,
             self.amount,
             self.metadata,
--- a/crates/pessimistic-proof-core/src/global_index.rs
+++ b/crates/pessimistic-proof-core/src/global_index.rs
@@ -23,6 +23,7 @@ pub struct GlobalIndex {
 impl GlobalIndex {
     const MAINNET_FLAG_OFFSET: usize = 2 * 32;
 
+    #[inline(always)]
     pub fn network_id(&self) -> NetworkId {
         if self.mainnet_flag {
             0
@@ -32,12 +33,18 @@ impl GlobalIndex {
     }
 
     pub fn hash(&self) -> Digest {
-        let global_index: U256 = (*self).into();
-        keccak256(global_index.as_le_slice())
+        let mut bytes = [0u8; 32];
+        bytes[0..4].copy_from_slice(&self.leaf_index.to_le_bytes());
+        bytes[4..8].copy_from_slice(&self.rollup_index.to_le_bytes());
+        if self.mainnet_flag {
+            bytes[8] |= 0x01;
+        }
+        keccak256(&bytes)
     }
 }
 
 impl From<GlobalIndex> for NullifierKey {
+    #[inline]
     fn from(value: GlobalIndex) -> Self {
         Self {
             network_id: value.network_id(),
--- a/crates/pessimistic-proof-core/src/local_balance_tree.rs
+++ b/crates/pessimistic-proof-core/src/local_balance_tree.rs
@@ -11,9 +11,9 @@ use crate::{
     ProofError,
 };
 
-/// The key is [`TokenInfo`] which can be packed into 192 bits (32 for network
+/// The key is [`TokenInfo`] which can be packed into 176 bits (16 for network
 /// id and 160 for token address).
-pub const LOCAL_BALANCE_TREE_DEPTH: usize = 192;
+pub const LOCAL_BALANCE_TREE_DEPTH: usize = 176;
 
 // TODO: This is basically the same as the nullifier tree, consider refactoring
 /// A commitment to the set of per-network nullifier trees maintained by the
--- a/crates/pessimistic-proof-core/src/local_exit_tree/mod.rs
+++ b/crates/pessimistic-proof-core/src/local_exit_tree/mod.rs
@@ -26,6 +26,11 @@ where
     /// computed by log2(leaf_count). After that, all values are zeroed out.
     #[serde_as(as = "[_; TREE_DEPTH]")]
     pub frontier: [H::Digest; TREE_DEPTH],
+
+    /// `empty_hash_at_height[i]` is the root of an empty Merkle tree of depth
+    /// `i`.
+    #[serde_as(as = "[_; TREE_DEPTH]")]
+    pub empty_hash_at_height: [H::Digest; TREE_DEPTH],
 }
 
 #[derive(Clone, Debug, Error, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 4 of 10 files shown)


crates/pessimistic-proof-core/src/local_state.rs
```diff
@@ -90,7 +90,6 @@ impl NetworkState {
             });
         }
 
-        // TODO: benchmark if BTreeMap is the best choice in terms of SP1 cycles
```
Nashtare, line 93: Note that I have benchmarked independently the switch from `BTreeMap` to `HashMap` (with preallocation) for all 4 sample sizes listed in the PR description, and it didn't bring any improvement (performances actually worsened negligibly).

crates/pessimistic-proof-core/src/bridge_exit.rs
```diff
@@ -70,18 +70,18 @@ const EMPTY_METADATA_HASH: Digest = Digest(hex!(
 
 pub fn bridge_exit_hasher(
     leaf_type: u8,
-    origin_network: u32,
+    origin_network: u16,
```
Freyskeyd, line 73: I think that the spec defines that `networkId` are `u32`, @hadjiszs do you think we could scope this to only the `program`?

crates/pessimistic-proof-core/src/local_balance_tree.rs
```diff
@@ -11,9 +11,9 @@ use crate::{
     ProofError,
 };
 
-/// The key is [`TokenInfo`] which can be packed into 192 bits (32 for network
+/// The key is [`TokenInfo`] which can be packed into 176 bits (16 for network
 /// id and 160 for token address).
-pub const LOCAL_BALANCE_TREE_DEPTH: usize = 192;
+pub const LOCAL_BALANCE_TREE_DEPTH: usize = 176;
```
Freyskeyd, line 16: This is going to break the network state persisted in the storage I guess.
Does the switch from `u32` to `u16` really brings a lot of improvement?

crates/pessimistic-proof-core/src/global_index.rs
```diff
@@ -32,12 +33,18 @@ impl GlobalIndex {
...
     pub fn hash(&self) -> Digest {
-        let global_index: U256 = (*self).into();
-        keccak256(global_index.as_le_slice())
+        let mut bytes = [0u8; 32];
+        bytes[0..4].copy_from_slice(&self.leaf_index.to_le_bytes());
+        bytes[4..8].copy_from_slice(&self.rollup_index.to_le_bytes());
+        if self.mainnet_flag {
+            bytes[8] |= 0x01;
+        }
+        keccak256(&bytes)
```
Ekleog-Polygon, line 42: Have you tried using `#[inline(always)]` on the trait method, rather than copying it here? I'd expect LLVM to be able to detect that the `as_le_slice` is a noop, though I could be wrong. If I'm wrong, I'd suggest having a `GlobalIndex::as_le_slice` function used by both this and the From impl to avoid code duplication

crates/pessimistic-proof-core/src/local_exit_tree/mod.rs
```diff
@@ -26,6 +26,11 @@ where
     /// computed by log2(leaf_count). After that, all values are zeroed out.
     #[serde_as(as = "[_; TREE_DEPTH]")]
     pub frontier: [H::Digest; TREE_DEPTH],
+
+    /// `empty_hash_at_height[i]` is the root of an empty Merkle tree of depth
+    /// `i`.
+    #[serde_as(as = "[_; TREE_DEPTH]")]
+    pub empty_hash_at_height: [H::Digest; TREE_DEPTH],
```
Ekleog-Polygon, line 33: Instead of having this variable, can we have a `const` table, be it generated from a `const fn`, generated by a build script, or just generated once and hardcoded with instructions to repro the generation?

I'd expect loading ~6k in program size to be faster than generating said 6k on each program run, though I could be wrong.

crates/pessimistic-proof-core/src/nullifier_tree.rs
```diff
@@ -44,15 +42,16 @@ pub struct NullifierKey {
...
+impl ToBits<48> for NullifierKey {
+    fn to_bits(&self) -> [bool; 48] {
+        let mut bits = [false; 48];
+        for i in 0..16 {
+            bits[i] = (self.network_id >> i) & 1 == 1;
+        }
+        for i in 16..48 {
+            bits[i] = (self.let_index >> (i - 16)) & 1 == 1;
+        }
+        bits
```
Ekleog-Polygon, line 54: I'm curious, did this change bring any improvement? It's something I'd expect LLVM to easily optimize out itself, though I'm always surprised by both optimizations done and not done.

This being said, IMO this new code is cleaner anyway, so :shrug:

crates/pessimistic-proof-core/src/utils/smt.rs
```diff
@@ -94,10 +108,10 @@ where
     where
         K: ToBits<DEPTH> + Copy,
     {
-        if !self.verify(key, old_value, root) {
+        let bits = key.to_bits();
+        if !self.verify_with_bits(bits, old_value, root) {
```
Ekleog-Polygon, line 112: Can we kill the `verify` function, or at least make it reuse the `verify_with_bits` function, to avoid code duplication? And same below.

crates/pessimistic-proof/src/local_exit_tree/mod.rs
```diff
@@ -104,10 +111,17 @@ where
     /// Creates a new [`LocalExitTree`] from its parts: leaf count, and
     /// frontier.
     pub fn from_parts(leaf_count: u32, frontier: [H::Digest; TREE_DEPTH]) -> Self {
+        let mut empty_hash_at_height = [H::Digest::default(); TREE_DEPTH];
+        for i in 1..TREE_DEPTH {
+            empty_hash_at_height[i] =
+                H::merge(&empty_hash_at_height[i - 1], &empty_hash_at_height[i - 1]);
+        }
+
```
Ekleog-Polygon, line 119: To avoid this duplication, have `new` call `from_parts`?

Ekleog-Polygon (commented):
Not approving yet due to the NetworkId question, but overall LGTM!

Please find a few review comments below :)

crates/pessimistic-proof-core/src/global_index.rs
Nashtare: Ah I haven't, went for manually doing it :) 
Let me benchmark both and see how they differ

crates/pessimistic-proof-core/src/local_balance_tree.rs
Nashtare: Roughly speaking, the PP program currently consists in:
- 40% serialization
- 60% SMT proofs verification and other tree operations

Hence, yes pruning 16 layers in the tree _does_ have a serious impact (it's the most noticeable change here, by far).

As for @Ekleog-Polygon suggestion, higher aritiy trees indeed have tradeoff. The inclusion / non-inclusion proofs would be larger (passing 3 siblings instead of 1 per layer, with a depth halved which incurs a 50% authentication proof size increase). On the other hand as pointed the rate of keccak256 is 1088, so can fit 4 elements.

Overall, there should be gains in the portion related to SMTs, with an additional overhead in the initial deserialization (but which could be greatly amortized by #586) so I'd be in favour of such :) (should we be ok for storage breaking changes). We could probably shave off another 10/20% cycles.

crates/pessimistic-proof-core/src/local_exit_tree/mod.rs
Nashtare: No you're definitely right, no need to recompute it once per program anyway, I'll tweak it.

crates/pessimistic-proof-core/src/nullifier_tree.rs
Nashtare: yeah the function call + branching incurs a very slight overhead compared to the newer version (and 64 / 48 bits seemed to high to just _inline_).

crates/pessimistic-proof-core/src/bridge_exit.rs
hadjiszs: hm I think we can't modify it there, the hash of the bridge exits should stay the same otherwise the re-computed local exit root will differ with everywhere else on the stack (L1, cdk, etc).

if we want to reduce the size taken for one `NetworkId`, it could only be done on the balance tree and the nullifier tree, because only the agglayer + pp compute those two trees

However I wonder how this can be upgraded given that the PP is already in mainnet, we'd have to create one PP which starts and finish its transition on trees with different depth, do you have some ideas for that @Nashtare ?

atanmarko (commented):
Moved to draft, as PR is quite outdated so some significant rebase is needed.

Issue #540, opened 2025-01-21
Update the `Certificate` to support the new fields


Freyskeyd:
## Description

The `Certificate` needs to be updated with 2 new fields:

- `aggchain_proof` which is a `Vec<u8>` that may need to be deserialized to `compressed` proof
- `aggchain_params` which is a `Digest`

Freyskeyd:
@iljakuklic could you take a look at this ticket with @hadjiszs ?

The goal would be to introduce this as non-breaking for both the API and the storage.
Maybe by introducing an enum "versioning" on the storage side?

PR #597, opened 2025-02-06
fix: certificate storage backwards compatibility

iljakuklic:
Related
* #540

## PR Checklist:


crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            aggchain_proof,
+        } = certificate;
+
+        CertificateV1 {
+            network_id: *network_id,
+            version: VersionTag,
+            height: *height,
+            prev_local_exit_root: *prev_local_exit_root,
+            new_local_exit_root: *new_local_exit_root,
+            // TODO get rid of the clones
```
0xaatif, line 186: Can we add an owner and tracking issue for the TODO?

crates/agglayer-storage/src/types/certificate/tests.rs
```diff
@@ -0,0 +1,89 @@
...
+#[case(CertificateV0::test0())]
+#[case(CertificateV1::test0())]
+#[case(CertificateV1::test1())]
+#[case(Certificate::new_for_test(74.into(), 998))]
+fn roundtrip_through_versioned(#[case] certificate: impl Codec + Into<Certificate>) {
+    let bytes = certificate.encode().unwrap();
+    let decoded = Certificate::decode(&bytes).unwrap();
+    let orig: Certificate = certificate.into();
+
+    // TODO: This should really compare the certificates directly but that requires adding
```
0xaatif, line 86: I don't think this comment needs to be a TODO, but if you want it that way, as above with the tracking issue and owner: `TODO(iljakuklic): https://github.com/...`

0xaatif (dismissed):

crates/agglayer-storage/src/types/certificate/mod.rs
iljakuklic: > Can we add an owner and tracking issue for the TODO?

I'm not sure if it makes sense to introduce an issue even if the branch is not merged yet and it's not clear how exactly will it look when merged. Will leave this conversation open and introduce an issue closer to the merge time if the comment is still there.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+
+const TAG_BEGIN: u64 = 1u64 << 63;
+
+/// A height field restricted to values < [TAG_BEGIN].
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(test, derive(Serialize))]
+#[serde(try_from = "Height")]
+struct V0Height(Height);
+
+impl V0Height {
```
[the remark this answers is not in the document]
iljakuklic: > Why not use the `NetworkId` so at the end we'll have the version first in the bytearray?

I did not want to take any network_id values that might clash with legit values for `network_id`. Using the high half of `Height` range seems much safer.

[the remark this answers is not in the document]
iljakuklic: > Yeah, but currently I think we're safe to use that. Maybe that's a question that we can ask @krlosMata about.
> 
> if the max network id on mainnet or cardona is below the bitmask that we're aiming, it is fine.

Happy to change that once the set of allowable `network_id`s is clear.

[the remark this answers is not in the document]
krlosMata: The `version`  is a very similar approach that Ethereum took on the [EOF spec](https://eips.ethereum.org/EIPS/eip-3541) & [this](https://eips.ethereum.org/EIPS/eip-3540).
They also add more fields to be more modular (like the `magic`).

In our case, you read 4 bytes, the `magic` number (`0xFFFFFFFF` --> max `networkID` number) --> then you start defining what you want to read afterwards.
This fully respect the protocol, although 2^31 networks are enough I think xD

crates/agglayer-types/src/lib.rs
```diff
@@ -219,10 +222,11 @@ pub struct Certificate {
     pub bridge_exits: Vec<BridgeExit>,
     /// List of imported bridge exits included in this state transition.
     pub imported_bridge_exits: Vec<ImportedBridgeExit>,
-    /// Signature committed to the bridge exits and imported bridge exits.
-    pub signature: Signature,
     /// Fixed size field of arbitrary data for the chain needs.
     pub metadata: Metadata,
+    /// Aggchain proof which is either one ECDSA or one SP1 plonk proof.
```
hadjiszs, line 227:
```suggestion
    /// Aggchain proof which is either one ECDSA or one SP1 stark proof.
```

iljakuklic: I changed it to just `SP1 proof`. Whether it is plonk or stark or whatnot could be considered an implementation detail. Let me know if you disagree @hadjiszs.

crates/agglayer-storage/src/types/certificate/mod.rs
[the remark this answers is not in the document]
iljakuklic: Switched to using the `network_id` field to differentiate between V0 and other versions.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+    imported_bridge_exit::ImportedBridgeExit,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::columns::{default_bincode_options, CodecError};
+
+/// A unit type serializing to a constant byte representing the storage version.
+#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize)]
+#[serde(try_from = "u8", into = "u8")]
+struct VersionTag<const VERSION: u8>;
```
iljakuklic, line 27: Based on the feedback so far, I changed the format again. Now all versions start with a one byte version tag.

In version 0, where backwards compatibility is required, the first byte happens to be the highest byte of network ID. This effectively limits the range of network IDs in v0 storage to [0, 2<sup>24</sup>-1]. The current network IDs fall into this range, so the highest byte (with value 0) acts as the version tag.

In subsequent versions, we just have the version byte followed by a straightforward encoding of the certificate, restoring the full range of network IDs.

This has some advantages:
* No awkward bit fiddling or offsetting
* The storage format version can be easily determined by looking at the first byte

However, in v0, it reserves a big chunk of `network_ids` to encode version. I am wondering @krlosMata what are the odds in your estimation that we will see network IDs >= 2<sup>24</sup> before this gets deployed? Can someone deliberately claim a network ID >= 2<sup>24</sup> at a reasonable cost?

krlosMata: it is ok to have a maximum of 2^24, although the protocol supports until 2^32.
I do not think it is an issue since the rollupID is assigned by the Sc on L1 incrementally.
If anyone attacks in some way, agglayer-node could just be refactored to support until 2^31 easily

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -6,6 +6,14 @@ pub enum CodecError {
...
         This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
     Serialization(#[from] bincode::Error),
+
+    #[error(r#"Certificate encoded to an empty byte sequence.
+        This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
+    CertEmpty,
+
+    #[error(r#"Unrecognized certificate storage format version {version}.
+        This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
+    BadCertVersion { version: u8 },
```
Freyskeyd, line 16: We can have this generic and reusable, I think.

```suggestion
    #[error(r#"Unrecognize storage format version {version}.
        This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
    BadVersion { version: u8 },
```

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -6,6 +6,14 @@ pub enum CodecError {
     #[error(r#"Serialization error: {0}
         This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
     Serialization(#[from] bincode::Error),
+
+    #[error(r#"Certificate encoded to an empty byte sequence.
+        This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
+    CertEmpty,
```
Freyskeyd, line 12: I would make this generic: 

```suggestion
    #[error(r#"Value encoded to an empty byte sequence.
        This is a critical bug that needs to be reported on `https://github.com/agglayer/agglayer/issues`"#)]
    EmptyValue,
```

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+//!
+//! Currently, we have two versions of certificate storage format.
+//! All begin with `network_id: 32`, followed by a 64-bit field which encodes
+//! certificate height and/or format version. In the V0 format, this field
+//! corresponds to certificate height, providing backwards compatibility.
+//!
+//! If the field is < [TAG_BEGIN], it is interpreted as height in storage format V0.
+//! If the field is >= [TAG_BEGIN], it is interpreted as an identifier of the storage
+//! format version. For versions > 0, the format is encoded in the former height field
+//! as `TAG_BEGIN + FORMAT_VERSION`. In this case, the height has to be stored elsewhere.
```
Freyskeyd, line 11: This doc needs to be refactored a bit

Freyskeyd (approved):
LGTM, but I'm still wondering if we shouldn't have 2 bytes for safety and further evolutions. wdyt?

iljakuklic:
> LGTM, but I'm still wondering if we shouldn't have 2 bytes for safety and further evolutions. wdyt?

More bytes can be introduced in subsequent (not yet allocated) versions as needed. We can, for example, decide that if the first byte is >= 128, we also read the second byte and branch further based on that.

crates/agglayer-storage/src/columns/mod.rs
Freyskeyd: Okey, then we can use `BadCertificateVersion`

Freyskeyd (approved):

crates/agglayer-storage/src/types/certificate/tests/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+#[case(CertificateV1::test1())]
+#[case(CertificateV1::from(&Certificate::new_for_test(74.into(), 998)))]
+fn encoding_roundtrip_consistent_with_into(#[case] orig: impl Into<Certificate> + Serialize) {
+    let bytes = default_bincode_options().serialize(&orig).unwrap();
+    let decoded = Certificate::decode(&bytes).unwrap();
+    let converted: Certificate = orig.into();
+
+    // This should really compare the certificates directly but that requires adding
+    // whole bunch of `Eq` impl to many types.
+    assert_eq!(format!("{converted:?}"), format!("{decoded:?}"));
```
Ekleog-Polygon, line 134: How bad would it be to just add them? TBH most things should likely be `Eq`, and `format!` is very fragile (eg. hashmaps are AFAIR not guaranteed to display in the same order — though they might by chance)

crates/agglayer-storage/src/types/certificate/tests/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+#[case("n15-cert_h2")]
+#[case("n15-cert_h3")]
+fn cert_in_v0_format_decodes(#[case] cert_name: &str) {
+    let from_json = sample_data::load_certificate(&format!("{cert_name}.json"));
+
+    let bytes = load_sample_certificate_bytes(&format!("encoded_v0-{cert_name}.hex"));
+    let from_bytes = Certificate::decode(&bytes).expect("v0 certificate to decode successfully");
+
+    // Again comparing debug output due to lack of `Eq`.
+    assert_eq!(format!("{from_bytes:?}"), format!("{from_json:?}"));
```
Ekleog-Polygon, line 149: Same here

crates/agglayer-types/src/lib.rs
```diff
@@ -356,17 +360,23 @@ impl Certificate {
...
+                    self.imported_bridge_exits
+                        .iter()
+                        .map(|exit| exit.global_index),
+                );
 
-        self.signature
-            .recover_address_from_prehash(&B256::new(combined_hash.0))
+                Ok(Some(signature.recover_address_from_prehash(&B256::new(
+                    combined_hash.0,
+                ))?))
```
Ekleog-Polygon, line 376:
```suggestion
                signature.recover_address_from_prehash(&B256::new(combined_hash.0)).map(Some)
```

crates/pessimistic-proof/Cargo.toml
```diff
@@ -26,6 +28,5 @@ rs_merkle = { version = "1.4", default-features = false }
 workspace = true
 
 [features]
-default = ["sp1"]
-sp1 = ["dep:sp1-sdk"]
+default = []
```
Ekleog-Polygon, line 31:
```suggestion
```

crates/agglayer-storage/src/stores/state/snapshots/agglayer_storage__stores__state__tests__n15-cert_h0.snap
```diff
@@ -3,4 +3,4 @@ source: crates/agglayer-storage/src/stores/state/tests.rs
 expression: hash
 snapshot_kind: text
 ---
-0ba4dffc54798c555b81589351fe02e3a826d2330f4a2d8f073b214caf4c3e25
+3abee04dfc4d848597acfb65f8a55bbb32c8989ccdfae9542a46957a5f56d0fb
```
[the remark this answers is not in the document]
iljakuklic: > Are these changes really fine? They look pretty weird to me. And actually, looking at how they get generated:
> 
> https://github.com/agglayer/agglayer/blob/e5ef31b033bf6e5812148623e5338a07bf6cbfe1/crates/agglayer-storage/src/stores/state/tests.rs#L244-L247
> 
> Why do we hash the encoded bytes instead of just asserting the encoded bytes? (which would make it easier to understand what changed)

Could do that. Was not done originally as the snapshot would have been massive, but I already did some full snapshots of the V0 format in this PR to test backwards compat.

> Overall, it seems weird to me that the same certificate would now encode differently, and it sounds likely to not be backward-compatible?
>
> Overall, it seems weird to me that the same certificate would now encode differently, and it sounds likely to not be backward-compatible?

This is intentional. All insertions/writes use the V1 format. In fact, it is no longer possible to write a certificate in the old format through the storage interface. The storage format distinction only exists internally in the storage layer, the rest of the code just uses the new `Certificate` type.

crates/agglayer-storage/src/types/certificate/tests/mod.rs
iljakuklic: Probably not too bad but these types live in the pessimistic proof code I did not want to touch. Also the transitive closure of all types where it would have to be added is quite large.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+            aggchain_proof,
+        } = certificate;
+
+        CertificateV1 {
+            version: VersionTag,
+            network_id: *network_id,
+            height: *height,
+            prev_local_exit_root: *prev_local_exit_root,
+            new_local_exit_root: *new_local_exit_root,
+            // TODO get rid of the clones <https://github.com/agglayer/agglayer/issues/618>
```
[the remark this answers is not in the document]
iljakuklic: Yea maybe it would be easier to just clone the whole certificate one level higher. It is a bit irritating though since the encode method gives us a reference and it ought to be possible to just encode this using references.

Ideally these fields would be references when serializing and owned types when deserializing. I tried that with generics but it became somewhat convoluted (lack of rank2 types does not help). Maybe a `Cow` would be acceptable though.

crates/agglayer-storage/src/types/certificate/tests/mod.rs
iljakuklic: Also the sp1 proof type does not provide `Eq`, so the serialization hack would have to be present at least there...

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+/// Type specifying the current certificate encoding format.
+type CurrentCertificate = CertificateV1;
+
+fn decode<T: for<'de> Deserialize<'de> + Into<Certificate>>(
+    bytes: &[u8],
+) -> Result<Certificate, CodecError> {
+    Ok(default_bincode_options().deserialize::<T>(bytes)?.into())
```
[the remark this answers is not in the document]
Freyskeyd: I think we already had this discussion before, and it makes more sens to me to keep using bincode for such low level storage.

crates/agglayer-storage/src/types/certificate/mod.rs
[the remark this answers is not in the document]
Ekleog-Polygon: TBH I also do think that we can live with a few clones here and there, at least until some profiling tells us we're wasting significant time on them — this is definitely not a blocker :)

crates/agglayer-storage/src/types/certificate/mod.rs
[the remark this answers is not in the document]
Ekleog-Polygon: My recollection of the conclusion of this discussion was we weren't going to be migrating everything anyway so the discussion was moot for now… but now we're actually migrating everything anyway 😅 

Anyway, we can just introduce a v2 whenever we change it again, and on this day we'll likely take some more time and make the storage use protobuf. And hopefully by that time we'll have actually migrated the old certificates in storage and killed the v0 code path

My point is basically that bincode was originally designed for servo's IPC while protobuf was (I think) originally designed by google to store huge amounts data long-term, hence the various design choices. Basically, protobuf is made to never need to migrate old data, while bincode is made for speed with a single version of the data.

crates/agglayer-storage/src/stores/state/snapshots/agglayer_storage__stores__state__tests__n15-cert_h0.snap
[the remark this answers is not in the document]
Ekleog-Polygon: Got it, thank you for the clarification!

crates/agglayer-storage/src/types/certificate/tests/mod.rs
Ekleog-Polygon: Ughhhh can we at least assert that the hash is the same, rather than the debug output? I'd feel really sad to be matching on Debug, and at least Hash should already be implemented basically everywhere considering we're already hashing certificates.

iljakuklic: Some lessons learned attempting to add `Eq` to various pp types.

* Since the sp1 proof type does not provide `Eq`, we would still have to compare serialized versions, or come up with another hack.
* With the hack, I'd be reluctant to just expose these impls, so would be better guarded by "testutils".
* Since some hacky `Eq` is needed somewhere, it's probably better be contained here in tests rather than a couple of crates deep.

Ekleog-Polygon: Yeah, I'm with you, I'm curious why SP1Proof does not provide even PartialEq, but I guess the only way to go would be `Hash`ing and then using `Eq`

iljakuklic: > Ughhhh can we at least assert that the hash is the same, rather than the debug output? I'd feel really sad to be matching on Debug, and at least Hash should already be implemented basically everywhere considering we're already hashing certificates.

Unfortunately, the sp1 proof type does not provide `Hash` impl either. The `Certificate::hash` method does not include all fields so not suitable for checking serialization.

iljakuklic:
Got this beauty from the CI:
> Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v4.0.2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions.

It complains about `v4.0.2` but suggests switching to `v3/v4`. That sounds a bit incoherent.

Ekleog-Polygon (approved):
Considering the clone only happened on a slow path anyway (just before writing to disk) I felt better about the option without the `Cow`. Because `Cow` incurs a systematic enum match, which might have more impact on performance than the occasional `clone()` just before writing to disk.

This being said, this should still be fine, and we can improve on performance later on :)

Ekleog-Polygon (approved):

Issue #563, opened 2025-01-28
Ensure crates can be built individually


iljakuklic:
The commands such as:

```sh
cargo build -p $SOME_CRATE
cargo nextest run -p $SOME_CRATE
```

often fail to build due to unsatisfied features.

There should be a CI job to check crates can be build individually. The job should run something like

```sh
for CRATE in $ALL_CRATES; do
   cargo check -p $CRATE || report_build_error "$CRATE"
   cargo check --tests -p $CRATE || report_build_error "$CRATE"
done
```

Maybe there are better ways to do this but even the above would be an improvement.

0xaatif:
Are there any example jobs which failed that you can point me to? So I can make a test :)

Ekleog-Polygon:
FWIW, the `cargo-hack` tool is basically made for that, and even more. We could introduce it to CI if this becomes a big issue :)

0xaatif:
Perhaps we `cargo metadata | jq .. | xargs cargo build-all-features ...`?

PR #603, opened 2025-02-10
test: build each crate standalone, and with each feature

0xaatif:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)
https://github.com/agglayer/agglayer/issues/563

## Additions and Changes

Please explain in detail, if applicable, which changes you performed in the PR and link them to issues. (If not applicable you can remove this whole section)

## Breaking changes

Please explain in detail which changes are breaking changes in the PR. (If not applicable you can remove this whole section)

## PR Checklist:


atanmarko (approved):

Ekleog-Polygon (commented):
`cargo hack` is usually pretty slow (because it needs to do *lots* of builds).

I'd suggest putting it in another job to not slow down CI too much :)

.github/workflows/test.yml
```diff
@@ -130,12 +130,15 @@ jobs:
...
-      - uses: taiki-e/install-action@nextest
-      - uses: taiki-e/install-action@protoc
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest,protoc,cargo-hack
 
       - name: Install Anvil
         uses: foundry-rs/foundry-toolchain@v1
 
+      - run: cargo hack --each-feature --all check
```
iljakuklic, line 140: Other steps in the job also have a descriptive name, could we add one?

iljakuklic:
> `cargo hack` is usually pretty slow (because it needs to do _lots_ of builds).
> 
> I'd suggest putting it in another job to not slow down CI too much :)

This only runs cargo check. Would be interesting to have some data to see how slow it really is. I'd appreciate if it was feasible to run this on PRs because I tend to use the likes of `cargo nextest run -p $PACKAGE` quite a lot but it's often broken.

.github/workflows/test.yml
0xaatif: I've moved this into its own job.

In general I think we're overzealous with adding `name: ...`, we end up with a mix of e.g `name: Checkout`, `name: Checkout sources`, etc, where just `uses: actions/checkout` communicates everything a reader needs

0xaatif:
I think this is blocked on https://github.com/agglayer/provers/pull/33

iljakuklic:
I merged in `main`, fixed up some things, so this now passes.

However, the whole feature checking job takes about 10 minutes on github CI. Does this seem acceptable?

atanmarko:
> However, the whole feature checking job takes about 10 minutes on github CI. Does this seem acceptable?

I think 10 minutes is OK. As the code-base and feature list grows, it will be more useful.

hadjiszs (approved):
> However, the whole feature checking job takes about 10 minutes on github CI. Does this seem acceptable?

fine by me 👍

atanmarko (approved):

PR #604, opened 2025-02-11
feat: make mock prover work with kurtosis-cdk

Freyskeyd:
# Description

This PR is updating agglayer to support mock verifier and is compatible with kurtosis-cdk in mock mode.

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/src/lib.rs
+++ b/crates/agglayer-config/src/lib.rs
@@ -111,6 +111,10 @@ pub struct Config {
     #[serde(default)]
     #[serde(skip_serializing_if = "is_false")]
     pub debug_mode: bool,
+
+    #[serde(default)]
+    #[serde(skip_serializing_if = "is_false")]
+    pub mock_verifier: bool,
 }
 
 impl Config {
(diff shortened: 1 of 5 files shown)


atanmarko (approved):

crates/agglayer-config/src/lib.rs
```diff
@@ -162,6 +166,7 @@ impl Config {
             prover_entrypoint: default_prover_entrypoint(),
             prover: Default::default(),
             debug_mode: false,
+            mock_verifier: false,
```
iljakuklic, line 169: It would be nicer for this to be a field like `verifier` of the type `enum` with the `Cpu` and `Mock` variants. Would show up as `verifier = "cpu"` or `verifier = "mock"` in the config.

Ekleog-Polygon (approved):

iljakuklic (approved):
One suggestion. Approving anyway in case this is urgent.

crates/agglayer-config/src/lib.rs
Freyskeyd: I first try this, but end up using a flag, as this should not be used to configure the verifier, just turn on/off the mock for tests. We'll not have something like `verifier = "gpu"` or `verifier = "network"`, at least for now. I'll stick with this format, and if we do want to configure the `verifier` at some point, we'll be able to and we'll deprecate this flag if needed.

PR #610, opened 2025-02-12
fix: skip native claims in witness generation

hadjiszs:
# Description

The imported bridge exits for the native token weren't skipped in the witness generation but only in the prover, leading to a discrepancy on the computed balance root.

## PR Checklist:


Freyskeyd (approved):

crates/agglayer-storage/src/stores/state/tests.rs
```diff
@@ -233,6 +233,43 @@ fn can_read(network_id: NetworkId, store: StateStore) {
     ));
 }
 
+#[test]
+fn import_native_tokens() {
+    let certificates: Vec<Certificate> = ["cert_h0.json", "cert_h1.json", "cert_h2.json"]
+        .iter()
+        .map(|p| data::load_certificate(p))
```
0xaatif, line 240: nit: surely this closure is redundant?

crates/agglayer-storage/src/stores/state/tests.rs
```diff
@@ -233,6 +233,43 @@ fn can_read(network_id: NetworkId, store: StateStore) {
...
+    let mut lns = LocalNetworkStateData::default();
+
+    for (idx, certificate) in certificates.iter().enumerate() {
+        info!(
+            "Certificate ({idx}|{}) | {}, nib:{} b:{}",
+            certificate.height,
+            certificate.hash(),
+            certificate.imported_bridge_exits.len(),
+            certificate.bridge_exits.len(),
+        );
```
0xaatif, line 252: nit: structured logging

crates/agglayer-storage/src/stores/state/tests.rs
```diff
@@ -233,6 +233,43 @@ fn can_read(network_id: NetworkId, store: StateStore) {
...
+
+        info!("Certificate {idx}: successful witness generation");
+        let initial_state = LocalNetworkState::from(lns.clone());
+
+        generate_pessimistic_proof(initial_state.into(), &multi_batch_header).unwrap();
+        info!("Certificate {idx}: successful native execution");
+
+        lns.apply_certificate(certificate, signer, l1_info_root)
+            .unwrap();
+        info!("Certificate {idx}: successful state transition, waiting for the next");
```
0xaatif, line 269: question: VERY surprised to see `tracing::info` logs in a `#[test]` - are you setting up a subscriber somewhere? Else I think that these should be `printlns`

0xaatif (approved):
Approving because you said this was urgent, but would like to see the comments addressed in a follow-up

atanmarko (approved):

crates/agglayer-storage/src/stores/state/tests.rs
Freyskeyd: Yes, we can do that on main.

crates/agglayer-storage/src/stores/state/tests.rs
Freyskeyd: As it is for test, I don't really mind have such type of logs. If we want to update those, let's do it on main.

crates/agglayer-storage/src/stores/state/tests.rs
Freyskeyd: We generally do that using `test_log::test(test)`, but in this particular case those logs are useless.
Will fix this on main, as it is not critical.

PR #622, opened 2025-02-18
fix: adding set latest proven/pending method on admin RPC (#569)

Freyskeyd:
# Description

This PR is porting back to main the changes made for 0.2.2 in PR #569
(cherry picked from commit 42721bd2b9f086d788a58bdd8bc10dfabaa4ee48)

## PR Checklist:


crates/agglayer-node/src/service/mod.rs
```diff
@@ -294,9 +294,9 @@ where
             .inspect_err(|e| error!("Failed to get latest pending certificate: {e}"))?;
 
         let certificate_id = [
-            settled_certificate_id_and_height,
-            proven_certificate_id_and_height,
             pending_certificate_id_and_height,
+            proven_certificate_id_and_height,
+            settled_certificate_id_and_height,
```
Ekleog-Polygon, line 299: AFAICT considering the `.max_by` call before this will not have any impact, unless we have multiple certificate IDs for the same height all on this path, which would be very wrong.

Ekleog-Polygon (approved):
LGTM, reviewed assuming there were no cherry-pick conflicts, which sounds likely considering this is basically only adding code :) (Did not review the details of the added code that have already been reviewed in the first PR)

crates/agglayer-node/src/service/mod.rs
Ekleog-Polygon: I'm not sure I understand, what do we gain by prioritizing the settled one? For the max height it should always be the same certificate ID regardless of which category we're looking at, should it not?

Freyskeyd: If, for any reason, we have for the same `network_id` and height a different `certificate_id`, we should **always** prioritize the settled one.

iljakuklic (approved):

crates/agglayer-node/src/service/mod.rs
Freyskeyd: If you take this as a product, it makes sense. But it doesn't when you have pre-existing data that *could* maybe already in this shape.
If this has been implemented and if the branch is tag as `fix`, it means that it was involved in an existing issue either on `devnet` or `testnet`, where we can't just "delete everything and rebuild".

FWIW, this route and this mechanism will be deprecated as soon as we migrate to gRPC.

PR #628, opened 2025-02-18
fix: skip native claims in witness generation (#610)

Freyskeyd:
# Description

This PR is porting back to main the changes made for 0.2.2 in PR #610
(cherry picked from commit 1031c18c17bb0cf14a0cd0a738e71681e5cbf18d)

## PR Checklist:


crates/agglayer-storage/src/stores/state/tests.rs
```diff
@@ -233,6 +233,43 @@ fn can_read(network_id: NetworkId, store: StateStore) {
...
+fn import_native_tokens() {
+    let certificates: Vec<Certificate> = ["cert_h0.json", "cert_h1.json", "cert_h2.json"]
+        .iter()
+        .map(|p| data::load_certificate(p))
+        .collect();
+
+    let mut lns = LocalNetworkStateData::default();
+
+    for (idx, certificate) in certificates.iter().enumerate() {
+        info!(
```
atanmarko, line 246: Do you have logs enabled in the test?

atanmarko (dismissed):

iljakuklic (dismissed):

atanmarko (approved):

crates/agglayer-storage/src/stores/state/tests.rs
Freyskeyd: We're using `test_log` when needed.

atanmarko: @Freyskeyd It is not fortunate to manually need to change every test we need the output from. For example if the CI build fails we can not see what happened. We should consider using `println` in the tests consistently.

PR #629, opened 2025-02-19
fix: github action proto ref for merge_group

Freyskeyd:
# Description

Fixes the ref+branch used when triggered in merge_group.

## PR Checklist:

### DIFF
--- a/.github/workflows/proto.yml
+++ b/.github/workflows/proto.yml
@@ -11,18 +11,33 @@ on:
 permissions:
   contents: read
   pull-requests: write
+
+env:
+  BREAKING_REF: branch=main
+
 jobs:
   buf:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


iljakuklic (approved):

Ekleog-Polygon (commented):
FWIW, here is the code at near that does basically the same:
https://github.com/near/nearcore/blob/master/.github/workflows/ci.yml#L76

I think that in this situation, the best would be to have `ref=pull_request.base.sha` in the pull request case, and `branch=main` in the merge group case. Does that make sense?

Ekleog-Polygon:
(it's basically what's implemented in the workflow at near, that I had originally implemented there in https://github.com/near/nearcore/pull/10090)

atanmarko (approved):

PR #662, opened 2025-02-26
docs: fix dead link in README.md

rebustron:
# Description

Hi! I fixes a dead link in the README.md file. The original link pointing to Succinct's documentation for generating proofs was outdated and has been updated to the correct URL.

## Additions and Changes

Updated the dead link from https://docs.succinct.xyz/docs/generating-proofs/prover-network/key-setup to https://docs.succinct.xyz/docs/sp1/generating-proofs/prover-network/key-setup.

## PR Checklist:

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added or updated tests that comprehensively prove my change is effective or that my feature works

PR #663, opened 2025-02-27
fix: make buf CI check hopefully work for external contributors' PRs

Ekleog-Polygon:
# Description

See [failing example](https://github.com/agglayer/agglayer/pull/662) with [its detailed logs](https://github.com/agglayer/agglayer/actions/runs/13550198886/job/37902593573?pr=662) — it failed thrice, so is most likely not an intermittent.

I'm not sure this will fix things, but it's the only hypothesis I have, and deleting this should hopefully not affect anything adversely. So let's just try it out!

## PR Checklist:


.github/workflows/proto.yml
```diff
@@ -20,8 +20,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.head_ref || github.event.pull_request.head.ref || github.ref }}
```
Freyskeyd, line 24: We should keep that, as this workflow is triggered on `pull_request_target`.
Also, I think that the solution is to define the `repository`, currently it defaults to `${{ github.repository }}` which could be changed to `${{github.event.pull_request.head.repo.full_name || github.repository }}` or something similar

Ekleog-Polygon: Good point! I hadn't noticed this was running on `pull_request_target`. Fixed, thank you :)

Freyskeyd: To be sure, could you test this change using `act` for the various trigger that we have? (merge_group, etc)

Freyskeyd: What I did recently was to check this using multiple events:

- `act merge_group`
- `act push`

And just looking at what branch/ref was used

hadjiszs (approved):
>I'm not sure this will fix things, but it's the only hypothesis I have, and deleting this should hopefully not affect anything adversely. So let's just try it out!

not entirely sure how to properly test on my side so approving to try it out as suggested 👍

Issue #666, opened 2025-02-27
Agglayer-node integration with kurtosis & aggsender


Freyskeyd:
## Description

After drafting a new RC or tag containing the recent changes on gRPC.

Test and help the integration of the aggsender with kurtosis.

The `aggsender` should be able to communicate with `agglayer-node` using gRPC. While also still providing JSON-rpc support.
 
agglayer version under test: `v0.3.0-rc.4`

## Goal

- [x] Test 0.2 integration with gRPC (having aggsender sending certificate using gRPC entrypoint)
- [ ] Test 0.2 integration with JSON-RPC (having aggsender sending certification using JSON-RPC entrypoint, possible option is to use an old version of aggsender)
- [ ] Test the different error returns by the gRPC services using the standard [ErrorDetails](https://google.aip.dev/193#error_model) spec

*Tickets and refacto linked to this integration need to be created and prioritized as subtask of this issue*

-------
## PRs

- [ ] #688

Ekleog-Polygon:
I'm not sure I understand, what does that issue encompass?

Freyskeyd:
I updated the descripton

Ekleog-Polygon:
Do I understand correctly that this should be mostly helping dev-tools and aggkit teams integrate with agglayer-node?

Freyskeyd:
> Thanks! Do I understand correctly that this should be mostly helping dev-tools and aggkit teams integrate with agglayer-node?

Yes, and addressing fixes or created tickets for improvments

Ekleog-Polygon:
Got it, thank you!

goran-ethernal:
@Freyskeyd Tried to run the new agglayer rc on kurtosis, and get this error:

```
missing field `aggchain-base-contract`
```
We need to modify the agglayer `.toml` file in kurtosis to include this address. Which contract is this?

Ekleog-Polygon:
If I understand [this code](https://github.com/agglayer/agglayer/pull/644/files#diff-1d79c2b8914a349020452b2fdb91f2a8c464a17a70bc510e33dc7d5379d10553R24-R29) correctly, it comes from [contracts/v2/lib/AggchainBase.sol](https://github.com/agglayer/agg-contracts-internal/blob/feature/ongoing-v0.3.0/contracts/v2/lib/AggchainBase.sol) :)

Ekleog-Polygon:
[...]

PR #688, opened 2025-03-05
fix: grpc always expects to be on the domain root

Ekleog-Polygon:
See https://stackoverflow.com/questions/40410392/grpc-endpoint-with-non-root-path In particular, the go library seems to not support prefixing calls with `/grpc`.

cc @goran-ethernal as you handled the go part of this investigation
cc @Freyskeyd as you wrote the initial code

Part of #666 

## PR Checklist:

### DIFF
--- a/crates/agglayer-config/src/rpc.rs
+++ b/crates/agglayer-config/src/rpc.rs
@@ -4,8 +4,12 @@ use jsonrpsee::core::TEN_MB_SIZE_BYTES;
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_with::serde_as;
 
-/// The default port for the local RPC server.
-const DEFAULT_PORT: u16 = 9090;
+/// The default port for the local gRPC server.
+const DEFAULT_GRPC_PORT: u16 = 9089;
+
+/// The default port for the local ReadRPC server.
+const DEFAULT_READRPC_PORT: u16 = 9090;
+
 /// The default port for the local admin RPC server.
 const DEFAULT_ADMIN_PORT: u16 = 9091;
 
@@ -14,12 +18,24 @@ const DEFAULT_ADMIN_PORT: u16 = 9091;
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
 #[serde(rename_all = "kebab-case")]
 pub struct RpcConfig {
-    /// If the `PORT` environment variable is set, it will take precedence over
-    /// the configuration file.
-    #[serde(default = "default_port", deserialize_with = "deserialize_port")]
-    pub port: u16,
+    /// The default port for the local gRPC server.
+    /// Overridden by `AGGLAYER_GRPC_PORT` environment variable, defaults to 9089.
+    #[serde(
+        default = "default_grpc_port",
+        deserialize_with = "deserialize_grpc_port"
+    )]
+    pub grpc_port: u16,
 
-    /// The port for `admin` RPC
+    /// The default port for the local ReadRPC server.
+    /// Overridden by `AGGLAYER_READRPC_PORT` environment variable, defaults to 9090.
+    #[serde(
+        default = "default_readrpc_port",
+        deserialize_with = "deserialize_readrpc_port"
+    )]
+    pub readrpc_port: u16,
+
+    /// The default port for the local AdminRPC server.
+    /// Overridden by `AGGLAYER_ADMIN_PORT` environment variable, defaults to 9091.
     #[serde(
         default = "default_admin_port",
         deserialize_with = "deserialize_admin_port"
@@ -64,7 +80,8 @@ pub struct RpcConfig {
 impl Default for RpcConfig {
     fn default() -> Self {
         Self {
-            port: default_port(),
+            grpc_port: default_grpc_port(),
+            readrpc_port: default_readrpc_port(),
             admin_port: default_admin_port(),
             host: default_host(),
             max_request_body_size: default_body_size(),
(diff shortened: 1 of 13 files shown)


atanmarko (approved):
I trust you on this one 😉

crates/agglayer-node/src/node.rs
```diff
@@ -293,7 +293,7 @@ impl Node {
         let router = axum::Router::new()
             .merge(health_router)
             .merge(json_rpc_router)
-            .nest("/grpc", grpc_router);
+            .merge(grpc_router);
```
Freyskeyd, line 296: If we do not nest here, It means that we have all the routes and router on the same entrypoint in the API. Maybe it's not a problem but I liked the idea of keeping things separated at least internally.

Another approach could have been something like this:

We could route the query based on the `Content-Type` header, keeping everything ordered on the server with good separation, while solving the `aggsender` problem.

The goal would be to keep the `nest("/grpc")` and have an axum middleware handling the job:

```rust

let router = axum::Router::new().nest("/grpc", grpc_router)...;

let middleware = axum::middleware::from_fn(grpc_route_middleware);
axum::server(listener, middleware.layer(router)).;

async fn grpc_route_middleware<B>(request: Request<B>, next: Next<B>) -> Response {
    let (mut parts, body) = request.into_parts();
    
    // Do something with the uri
    let accept_value = HeaderValue::from_static("application/grpc");
    if Some(&accept_value) == parts.headers.get("Accept")
        || Some(&accept_value) == parts.headers.get("Content-Type")
    {
        uri = format!("/grpc/{uri}");
    }

    next.run(request).await
}
```

I let you decide what's best and more futureproof for following iteration and depreciation of some part of this entrypoint.

goran-ethernal:
Just to leave the update here, I tried the `grpc` communication from `aggsender` to `agglayer` with these changes, and everything works as expected. Like @Ekleog-Polygon said, `google.golang.org/grpc` does not support adding suffixes to the calls for some reason 😢.

crates/agglayer-node/src/node.rs
Ekleog-Polygon: What would be best to me, as well as simplest, would be to just have a second port for the gRPC API. It seems to be the officially-recommended way. And it would make it possible to eventually expose it straight on `.443` with a reverse-proxy doing eg. hostname-based dispatch or any other mechanism as preferred by ops without any changes on our end..

If this plan looks good to you, I'll:
- Rename the current "RPC port" into "ReadRPC port", it'd keep JSON-RPC and /health
- Add a "gRPC port" defaulting 9089, that'd serve only the gRPC endpoints

WDYT about that plan?

Ekleog-Polygon: (And if my long-term hopes are met, we can eventually deprecate all of the ReadRPC port and hopefully someday just kill all that code to simplify our codebase… but that's not contained in the plan as I do understand it's still controversial for various reasons :))

Freyskeyd: > What would be best to me, as well as simplest, would be to just have a second port for the gRPC API. It seems to be the officially-recommended way. And it would make it possible to eventually expose it straight on `.443` with a reverse-proxy doing eg. hostname-based dispatch or any other mechanism as preferred by ops without any changes on our end.
> 
> If this plan looks good to you, I'll:
> 
> * Rename the current "RPC port" into "ReadRPC port", it'd keep JSON-RPC and /health
> * Add a "gRPC port" defaulting 9089, that'd serve only the gRPC endpoints
> 
> WDYT about that plan?

That's something that I wanted to avoid at first to not have too many changes on the infra side, but yes I agree that maybe we can switch to a dedicated port.

crates/agglayer-config/src/rpc.rs
```diff
@@ -14,12 +18,24 @@ const DEFAULT_ADMIN_PORT: u16 = 9091;
...
+        default = "default_grpc_port",
+        deserialize_with = "deserialize_grpc_port"
+    )]
+    pub grpc_port: u16,
 
-    /// The port for `admin` RPC
+    /// The default port for the local ReadRPC server.
+    /// Overridden by `AGGLAYER_READRPC_PORT` environment variable, defaults to 9090.
+    #[serde(
+        default = "default_readrpc_port",
```
Freyskeyd, line 32: Do we want to have an `alias = "port"` here for backward compatibility?

Freyskeyd (approved):

goran-ethernal:
> I created #690 to not forget to update kurtosis-cdk config with those changes

@Freyskeyd I am not sure what to do with that `aggchain-base-contract` config parameter on `agglayer`. In our testing in kurtosis, we just put the default `aggchain-base-contract = "0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"`, but maybe some support needs to be added in kurtosis for these contracts. We will need them for further integration testing once the prover is complete.

crates/agglayer-config/src/rpc.rs
Ekleog-Polygon: I'd rather not: better to fail loudly than silently accept deprecated options, especially today as we're still in a very centralized, easy-to-change setup.

Landing as-is, let me know if you think it's important and we can add the alias in a follow-up PR!

Freyskeyd:
> > I created #690 to not forget to update kurtosis-cdk config with those changes
> 
> @Freyskeyd I am not sure what to do with that `aggchain-base-contract` config parameter on `agglayer`, because on the new RC it won't start without providing this address. In our testing in kurtosis, we just put the default `aggchain-base-contract = "0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"`, but maybe some support needs to be added in kurtosis for these contracts. We will need them for further integration testing once the prover is complete.

I'll add that to the ticket

Ekleog-Polygon:
Just submitted https://github.com/0xPolygon/kurtosis-cdk/pull/521 that'll adjust Kurtosis CDK to the new RPC port setup :)

crates/agglayer-config/tests/snapshots/auth_config__auth_kebab_case.snap
```diff
@@ -17,7 +18,8 @@ outputs = []
 format = "pretty"
 
 [rpc]
-port = 9090
+grpc-port = 9089
+readrpc-port = 9090
```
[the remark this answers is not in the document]
Ekleog-Polygon: See https://github.com/agglayer/agglayer/pull/688#discussion_r1983378413 😄

PR #689, opened 2025-03-05
feat: add pp migration logic

### LINKED ISSUE #674: Add the `height` and `networkId` to the pessimistic-proof root generation

hadjiszs:
# Description

- [x] Fetch last settled pessimistic root from L1
- [x] Infer the version and migration steps based on signature and last settled pp root
- [x] Add tests
  - [x] The `happy_path` test considers a `prev_pessimistic_root` at `[0u8; 32]` fetched from L1 (as it is in practice for newly bootstrapped network in L1). The `happy_path` test passes which handles then at least this case
  - [x] Add tests for the 4 cases `v(2|3) -> v(2|3)`

Fixes #674 

## PR Checklist:


crates/pessimistic-proof-core/src/local_state/commitment.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub origin_network: NetworkId,
+}
+
+impl PessimisticRoot {
+    /// Infer the version of the provided pessimistic root.
+    pub fn infer_pp_root_version(&self, pp_root: Digest) -> Result<PPRootVersion, ProofError> {
+        // NOTE: Return v2 to trigger the migration
+        if pp_root.0 == [0u8; 32] {
+            return Ok(PPRootVersion::V2);
+        }
```
invocamanman, line 37: We must assert here that the fields/roots aare emtpy, otherwise any "inital" value is psosible, that menas that:
```
 if self.balance_root != EMPTY_LER
                || self.nullifier_root != EMPTY_LER
                || self.ler_leaf_count != 0
                || self.height != 0 {
                return Err(ProofError::Unknown(
                    "Empty pp_root provided but internal fields are non-zero".to_string(),
                ));
            }
````

Comment (hadjiszs) [reply crates/pessimistic-proof-core/src/local_state/commitment.rs:37]: good catch 👍

File: crates/agglayer-aggregator-notifier/src/certifier/mod.rs [hunk 2]
```diff
@@ -278,6 +277,12 @@ where
             .await
             .map_err(|_| CertificationError::TrustedSequencerNotFound(network_id))?;
 
+        let prev_pessimistic_root = self
+            .l1_rpc
+            .get_prev_pessimistic_root(*network_id)
+            .await
+            .map_err(|_| CertificationError::LastPessimisticRootNotFound(network_id))?;
+
```
Freyskeyd, line 285: Maybe we should have a method that returns both params, instead of doing 2 RPC calls

crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -254,21 +295,18 @@ impl NetworkState {
...
-                    ),
+                    commit_imported_bridge_exits: imported_hash,
                 };
 
                 sp1_zkvm::lib::verify::verify_sp1_proof(
                     aggchain_vkey,
                     &aggchain_proof_public_values.hash().into(),
                 );
+
+                PPRootVersion::V2
```
Freyskeyd, line 306: Shouldn't it be `v3`?

crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -211,37 +203,86 @@ impl NetworkState {
...
                 }
+
+                target_pp_root_version
             }
             #[cfg(not(target_os = "zkvm"))]
             AggchainData::Generic { .. } => {
                 // NOTE: No stark verification in the native rust code due to
                 // the sp1_zkvm::lib::verify::verify_sp1_proof syscall
                 warn!("verify_sp1_proof is not callable outside of SP1");
+                PPRootVersion::V2
```
Freyskeyd, line 285: Shouldn't it be `v3`?

crates/pessimistic-proof-core/src/proof.rs
```diff
@@ -142,87 +154,69 @@ pub const EMPTY_LER: Digest = Digest(hex!(
...
-        });
-    }
+        PessimisticRoot {
+            balance_root: network_state.balance_tree.root,
+            nullifier_root: network_state.nullifier_tree.root,
+            ler_leaf_count: network_state.exit_tree.leaf_count,
+            height,
+            origin_network: batch_header.origin_network,
+        }
+        .compute_pp_root(target_pp_root_version)
```
Freyskeyd, line 205: It should always be `v3` here right?

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
hadjiszs: yeah, hard to name it tho, maybe we could have one method which return the entire rollup data structure which is fetched entirely anyway

crates/pessimistic-proof-core/src/proof.rs
hadjiszs: after migration yes (i.e., right after that the aggsender adds the height in its signature), but before that it'll stay v2

crates/pessimistic-proof-core/src/local_state/mod.rs
hadjiszs: same as for https://github.com/agglayer/agglayer/pull/689#discussion_r1987184945

crates/pessimistic-proof-core/src/local_state/commitment.rs
hadjiszs: actually not entirely fixed, I need to give the proper last pp root in `apply_certificate()` especially for the tests

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
hadjiszs: ah not exactly the same actually, in one case we ask for one `RollupDataReturn` and in the other case the `RollupDataReturnV2`, so not entirely sure we can make single call there

crates/pessimistic-proof-core/src/local_state/commitment.rs
hadjiszs: done

atanmarko (commented):
LGTM with some nits

Freyskeyd (approved):

crates/pessimistic-proof-test-suite/src/bin/ppgen.rs
```diff
@@ -74,7 +75,12 @@ pub fn main() {
...
 
     let multi_batch_header = old_state
-        .make_multi_batch_header(&certificate, state.get_signer(), l1_info_root)
+        .make_multi_batch_header(
+            &certificate,
+            state.get_signer(),
+            l1_info_root,
+            PessimisticRootInput::Computed(PPRootVersion::V2),
+        )
         .unwrap();
```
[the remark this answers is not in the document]
hadjiszs: hm I guess in a dedicated PR for such changes on `ppgen`

crates/pessimistic-proof-core/src/imported_bridge_exit.rs
```diff
@@ -38,7 +38,7 @@ pub struct L1InfoTreeLeaf {
 
 impl L1InfoTreeLeaf {
     pub fn hash(&self) -> Digest {
-        self.inner.hash()
+        self.inner.hash(keccak256_combine([self.mer, self.rer]))
```
invocamanman, line 41: but you need the hash leaf to verify it against the l1 info root right?¿ i think this won't work

atanmarko (approved):

atanmarko (approved):

atanmarko (approved):

atanmarko (approved):

crates/pessimistic-proof-core/src/multi_batch_header.rs
```diff
@@ -27,6 +27,11 @@ where
 {
     /// Network that emitted this [`MultiBatchHeader`].
     pub origin_network: NetworkId,
+    /// Current height of the L2 chain.
+    pub height: u64,
```
Freyskeyd, line 31: It is not the L2 chain height (this can refer to the number of blocks) while here, it is the height of the network in regard of the agglayer based on how many certificates it receives

Freyskeyd (approved):

crates/pessimistic-proof-core/src/multi_batch_header.rs
hadjiszs: yeah misleading indeed, what about this:

```diff
-    /// Current height of the L2 chain.
+    /// Current certificate height of the L2 chain.
```

also related thread:  https://github.com/agglayer/agglayer/pull/689#discussion_r1987575045

Freyskeyd: This is fine solution for me

Freyskeyd (approved):

PR #680, opened 2025-03-03
build(deps): make dependabot update prover packages in one PR

iljakuklic:
## PR Checklist:

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added or updated tests that comprehensively prove my change is effective or that my feature works

PR #702, opened 2025-03-12
build: more granular dependabot groups, take 2

iljakuklic:
Three groups have been introduced:
* `provers` for keeping up to date with the `agglayer/provers` repo
* `sp1` for which we have a different update policy from other external dependencies, and updating it often fails, making dependabot PRs way less useful
* `external`, previously called `all`, for the remaining external dependencies. Major version updates still get separate PRs.

Looking at the dependabot logs, the failures we've seen may not have been a result of #680 (reverted #685), so I think this calls for another attempt. I have also changed the indentation a bit to be in line with most yaml syntax examples online, and possibly friendlier to parsers.

## PR Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,14 +4,31 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
     groups:
-      all:
+      provers:
         applies-to: version-updates
         patterns:
-        - "*"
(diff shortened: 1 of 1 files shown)


Ekleog-Polygon (approved):
LGTM! Let's try it out and see if it works better :)

PR #710, opened 2025-03-13
ci: switching to agglayer e2e

praetoriansentry:
# Description

This is a minor refactor to run the kurtosis tests in the `agglayer/e2e` repo.

## PR Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,6 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   REGISTRY: ghcr.io
-  KURTOSIS_CDK_GIT_REF: v0.2.11
   ZKEVM_CONTRACT_GIT_REF: feature/ongoing-v0.3.0-docker
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -177,67 +176,34 @@ jobs:
...
+          "deploy_cdk_bridge_ui": false
+        },
+          "args": {
+            "consensus_contract_type": "pessimistic",
+            "erigon_strict_mode": false,
+            "enable_normalcy": true,
+            "agglayer_prover_sp1_key": "",
+            "agglayer_prover_primary_prover": "mock-prover",
+            "sequencer_type": "erigon",
+            "zkevm_contracts_image": "leovct/zkevm-contracts:v10.0.0-rc.3-fork.12"
```
Freyskeyd, line 204: This could be use as ENV and also use in our integration tests instead of manually building the docker image.

Freyskeyd (approved):

Freyskeyd:
Question, what are the tests running by the current setup?

.github/workflows/test.yml
```diff
@@ -177,67 +176,34 @@ jobs:
...
-
-      - name: Expose RPC URL
-        run: echo "ETH_RPC_URL=$(kurtosis port print cdk-v1 cdk-erigon-node-001 http-rpc)" >> $GITHUB_ENV
-
-      - name: Check that batches are progressing
-        continue-on-error: true
-        env:
-          ETH_RPC_URL: ${{ env.ETH_RPC_URL }}
-          TIMEOUT_SECONDS: 900
-        run: ./.github/scripts/monitor-verified-batches.sh --rpc-url "$ETH_RPC_URL"
```
Ekleog-Polygon, line 237: Just to confirm, do we already have something equivalent to monitor-verified-batches.sh in the e2e repo? (Disclaimer: I haven't been able to figure out where this script actually lives, it looks like I have it in neither of my kurtosis-cdk nor my agglayer checkouts…)

.github/workflows/test.yml
praetoriansentry: Yes - There is a separate CI job that builds this, but because the repo has a different name (temporarily) it had to be manually adjusted last night.

praetoriansentry:
> Question, what are the tests running by the current setup?

- An L1 to L2 bridge that confirms the balance increases (this is necessary for the next text)
- An L2 to L1 bridge that confirms the balance increases
- Simple tests of some of the RPCs

image

The tests are defined by this file at the moment:
https://github.com/agglayer/e2e/blob/main/tests/agglayer/bridges.bats

.github/workflows/test.yml
praetoriansentry: Yeah there are a few things like it, but I think the approach for test validation can be a bit more precise depending on the test case. E.g. in the default cases that we're running here, we're doing and L2 to L1 bridge. It's possible that the number of verified batches could increase even if the bridge were to fail for some reason.

praetoriansentry: For what it's worth that file use to to live [here](https://github.com/0xPolygon/kurtosis-cdk/blob/39f708f421d5569d80f76d25250994d7036b3311/.github/scripts/monitor-verified-batches.sh) but it looks like it was [renamed](https://github.com/0xPolygon/kurtosis-cdk/commit/2edcc2f06baf78aafbbf5652db47c5ad4f1fda17#diff-4d2f9a0c645cd92237a8a4b5ce67be7977f58450b934efa4918bd2afe2d997a8).

I would guess this step was passing because of the `continue-on-error: true` setting which:
> Prevents a job from failing when a step fails. Set to true to allow a job to pass when this step fails.

PR #715, opened 2025-03-14
refactor: switch vkey to big endian

Freyskeyd:
# Description

Refactorisation of how we proceed `aggchain-vkey` to match and have a standard format accros various vkeys.

## PR Checklist:

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
@@ -375,8 +375,8 @@ where
 
                 let agglayer_types::aggchain_proof::Proof::SP1Stark(sp1_reduce_proof) = proof;
 
-                let proof_vk_hash = agglayer_contracts::aggchain::AggchainVkeyHash::from_u32_array(
-                    sp1_reduce_proof.vk.hash_u32(),
+                let proof_vk_hash = agglayer_contracts::aggchain::AggchainVkeyHash::new(
+                    sp1_reduce_proof.vk.hash_bytes(),
                 );
 
                 if aggchain_vkey != proof_vk_hash {
@@ -386,7 +386,7 @@ where
                     });
                 }
 
-                Some(aggchain_vkey.as_u32_array())
+                Some(sp1_reduce_proof.vk.hash_u32())
             }
         };
 
--- a/crates/agglayer-primitives/src/lib.rs
+++ b/crates/agglayer-primitives/src/lib.rs
@@ -3,3 +3,14 @@ pub use alloy_primitives::{address, ruint, Address, SignatureError, B256, U256, 
 pub use crate::signature::Signature;
 
 mod signature;
+
+/// Utility method for converting u32 words to bytes in big endian.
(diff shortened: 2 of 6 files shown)


crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -386,7 +386,7 @@ where
                     });
                 }
 
-                Some(aggchain_vkey.as_u32_array())
+                Some(sp1_reduce_proof.vk.hash_u32())
```
Ekleog-Polygon, line 389: Could we push uniformity further and stop using `hash_u32` altogether in favor of `hash_bytes`?

crates/agglayer-primitives/src/lib.rs
```diff
@@ -3,3 +3,14 @@ pub use alloy_primitives::{address, ruint, Address, SignatureError, B256, U256,
 pub use crate::signature::Signature;
 
 mod signature;
+
+/// Utility method for converting u32 words to bytes in big endian.
+pub fn words_to_bytes_be(words: &[u32; 8]) -> [u8; 32] {
```
Ekleog-Polygon, line 8: We could use [`byteorder::ByteOrder::BigEndian::write_u32_into`](https://docs.rs/byteorder/1.5.0/byteorder/trait.ByteOrder.html#tymethod.write_u32_into) in order to not write this function ourselves — we already depend on the byteorder crate ten times throughout our dependency graph, once more should be fine 😄

Ekleog-Polygon (approved):
LGTM! I just left a few comments for potential further improvements :)

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
iljakuklic: Unfortunately, the sp1 zkvm-side verification function expects the hash in the `hash_u32` format. It seems to be the main working format from which the other representations are derived for presentation purposes.

crates/pessimistic-proof-core/src/aggchain_proof.rs
```diff
@@ -53,7 +52,7 @@ impl AggchainData {
                 aggchain_vkey,
             } => keccak256_combine([
                 &(ConsensusType::Generic as u32).to_be_bytes(),
-                words_to_bytes_le(aggchain_vkey).as_slice(),
+                words_to_bytes_be(aggchain_vkey).as_slice(),
```
iljakuklic, line 55: I don't have the crate dependency graph fresh in mind but could we somehow reuse `aggkit_prover_types::vkey_hash::VKeyHash` type which already has methods for these conversions? If not, at least add an issue for it, so we can reuse the logic.

Freyskeyd: No we can't. Aggkit and pessimistic-proof are two different domains.
We can move this when we'll have a shared repo with everything related to interop

hadjiszs (approved):

iljakuklic (approved):

PR #723, opened 2025-03-20
feat: adding pp-vkey-selector to proof bytes

Freyskeyd:
# Description

This PR adds the `ppVkeySelector` to the proof sent to L1.

cc @krlosMata 

## PR Checklist:


atanmarko (approved):
Approved, although I am not a fan of manual `version.rs` generation.

crates/pessimistic-proof-core/Cargo.toml
```diff
@@ -41,4 +41,6 @@ rand = "0.9.0"
 
 [build-dependencies]
 semver = "1.0"
-toml = "0.8.20"
+toml.workspace = true
+# semver = "1.0"
+# toml = "0.8.20"
```
atanmarko, line 46: Remove

crates/agglayer-aggregator-notifier/src/packer/mod.rs
```diff
@@ -160,14 +160,19 @@ where
                 return Err(Error::InternalError("Unable to find the proof".to_string()));
             };
 
+        let mut proof_with_selector = pessimistic_proof::core::PESSIMISTIC_PROOF_PROGRAM_VERSION
+            .to_be_bytes()
```
Ekleog-Polygon, line 164: Can we have an intermediate variable like `proof_with_selector: [u8; 2] = VERSION.to_be_bytes()`? This would avoid someone in the future changing the type of PESSIMISTIC_PROOF_PROGRAM_VERSION and breaking everything.

Or even better, just have an alternative global, like `const PESSIMISTIC_PROOF_PROGRAM_SELECTOR: [u8; 2] = PESSIMISTIC_PROOF_PROGRAM_VERSION.to_be_bytes();`

crates/pessimistic-proof-core/build.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    let dest_path = Path::new(&env::var_os("OUT_DIR").expect("OUT_DIR not set")).join("version.rs");
+    fs::write(
+        &dest_path,
+        format!(
+            "pub const PESSIMISTIC_PROOF_PROGRAM_VERSION: u16 = {};\n",
+            major_version
+        ),
+    )
+    .expect("Failed to write pessimistic-proof-core version.rs");
```
Ekleog-Polygon, line 35: I'm also not a huge fan of this, it'd feel better to me to just expose it from the pessimistic-proof-program's lib.rs without any build script. But it should be good enough for now :)

Ekleog-Polygon (approved):
LGTM! The below is just for future-proofing, we can definitely land as-is 😄

crates/pessimistic-proof-core/build.rs
Freyskeyd: but there is no link between `proof-core` and `proof-program`, not in the right direction I mean

Ekleog-Polygon: Oh right, this is before the new crates split. Once we have the new crates split, we can have a dep from *-proof to *-proof-program that'd literally just retrieve the program version. Hopefully one day artifact dependencies will even stabilize in cargo, but we're not there yet 🥲 

Anyway, should not be a problem for landing!

Ekleog-Polygon (approved):

Ekleog-Polygon:
Attempting another land, hoping it's an intermittent

Ekleog-Polygon:
Let's do one last attempt, if it fails again here it might actually be due to this PR and not the same intermittent as other PRs have been seeing

atanmarko (approved):

Ekleog-Polygon (approved):

Ekleog-Polygon (approved):
I'm curious, do you know why we need to do all these changes for this specific PR? It seems to me like it should be mostly orthogonal to these matters, but it's hitting all the CI test issues

PR #730, opened 2025-03-26
ci: push docker images when updaing `main`

iljakuklic:
Here are some CI updates. There are two changes, pretty much independent of each other. We can pick which we want.

I am not able to test them properly. Running the workflow manually on this branch runs into an image upload permission issue. Running it manually on main works but that does not include the changes from this PR.

## PR Checklist:

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -61,6 +61,7 @@ jobs:
             type=ref,event=pr
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
+            type=raw,value=latest,enable={{is_default_branch}}
 
       - name: Build${{ inputs.push == true && '-Push' || '' }}
         uses: docker/build-push-action@v6
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -61,7 +61,6 @@ jobs:
             type=ref,event=pr
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
-            type=raw,value=latest,enable={{is_default_branch}}
 
       - name: Build${{ inputs.push == true && '-Push' || '' }}
         uses: docker/build-push-action@v6
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -213,42 +213,6 @@ jobs:
       - run: exit 1
         if: ${{ contains(fromJSON('["failure", "cancelled"]'), needs.call-cdk-e2e-workflow.result) }}
 
-  pushing-docker-image:
-    name: Docker | Publish image to registry
-    runs-on: ubuntu-latest
-    if: ${{ contains('["merge_group", "workflow_dispatch"]', github.event_name) }}
-    permissions:
-      packages: write
-    needs:
-      - docker-build-local
-      - check-e2e-result
-      - check_elf
-      - unit
-      - integrations
-    steps:
-      - name: Download artifact
-        uses: actions/download-artifact@v4
-        with:
-          name: "docker-image"
-          path: "/tmp"
-
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
-        with:
-          registry: ${{ env.REGISTRY }}
-          username: ${{ github.actor }}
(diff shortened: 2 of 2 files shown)


.github/workflows/docker-build.yml
```diff
@@ -61,6 +61,7 @@ jobs:
             type=ref,event=pr
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
+            type=raw,value=latest,enable={{is_default_branch}}
```
iljakuklic, line 64: This adds the `latest` tag to images build from `main`. They are already tagged by the branch name, i.e. `main`. It may be sensible to keep the original behaviour, i.e. have the `latest` tag point to the latest versioned release.

.github/workflows/test.yml
```diff
@@ -172,7 +172,7 @@ jobs:
     uses: ./.github/workflows/docker-build.yml
     secrets: inherit
     with:
-      push: false
+      push: true
```
iljakuklic, line 175: This pushes the image.

I am somewhat confused. I triggered a test build manually on `main` earlier today. It does not include this change and the image has [still been uploaded](https://github.com/agglayer/agglayer/pkgs/container/agglayer/382120050?tag=main) for some reason.

iljakuklic: Turns out there is a separate step to push the image but it is not triggering for `main` merge, only for manual run. Testing a fix now.

.github/workflows/docker-build.yml
iljakuklic: removed

atanmarko (approved):

.github/workflows/test.yml
```diff
@@ -216,7 +216,7 @@ jobs:
   pushing-docker-image:
     name: Docker | Publish image to registry
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/main' || ( github.event_name == 'workflow_dispatch' )
+    if: ${{ contains('["merge_group", "workflow_dispatch"]', github.event_name) }} # skip-merge queue
```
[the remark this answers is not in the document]
iljakuklic: > or at least have all the tests as dependencies, to be sure it runs as the last step of the merge group

Updated the PR to do this and seems to work. For development, there might be some merit in building docker images even for stuff that does not pass all the tests. However, not under the `main` tag.

.github/workflows/test.yml
```diff
@@ -216,12 +216,15 @@ jobs:
...
+    if: ${{ contains('["merge_group", "workflow_dispatch"]', github.event_name) }}
     permissions:
       packages: write
     needs:
       - docker-build-local
-      - call-cdk-e2e-workflow
+      - check-e2e-result
+      - check_elf
+      - unit
+      - integrations
```
Ekleog-Polygon, line 227: This is just missing `isolated-feature-checks`.

Also, maybe move it to the bottom of the file, so that it's a bit more obvious it needs to run last?

iljakuklic: > This is just missing `isolated-feature-checks`.

I omitted that intentionally as the check is sort of independent of the functionality of the built image.

> Also, maybe move it to the bottom of the file, so that it's a bit more obvious it needs to run last?

Ok.

Ekleog-Polygon: AFAIU it still means we'd end up with a docker image pushed that wouldn't actually reflect main, if a PR landed but failed the isolated-feature-checks, right?

Ekleog-Polygon (commented):
(commenting to handle the review request for now, see my comment above)

.github/workflows/test.yml
iljakuklic: > AFAIU it still means we'd end up with a docker image pushed that wouldn't actually reflect main, if a PR landed but failed the isolated-feature-checks, right?

Yea probably, good point.

iljakuklic: Added the feature checks. The list seems it will be somewhat difficult to maintain. However, the `push` trigger condition did not work previously. It's probably fixable but I'm don't want to spend ages on this.

Testing job: https://github.com/agglayer/agglayer/actions/runs/14086454150

Ekleog-Polygon (approved):
LGTM, thank you!

PR #734, opened 2025-03-27
ci: docker image build on `push` to `main`

iljakuklic:
This changes how Docker images are pushed to the registry by the test CI workflow.

* On PR updates, the docker build is skipped.
* On merge queue run, an image is built and used in e2e tests but not pushed.
* On push to `main`, the image is rebuilt (which should be fast due to caching) and pushed under the `main` tag.
* Triggering the workflow manually builds and pushes the image under the branch name (which may be `main`). It does not wait for the tests to complete, which may be useful for diagnosing issues.

## PR Checklist:


Ekleog-Polygon (commented):
I'm not convinced about this:
> On merge queue run, an image is pushed under a tag with the format of gh-readonly-queue-main-pr-$PR_NUMBER-$HEAD_SHA1. This does not wait for the tests to pass so may be used for debugging issues.

We're probably charged by the image stored, so having as many main builds as we want should be fine, but if we have multiple images per PR it could end up very expensive?

This being said, I'm not seeing it in the code, so I might be missing something?

iljakuklic:
> We're probably charged by the image stored, so having as many main builds as we want should be fine, but if we have multiple images per PR it could end up very expensive?

Hmm, maybe only push images for the `push` event then, will give it a go.

> This being said, I'm not seeing it in the code, so I might be missing something?

It's already behaving that way but conditioned upon the test jobs by the `needs` dependecies.

iljakuklic:
Tested the branch push path (although not on `main`), PR update path, manual trigger path. Did not test the merge queue path, not sure it can be easily done without attempting to merge. It now seems to work, hopefully without being too hard on image hosting.

I am not mega happy with the number of places that do stuff conditionally on the github event type. It's becoming slightly spahetti-esque.

.github/workflows/test.yml
```diff
@@ -158,7 +161,7 @@ jobs:
       docker-image-override: agglayer_image
       docker-tag: ${{ needs.docker-build-local.outputs.tags }}
       kurtosis-cdk-ref: 2527f38a96cedba1bff6cc94848ae77de8d44a37 # jihwan/integrations
-      docker-artifact-name: docker-image
+      docker-artifact-name: ${{ github.event_name == '"workflow_dispatch"' && 'docker-iamge' || '' }}
```
Ekleog-Polygon, line 164:
```suggestion
      docker-artifact-name: ${{ github.event_name == 'workflow_dispatch' && 'docker-image' || '' }}
```
This looks weird to me: why would we not override the `agglayer_image` when running from push?

.github/workflows/test.yml
```diff
@@ -140,16 +140,19 @@ jobs:
...
     name: E2E Tests | Docker build
-    if: ${{ contains('["merge_group", "workflow_dispatch"]', github.event_name) }} # skip-merge queue
+    if: ${{ contains('["merge_group", "workflow_dispatch", "push"]', github.event_name) }}
     uses: ./.github/workflows/docker-build.yml
     secrets: inherit
+    permissions:
+      packages: write
     with:
-      push: false
+      push: ${{ contains('["workflow_dispatch", "push"]', github.event_name) }}
```
Ekleog-Polygon, line 149: Uhhhh this will push the image before running the tests. I'm not sure I understand why you removed the pushing-docker-image job?

iljakuklic: > Uhhhh this will push the image before running the tests. I'm not sure I understand why you removed the pushing-docker-image job?

The PR merge flow is:
* Run the merge group trigger
  * Builds a docker image locally, saves as an artifact
  * e2e tests pick the image from the artifact
  * All the other checks are run in parallel
  * (The image is NOT pushed as a part of the merge group checks, hence the missing `pushing-docker-image` job. The jobs that need the image communicate via artifacts.)
* If the above passes, the `push` event on `main` is triggered:
  * Unit tests are run
  * Docker image is built (from cache) and pushed (by setting `push: true` here instead of using the `pushing-docker-image step`)

The manual trigger flow is as follows:
* On manual trigger
  * Builds a docker image and pushes it to ghcr (setting `push: true` here)
  * e2e tests pick the image from ghcr
  * All other checks run in parallel

.github/workflows/test.yml
iljakuklic: > This looks weird to me: why would we not override the `agglayer_image` when running from push?

When running from push, the image gets built and pushed to ghcr and subsequently picked up from there. When running form merge queue, it gets picked from artifacts.

This has to match what the docker build job does. We now have a number of conditionals that have to be consistent. This is one more such conditional. That's what I meant when I pointed out in the other comment that the code is becoming spaghetti-esque.

iljakuklic: ofc I got the condition the other way around :facepalm:

Issue #703, opened 2025-03-12
Switch native execution for PP to sp1 execute


Freyskeyd:
## Description

In order to check the `aggchain-proof` during native execution, we need to 

- [ ] switch from calling the rust function to a `client.execute()`
- [ ] add cross checks on target roots between witness generation and native execution given that the target checks got removed from the prover execution in #739

PR #739, opened 2025-03-31
refactor: mv target roots check outside of the program

invocamanman:
# Description
- Refactor code to avoid multiple sources of truth by removing the `target` and `previous_` values from the `multiBatchHeader`.

### Bug fix (non-breaking change which fixes an issue)

- Fix edge case were no exit was inserted, having and empty LER as an output, making impossible to verify, since in next iteration, `Digest::default()` will be expected

## PR Checklist:


crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -410,6 +410,8 @@ where
         let _ = generate_pessimistic_proof(initial_state.clone().into(), &multi_batch_header)
             .map_err(|source| CertificationError::NativeExecutionFailed { source })?;
 
+        // TODO: Cross check the roots between witness generation and native execution
```
hadjiszs, line 413: will have to be done together with #703

crates/pessimistic-proof-program/elf/riscv32im-succinct-zkvm-elf
Ekleog-Polygon, line 1: Hmmmmm -98% of the size of the binary? Is that normal?

crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -69,28 +63,6 @@ impl NetworkState {
...
-                declared: multi_batch_header.prev_balance_root,
-            });
-        }
-
-        if self.nullifier_tree.root != multi_batch_header.prev_nullifier_root {
-            return Err(ProofError::InvalidPreviousNullifierRoot {
-                computed: self.nullifier_tree.root,
-                declared: multi_batch_header.prev_nullifier_root,
-            });
-        }
```
Ekleog-Polygon, line 92: IIUC this removes the provided previous * roots check, and instead just assumes that the agglayer and PP agree on them.

Can we just get some logs somewhere, so that if the two ever get out of sync it'll be easy to debug?

Also, this PR title claims the check is moved outside of the program, but I didn't see the new place that actually does the checks when reviewing. Did I miss it, or did you forget to add the new ones maybe?

crates/pessimistic-proof-core/src/proof.rs
```diff
@@ -165,63 +173,156 @@ pub fn generate_pessimistic_proof(
...
+    // Compute the hash of the imported bridge exits
+    let imported_hash = commit_imported_bridge_exits(
+        multi_batch_header
+            .imported_bridge_exits
+            .iter()
+            .map(|(exit, _)| exit.global_index),
+    );
+
+    // Verify the aggchain proof which can be either one signature or one sp1 proof.
+    // NOTE: The STARK is verified exclusively within the SP1 VM.
```
Ekleog-Polygon, line 249: Note to self: when trying it out, we'll certainly need to disable deferred proof validation just like [in this code](https://github.com/agglayer/provers/pull/157)

crates/pessimistic-proof-core/src/proof.rs
```diff
@@ -165,63 +173,156 @@ pub fn generate_pessimistic_proof(
...
+                    == verify_signature(
+                        signature_commitment.commitment(PPRootVersion::V2),
+                        signature,
+                    )?
+                {
+                    PPRootVersion::V2
+                } else {
+                    return Err(ProofError::InvalidSignature);
+                }
+            };
```
Ekleog-Polygon, line 282: Could be cleaner written this way, especially for the day we'll have V4?

```rust
let target_pp_root_version = [PPRootVersion::V3, PPRootVersion::V2].map(|v| Ok(*signer == verify_signature(v, signature)?)).find(|r| r != Ok(false)).ok_or(ProofError::InvalidSignature)??;
```

crates/pessimistic-proof-core/src/proof.rs
```diff
@@ -165,63 +173,156 @@ pub fn generate_pessimistic_proof(
...
+
+            match (base_pp_root_version, target_pp_root_version) {
+                // From V2 to V2: OK
+                (PPRootVersion::V2, PPRootVersion::V2) => {}
+                // From V3 to V3: OK
+                (PPRootVersion::V3, PPRootVersion::V3) => {}
+                // From V2 to V3: OK (migration)
+                (PPRootVersion::V2, PPRootVersion::V3) => {}
+                // Inconsistent signed payload.
+                _ => return Err(ProofError::InconsistentSignedPayload),
```
Ekleog-Polygon, line 292: Can we add a `to_usize` function on `PPRootVersion` to avoid too much noise, so we can then just compare the integers instead of a big match?

crates/pessimistic-proof-program/elf/riscv32im-succinct-zkvm-elf
hadjiszs: `Bin 352988 -> 344808 bytes`

Github UI tells `Binary file modifiedBIN -7.99 KB (98%)`

I assume that this 98% is more to say "this file is ~unchanged at 98%"

crates/pessimistic-proof-core/src/proof.rs
hadjiszs: I would still have such match no matter what no? I simply thought to explicitly outline the valid cases of version transitions from `base_version` to `target_version`

crates/pessimistic-proof-core/src/proof.rs
hadjiszs: This code doesnt come from this PR, it has just been moved. The snippet doesnt work as is as it should return also the matched version, I have tried quickly but then it felt less cleaner than just in a unrolled way so I'll keep it as is for now 👍

atanmarko (approved):

atanmarko (approved):

PR #751, opened 2025-04-07
feat: align imported bridge exit and signature commitments

hadjiszs:
# Description

Fixes https://github.com/agglayer/agglayer/issues/754

Aligns the commitment on the imported bridge exits which is involved in the signature and in the aggchain proof public values.

The following emphasize the difference between `V2` and `V3` for both commitments:
- imported bridge exit commitment
- certificate signature commitment (which involves the imported bridge exit commitment)

### Imported bridge exit commitment version

The `V3` of this commitment is introduced in the `interop` repo in https://github.com/agglayer/interop/pull/12

```rust
/// The values which compose the commitment on the imported bridge exits.
pub struct ImportedBridgeExitCommitmentValues {
    pub claims: Vec<GlobalIndexWithLeafHash>,
}

impl ImportedBridgeExitCommitmentValues {
    /// Returns the expected signed commitment for the provided version.
    pub fn commitment(&self, version: Version) -> Digest {
        match version {
            V2 => {
                // Commits solely to the global index of each imported bridge exit. Designed
                // prior to having any notion of aggchain proof.
                keccak256_combine(
                    self.claims
                        .iter()
                        .map(|ibe| keccak256(ibe.global_index.as_le_slice())),
                )
            }
            V3 => {
                // Adds the bridge exit hashes in the commitment to ensure that the aggchain
                // proof and PP talk about the exact same set of imported bridge exits.
                keccak256_combine(self.claims.iter().map(|ibe| {
                    [
                        ibe.global_index.as_le_slice(),
                        ibe.bridge_exit_hash.as_slice(),
                    ]
                    .concat()
                }))
            }
        }
    }
}
```

This commitment is involved in the certificate signature and also in the aggchain proof.

### Certificate signature version

```rust
/// The values which compose the signature.
pub struct SignatureCommitmentValues {
    pub new_local_exit_root: Digest,
    pub commit_imported_bridge_exits: ImportedBridgeExitCommitmentValues,
    pub height: u64,
}

impl SignatureCommitmentValues {
    /// Returns the expected signed commitment for the provided version.
    pub fn commitment(&self, version: Version) -> Digest {
        let imported_bridge_exit_commitment = self.commit_imported_bridge_exits.commitment(version);
        match version {
            V2 => keccak256_combine([
                imported_bridge_exit_commitment.as_slice(),
            ]),
            V3 => keccak256_combine([
                imported_bridge_exit_commitment.as_slice(),
                self.height.to_le_bytes().as_slice(),
            ]),
        }
    }
}
```

## Misc

Also now uses the `AggchainProofPublicValues` structure moved in the `interop` repository.

## PR Checklist:


crates/pessimistic-proof-core/src/proof.rs
```diff
@@ -6,23 +6,29 @@ use serde::{Deserialize, Serialize};
...
-        commitment::PPRootVersion, commitment::PessimisticRoot, commitment::SignatureCommitment,
+        commitment::PessimisticRoot, commitment::SignatureCommitmentValues,
         commitment::StateCommitment, NetworkState,
     },
     multi_batch_header::MultiBatchHeader,
 };
 
+/// Version of the commitment on the imported bridge exits to verify the
+/// aggchain proof.
+pub const IMPORTED_BRIDGE_EXIT_COMMITMENT_VERSION: CommitmentVersion = CommitmentVersion::V2;
```
invocamanman, line 30: should be v3 right?

hadjiszs: yeah as discussed f2f taking it back to V3 as I set it in the PR in the `provers` side.

rational: the commitment in the certificate signature and the one in the aggchain proof public values does not need to be in the same version, so keeping it v2 or v3 changes nothing in terms of easing the integration process

invocamanman (dismissed):

Cargo.toml
```diff
@@ -115,8 +115,8 @@ sp1-primitives = "=4.1.4"
...
-agglayer-primitives = { git = "https://github.com/agglayer/interop.git", rev = "v0.2.0" }
-agglayer-tries = { git = "https://github.com/agglayer/interop.git", rev = "v0.2.0" }
-agglayer-interop = { git = "https://github.com/agglayer/interop.git", rev = "v0.2.0" }
-unified-bridge = { git = "https://github.com/agglayer/interop.git", rev = "v0.2.0" }
-agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", rev = "v0.2.0" }
+agglayer-primitives = { git = "https://github.com/agglayer/interop.git", branch = "feat/commit-imported" }
+agglayer-tries = { git = "https://github.com/agglayer/interop.git", branch = "feat/commit-imported" }
+agglayer-interop = { git = "https://github.com/agglayer/interop.git", branch = "feat/commit-imported" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "feat/commit-imported" }
+agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", branch = "feat/commit-imported" }
```
Ekleog-Polygon, line 122: Can we make a tag and point to it? Pointing to a feature branch sounds weird to me

crates/agglayer-types/src/lib.rs
```diff
@@ -291,11 +292,12 @@ impl Default for Certificate {
         let network_id = Default::default();
         let wallet = Self::wallet_for_test(network_id);
         let exit_root = LocalExitTree::<Keccak256Hasher>::default().get_root();
+        let height = Default::default();
```
Ekleog-Polygon, line 295: `height` is `u64`, can we just set it to `0` instead of the weird `Default::default`? Or if we really want to be ready for the day `Height` will be a newtype wrapper, we could use `Height::MIN` that'll work with `u64` and at least has semantic meaning.

EDIT: Actually this is an `impl Default for Certificate`… that also sounds wrong and we should have something like `Certificate::for_test()` instead. That's outside the scope of this PR… but it makes me sad 🥲

crates/agglayer-types/src/lib.rs
```diff
@@ -422,12 +430,8 @@ impl Certificate {
         match self.aggchain_data {
             AggchainData::ECDSA { signature } => {
                 // retrieve signer
-                let combined_hash = signature_commitment(
-                    self.new_local_exit_root,
-                    self.imported_bridge_exits
-                        .iter()
-                        .map(|exit| exit.global_index),
-                );
+                let version = CommitmentVersion::V2;
```
Ekleog-Polygon, line 433: I'm curious, you're setting `version` as a variable twice now before using it… why?

crates/pessimistic-proof-core/src/proof.rs
```diff
@@ -6,23 +6,29 @@ use serde::{Deserialize, Serialize};
...
-        commitment::PPRootVersion, commitment::PessimisticRoot, commitment::SignatureCommitment,
+        commitment::PessimisticRoot, commitment::SignatureCommitmentValues,
         commitment::StateCommitment, NetworkState,
     },
     multi_batch_header::MultiBatchHeader,
 };
 
+/// Version of the commitment on the imported bridge exits to verify the
+/// aggchain proof.
+pub const IMPORTED_BRIDGE_EXIT_COMMITMENT_VERSION: CommitmentVersion = CommitmentVersion::V3;
```
Ekleog-Polygon, line 30: I'm not sure I understand the comment, maybe you could expand on it a tiny bit? My current understanding is this is basically `LATEST_COMMITMENT_VERSION`, which could then be an associated `const fn latest()` of `CommitmentVersion`, but maybe I'm missing some semantics here?

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/lib/Cargo.toml
```diff
@@ -11,4 +11,4 @@ tiny-keccak = { git = "https://github.com/sp1-patches/tiny-keccak", tag = "patch
     "keccak",
 ] }
 
-pessimistic-proof-core = { path = "../../../pessimistic-proof-core/" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "feat/commit-imported" }
```
Ekleog-Polygon, line 14: Same here, we should be using an officially-released version

crates/pessimistic-proof-test-suite/src/forest.rs
```diff
@@ -209,12 +213,17 @@ impl Forest {
 
         let new_local_exit_root = self.state_b.exit_tree.get_root();
 
-        let (_combined_hash, signature, _signer) =
-            compute_signature_info(new_local_exit_root, &imported_bridge_exits, &self.wallet);
+        let height = 0;
+        let (_combined_hash, signature, _signer) = compute_signature_info(
```
Ekleog-Polygon, line 217: Just thinking out loud: would it make sense to have `compute_signature_info` be a method of `Certificate`? Or do we also use it in other contexts where there's no `Certificate` to call it on?

Ekleog-Polygon (changes requested):
LGTM! My only negative comment down there is that we probably shouldn't point our cross-repo dependencies to a feature branch; aside from that I mostly have a few questions down below 😄

atanmarko (commented):
LGTM

crates/agglayer-types/src/lib.rs
hadjiszs: not sure to see which two occurrences you're referring to?

crates/pessimistic-proof-test-suite/aggchain-proof-ecdsa-example/lib/Cargo.toml
hadjiszs: yeah, that's mainly for integration and CI to pass the 3 PRs, but then definitely will move to proper rev once the PR in interop get merged first

crates/pessimistic-proof-test-suite/src/forest.rs
hadjiszs: requires bit of changes in the tests which I'd prefer to not touch too much for now

crates/pessimistic-proof-core/src/proof.rs
hadjiszs: There are two commitment versions on the imported bridge exits defined in the interop linked PR.

This commitment is involved in the aggchain proof public values: that's the `commit_imported_bridge_exits` field.

The constant `IMPORTED_BRIDGE_EXIT_COMMITMENT_VERSION` defines which commitment version is expected as part of the aggchain proof public values. I'll check how I can expand a bit this constant comment description

atanmarko (approved):

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: This one and the one here: https://github.com/agglayer/agglayer/pull/751/files/382024f895a733ac06cec32adc6a4e15c5c4d010#diff-a5ff868adb803b1ad8675ca67c1b9d46b0c8c99c8842684ff887070c1a6c96d8R321

Basically I don't really understand why you're making a variable for a single use, considering the name doesn't serve any documentation purposes, as it's already in `CommitmentVersion`?

Ekleog-Polygon (dismissed):

crates/agglayer-types/src/lib.rs
hadjiszs: ah that's because the signature here is purely V2 but only for now, but then will need to verify either a V2 or V3 in the same fashion as what we do in the PP

atanmarko (approved):

Ekleog-Polygon (approved):

PR #764, opened 2025-04-16
feat: add script to build the docker images

Ekleog-Polygon:
## PR Checklist:

### DIFF
--- a/scripts/build-docker-images
+++ b/scripts/build-docker-images
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+docker build -t agglayer:local .


scripts/build-docker-images
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+docker build -t agglayer:local .
```
atanmarko, line 3: Not sure about use-fulness of this? Maybe to add script argument for tag?

Ekleog-Polygon: The context is https://0xpolygon.slack.com/archives/C08K3PSSLM8/p1744774243892639 ; the goal being to have the same script everywhere because every repo has a different way of building the docker images — ours is relatively simple but there are other repos where even after trying for a while I didn't manage to build their docker images.

As for the name, I'd say we could pick any image name and it doesn't matter? It's just a way to try to standardize building the docker images across repos

atanmarko: Ok, then :)

atanmarko (approved):
Nit: add a comment that it should be executed from the project root.

Ekleog-Polygon:
Good point! I just pushed such a comment 😄

iljakuklic (approved):

PR #769, opened 2025-04-17
fix: json serialization backwards compat for `Certificate`

iljakuklic:
## PR Checklist:


crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -95,17 +95,23 @@ impl From<CertificateV0> for Certificate {
             metadata,
         } = certificate;
 
+        let l1_info_tree_leaf_count = imported_bridge_exits
+            .iter()
+            .map(|ibe| ibe.l1_leaf_index() + 1)
+            .max()
+            .unwrap_or(0);
```
iljakuklic, line 102: I am not 100% sure the default of 0 will behave correctly here.

hadjiszs: yeah I might need to re-introduce the default l1 info tree leaf entry in the edge case where we receive an old certificate without any imported bridge exits

hadjiszs (approved):

PR #771, opened 2025-04-18
feat: switch to sp1 execute and add root cross checks

hadjiszs:
# Description

Fixes https://github.com/agglayer/agglayer/issues/703

## PR Checklist:


crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -408,14 +445,24 @@ where
...
-        let _ = generate_pessimistic_proof(initial_state.clone().into(), &multi_batch_header)
-            .map_err(|source| CertificationError::NativeExecutionFailed { source })?;
+        let targets_witness_generation: StateCommitment = {
+            let ns: LocalNetworkState = state.clone().into();
+            NetworkState::from(ns).get_state_commitment()
+        };
 
-        // TODO: Cross check the roots between witness generation and native execution
+        // Perform the native PP execution without the STARK verification in order to
+        // cross check the target roots.
```
Ekleog-Polygon, line 454: Do I understand correctly that we now run the proof code 3 times:
1. Inside the proving system
2. In the SP1 VM, in execute mode
3. Outside the SP1 VM

If so, why do we keep execution number 3, instead of just keeping number 1 and 2?

hadjiszs: >Do I understand correctly

Yeah

>If so, why do we keep execution number 3, instead of just keeping number 1 and 2?

Here is a small support:

image

With

```rust
pub struct StateCommitment {
    pub exit_root: Digest,
    pub ler_leaf_count: u32,
    pub balance_root: Digest,
    pub nullifier_root: Digest,
}


pub struct PessimisticProofOutput {
    /// The previous local exit root.
    pub prev_local_exit_root: Digest,
    /// The previous pessimistic root.
    pub prev_pessimistic_root: Digest,
    /// The l1 info root against which we prove the inclusion of the imported
    /// bridge exits.
    pub l1_info_root: Digest,
    /// The origin network of the pessimistic proof.
    /// The aggchain hash.
    pub aggchain_hash: Digest,
    /// The new local exit root.
    pub new_local_exit_root: Digest,
    /// The new pessimistic root.
    pub new_pessimistic_root: Digest,
}
```

I wanted to add such granularity of error for faster debug from logs.

If I drop (3), then the only thing that I could do (and that wont be nice for debugging) would be to construct one `PessimisticProofOutput` from the witness generation step, and to compare it with the one from (2). Consequently, any issue on the roots would be opaque because of the `pessimistic_root` which hash the roots together, hence the intention of having the step 3 in order to make comparisons on `StateCommitment` prior to executing anything on sp1.

In other words, if we drop 3, then we would be able to know that there is a mismatch on the pessimistic root, but wont be able to know if the mismatch is for example on the balance root, or the nullifier root, but with (3) we can.

hadjiszs: >Is the SP1 VM unable to output debug logs while executing?

I'm open with the approach of only logging, but so far within sp1 I've only manage doing logging attempts with eprintln without tracing, not sure how it would play out with datadog and such, maybe @iljakuklic @Freyskeyd have ideas

Ekleog-Polygon: When you tried with tracing, had you built and configured a new tracing subscriber in the main function of the proof program?

hadjiszs: I am getting panic upon setting up the subscriber with `time not implemented on this platform`. Not sure that it is trivial to setup tracing/logging within the zkvm and could probably be another task in itself. Alternative would be println as mentioned above, but not sure how it plays out with datadog etc.

That being said, the rust native execution is already present in `main`, this PR is mainly about bringing more checks alongside the sp1 zkvm native execution, and I am not entirely sure that it is very critical to remove the rust native call to `generate_pessimistic_proof` but I might have forgotten details about the crates refactoring that we discussed 1~2 months ago

Ekleog-Polygon: You're right it's not critical; I was just hoping to not pile on more tech debt 🥲

Can we maybe just use the native execution only for logging for now then, instead of adding a debug-only check?

As part of the crates refactor we discussed killing all native execution; and if we add more things that rely on it it'll make hesitate to kill it later on

hadjiszs: > You're right it's not critical; I was just hoping to not pile on more tech debt 🥲

I don't think we are piling into more tech debt given that the rust native call is still there and even more required in `main` than after this PR, I mean that's not like we removed it in `main` and that we're re-introducing it here I believe

> Can we maybe just use the native execution only for logging for now then, instead of adding a debug-only check?

yeah done 👍

Ekleog-Polygon (approved):
LGTM, thank you!

PR #790, opened 2025-05-01
test: certificate status storage encoding snapshots

iljakuklic:
Check storage encoding is not accidentally broken by adding / reordering enum arms in `CertificateStatus` or various error types.

## PR Checklist:


crates/agglayer-types/src/lib.rs
```diff
@@ -166,12 +166,12 @@ pub enum CertificateStatusError {
...
     /// Failure on the proof verification.
     #[error("proof verification failed")]
-    ProofVerificationFailed(#[from] ProofVerificationError),
+    ProofVerificationFailed(#[source] ProofVerificationError),
 
     /// Failure on the pessimistic proof witness generation from the
     /// [`LocalNetworkStateData`] and the provided [`Certificate`].
     #[error(transparent)]
-    TypeConversionError(#[from] Error),
+    TypeConversionError(Error),
```
Ekleog-Polygon, line 174: Do we not want this to be `#[source]` still?

Ekleog-Polygon (approved):

crates/agglayer-types/src/lib.rs
iljakuklic: > Do we not want this to be `#[source]` still?

That's not compatible with `#[error(transparent)]` so I added an error string.

Ekleog-Polygon (approved):
LGTM, thank you! And sorry, I don't know how, when reviewing, I missed the `error(transparent)` line just above 😅

PR #812, opened 2025-05-21
fix: clean up network task after a failure

iljakuklic:
## PR Checklist:

### DIFF
--- a/crates/agglayer-certificate-orchestrator/src/lib.rs
+++ b/crates/agglayer-certificate-orchestrator/src/lib.rs
@@ -19,7 +19,7 @@ use agglayer_storage::{
 };
 use agglayer_types::{CertificateId, Height, NetworkId};
 use arc_swap::ArcSwap;
-use futures_util::{stream::FuturesUnordered, FutureExt, Stream, StreamExt};
+use futures_util::{stream::FuturesUnordered, FutureExt, Stream, StreamExt, TryFutureExt};
 use network_task::{NetworkTask, NewCertificate};
 use tokio::{
     sync::mpsc::{self, Receiver},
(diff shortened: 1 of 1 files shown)


crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -45,8 +45,9 @@ const MAX_POLL_READS: usize = 1_000;
 pub type EpochPackingTasks =
     FuturesUnordered<Pin<Box<dyn Future<Output = Result<(), Error>> + Send + 'static>>>;
 
-pub type NetworkTasks =
-    FuturesUnordered<Pin<Box<dyn Future<Output = Result<NetworkId, Error>> + Send + 'static>>>;
+pub type NetworkTasks = FuturesUnordered<
+    Pin<Box<dyn Future<Output = Result<NetworkId, (NetworkId, Error)>> + Send + 'static>>,
```
Ekleog-Polygon, line 49: Would this type make more sense?

```suggestion
    Pin<Box<dyn Future<Output = (NetworkId, Result<(), Error>)> + Send + 'static>>,
```

Ekleog-Polygon (approved):

crates/agglayer-certificate-orchestrator/src/lib.rs
iljakuklic: > Would this type make more sense?

I agree on its face but it would involve either changing the return type of the network task itself or somewhat more awkward postprocessing of the result of the task future.

Imho better keep this as a fix and do various cleanups as a part of a bigger refactor. I can imagine we'd want to make the network task into a proper tokio task at some point, to avoid the manual future juggling we have now. And this whole thing will probably go away when that happens.

Ekleog-Polygon: Sounds good to me!

PR #814, opened 2025-05-22
chore: Remove unused structures

c410-f3r:
# Description

Removes fields and functions that are not used anywhere. It shouldn't be too difficult to recover them from the GIT history if needed in the future, but it's unlikely.

## PR Checklist:

### DIFF
--- a/crates/agglayer-storage/src/columns/proof_per_certificate/tests.rs
+++ b/crates/agglayer-storage/src/columns/proof_per_certificate/tests.rs
@@ -40,7 +40,7 @@ fn can_parse_value() {
 #[derive(Deserialize)]
 struct SP1ProofWithPublicValuesV3 {
     pub proof: sp1_sdk::SP1Proof,
-    #[allow(unused)]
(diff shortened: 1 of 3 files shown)


crates/agglayer-storage/src/columns/proof_per_certificate/tests.rs
```diff
@@ -40,8 +40,6 @@ fn can_parse_value() {
 #[derive(Deserialize)]
 struct SP1ProofWithPublicValuesV3 {
     pub proof: sp1_sdk::SP1Proof,
-    #[allow(unused)]
-    pub stdin: sp1_sdk::SP1Stdin,
```
iljakuklic, line 44: I don't think this change is correct. It changes the serialization format which is used to load the values of this type from files.

c410-f3r: Thank you for the feedback.

This specific field now has a "reason" declaration to avoid future accidental removals

PR #817, opened 2025-05-23
fix: clippy 0.1.88

atanmarko:
Fixes for clippy version 0.1.88, mostly printout formatting.


iljakuklic (approved):
Would be nice to apply the inline formatting to logs too. Unfortunately, it seems clippy does not do that.

PR #818, opened 2025-05-23
refacor: use shared bincode options

iljakuklic:
# Description

* depends on https://github.com/agglayer/interop/pull/26
* [ ] wait for interop release

## PR Checklist:


atanmarko (approved):

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -196,7 +192,7 @@ where
 
         let request = GenerateProofRequest {
             stdin: Some(Stdin::Sp1Stdin(
-                default_bincode_options()
+                bincode::default()
```
Ekleog-Polygon, line 195: Nit: we have a `bincode::sp1`, it might be good to add `bincode::sp1stdin` so that we avoid thinking "why not sp1 here?" — exactly my thought process as I was reading this

crates/agglayer-storage/src/types/certificate/tests/snapshots/agglayer_storage__types__certificate__tests__aggdata_v1_01.snap
```diff
@@ -3,4 +3,4 @@ source: crates/agglayer-storage/src/types/certificate/tests/mod.rs
...
...000000000143421305623730095048800168872402096980078569670187537634cc321d435ea2e421d87e97032424500e489aa01a61b9bb6da39f75000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000523cf1e0096e2d902927946e633ab8ac50ff1a805b62a642220893c33a5680db002015042bef52c006dd64c800308ec805f273b7341b15a036fc3ebf56e5a594077c514b0f6ebea522830a240e8a7e8648dc416e3a2b987367eb006d0000000000000002000000000000000750726f6772616d000000000000001300000001000000000000000e0000000000080000000000000000000442797465000000000000001000000001000000000000000b00000000000100000000000000000002000000000000000750726f6772616d000000000000000000000000000000044279746500000000000000010000000000000005312e322e3358585858585858585858585858585858585858585858585858585858585858580000000000000020787878787878787878787878787878787878787878787878787878787878787800000000000000209a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a00
```
Ekleog-Polygon: These changes… do not look normal to me? afaict all the changes should be impact-free on the storage, should they not?

iljakuklic: > These changes… do not look normal to me? afaict all the changes should be impact-free on the storage, should they not?

Yea correct, this should not change the storage format.

It happens for this reason:
* The snapshot uses certificate with the mock proof
* The mock proof also contains vkey of the ELF the mock prover was initialised with.
* The mock prover is (quite stupidly) initialised with PP elf
* So every time PP elf changes, the shanpshot here changes

I double checked this is the case.

It would be better to initialize the snapshot with some more stable ELF, e.g. just an empty `main`. I changes this to the ecdsa example elf in one of the other PRs but maybe it would be better to do it as a separate change to make the diff on this PR clean.

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
iljakuklic: > Nit: we have a `bincode::sp1`, it might be good to add `bincode::sp1stdin` so that we avoid thinking "why not sp1 here?" — exactly my thought process as I was reading this

Unfortunately, this would be misleading. Passing data to `Sp1Stdin` by default uses `bincode::sp1` (little endian) but not in this case, where we stick to our default encoding (big endian), for whatever reason.

crates/agglayer-storage/src/types/certificate/tests/snapshots/agglayer_storage__types__certificate__tests__aggdata_v1_01.snap
iljakuklic: > It would be better to initialize the snapshot with some more stable ELF, e.g. just an empty `main`.

Done here:
* #842

iljakuklic: > > It would be better to initialize the snapshot with some more stable ELF, e.g. just an empty `main`.
> 
> Done here:
> * #842

This PR is now based on #842 so the diff is now clean.

crates/agglayer-storage/src/types/certificate/tests/mod.rs
```diff
@@ -9,6 +9,8 @@ use sp1_sdk::Prover;
 use super::*;
 use crate::columns::Codec;
 
+const EMPTY_ELF: &[u8] = include_bytes!("empty.elf");
```
Ekleog-Polygon: Will we need to rebuild the empty elf on sp1 version upgrades? And if yes, can we have some script to not have to search how to rebuild it next time?

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
Ekleog-Polygon: Got it, while I'm sad about that I guess changing it would be harder than it's worth 🥲

crates/agglayer-storage/src/types/certificate/tests/mod.rs
iljakuklic: > Will we need to rebuild the empty elf on sp1 version upgrades? And if yes, can we have some script to not have to search how to rebuild it next time?

I don't know. Will add a script when it turns out we do, as it is not completely trivial with the current tooling we have.

Freyskeyd (approved):

PR #819, opened 2025-05-23
feat: separate the logic currently inside NetworkTask out towards CertificateTask's

Ekleog-Polygon:
Part of [Tech debt: Improve the certificate flow in the orchestrator](https://github.com/agglayer/protocol-research/issues/148#top)#148

Currently, the NetworkTask handles both orchestration of certificate proving and the certificate proving itself.

This PR is a first step towards simplifying that code, making certificate proving into a separate task, and leaving only orchestration inside NetworkTask.

Future work is:
- Removing certificate storage writes from everywhere but the CertificateTask itself. This will remove the need for locks that is in `// TODO` at a few places.
- Updating the network storage so that instead of storing `network_id -> state` we store `network_id -> height -> state`. This will allow us to process multiple certificates in parallel for a given network, as well as help us recover from poweroff much more easily. It will probably be essential for preconfs, as we will need a way to validate that a certificate relying on a preconf (and thus not yet present in settled storage) is indeed valid.
- If we want to process multiple certificates in parallel indeed, introduce some additional orchestration inside the network task, so that we could still limit to eg. 1 certificate undergoing settlement at a time.

## PR Checklist:


iljakuklic (approved):
Looks good. Some minor nits and observations below.

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -138,6 +173,7 @@ where
 
         loop {
             tokio::select! {
+                // TODO (IN ANOTHER PR): move cancellation token to make_progress, have make_progess return ControlFlow?
```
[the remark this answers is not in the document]
Ekleog-Polygon: Good point! TBH in my network task refactors, after the storage changes I think I'll become able to remove `make_progress` altogether and have `NetworkTask` do little more than just looping on its incoming message queue and dispatching messages as required 😄

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -250,459 +286,90 @@ where
...
-                    CertificationError::LastPessimisticRootNotFound(network_id) => {
-                        CertificateStatusError::LastPessimisticRootNotFound(network_id)
-                    }
-                    CertificationError::ProofVerificationFailed { source } => {
-                        CertificateStatusError::ProofVerificationFailed(source)
-                    }
-                    CertificationError::L1InfoRootNotFound(_certificate_id, l1_leaf_count) => {
-                        CertificateStatusError::L1InfoRootNotFound(l1_leaf_count)
+        loop {
+            tokio::select! {
```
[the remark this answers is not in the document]
Ekleog-Polygon: Hmm it would probably make sense; but with the various `return`/`continue`s sprinkled around it'd take a bit more effort. After the storage refactoring, as per my other comment I'm hoping to kill `make_progress` altogether, at which point there should no longer be any `return` sprinkled around (beyond just returning a `Result` to handle unrecoverable errors).

Would you be fine with landing this as is currently, and making sure the last changes to network_task.rs are actually good? 😄

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,263 @@
...
+            trace!("Retrieving local network state");
+            let (response, state) = oneshot::channel();
+            self.network_task
+                .send(NetworkTaskMessage::GetLocalNetworkStateBeforeHeight { height, response })
+                .await
+                .map_err(send_err)?;
+            let state = state.await.map_err(recv_err)?;
+
+            // Actually certify
+            debug!("Starting certification");
```
atanmarko, line 127: Add some more info to debug message, e.g. height, certificate id, network_id...here and on other places

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,263 @@
...
+            let certifier_output = self
+                .certifier_client
+                .certify(*state, network_id, height)
+                .await?;
+            debug!("Proof certification completed");
+
+            // Record the certification success
+            self.header.status = CertificateStatus::Proven;
+            self.state_store
+                .update_certificate_header_status(&certificate_id, &CertificateStatus::Proven)?;
```
atanmarko, line 137: Should we first update state store and then mark the header status proven?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,263 @@
...
+                })
+                .await
+                .map_err(send_err)?;
+        } else {
+            // TODO: once we store network_id -> height -> state and not just network_id ->
+            // state, we should not need this any longer, because the state will
+            // already be recorded.
+
+            // Retrieve local network state
+            trace!("Retrieving local network state");
```
atanmarko, line 152: Add more info to the traces, cert id... By itself this one does not say much.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,263 @@
...
+                CertificateStatusError::SettlementError(
+                    "Settled certificate header has no settlement tx hash".into(),
+                )
+            })?;
+            if !self
+                .settlement_client
+                .transaction_exists(tx_hash.0.into())
+                .await?
+            {
+                return Err(CertificateStatusError::SettlementError(
```
atanmarko, line 217: We should probably refactor `SettlementError` to be self sufficient, returning more info. What if we process multiple certificates in parallel, would we know what certificate "Settlement transaction not found" happened from the logs?

crates/agglayer-certificate-orchestrator/src/network_task.rs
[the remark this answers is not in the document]
iljakuklic: > Would you be fine with landing this as is currently, and making sure the last changes to network_task.rs are actually good? It'd avoid a refactoring now that'll only be deleted in 1-2 PRs 😄

Sounds good to me

hadjiszs (approved):

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -255,459 +287,90 @@ where
...
-                        CertificateStatusError::LastPessimisticRootNotFound(network_id)
-                    }
-                    CertificationError::ProofVerificationFailed { source } => {
-                        CertificateStatusError::ProofVerificationFailed(source)
+        loop {
+            tokio::select! {
+                msg = receiver.recv() => match msg {
+                    None => {
+                        error!(height = *next_expected_height, %certificate_id, "Certificate task channel closed");
+                        return Err(Error::InternalError("Certificate task channel closed".into()));
```
Freyskeyd, line 315: In case of a certificate already settled (which return `Ok(())` and drop the sender), it means that we end up here, right?

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -255,459 +287,90 @@ where
...
                     }
-                    CertificationError::L1InfoRootNotFound(_certificate_id, l1_leaf_count) => {
-                        CertificateStatusError::L1InfoRootNotFound(l1_leaf_count)
+                    Some(NetworkTaskMessage::GetLocalNetworkStateBeforeHeight { response, .. }) => {
+                        let state = self.local_state.clone();
+                        response.send(state).map_err(|_| {
```
Freyskeyd, line 319: If we can't send the `state`, can the certificate task still be alive ? If yes, how can we kill it?

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -255,459 +287,90 @@ where
...
-                    CertificationError::Deserialize { source } => {
-                        let error_msg = format!(
-                            "Deserialization error happened in the certification process of \
-                             {certificate_id}",
+                    Some(NetworkTaskMessage::CertificateSettled { settled_certificate, .. }) => {
+                        self.at_capacity_for_epoch = true;
+                        self.latest_settled = Some(settled_certificate);
+                        *next_expected_height += 1;
+                        debug!(%certificate_id, "Certification process completed");
+                        let Some(new) = self.pending_state.take() else {
```
Freyskeyd, line 342: This should be done before altering the different fields above, right?
If we end up having no state to update in DB, we'll still have the `next_expected_height` updated

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: As discussed this morning, this is all automatically added by this span, both here and everywhere else 😄 

https://github.com/agglayer/agglayer/pull/819/files/4a6ef09e3b046c930a6a6216a52d578e1117d317#diff-6670f8d5c0a95267f210b07afd4dd274126416037a4f01cb61920ef0fdad5da9R45-R53

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,263 @@
...
+        // TODO: when all the storage related to this cert is only ever handled from the
+        // certificate task, the certificate task should be the one to start
+        // with storing the certificate if needed.
+
+        debug!(initial_status = ?self.header.status, "Processing certificate");
+
+        // Skip if there's nothing to do
+        if self.header.status == CertificateStatus::Settled {
+            warn!("Built a CertificateTask for a certificate that is already settled");
+            return Ok(());
```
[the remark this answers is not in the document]
Ekleog-Polygon: This is currently true. This being said, I had actually hesitated making that an `assert!`, before I remembered that security team doesn't like panics. IMO if this code path is being reached, it means there's a bug in our code.

I'm also hoping that the future refactorings will make the NetworkTask code (much) simpler, at which stage it'd be more obvious we're not calling this code path.

Does that make sense, or do you think it'd be better to have the CertificateTask lie and claim it settled the certificate again? TBH I'm a bit more concerned about the "lie" option, but considering I don't think this code path will ever be touched, I'm not strongly against it

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,263 @@
...
+                .witness_generation(&self.certificate, &mut state)
+                .await
+                .map_err(|error| {
+                    error!(%certificate_id, ?error, "Failed recomputing the new state for already-proven certificate");
+                    error
+                })?;
+            debug!("Recomputing new state completed");
+
+            // Record the certification success
+            self.network_task
```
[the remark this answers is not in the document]
Ekleog-Polygon: This is an interesting question.

For now, we have a single inflight cert, so if we have a Candidate cert, then it's also the latest_proven_certificate. It means that re-setting it as latest proven is not going to be a problem.

If we had multiple inflight certs, it could be an issue: we'd then possibly revert the latest proven cert. However, we won't have multiple inflight certs until we fix the storage, and by the time we fix the storage we'll no longer need this whole code block.

I've edited the comments and added a todo to make things clear, is that good enough for you for now? Otherwise I could introduce a new NetworkTaskMessage::UpdateLocalNetworkState, but it'd need to be removed at the time of the storage refactor; and AFAICT it wouldn't help until then.

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -255,459 +287,90 @@ where
...
-        {
-            Ok(CertifierOutput {
-                height,
+        let bridge_exit_hashes = certificate
+            .bridge_exits
+            .iter()
+            .map(|exit| exit.hash())
+            .collect::<Vec<Digest>>();
+        tokio::spawn(
+            CertificateTask::new(
```
iljakuklic, line 299: The handle of the certificate task is immediately dropped. Maybe it should be polled in the loop below so we can detect and log any potential panics. Also maybe await it at the end when the loop exits for the same reason + to have the task lifetimes properly nested which should ease the reasoning about them somewhat.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,282 @@
...
+                .map_err(send_err)
+                .unwrap_or_else(|error| {
+                    error!(?error, "Failed to send certificate error message");
+                });
+        }
+    }
+
+    /// Process a certificate, not doing any specific error handling except for
+    /// returning it
+    async fn process_impl(&mut self) -> Result<(), CertificateStatusError> {
```
iljakuklic, line 100: Would a structure like this be perhaps slightly clearer?

```rust
fn process_impl(&mut self) -> Result<(), _> {
    match self.cert_status {
        Pending => self.process_pending()?,
        Candidate => self.process_candidate()?,
        Proven => self.process_proven()?,
        Settled => log!("already done"),
        InError { error } => log!("oops, we got {error}"),
    }
}

fn process_pending(&mut self) -> Result<(), _> {
    // do processing
    self.set_status(Candidate);
    self.process_candidate()
}

fn process_candidate(&mut self) -> Result<(), _> {
    // do processing
    self.set_status(Proven);
    self.process_proven()
}

fn process_proven(&mut self) -> Result<(), _> {
    // do processing
    //self.set_status(Settled);  // done elsewhere for now
    Ok(())
}
```

I am not sure, however, how that interacts with settlement task doing the status update by itself. And I may be missing some other considerations.

crates/agglayer-types/src/lib.rs
```diff
@@ -259,6 +264,32 @@ impl std::fmt::Display for CertificateStatus {
...
+            InError { .. } => 4,
+        }
+    }
+}
+
+impl Ord for CertificateStatus {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.as_order_number().cmp(&other.as_order_number())
+    }
+}
```
iljakuklic, line 291: Since `InError` is not in the usual flow, perhaps this should stay as `PartialOrd` only with `InError` being not comparable to the other ones? Seems conceptually cleaner to me but it may complicate handling at use sites. somewhat.

Or perhaps if the other suggestion turns out to be feasible, the `Ord` for `CertificateStatus` will not be needed after all.

Ekleog-Polygon: I actually started with only `PartialOrd`, and then added `Ord` for convenience and to be sure we don't forget to add new status types to the right place in the chain 😅 

But as you say, as soon as the other task is done we can remove the whole `PartialOrd` implementation!

atanmarko:
LGTM

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,282 @@
...
+                .send(NetworkTaskMessage::CertificateErrored {
+                    height: self.header.height,
+                    certificate_id: self.header.certificate_id,
+                    error,
+                })
+                .await
+                .map_err(send_err)
+                .unwrap_or_else(|error| {
+                    error!(?error, "Failed to send certificate error message");
+                });
```
iljakuklic, line 111: If the task handle was checked in the network task message loop, the messages sent at "tail position" (just before the certificate task exits) could be passed in the return value instead. This applies to `CertificateErrored` here and `CertificateSettled` below. WDYT?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,282 @@
...
+            self.state_store
+                .update_certificate_header_status(&certificate_id, &CertificateStatus::Proven)?;
+            self.network_task
+                .send(NetworkTaskMessage::CertificateProven {
+                    height,
+                    certificate_id,
+                    new_state: Box::new(certifier_output.new_state),
+                })
+                .await
+                .map_err(send_err)?;
```
iljakuklic, line 186: Nit: this pattern could be extracted into a helper method which does the `map_err` internally:
```suggestion
            self.notify_network_task(NetworkTaskMessage::CertificateProven {
                    height,
                    certificate_id,
                    new_state: Box::new(certifier_output.new_state),
                })
                .await?;
```

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -255,459 +287,95 @@ where
...
-            self.local_state = new;
 
-            // Store the current state
-            let new_leaves = certificate
-                .bridge_exits
-                .iter()
-                .map(|exit| exit.hash())
-                .collect::<Vec<Digest>>();
+        task.await
+            .map_err(|e| Error::InternalError(format!("Certificate task panicked: {e}")))?;
```
iljakuklic, line 381: I don't think the error will be ever propagated here. The loop above early exits the function if the message channel returns `None`. In general, this await is not performed if anything in the loop exits early. Maybe the message processing loop should be in a separate function so this is shielded form early exits.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: It would be possible right now, but one of my goals is to make it possible to have multiple inflight `CertificateTask`s, which is also the reason why I'm passing `certificate_id` in all of the `NetworkTaskMessage`s even though the `NetworkTask` actually already knows it and doesn't actually use it.

If we switch to a world where we have like 5 inflight certificates (one pending settlement confirmation, 3 undergoing proving in parallel and 1 doing native execution), then the `select!` to handle all these task handles would start to get quite messy. OTOH, as the `NetworkTaskMessage` channel is MPSC, in that future we'll be able to just have one `Sender` per `CertificateTask`, and by just reading the channel the network task should be able to handle all the orchestration 😄

crates/agglayer-certificate-orchestrator/src/network_task.rs
Ekleog-Polygon: Good catch! My ongoing work on the next PR already touches the message processing loop quite a bit so… in order to avoid too messy merge conflicts, would you be fine with landing as-is despite the returned error message not being the best, and I'll make that refactoring in the next PR, or even better a follow-up PR because the next one is already way too big? (and same for your other refactoring suggestion)

iljakuklic (approved):

atanmarko (approved):

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,282 @@
...
+impl<StateStore, CertifierClient, SettlementClient>
+    CertificateTask<StateStore, CertifierClient, SettlementClient>
+where
+    StateStore: StateReader + StateWriter,
+    CertifierClient: Certifier,
+    SettlementClient: EpochPacker,
+{
+    pub fn new(
+        certificate: Certificate,
+        header: CertificateHeader,
```
Freyskeyd, line 43:
```suggestion
        certificate: Certificate,
        header: CertificateHeader,
```

This method could encapsulate the db call to fetch the `CertificateHeader`, to avoid the fact that we could initiate a `CertificateTask` with a header that is different from the certificate passed.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -0,0 +1,282 @@
...
+        if self.header.status < CertificateStatus::Proven {
+            return Err(CertificateStatusError::InternalError(
+                "Trying to settle a non-proven certificate".into(),
+            ));
+        }
+        let settled_certificate = if self.header.status < CertificateStatus::Candidate {
+            debug!("Starting certificate settlement");
+            let result = self
+                .settlement_client
+                .settle_certificate(certificate_id)
```
Freyskeyd, line 238: Shouldn't we verify the proof before this?
We're not sure that the inserted proof is still valid in regard to the current program.

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -255,459 +287,95 @@ where
...
-                Ok(())
-            }
+                header,
+                sender,
+                self.state_store.clone(),
+                self.certifier_client.clone(),
+                self.settlement_client.clone(),
+            )
+            .process(),
+        );
```
Freyskeyd, line 311: Should we instrument this spawn?

crates/agglayer-certificate-orchestrator/src/network_task.rs
Freyskeyd: Maybe you can just move the `let Some(new)...` at the top of the block?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Good point! It'll be a bit bigger, so in order to not cause too many conflicts with [feat: refactor settlement logic and use strong types for agglayer-types](https://github.com/agglayer/agglayer/pull/835/files/80f39313458058dc88bcf7d708cf279744980874#top)#835, I'll write that down as a follow-up I'll do after these two, if that's fine with you 😄

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Hmmmmm so that's actually a great question: what do we want to do when the current program changes? Currently, we're re-generating the proof without thinking about it any more about it. However, AFAIR the L1 can handle multiple program versions based on the selector, to deal with migrations properly.

So even if the current program no longer accepts the proof, the old program (that L1 still has) should still accept it. And considering it entered rocksdb's ready-for-settlement state, it means that some previous version of agglayer-node has already validated the proof with the old version.

So I'm thinking that this behavior is actually correct. This being said, I may be completely misremembering L1's behavior on old proofs! WDYT?

crates/agglayer-certificate-orchestrator/src/network_task.rs
Ekleog-Polygon: It is actually already instrumented as part of the `fn process`'s `#[instrument]`

crates/agglayer-certificate-orchestrator/src/network_task.rs
Ekleog-Polygon: Done! I don't think it'll change anything (because PersistenceError will kill the network task anyway, which is probably better because we haven't tested edge cases at all), but it's true that it'll help us fail a bit earlier 😄

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: (And for context: re-generating a proof might make a previously-valid cert now-invalid, if we changed PP in a non-backward-compatible way — so I'm thinking the new behavior, of letting L1 validate the certificate is indeed old-PP-valid, is actually better than the previous one)

Freyskeyd: > However, AFAIR the L1 can handle multiple program versions based on the selector, to deal with migrations properly.

As of the time of this message, it's not really active, as it requires the deployment of a new version of the contract and a new `rollupTypeId` deployment for all the chains.

***After this line, I will assume that we've the two mandatory elements in place on L1 (contract upgrade + all chain migrated to new `rollupTypeId`)***

-----------------

> So even if the current program no longer accepts the proof, the old program (that L1 still has) should still accept it.

It depends if the associated `PPVkeyRoute` on L1 is frozen or not. If it is `frozen` we may need to remove the proof as it'll never be able to settle.

> And considering it entered rocksdb's ready-for-settlement state, it means that some previous version of agglayer-node has already validated the proof with the old version.

True, but we're currently not tracking the `PPVkeySelector` with the proof, making any "old" proof using the latest `PPVkeySelector`. If we want to support that, we should have another PR that introduces this changes to add those bytes to the proofs.

> So I'm thinking that this behavior is actually correct. This being said, I may be completely misremembering L1's behavior on old proofs! WDYT?

It is correct if everything is in place yes, until we've what I mention above, it is not.

crates/agglayer-certificate-orchestrator/src/network_task.rs
Freyskeyd: I don't think the instrumentation is automatically propagated to tasks, as per [this](https://github.com/tokio-rs/tracing/blob/master/examples/examples/tokio-spawny-thing.rs#L25) and [this](https://github.com/tokio-rs/tokio/discussions/6008#discussioncomment-7005633). It may have changed on a recent version, but I don't think so.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Do I understand correctly that the two contracts points are already required if we ever want to upgrade the PP anyway?

Also, I'm sad to hear that we don't have the ppvkeyselector in storage yet 🥲 I'll look to adjust this PR accordingly tomorrow, thank you for the explanation! (And I'll add the "adding selector to storage" on my stash of things to change for the upcoming storage refactorings)

crates/agglayer-certificate-orchestrator/src/network_task.rs
Ekleog-Polygon: Right, what I mean is that [this instrument](https://github.com/agglayer/agglayer/pull/819/files/c848d747771c6c26bd9c1613f8353b684edeb1dc#diff-6670f8d5c0a95267f210b07afd4dd274126416037a4f01cb61920ef0fdad5da9R52-R60) already contains AFAICT all the data we could want, and it's already running inside the certificate task, so it doesn't need propagating through tasks 😄

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Freyskeyd: > Do I understand correctly that the two contracts points are already required if we ever want to upgrade the PP anyway?

Yes and no, it's up to the contract team to decide that, in regard to the different audit etc.

> Also, I'm sad to hear that we don't have the ppvkeyselector in storage yet 🥲 I'll look to adjust this PR accordingly tomorrow, thank you for the explanation! (And I'll add the "adding selector to storage" on my stash of things to change for the upcoming storage refactorings)

Could you create a ticket in our backlog so that we all keep track of this?
I think it can be part of a smaller PR to tackle adding the PPSelector.

crates/agglayer-certificate-orchestrator/src/network_task.rs
Freyskeyd: Okey, we'll review this when we work on the traces tho :)

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Done for the storage issue: [Add PPVKeySelector to storage](https://github.com/agglayer/agglayer/issues/856#top)#856.

This being said, I started updating the PR, but doing so got me thinking about it some more. And I think this is actually a problem we cannot deal with the way our code is currently designed, be it before or after this PR:
- We need to be able to deal with Candidate certificates, for which settlement has already been submitted to L1 with a previous PP version
- For these, we will need to compute the next state for the network, for the next certificate height. However, we only have the new PP version now. So we cannot actually compute the same state as the one that L1 will now assume the network to have

So I think that, as things currently stand, we cannot update PP in any non-backward-compatible way anyway in agglayer-node. For that, we would need at least [[Design] Get agglayer-node's network state storage ready for preconfs](https://github.com/agglayer/protocol-research/issues/149#top)#149, so that we can store the new storage state at the same time as we compute the proof. And considering this storage refactoring should be significantly helped by this PR, I think we should land this basically as-is and focus on the storage refactorings much more — they'll naturally get this issue fixed.

TL;DR: I agree that the code I'm writing is broken, but the previous code was basically just as broken (except for the very short window of time where a certificate was Proven but not Candidate yet — so the ~30s of time a power failure could cause problem would only become a ~30.1s on average). This crash also needs to be immediately followed by a PP upgrade before the reboot, for it to cause any problem.

So I'd say probably we can keep this PR as-is for now, and work faster on [[Design] Get agglayer-node's network state storage ready for preconfs](https://github.com/agglayer/protocol-research/issues/149#top)#149 so we can solve the issue for all of the 30.1s the crash would be problematic.

WDYT?

Ekleog-Polygon: After discussion: in order to avoid any issue with the additional second, let's unconditionally re-prove the certificate on startup for now, like we were doing before this PR.

Ekleog-Polygon: Done 🎉 This should be ready to land 😄

Ekleog-Polygon:
Now I'm sad, looks like it'll merge conflict again, I'll need to merge again, and ping for reviews again 🥲 

Well, if you can have a look at the latest changes already, at least the merge conflict resolution should be easy to review and we can land this one soon after #826 lands!

atanmarko (approved):

crates/agglayer-certificate-orchestrator/src/network_task.rs
Ekleog-Polygon: Note to self: I'll have to come back to this after you're done with your current work

PR #823, opened 2025-05-27
feat: typed tree root types

atanmarko:
# Description

Use exit root types.

Fixes https://github.com/agglayer/security/issues/66

## PR Checklist:


crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -96,8 +96,8 @@ impl From<CertificateV0> for Certificate {
         Certificate {
             network_id: network_id.into(),
             height,
-            prev_local_exit_root,
-            new_local_exit_root,
+            prev_local_exit_root: prev_local_exit_root.into(),
+            new_local_exit_root: new_local_exit_root.into(),
```
Ekleog-Polygon: Can we adjust CertificateV0 to also use the new wrapper type? AFAICT it should be perfectly compatible with the old one so we should be fine adjusting CertificateV0

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -143,8 +143,8 @@ impl From<CertificateV1<'_>> for Certificate {
         Certificate {
             network_id,
             height,
-            prev_local_exit_root,
-            new_local_exit_root,
+            prev_local_exit_root: prev_local_exit_root.into(),
+            new_local_exit_root: new_local_exit_root.into(),
```
Ekleog-Polygon: Same for CertificateV1

crates/agglayer-types/src/lib.rs
```diff
@@ -303,15 +309,17 @@ impl Default for Certificate {
     fn default() -> Self {
         let network_id = NetworkId::ETH_L1;
         let wallet = Self::wallet_for_test(network_id);
-        let exit_root = LocalExitTree::<Keccak256Hasher>::default().get_root();
+        let local_exit_root = LocalExitTree::<Keccak256Hasher>::default()
+            .get_root()
+            .into();
```
Ekleog-Polygon, line 314: Hmmm… Why does `LocalExitTree::get_root()` not directly return a `LocalExitRoot`-typed `Digest`?

crates/agglayer-types/src/lib.rs
```diff
@@ -370,14 +378,16 @@ impl Certificate {
     #[cfg(any(test, feature = "testutils"))]
     pub fn new_for_test(network_id: NetworkId, height: Height) -> Self {
         let wallet = Self::wallet_for_test(network_id);
-        let exit_root = LocalExitTree::<Keccak256Hasher>::default().get_root();
-        let (_, signature, _signer) = compute_signature_info(exit_root, &[], &wallet, height);
+        let local_exit_root = LocalExitTree::<Keccak256Hasher>::default()
+            .get_root()
+            .into();
```
Ekleog-Polygon, line 383: Same here

crates/agglayer-types/src/lib.rs
```diff
@@ -558,16 +568,16 @@ impl LocalNetworkStateData {
         let prev_pessimistic_root = match prev_pp_root {
             PessimisticRootInput::Fetched(settled_from_l1) => settled_from_l1,
             PessimisticRootInput::Computed(version) => PessimisticRoot {
-                balance_root: self.balance_tree.root,
-                nullifier_root: self.nullifier_tree.root,
+                balance_root: self.balance_tree.root.into(),
+                nullifier_root: self.nullifier_tree.root.into(),
```
Ekleog-Polygon, line 572: As the *root types are in interop-types, can we make use of them from the PP program too and avoid this conversion?

crates/agglayer-types/src/lib.rs
```diff
@@ -558,16 +568,16 @@ impl LocalNetworkStateData {
...
                 ler_leaf_count: self.exit_tree.leaf_count(),
                 height: certificate.height,
                 origin_network: certificate.network_id,
             }
             .compute_pp_root(version),
         };
 
-        let prev_local_exit_root = self.exit_tree.get_root();
+        let prev_local_exit_root = self.exit_tree.get_root().into();
```
Ekleog-Polygon, line 580: Same here, this shouldn't require an `.into` as it should already be the right type

crates/agglayer-types/src/lib.rs
```diff
@@ -680,11 +690,11 @@ impl LocalNetworkStateData {
                 .collect::<Result<Vec<_>, Error>>()?;
 
         // Check that the certificate referred to the right target
-        let computed = self.exit_tree.get_root();
+        let computed: LocalExitRoot = self.exit_tree.get_root().into();
```
Ekleog-Polygon: Same here.

Also, that's a matter of taste, but when the conversion is really required I usually find this version easier to read. WDYT about the difference? (though I'll always be sad that I cannot `.into::<LocalExitRoot>()`, which'd be the cleanest version IMO)
```suggestion
        let computed = LocalExitRoot::from(self.exit_tree.get_root());
```

crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -31,10 +32,10 @@ impl NetworkState {
     /// Returns the roots.
     pub fn get_state_commitment(&self) -> StateCommitment {
         StateCommitment {
-            exit_root: self.exit_tree.get_root(),
+            exit_root: self.exit_tree.get_root().into(),
             ler_leaf_count: self.exit_tree.leaf_count,
-            balance_root: self.balance_tree.root,
-            nullifier_root: self.nullifier_tree.root,
+            balance_root: LocalBalanceRoot::new(self.balance_tree.root),
+            nullifier_root: LocalNullifierRoot::new(self.nullifier_tree.root),
```
Ekleog-Polygon, line 38: I'd think these three should already be of the right type, as they're respectively of types `LocalExitTree`, `LocalBalanceTree` and `NullifierTree`?

crates/pessimistic-proof/src/local_state.rs
```diff
@@ -48,10 +49,10 @@ impl StateCommitment {
 impl From<StateCommitment> for pessimistic_proof_core::local_state::commitment::StateCommitment {
     fn from(commitment: StateCommitment) -> Self {
         Self {
-            exit_root: commitment.exit_root,
+            exit_root: LocalExitRoot::new(commitment.exit_root),
             ler_leaf_count: commitment.ler_leaf_count,
-            balance_root: commitment.balance_root,
-            nullifier_root: commitment.nullifier_root,
+            balance_root: LocalBalanceRoot::new(commitment.balance_root),
+            nullifier_root: LocalNullifierRoot::new(commitment.nullifier_root),
```
Ekleog-Polygon, line 55: These three values inside `commitment` should also already be of the right types, I think

crates/pessimistic-proof/src/proof.rs
```diff
@@ -82,7 +83,7 @@ mod tests {
     fn empty_tree_roots() {
         let empty_state = LocalNetworkState::default();
 
-        let ler = empty_state.exit_tree.get_root();
+        let ler = LocalExitRoot::new(empty_state.exit_tree.get_root());
```
Ekleog-Polygon, line 86: And this is the last place where I think the type should already be the right one without any conversion

crates/agglayer-storage/src/types/certificate/mod.rs
atanmarko: Hey @Ekleog-Polygon , purpose of this PR is to resolve https://github.com/agglayer/security/issues/66, so changing Certificate (which has much more impact on the project) is out of scope of this PR. Feel free to open a separate ticket for that.

crates/agglayer-storage/src/types/certificate/mod.rs
atanmarko: Repeating above answer

crates/agglayer-types/src/lib.rs
atanmarko: Yes, I did that implementation first, but as I remember, there you get into an issue with the constraints of the `LocalExitTree` hasher type (as `H::Digest` is used). As resolving that would have more repercussions, I decided to leave it as it is.

crates/agglayer-types/src/lib.rs
atanmarko: Same answer as above

crates/agglayer-types/src/lib.rs
atanmarko: We can, but it is out of scope of this PR, because it affects the `Smt` `Hasher` type constraints, which would have more impact on the other code. Please open a related ticket for that change.

crates/agglayer-types/src/lib.rs
atanmarko: Same answer as above.

crates/agglayer-types/src/lib.rs
atanmarko: Right, better to use `from` here.

crates/pessimistic-proof/src/local_state.rs
atanmarko: @Ekleog-Polygon I agree they should be the same type, but it is out of scope of this PR, as it puts different requirements on the Tree hasher type. So either we would need to make  `LocalBalanceRoot` and `LocalNullifierRoot` generic over the hasher type, or to limit the `NullifierTree` and `LocalBalanceTree` to the `Digest` of the root types. I am not sure if that action pays out. Please open a related ticket for that if you think we should go in that direction.

crates/pessimistic-proof/src/proof.rs
atanmarko: Well, same answer as above.

crates/agglayer-storage/src/types/certificate/mod.rs
iljakuklic: > Hey @Ekleog-Polygon , purpose of this PR is to resolve [agglayer/security#66](https://github.com/agglayer/security/issues/66), so changing Certificate (which has much more impact on the project) is out of scope of this PR. Feel free to open a separate ticket for that.

I don't quite understand. The LER type in `agglayer_types::Certificate` is already changed. As for `CertifivateV{0,1}`, these types are local to the current module. It should be pretty low impact to change the LER types here.

atanmarko: @iljakuklic Looking again at the issue, and you are totally right. Fixing this in next commit.

crates/agglayer-storage/src/columns/epochs/transaction_hash_per_certificate_index/mod.rs
```diff
@@ -9,6 +9,7 @@ use crate::columns::PER_EPOCH_TRANSACTION_HASH_PER_CERTIFICATE_INDEX;
 /// | key                | value    |
 /// | --                 | --       |
 /// | `CertificateIndex` | `Hash`   |
+#[allow(unused)]
 pub struct TransactionHashPerCertificateIndexColumn;
```
iljakuklic: Hmm is this db column actually used? I don't see any `db.(get|put|keys)<TransactionHashPerCertificateIndexColumn>`. For transaction hash where a certificate was settled, there is a field in `CertificateHeader` IIRC.

iljakuklic (approved):

crates/agglayer-storage/src/columns/epochs/transaction_hash_per_certificate_index/mod.rs
atanmarko: @iljakuklic Clippy says it is not used, but I didn't want to interfere in the db business in this PR :)

iljakuklic: Filed a PR to clean this up:
* #845

iljakuklic (approved):

Issue #821, opened 2025-05-26
Cross check aggchain proof public values with L1


hadjiszs:
# Problem

Often the `InvalidProof` errors observed upon L1 settlement of the PP are usually due to mismatch on public values which are painful and repetitive to debug manually:

- Check the public values expected by the aggchain proof from the aggkit prover logs
- Check the public values expected by the L1 manually with `cast` commands against functions/variables which are not necessarily well documented

# Solution

- Submit with the Certificate the unrolled[^1] aggchain proof public values expected by the aggchain proof
- Having the agglayer to fetch the public values expected by the L1 (e.g., range/aggregation vkeys, prover/trusted sequencer addresses, aggregation proof public values, output roots, etc)
- Trigger dedicated error for each mismatch

[^1]: "unrolled" because some of the public values are aggregation of several underlying values, such as the `aggchain_params` which is the hash of multiple values.

hadjiszs:
Done through #826 #829 

Stopping on the aggchain proof public values level, not going down further with the FEP aggregation proof public values given that the agglayer needs to remain agnostic from the specific intricacies of the aggchain params

PR #826, opened 2025-05-29
feat: cross check the aggchain hash with the L1

hadjiszs:
# Description

The `aggchain_hash` depends on a lot of parameters, which makes it easy for it to end up different from what the L1 expects.

So far, this kind of mismatch would only show up as an `InvalidProof` error during L1 settlement. With this change, it’ll be caught earlier on the agglayer as an `AggchainHashMismatch` error, before we generate a PP.

First shot for https://github.com/agglayer/agglayer/issues/821

Note: not unrolling all the underlying parameters of the `aggchain_hash` because several ones are aggchain type dependent and still unsure whether we want to introduce checks which depend on the aggchain type on the agglayer side.

## PR Checklist:


crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -392,6 +388,13 @@ where
             }
         };
 
+        // Fetching rollup contract address
+        let rollup_address = self
+            .l1_rpc
+            .get_rollup_contract_address(network_id.to_u32())
+            .await
+            .map_err(|source| CertificationError::RollupContractAddressNotFound { source })?;
```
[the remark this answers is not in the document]
hadjiszs: That's moved code but I changed this error accordingly then, I see many other occurrences of those but didn't touch them as not really in the scope of this PR

crates/agglayer-contracts/src/contracts/aggchain_base.rs
```diff
@@ -21,6 +21,28 @@ pub mod aggchain_base {
                         ::ethers::core::abi::ethabi::Function {
                             name: ::std::borrow::ToOwned::to_owned("AGGCHAIN_TYPE"),
                             inputs: ::std::vec![],
+                            outputs: ::std::vec![
```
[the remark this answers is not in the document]
hadjiszs: I don't know if there is a task for that, and not sure what's the pros of switching to the macro compared to what is done in `agglayer-contracts`, happy if you can populate one ticket if you have opinions on that 👍

Freyskeyd (approved):

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -188,11 +188,7 @@ where
             });
         }
 
-        info!(
-            "Successfully executed the PP in SP1 for the Certificate {}: {} sp1-cycles ",
-            report.cycle_tracker.values().sum::<u64>(),
-            certificate_id
-        );
+        info!("Successfully executed the PP in SP1 for the Certificate");
```
Ekleog-Polygon: Nit: can we rename the `hash` trace field of this function into `certificate_id`? I was about to comment that we're losing valuable info until I noticed it 😅

Ekleog-Polygon (approved):
LGTM! My only nit is not actually on code in this PR, feel free to ignore and we can address it next time we see it arise 😄

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -198,12 +200,14 @@ pub enum AggchainDataV1<'a> {
...
         proof: Cow<'a, Proof>,
         aggchain_params: Digest,
+        public_values: Option<Box<AggchainProofPublicValues>>,
     },
 
     GenericWithSignature {
         proof: Cow<'a, Proof>,
         aggchain_params: Digest,
         signature: Cow<'a, Box<Signature>>,
+        public_values: Option<Box<AggchainProofPublicValues>>,
```
iljakuklic: Unfortunately, adding fields breaks storage binary backwards compatibility. You have to introduce a new enum arm, like `GenericWithPublicValues`. Two arms should not be needed since the new one can have the signature field as an `Option`.

Regarding the arm names. Since this enum is only used for versioning storage binary encoding, I would not mind renaming the arms to just something like `GenericV0`, `GenericV1` and `GenericV2`. I'm afraid that at this rate, we will run out of descriptive names quite soon anyway :laughing:.

iljakuklic: Filed a PR with tests that would have detected this issue:
* #834

hadjiszs: thanks for the explanations 👍 normally good now

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s01_be002.snap
```diff
@@ -3,45 +3,41 @@ source: crates/pessimistic-proof-test-suite/tests/cycle-tracker.rs
...
-     521 secp256k1_add
-       8 commit
-       8 commit_deferred_proofs
-       8 hint_len
-       8 hint_read
-       4 enter_unconstrained
-       1 halt
-       1 write
+gas: 25718188
+opcode counts (23846619 total instructions):
```
iljakuklic: The cycles increased significantly but that might be solved by merging in up-to-date changes from main.

iljakuklic (approved):

.github/workflows/test.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          repository: agglayer/agg-contracts-internal
+          repository: agglayer/agglayer-contracts
```
Freyskeyd, line 60: Maybe we should also update the ref for building the image?

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -461,10 +477,54 @@ where
...
+        } = &certificate.aggchain_data
+        {
+            // Consistency check across these 2 sources:
+            //
+            // - Public values expected by the proof (i.e., the valid ones to succeed the
+            //   proof verification, provided as metadata in the Certificate as-is)
+            //
+            // - Public values expected by the PP (i.e., the ones used to verify the
+            //   aggchain proof in the PP)
+            debug!(%certificate_id, "Aggchain proof public values expected by the received aggchain proof: {pv_from_proof:?}");
```
Freyskeyd:
```suggestion
            debug!("Public values expected by the certificate's aggchain-proof: {pv_from_proof:?}");
```

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -187,11 +190,7 @@ where
             });
         }
 
-        info!(
-            "Successfully executed the PP in SP1 for the Certificate {}: {} sp1-cycles ",
-            report.cycle_tracker.values().sum::<u64>(),
-            certificate_id
-        );
+        info!("Successfully executed the PP in SP1 for the Certificate");
```
Freyskeyd:
```suggestion
        info!("Successfully executed the PP program locally");
```

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -461,10 +477,54 @@ where
...
+                        .iter()
+                        .map(|(exit, _)| exit.to_indexed_exit_hash())
+                        .collect(),
+                }
+                .commitment(CommitmentVersion::V3),
+                aggchain_params: *aggchain_params,
+            };
+
+            if **pv_from_proof != pv_from_pp_witness {
+                error!(%certificate_id, "Mismatch on the aggchain proof public values.");
```
Freyskeyd:
```suggestion
                error!("Mismatch on the aggchain-proof public values.");
```

No strong opinion on `aggchain-proof` vs `aggchain proof`

.github/workflows/test.yml
hadjiszs: Updated it to the latest `v11.0.0-rc.1` (It seems to include sp1 v5 verifier normally, right @ignasirv ?)

Issue #661, opened 2025-02-26
Use strong types to semantically differentiate plain data


iljakuklic:
To enhance the safety and robustness, various plain data types should be replaced with strongly typed wrappers, to semantically differentiate between them. This applies mostly to numbers and hashes.

The value we get from tooling will be improved too:

* Compiler checking for potential mismatches.
* Self-documenting function signatures. (Rustdoc is capable of looking up functions from signature, precise types help there.)
* I also noticed that lately rust language server does some light type directed search to rank completions.

An incomplete list (feel free to add):

* Booleans (`bool`)
  * a custom enum is often more expressive and readable, e.g. `ForceUpdate::{Yes, No}`.
* Numbers (`u32`, `u64` for most part)
  * mostly various IDs and indices and counters
  * `NetworkId`, `RollupId`
  * `RollupIndex`, `ExitIndex`
  * `EpochNumber`, `CertificateHeight`, `L1BlockHeight`, ...
* Hashes (`H256`, `[u8; 32]`, `agglayer_types::Digest`)
  * `${WHATEVER}ExitRoot`, `L1InfoRoot`
  * `CertificateId`

Related:
* Mentioned in https://github.com/agglayer/agglayer/pull/644#discussion_r1971744952

Ekleog-Polygon:
I need a new emoji with 1000 thumbs up for this comment 👍 👍 👍

iljakuklic:
Another type which could benefit from this is the proofs. We have `PessimisticProof`, `AggchainProof`, `BridgeProof`, `ExecutionProof`, etc. There are also corresponding proving keys and verification keys that could be tagged with which kind of proof they correspond to.

atanmarko:
Good idea.

PR #835, opened 2025-06-04
feat: refactor settlement logic and use strong types for agglayer-types

Ekleog-Polygon:
Fixes https://github.com/agglayer/protocol-research/issues/148
Fixes https://github.com/agglayer/agglayer/issues/661

Builds on top of https://github.com/agglayer/agglayer/pull/819 ; left in draft until it lands

## PR Checklist:


crates/agglayer-types/src/lib.rs
iljakuklic, line 1: Better as a follow up task but this file really deserves to be broken down into smaller modules in separate files.

crates/agglayer-types/src/lib.rs
```diff
@@ -23,18 +30,133 @@ use pessimistic_proof::{
...
+    Ord,
+    PartialEq,
+    PartialOrd,
+    derive_more::Display,
+    serde::Deserialize,
+    serde::Serialize,
+)]
+#[cfg_attr(feature = "testutils", derive(arbitrary::Arbitrary))]
+#[serde(transparent)]
+pub struct EpochNumber(pub u64);
```
iljakuklic: For types that are in effect "counters", I like to introduce a method like `.next()` or `.increment()` for incrementing it by one. Applies to `EpochNumber`, `Height`, `CertificateIndex`.

crates/agglayer-types/src/lib.rs
```diff
@@ -23,18 +30,133 @@ use pessimistic_proof::{
...
+        Height(self.0 - rhs)
+    }
+}
+
+impl Sub<&Height> for Height {
+    type Output = u64;
+
+    fn sub(self, rhs: &Height) -> Self::Output {
+        self.0 - rhs.0
+    }
```
iljakuklic: While we're introducing stronger types: The subtraction here might underflow. Do we want `Output = Result<u64, NonZeroU64>` to represent both positive and negative result without a panic or a silent wrap-around?

Similar for `Add`. Probably not possible with `AddAssign`. One advantage of introducing `.next()` or `.increment()` as suggested above is that it is quite hard to reach overflow using it so can be made infallible.

crates/agglayer-types/src/lib.rs
```diff
@@ -23,18 +30,133 @@ use pessimistic_proof::{
...
+impl Sub<u64> for Height {
+    type Output = Height;
+
+    fn sub(self, rhs: u64) -> Self::Output {
+        Height(self.0 - rhs)
+    }
+}
+
+impl Sub<&Height> for Height {
+    type Output = u64;
```
iljakuklic: If you want to go super-strong types, this could return something like `HeightDistance` :slightly_smiling_face:

crates/agglayer-types/src/lib.rs
```diff
@@ -259,6 +383,32 @@ impl std::fmt::Display for CertificateStatus {
...
 
+impl PartialOrd for CertificateStatus {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl CertificateStatus {
+    // Only ever used as implementation for Ord, feel free to change it
+    fn as_order_number(&self) -> usize {
```
iljakuklic: Could this function be moved inside the body of `CertificateStatus::cmp`, so it's even more private?

crates/agglayer-types/src/lib.rs
```diff
@@ -259,6 +383,32 @@ impl std::fmt::Display for CertificateStatus {
...
+            InError { .. } => 4,
+        }
+    }
+}
+
+impl Ord for CertificateStatus {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.as_order_number().cmp(&other.as_order_number())
+    }
+}
```
iljakuklic: Given that all `InError` values all compare as equal to each other, it probably breaks `BTreeMap`s, `binary_search` and other stuff that relies on `Ord` that does not "collapse" values like this. To be on the safe side, it might be a good idea to turn this into a method.
```suggestion
impl CertificateStatus {
    fn comes_strictly_before(&self, other: &Self) -> bool {
        self.as_order_number() < other.as_order_number()
    }
}
```

crates/agglayer-types/src/lib.rs
```diff
@@ -399,15 +549,15 @@ impl Certificate {
...
             self.network_id.to_be_bytes().as_slice(),
-            self.height.to_be_bytes().as_slice(),
+            self.height.0.to_be_bytes().as_slice(),
             self.prev_local_exit_root.as_slice(),
             self.new_local_exit_root.as_slice(),
             commit_bridge_exits.as_slice(),
             commit_imported_bridge_exits.as_slice(),
-            self.metadata.as_slice(),
-        ])
+            self.metadata.0.as_slice(),
```
iljakuklic: Would be nice to expose `to_be_bytes` and `as_slice` on the wrapper types directly to avoid accessing the internals. I'm sure there are more cases like this.

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Added as a // TODO so I don't forget it when I do my round of all the added TODOs after this PR 😄

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: TBH I think that might be overkill — it's currently used at exactly one place to check that we're not putting too many certs in an epoch 😅

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: This is actually coming from https://github.com/agglayer/agglayer/pull/819 ; but you're right that I should be able to refactor that out and just delete the `Ord` implementation as I initially promised. I'll add that as a // TODO in order to not make this PR even bigger than it already is, hopefully #819 can land soon and we stop seeing the diff for both PRs here 😅

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Makes sense, though I'm planning on removing it altogether in (one of) the next PRs; would that solution be fine with you?

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Definitely — actually thinking about it some more I should probably just expose `Deref` and make the inner value private 😄

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: I did replace `Sub` with a `distance_since` method, hopefully it should be enough 😄

crates/agglayer-grpc-types/src/compat/v1/certificate_header.rs
```diff
@@ -24,16 +25,19 @@ impl From<CertificateHeader> for v1::CertificateHeader {
...
             new_local_exit_root: Some(value.new_local_exit_root.into()),
-            metadata: Some(value.metadata.into()),
+            metadata: Some((*value.metadata).into()),
             status: status.into(),
             error,
-            settlement_tx_hash: value.settlement_tx_hash.map(Into::into),
+            settlement_tx_hash: value
+                .settlement_tx_hash
+                .map(Digest::from)
+                .map(FixedBytes32::from),
```
iljakuklic: Nit: maps could be fused:
```suggestion
                .map(|h| FixedBytes32::from(Digest::from(h))),
```

crates/agglayer-jsonrpc-api/src/tests/get_latest_known_certificate_header.rs
```diff
@@ -20,9 +22,9 @@ async fn returns_the_pending_certificate_header() {
 
     let network_id = 1.into();
 
-    let settled_certificate = Certificate::new_for_test(network_id, 0);
-    let proven_certificate = Certificate::new_for_test(network_id, 1);
-    let pending_certificate = Certificate::new_for_test(network_id, 2);
+    let settled_certificate = Certificate::new_for_test(network_id, Height::ZERO);
+    let proven_certificate = Certificate::new_for_test(network_id, Height::ONE);
+    let pending_certificate = Certificate::new_for_test(network_id, Height::new(2));
```
iljakuklic: Seems a bit arbitrary to have a constant for `ONE` but not for `2`. Is there a reason for that? I understand `ZERO`, that's kind of standard.

crates/agglayer-storage/src/columns/certificate_header/tests.rs
```diff
@@ -1,11 +1,11 @@
-use agglayer_types::CertificateId;
+use agglayer_types::{CertificateId, CertificateIndex, EpochNumber, Height, Metadata};
 
 use super::{Key, Value};
 use crate::columns::Codec as _;
 
 #[test]
 fn can_parse_key() {
-    let key: CertificateId = [1; 32].into();
+    let key = CertificateId::new([1; 32].into());
```
iljakuklic, line 8: Is `CertificateId` constructible directly from `[u8; 32]`? It probably should be, there is a number of occurrences where it is constructed from `[u8; 32]` via some other type (`Digest`?). Same for `Metadata` and possibly some other types.

crates/agglayer-types/src/lib.rs
```diff
@@ -23,18 +31,223 @@ use pessimistic_proof::{
...
+}
+
+impl From<CertificateId> for Digest {
+    fn from(cert_id: CertificateId) -> Self {
+        cert_id.0
+    }
+}
+
+impl Deref for CertificateId {
+    type Target = Digest;
```
iljakuklic: I am not quite convinced by adding `Deref`s. It introduces implicit conversions, largely reducing the benefit of introducing strong types to begin with.

crates/agglayer-types/src/lib.rs
```diff
@@ -23,18 +31,223 @@ use pessimistic_proof::{
...
+    }
+
+    pub fn increment(&mut self) {
+        *self = self.next();
+    }
+
+    pub fn distance_since(&self, o: &Height) -> u64 {
+        self.0
+            .checked_sub(o.0)
+            .expect("Subtracting to negative values")
```
iljakuklic: Hm, do we want this operation to be panicking if the operands are the wrong way around?

crates/agglayer-types/src/lib.rs
iljakuklic, line 1: Nit: quite a few inherent methods in this file could be `const fn`s.

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -132,7 +133,7 @@ where
         clock: ClockRef,
         data_receiver: Receiver<(NetworkId, Height, CertificateId)>,
         cancellation_token: CancellationToken,
-        epoch_packing_task_builder: E,
+        epoch_packing_task_builder: Sc,
```
iljakuklic: Would this be a more appropriate name, given the generic bounds?
```suggestion
        settlement_client: Sc,
```

iljakuklic (commented):
Another pass over the strong type changes

crates/agglayer-jsonrpc-api/src/tests/get_latest_known_certificate_header.rs
Ekleog-Polygon: For example in `crates/agglayer-storage/src/stores/per_epoch/tests.rs`, we `match` on `Height(1)`. As we can't use a `const fn` to match, I pretty much arbitrarily decided that `ONE` would be fine to add, in order to not have the code in this test become much more complex due to the need to unwrap 4 layers of nested structs.

Does that make sense to you?

crates/agglayer-storage/src/columns/certificate_header/tests.rs
Ekleog-Polygon: Agreed, but IMO we should share that across all our strong hash types, including the ones in interop, so I didn't want to spend too much effort here considering strong types here were mostly done en-passant.

Does that make sense to you, or do you think we should have good strong types from the beginning, without requiring the intermediary cast to `Digest` for creation?

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: IMO the point of strong types here is not to mix certificate ids with eg. exit roots. Deref allows us to still be able to call all Digest functions, like `as_slice` & co, which can be pretty useful on basically any strong hash type.

For full disclosure, my first iteration here had the Digest as a public variable, and then I thought it'd be better to have it private but still expose Deref (without DerefMut for now because we should probably never need it).

Is there a specific implicit conversion you'd be particularly concerned about?

(On my end I'd think we may want to keep things as-is for now, until we make some strong-hash-type macro in interop that we could then use to automatically implement all the important Digest functions)

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Here I basically reused the previous debug builds' behavior — we had been using a bare `-`, which would panic in debug mode.

To be fully honest I'm thinking we may want to drop this function altogether and deal with the rate limiting check another way, but with the size of this PR I didn't want it to get even bigger 😅 

Would you feel better with a simple `-` instead of `checked_sub`?

crates/agglayer-jsonrpc-api/src/tests/get_latest_known_certificate_header.rs
iljakuklic: Well I don't think using a comparison would be *that* much worse. Seems like introducing a constant with no special meaning. Maybe gate it under `testutils` as a compromise?

crates/agglayer-types/src/lib.rs
iljakuklic: Ok, let's have it as in this PR for now but log it as an issue somewhere. This is already a big improvement.

crates/agglayer-types/src/lib.rs
iljakuklic: > Here I basically reused the previous debug builds' behavior — we had been using a bare `-`, which would panic in debug mode.

That's a fair point. I see the general issue of wrapping arithmetic is already tracked, so we're good.

> Would you feel better with a simple `-` instead of `checked_sub`?

I don't feel particularly strongly about that.

crates/agglayer-jsonrpc-api/src/tests/get_latest_known_certificate_header.rs
Ekleog-Polygon: Good point! I always forget about the if clause to match arms, not sure why 😅 Will do that, thanks!

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Opened in [Tech debt: introduce a macro for creating strong hash types](https://github.com/agglayer/interop/issues/39#top)#39 😄

crates/agglayer-types/src/lib.rs
iljakuklic, line 1: Since you already pulled in `derive_more` as a dependency, quite a few of the `From`, `AsRef`, `Deref`, `Display` impls could be `#[derive]`d. `FromStr` could also be thrown in for some of them for good measure.

iljakuklic (commented):
Workflow question: How does this relate to #819? Seems a number of changes is duplicate among the two PRs. Could #819 be made target of this branch?

Ekleog-Polygon:
This PR actually builds on top of #819 (as per the PR description 😝). So yes, all changes from #819 are also duplicated here, it's also the reason why I've left it as draft until #819 lands and we can easily review the diff between this and main-then. This being said, thank you very much for your reviews, that are helping me start to improve on this PR even before #819 lands 😄 

(As for making #819 the target of this branch, last time I tried that when I merged github completely broke my PR and I had to reopen it, so I decided to just keep both targeting main, leave this one as draft, and just merge main once #819 will have merged-in; that should hopefully work fine)

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Good point for `From` and `Deref`! For `AsRef`, considering the issue to remove `Deref` I'd think it's probably better not to actually have `AsRef`? And `Display` IMO doesn't deserve being impl'd until we get a reasonable use case that'd be made more convenient by it — if we start displaying eg. Heights and CertificateIndexes the same, a human could start making the same mistake as a computer would if we had weak types. Same for `FromStr`, I'm thinking it'd be better to wait until we make the macro in interop before we think of implementing it on each independent type here.

I've just finished implementing your suggestions for `From` and `Deref` for now 😄 WDYT?

crates/agglayer-types/src/lib.rs
```diff
@@ -25,22 +32,183 @@ use pessimistic_proof::{
     LocalNetworkState, ProofError,
 };
 use serde::{Deserialize, Serialize};
-use unified_bridge::CommitmentVersion;
+use unified_bridge::{CommitmentVersion, LocalExitTree, LocalExitTreeError};
+
+// TODO: consider moving these newtype structs to interop some day, so that the
+// proofs could use eg. CertificateId and Height at least
```
Freyskeyd: I'm not sure about that, `CertificateId` and `Height` are linked to `agglayer-node` domain, not `interop`.

crates/agglayer-types/src/lib.rs
```diff
@@ -452,15 +623,15 @@ impl Certificate {
         let commit_imported_bridge_exits =
             keccak256_combine(self.imported_bridge_exits.iter().map(|exit| exit.hash()));
 
-        keccak256_combine([
+        CertificateId(keccak256_combine([
             self.network_id.to_be_bytes().as_slice(),
-            self.height.to_be_bytes().as_slice(),
+            self.height.0.to_be_bytes().as_slice(),
```
Freyskeyd: Maybe we could use the new `as_u64` ?

```suggestion
            self.height.as_u64().to_be_bytes().as_slice(),
```

crates/pessimistic-proof-test-suite/tests/e2e_local_pp.rs
```diff
@@ -78,7 +78,7 @@ fn pp_root_migration_helper(
         balance_root: new_state.balance_tree.root.into(),
         nullifier_root: new_state.nullifier_tree.root.into(),
         ler_leaf_count: new_state.exit_tree.leaf_count(),
-        height: certificate.height + 1,
+        height: certificate.height.as_u64() + 1,
```
Freyskeyd, line 81:
```suggestion
        height: certificate.height.increment(),
```

crates/agglayer-storage/src/types/mod.rs
```diff
@@ -16,13 +16,13 @@ macro_rules! default_codec_impl {
...
-    EpochSynchronization,
+    EpochSynchronization, // Actually unused, kept for storage backward compatibility
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 pub enum MetadataValue {
-    LatestSettledEpoch(u64),
-    EpochSynchronization(u64),
+    LatestSettledEpoch(EpochNumber),
+    EpochSynchronization(u64), // Actually unused, kept for storage backward compatibility
```
Freyskeyd, line 25: Maybe something to double-check with main/test/dev-net storage, but I think we can remove this.

crates/agglayer-storage/src/stores/per_epoch/tests.rs
```diff
@@ -71,38 +71,53 @@ type EndCheckpointState = CheckpointState;
...
-    EndCheckpointState::WithCheckpoint(vec![(NetworkId::new(0), 1)]),
+    StartCheckpointState::WithCheckpoint(vec![(NetworkId::new(0), Height::new(1))]),
+    EndCheckpointState::WithCheckpoint(vec![(NetworkId::new(0), Height::new(1))]),
     |result: Result<_, Error>| {
-        matches!(result, Err(Error::CertificateCandidateError(crate::error::CertificateCandidateError::UnexpectedHeight(_, 3, 1))))
-    }, 3, None, None)]
+        matches!(
+            result,
+            Err(Error::CertificateCandidateError(crate::error::CertificateCandidateError::UnexpectedHeight(_, h3, h1)))
+                if h1 == Height::new(1) && h3 == Height::new(3)
```
Freyskeyd, line 109: What's the reason behind instantiating a new `Height` and not use `as_u64()` in such case ?

```suggestion
                if h1.as_u64() == 1 && h3.as_u64() == 3
```

crates/agglayer-types/src/lib.rs
```diff
@@ -25,22 +32,183 @@ use pessimistic_proof::{
...
+    pub const fn new(epoch: u64) -> EpochNumber {
+        EpochNumber(epoch)
+    }
+
+    pub const fn next(&self) -> EpochNumber {
+        EpochNumber(self.0.checked_add(1).expect("Epoch number overflow"))
+    }
 
-pub type EpochNumber = u64;
+    pub const fn increment(&mut self) {
```
Freyskeyd, line 68: Could this `increment` also return the new value?

crates/agglayer-types/src/lib.rs
```diff
@@ -25,22 +32,183 @@ use pessimistic_proof::{
...
+
+    pub const fn new(height: u64) -> Height {
+        Height(height)
+    }
+
+    pub const fn next(&self) -> Height {
+        Height(self.0.checked_add(1).expect("Height overflow"))
+    }
+
+    pub const fn increment(&mut self) {
```
Freyskeyd, line 171: Could this `increment` also return the new value?

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
```diff
@@ -24,7 +24,7 @@ mod tests;
 const MAX_EPOCH_ASSIGNMENT_RETRIES: usize = 5;
 
 #[derive(Default, Clone)]
-pub struct EpochPackerClient<StateStore, PendingStore, PerEpochStore, RollupManagerRpc> {
+pub struct EthersSettlementClient<StateStore, PendingStore, PerEpochStore, RollupManagerRpc> {
```
Freyskeyd: I would prefer to not have `Ethers` here, either `SettlementClient` or `SettlementLayerClient`.

```suggestion
pub struct SettlementLayerClient<StateStore, PendingStore, PerEpochStore, RollupManagerRpc> {
```

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
```diff
@@ -0,0 +1,31 @@
...
+/// and to submit them in a settlement tx to the L1.
+#[cfg_attr(
+    any(test, feature = "testutils"),
+    mockall::automock(
+        type PerEpochStore=MockPerEpochStore;
+        type Provider=ethers::providers::MockProvider;
+    )
+)]
+#[async_trait::async_trait]
+pub trait SettlementClient: Unpin + Send + Sync + 'static {
```
Freyskeyd, line 16: What do you think of having the `trait` being `SettlementLayer` and the `client` `SettlementLayerClient` ?
```suggestion
pub trait SettlementLayer: Unpin + Send + Sync + 'static {
```

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
```diff
@@ -204,22 +189,18 @@ where
...
         let gas_estimate = contract_call.estimate_gas().await.map_err(|error| {
             let error_str =
                 RollupManagerRpc::decode_contract_revert(&error).unwrap_or(error.to_string());
 
             error!(
                 error_code = %error,
                 error = error_str,
-                hash,
-                "Failed to settle the certificate {certificate_id}: {}", error_str
+                "Failed to settle certificate"
```
Freyskeyd: Maybe we can add a context to this error to define that we couldn't estimate the gas ?
I'm afraid that the lib doesn't return a really contextualized error

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -231,76 +233,80 @@ where
...
+                .update_settlement_tx_hash(&certificate_id, settlement_tx_hash)?;
+            self.set_status(CertificateStatus::Candidate)?;
+            debug!(tx_hash = ?self.header.settlement_tx_hash, "Submitted certificate for settlement");
+        }
+
+        // Third, wait for settlement to complete
+        if self.header.status < CertificateStatus::Candidate {
+            return Err(CertificateStatusError::InternalError(
+                "Trying to wait on a non-settling certificate".into(),
+            ));
```
Freyskeyd, line 262: nit: Maybe we could have an `ensure_status` macro or something like that

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -231,76 +233,80 @@ where
...
+                .send(NetworkTaskMessage::CertificateReadyForSettlement {
+                    height,
+                    certificate_id,
+                    settlement_submitted_notifier,
+                })
+                .await
+                .map_err(send_err)?;
+
+            let settlement_tx_hash = settlement_submitted.await.map_err(recv_err)??;
+            fail::fail_point!("certificate_task::about_to_record_candidate");
```
Freyskeyd: The convention that I followed for the fail points was to use some kind of path to the point, to better contextualize them, do you think it's possible to follow the same pattern?

```suggestion
            fail::fail_point!("certificate_task::process_impl::about_to_record_candidate");
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -231,76 +233,80 @@ where
...
+            self.header.settlement_tx_hash = Some(settlement_tx_hash);
+            self.state_store
+                .update_settlement_tx_hash(&certificate_id, settlement_tx_hash)?;
+            self.set_status(CertificateStatus::Candidate)?;
```
Freyskeyd: I think that the `update_settlement_tx_hash` is updating the `status of the `header` to `candidate` already.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -231,76 +233,80 @@ where
...
+            debug!(tx_hash = ?self.header.settlement_tx_hash, "Submitted certificate for settlement");
```
Freyskeyd:
```suggestion
            debug!(settlement_tx_hash = ?self.header.settlement_tx_hash, "Submitted certificate for settlement");
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -231,76 +233,80 @@ where
...
         }
-        // If we're not in the shortcut path of `settle_certificate`, wait for settlement
-        // TODO: once the storage is only ever touched from here, we can make this a regular `if`
-        // For now, the only way to reach this stage is to boot with an already-candidate
-        // certificate. So we know that `self.header.settlement_tx_hash` is set.
-        else if self.header.status < CertificateStatus::Settled {
-            debug!("Resuming certificate settlement");
-            let tx_hash = self.header.settlement_tx_hash.ok_or_else(|| {
+        if self.header.status < CertificateStatus::Settled {
+            debug!(tx_hash = ?self.header.settlement_tx_hash, "Waiting for certificate settlement to complete");
```
Freyskeyd:
```suggestion
            debug!(settlement_tx_hash = ?self.header.settlement_tx_hash, "Waiting for certificate settlement to complete");
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -231,76 +233,80 @@ where
...
+                })
+                .await
+                .map_err(send_err)?;
+
+            let (epoch_number, certificate_index) =
+                settlement_complete.await.map_err(recv_err)??;
+            let settled_certificate =
+                SettledCertificate(certificate_id, height, epoch_number, certificate_index);
+            self.set_status(CertificateStatus::Settled)?;
+            debug!(tx_hash = ?settlement_tx_hash, ?settled_certificate, "Certificate settlement completed");
```
Freyskeyd:
```suggestion
            debug!(settlement_tx_hash = ?settlement_tx_hash, ?settled_certificate, "Certificate settlement completed");
```

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
```diff
@@ -0,0 +1,31 @@
+use agglayer_types::{CertificateId, CertificateIndex, EpochNumber, SettlementTxHash};
+use ethers::providers::JsonRpcClient;
+
+use crate::Error;
+
+/// Epoch Packer used to gather all the proofs generated on-the-go
```
Freyskeyd:
```suggestion
/// Settlement Client used to gather all the proofs generated on-the-go
```

crates/agglayer-types/src/lib.rs
```diff
@@ -25,22 +32,183 @@ use pessimistic_proof::{
...
+    derive_more::From,
+    serde::Deserialize,
+    serde::Serialize,
+)]
+#[cfg_attr(feature = "testutils", derive(arbitrary::Arbitrary))]
+#[serde(transparent)]
+pub struct Metadata(Digest);
+
+impl Metadata {
+    pub const ZERO: Metadata = Metadata(Digest::ZERO);
```
Freyskeyd, line 202: nit: what do you think of using `empty` instead of `zero`, as `metadata` is basically a container for chain data?

```suggestion
    pub const EMPTY: Metadata = Metadata(Digest::ZERO);
```

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -29,13 +30,30 @@ pub(crate) struct NewCertificate {
...
-        response: oneshot::Sender<Box<LocalNetworkStateData>>,
+        response: oneshot::Sender<Result<Box<LocalNetworkStateData>, CertificateStatusError>>,
     },
-    CertificateProven {
+    CertificateExecuted {
         height: Height,
         certificate_id: CertificateId,
         new_state: Box<LocalNetworkStateData>,
     },
+    CertificateProven {
```
Freyskeyd, line 51: Maybe we could have some doc for those variants ?

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Uhh as far as I understand they're both also used on the L2 side, if only for generation of the certificate and signature of it… are they not? (Though maybe that's in the part that's currently still in go, I haven't really checked)

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Good point! I'll also directly apply your other suggestions; you're totally right that I added `as_u64` after the fact and didn't check that (within the private domain) the .0 weren't used any longer. Would probably have caught it whenever we split lib.rs into multiple files, which would have constrained the private domain much better 😄

crates/pessimistic-proof-test-suite/tests/e2e_local_pp.rs
Ekleog-Polygon: This would actually be AFAICT we need the cast to u64 anyway. Do you feel this change would be better? TBH I'm finding it a bit on the heavier side for tests code, but if you prefer it that's fine with me 😄 
```suggestion
        height: certificate.height.next().as_u64(),
```

crates/agglayer-storage/src/types/mod.rs
Ekleog-Polygon: TBH removing it would probably be more effort than it's worth right now, especially considering we're looking at bigger storage refactorings soon-ish. I'll add that to my checklist for the storage refactoring, as soon as I can I'll open a proper issue with actionable next steps for that!

(Also, if we remove it I'd vote for removing it in a follow-up PR anyway: this way if we need to revert the change it'll be easier than mixing it up here)

crates/agglayer-storage/src/stores/per_epoch/tests.rs
Ekleog-Polygon: The only reason is I'm finding it cleaner to say "this is a Height comparison" than "this is an u64 comparison" — this being said, in practice the two options should be basically equivalent. LMK if you prefer the other one!

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: I'm a bit scared people would be confused and think it'd be equivalent to `+ 1`, whereas it's actually equivalent to `+= 1`. So I'd rather do like Rust for `AssignEq`, and not return the incremented value — people can write `x.increment(); x` if they want to get that behavior as easily.

This being said, there's also `.next()` if one needs to just get the next value 😄

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Let's continue the discussion on the above thread and spread the conclusion here: https://github.com/agglayer/agglayer/pull/835#discussion_r2168782447

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
Ekleog-Polygon: I'm fine with any name! That said, I kind of liked the `SettlementClient` name — objects that implement this trait would be clients, not layers. As for the struct that impls it, maybe `RpcSettlementClient` would be a better name, to make it explicit that it's a client designed to settle through RPC?

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
Ekleog-Polygon: Nice catch, thanks! I hadn't actually read the context, but the error message is indeed pretty misleading

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: I'm planning on reworking the certificate task to be more like https://github.com/agglayer/agglayer/pull/819#discussion_r2111744145 in a follow-up PR. Does that sound good enough to you?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: My reasoning was that semantic paths ("which task? -> what stop of the task?") would be more resilient than physical paths against eg. refactorings that could change the physical path. This being said, I'm totally fine with changing the pattern here if you think the current one is better! (Or I could also move all to semantic paths, it's just 12 fail points for now, should be reasonably easy to change in any direction)

crates/agglayer-types/src/lib.rs
```diff
@@ -25,22 +32,183 @@ use pessimistic_proof::{
...
+
+impl EpochNumber {
+    pub const ZERO: EpochNumber = EpochNumber(0);
+    pub const ONE: EpochNumber = EpochNumber(1);
+
+    pub const fn new(epoch: u64) -> EpochNumber {
+        EpochNumber(epoch)
+    }
+
+    pub const fn next(&self) -> EpochNumber {
```
iljakuklic, line 64: Elaborating on the discussion below on the potential misunderstanding whether given methods updates the value in place or returns the new value. A useful thing to do in this case is to require the user to do something with the result.

```suggestion
    #[must_use = "The value of the next epoch is returned but not used"]
    pub const fn next(&self) -> EpochNumber {
```

crates/agglayer-types/src/lib.rs
```diff
@@ -25,22 +32,183 @@ use pessimistic_proof::{
...
+
+    pub const fn as_u64(&self) -> u64 {
+        self.0
+    }
+
+    pub const fn new(height: u64) -> Height {
+        Height(height)
+    }
+
+    pub const fn next(&self) -> Height {
```
iljakuklic, line 167: Same here
```suggestion
    #[must_use = "The value of the next height is returned but not used"]
    pub const fn next(&self) -> Height {
```

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: TBH I'm thinking it'd go against [Tech debt: introduce a macro for creating strong hash types](https://github.com/agglayer/interop/issues/39#top)#39 in the future. Also, ZERO is a zero hash, which is not actually the hash of empty metadata.

This being said, maybe NULL would be a better name? TBH I really don't like that we're using all-zero hashes so much, it really feels bad to me, and I'd love if we didn't need this constant at all

crates/agglayer-types/src/lib.rs
```diff
@@ -798,3 +969,53 @@ impl LocalNetworkStateData {
...
+    fn as_ref(&self) -> &Digest {
+        &self.0
+    }
+}
+
+impl fmt::Display for SettlementTxHash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
```
iljakuklic: I believe these and a bunch of other impls in this file could be `derive_more`d.

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Great point! I had thought of one of the mistakes but not the other 😄

crates/agglayer-aggregator-notifier/src/lib.rs
```diff
@@ -3,8 +3,8 @@ const ELF: &[u8] =
     include_bytes!("../../pessimistic-proof-program/elf/riscv32im-succinct-zkvm-elf");
 
 mod certifier;
-mod packer;
 mod proof;
+mod settlement_client;
 
 pub use certifier::CertifierClient;
-pub use packer::EpochPackerClient;
+pub use settlement_client::EthersSettlementClient;
```
atanmarko: `ethers` is a well know library name and may introduce confusion (especially as we are switching to alloy).

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
atanmarko: Plus one for `SettlementLayerClient`

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
```diff
@@ -81,21 +78,15 @@ where
             self.state_store.get_certificate_header(&certificate_id)?
         {
             if status == CertificateStatus::Settled {
-                error!(
-                    hash,
-                    "The certificate {} is already settled", certificate_id
-                );
+                error!("Certificate is already settled");
```
atanmarko, line 81: At what span level do we instrument `certificate_id`, not in this function span? Would it be easily spot-able in the debug logs?

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
```diff
@@ -310,31 +262,22 @@ where
 
         match receipt.status {
             Some(n) if n == U64::zero() => {
-                warn!(
-                    hash,
-                    "The transaction failed to settle the certificate {}", certificate_id
-                );
+                warn!("Transaction failed to settle");
```
atanmarko, line 265: Printing tx hash here would help a lot, also on a few other places in this function.

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
```diff
@@ -354,73 +297,33 @@ where
...
 
                             return Err(Error::SettlementError {
                                 certificate_id,
                                 error: error_msg,
                             });
                         }
                         Err(error) => warn!(
-                            "Failed to add the certificate {} to the epoch (retrying): {}",
-                            certificate_id, error
+                            %error, "Failed to add the certificate to the epoch (retrying)"
```
atanmarko, line 311: Should we use a debug error print in this kind of logs?

crates/agglayer-types/src/lib.rs
atanmarko: Yeah, I would go with current version.

iljakuklic (dismissed):

crates/agglayer-aggregator-notifier/src/lib.rs
Ekleog-Polygon: 👍 marking as resolved as we're discussing this above

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
Ekleog-Polygon: We're instrumenting [here](https://github.com/agglayer/agglayer/pull/835/files/726303d3c27db3ccc7696444989ea4bcd40e4482#diff-bc3414834dec73ed0ed577d47b7e9d1d58d94232db6b0e8f7455165a6e0fdf00R67), at INFO level as INFO is the default level for tracing::instrument logs. Do you think there's anyone who could turn off INFO logging but not ERROR?

As for how spot-able it is, here's what the logs look like in our tests:
```
      2025-06-27T13:25:02.448300Z ERROR agglayer_aggregator_notifier::settlement_client: demo
        at crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs:72
        in agglayer_aggregator_notifier::settlement_client::submit_certificate_settlement with certificate_id: CertificateId(0c672404c808c995704c9be774fb8618030327645719b15a317ad1c0bb56d777)
        in agglayer_certificate_orchestrator::network_task::NetworkTask::run with network_id: 2
```

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
Ekleog-Polygon: Similarly, it's already in the instrument span at the top of the function

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
Ekleog-Polygon: IMO it feels like the pre-existing warn was good: sure we can retry, but if it failed once it's still an indication that something went wrong somewhere

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
Ekleog-Polygon: Renamed to `RpcSettlementClient` as per the slack poll

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
Ekleog-Polygon: Name adjusted to SettlementClient/RpcSettlementClient as per the slack poll

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Renamed the fail point to a semantic/physical path mix as per the slack poll

atanmarko (dismissed):

crates/agglayer-types/src/lib.rs
Freyskeyd: The `Certificate` itself is only linked to the `agglayer-node` meaning that it should stick to `agglayer-node` domain.
L2 chains are building the Certificate using the gRPC definition that is part of `agglayer-node` scope, if at some point we want to expose more capabilities for the `client` (such as pre-calculating the `CertificateId`), we'll expose those methods as part of a lib, most likely `agglayer-grpc-client`.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Freyskeyd: Okey, we could also explore a refactorisation where we have a clear state machine based on the status maybe?

crates/agglayer-types/src/lib.rs
Freyskeyd: Yes, the `hash` of :

- `EMPTY` is `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`
- `ZERO` is `0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563`

Based on how `aggsender` is [serializing](https://github.com/agglayer/aggkit/blob/6174b820e35cbc7ba14fbfba3c8bd3dbdc90531c/aggsender/types/certificate_metadata.go#L80C7-L80C38) the metadata, `ZERO` makes more sense indeed.

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
```diff
@@ -54,24 +54,21 @@ impl<StateStore, PendingStore, PerEpochStore, RollupManagerRpc>
...
     RollupManagerRpc: RollupContract + Settler + Send + Sync + 'static,
     PerEpochStore: PerEpochWriter + PerEpochReader + 'static,
 {
-    type PerEpochStore = PerEpochStore;
     type Provider = <<RollupManagerRpc as Settler>::M as ethers::providers::Middleware>::Provider;
 
-    #[instrument(skip_all, fields(hash, network_id, calldata), level = "debug")]
-    async fn settle_certificate(
+    #[instrument(skip(self), fields(network_id, calldata), level = "debug")]
+    async fn submit_certificate_settlement(
```
Freyskeyd, line 68: In this case, does `certificate_id` will be displayed or debuged in the log?

atanmarko (approved):

crates/agglayer-types/src/lib.rs
iljakuklic: Agreed on the general point of avoiding exposing the `ZERO` constant on strongly typed digests. If a zero hash has a special meaning for particular type, a constant with a descriptive name should be introduced selectively.

crates/agglayer-aggregator-notifier/src/settlement_client/mod.rs
Ekleog-Polygon: Yes, it will appear like this: https://github.com/agglayer/agglayer/pull/835#discussion_r2172046122

PR #839, opened 2025-06-09
fix: restore `pessimisitic_proof_core::Error` storage compatibility with 0.2.5

iljakuklic:
# Description

This restores backwards compatibility of `pessimisitic_proof_core::proof::Error` with format produced by agglayer node `v0.2.5`.

## Breaking changes

This restores storage compatibility with v0.2.5 but breaks all the prior versions of the v0.3.*, including all the release candidates. That is because the issue was introduced shortly after the v0.2.5 release.

Unfortunately, we should not let it go because the mainnet already includes some certificates in the original encoding format in the storage. It seems that so far, `bali` does not include any certificates in the new breaking format, so this update should not break anything there as of 6th June. However, it is an ongoing network so a certificate like that might show up between then and the time the fix is deployed. In that case, a manual intervention (such as removing the offending certificate from the storage) will be required.

## PR Checklist:


Ekleog-Polygon (approved):
Haven't checked that the changes do make the structs backward-compatible, but considering the effort you've been pouring into it recently I trust you on it. Thank you! 😄

PR #841, opened 2025-06-09
fix: replace deprecated #[clap(...)] with #[command(...)] and #[arg(.…)]

Himess:
# Description

In **clap** version 4.x and above, the use of `#[clap(...)]` is no longer valid.  
Many instances of `#[clap(...)]` are still present in the project.  
These should be replaced with `#[arg(...)]` or `#[command(...)]`.

Error : 

Pasted Graphic 3

Fix : 

Pasted Graphic 4


atanmarko:
Thanks for the PR and effort. We are still waiting some clarifications from the legal team before we could accept external contributions.

atanmarko (commented):
Thanks for the effort. We are still waiting some clarifications from the legal team before we could accept external contributions.

iljakuklic (approved):

PR #849, opened 2025-06-16
feat: accounting on U512 instead of U256 on balances

hadjiszs:
# Description

In the circuit, the sum of imported bridge exits is performed using `U512` values before subtracting the bridge exit amount. This allows importing large amounts, as long as enough is consumed in the same state transition to bring the total balance back within valid `U256` limits.

Witness generation currently constrains this sum to `U256`, which can reject transitions that would have been considered valid by the circuit.

This PR removes that constraint to align witness generation with the circuit logic.

## PR Checklist:

### DIFF
--- a/crates/agglayer-types/src/lib.rs
+++ b/crates/agglayer-types/src/lib.rs
@@ -6,7 +6,9 @@ use agglayer_interop_types::{
     ImportedBridgeExitCommitmentValues, TokenInfo,
 };
 pub use agglayer_primitives::Digest;
-use agglayer_primitives::{keccak::Keccak256Hasher, FromBool, Hashable, SignatureError};
+use agglayer_primitives::{
+    keccak::Keccak256Hasher, ruint::UintTryFrom, FromBool, Hashable, SignatureError,
+};
 use agglayer_tries::{error::SmtError, roots::LocalExitRoot, smt::Smt};
 use pessimistic_proof::{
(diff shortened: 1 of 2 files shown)


hadjiszs:
> Do we anticipate any need to go beyond u256 in the middle of a balance computation? I'd expect 2**256 to already be way more than enough for any practical use case?

It relates to this https://github.com/agglayer/agglayer/issues/242

The change was made in the circuit but not on the witness generation at that time

hadjiszs:
> even just bridging out a token they don't have would be exactly as simple a way to make that happen.

my understanding is that in this example, we **do** want the PP to be invalid if the token isn’t originated from the chain and the balance goes below zero, as opposed to the case exposed where the PP would have no reason to be invalid, basically that's over-constrained.

---

Another example to illustrate the intention here I think is the case of bootstrapping a chain from genesis. In such a scenario, we might want to generate a PP for a very large block range.

In the witness generation and the PP logic, what we do is:

1. Sum all the imported bridge exits per token
2. Subtract all the bridge exits per token

If the block range is very large, step (1) can overflow a `U256`. But there’s no strong reason to constrain this sum to `U256`, since all we really care about is that **after step (2)**, the resulting balance is within `0..U256::MAX`, hence the use of `U512` for step (1) only.

iljakuklic (approved):

PR #857, opened 2025-06-23
feat: use U256 instead of global index

hadjiszs:
# Description

Use U256 directly instead of GlobalIndex in order to preserve the values of the padded unused bits.

Companion of:
- https://github.com/agglayer/interop/pull/44
- https://github.com/agglayer/provers/pull/244

## PR Checklist:


Cargo.toml
```diff
@@ -44,11 +44,11 @@ pessimistic-proof-core = { path = "crates/pessimistic-proof-core" }
...
-agglayer-interop = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.0" }
-agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.0" }
-agglayer-primitives = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.0" }
-agglayer-tries = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.0" }
-unified-bridge = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.0" }
+agglayer-interop = { git = "https://github.com/agglayer/interop.git", branch = "feat/gi-unused-bits" }
+agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", branch = "feat/gi-unused-bits" }
+agglayer-primitives = { git = "https://github.com/agglayer/interop.git", branch = "feat/gi-unused-bits" }
+agglayer-tries = { git = "https://github.com/agglayer/interop.git", branch = "feat/gi-unused-bits" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "feat/gi-unused-bits" }
```
hadjiszs, line 51: todo: points to release tag once ready

crates/agglayer-types/src/lib.rs
```diff
@@ -105,7 +105,7 @@ pub enum Error {
     )]
     NullifierPathGenerationFailed {
         source: SmtError,
-        global_index: GlobalIndex,
+        global_index: U256,
```
hadjiszs, line 108: @iljakuklic such change is breaking for the serialization/deserialization iirc right? In that case should I just create a new variant with `U256` instead of `GlobalIndex`?

iljakuklic: > @iljakuklic such change is breaking for the serialization/deserialization iirc right? In that case should I just create a new variant with `U256` instead of `GlobalIndex`?

Yep, a new variant is required here.

Once we get the up-to-date `interop` dependency, I will be able to create a PR taking a snapshot of the structure of these types which will make it easier to spot the incompatibilities.

hadjiszs:
Not needed anymore: smart contracts enforce these bits to be zero

PR #858, opened 2025-06-23
chore: adding release strategy

Freyskeyd:
# Description

This PR introduces a release strategy to standardize and automate the release process for the `agglayer` project.


RELEASE.md
```diff
@@ -0,0 +1,115 @@
...
+
+The release process begins when a codeowner determines that the codebase is ready for a new release. This decision is typically made after evaluating completed features, bug fixes, and overall code stability. Codeowners initiate the release through GitHub's manual workflow dispatch interface, selecting only the target branch (`main` for minor releases or `release/vX.Y` for patch releases). 
+
+The system automatically determines the appropriate version number by reading the current version from `Cargo.toml` and applying our semantic versioning rules. This eliminates manual version input errors and ensures consistency with our established versioning strategy.
+
+### Step 2: Automated Release Preparation
+
+Once triggered, our automated system creates a dedicated release branch following our naming convention (`release/vX.Y` for minor releases or `release/vX.Y.Z` for patches). The system then performs several critical tasks: it reads the current version from `Cargo.toml` to determine the release version, generates a comprehensive changelog using `git-cliff` based on our conventional commit history, and creates a pull request tagged with the `release` label.
```
atanmarko: Not sure that we need separate  `release/vX.Y.Z` branch for patches. Probably one `release/vX.Y` is enough, with tag for every patch that is released on that branch.

RELEASE.md
```diff
@@ -0,0 +1,115 @@
...
+
+This automation ensures that all release preparations follow identical procedures, reducing human error and maintaining consistency across all releases. The generated changelog provides a clear overview of what changes will be included in the release, enabling informed review decisions.
+
+### Step 3: Review and Quality Gate
+
+The release PR enters a critical review phase where codeowners examine the generated changelog, verify the version bump logic, and ensure all intended changes are properly captured. During this review period, our automated systems prevent any other pull requests from merging to the target branch, creating a controlled environment that eliminates race conditions and ensures the release content matches the reviewed changelog.
```
atanmarko, line 45: > During this review period, our automated systems prevent any other pull requests from merging to the target branch...

This is a bit unclear to me. Do we create a release branch after we merge the release PR to main? What is the target branch in this case - main branch or release PR temporary  branch. Is release PR temporary  branch deleted when it is merged to main (and then we create `release/vX.Y` branch starting from tagged main commit)?

RELEASE.md
```diff
@@ -0,0 +1,115 @@
...
+
+If the base branch receives updates during the review process, our automation automatically updates the release PR by rebasing or merging the changes and regenerating the changelog. This keeps the release current while maintaining the integrity of the review process. The PR can only be merged after receiving explicit codeowner approval.
```
atanmarko, line 47: So in this case, we are talking about temporary release PR branch - not the `release/vX.Y` branch?

RELEASE.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Main Branch Philosophy
+
+The main branch serves as our primary development artery, representing the latest state of our codebase and the foundation for all future releases. This branch maintains a continuous flow of new features, improvements, and non-breaking changes that have passed through our merge group validation process. 
+
+The main branch follows a strict versioning discipline where the version number in `Cargo.toml` always reflects the next intended minor release in the format `vX.Y.0`. This convention ensures that every commit on main is clearly associated with an upcoming feature release, eliminating ambiguity about which version will contain specific changes.
+
+Following each minor release from main, our automation immediately increments the version to the next minor version (e.g., from `v1.2.0` to `v1.3.0`), ensuring that subsequent development work is properly versioned and the branch remains ready for the next development cycle. This automatic versioning prevents version conflicts and maintains a clear progression of development milestones.
```
atanmarko, line 71: We should cover here also the current `v0.major.minor` status - and when do we plan to switch to `v1.minor.patch`

RELEASE.md
```diff
@@ -0,0 +1,115 @@
...
+
+Release Candidates are particularly valuable for releases that include:
+- Significant new features or architectural changes
+- Updates that affect critical system components
+- Releases following extended development periods
+- Changes that require stakeholder or user validation before general availability
+
+### Release Candidate Workflow
+
+Release Candidates are created using the same workflow dispatch process as regular releases, with an additional parameter to designate the release as a candidate. When the "Release Candidate" option is selected, the system bypasses the standard PR creation and review process, instead creating a direct git tag with the appropriate RC suffix (e.g., `v1.2.0-rc.1`) and immediately publishing it as a pre-release on GitHub.
```
atanmarko, line 107: I think would be great to always use 2 digit number for release candidate versioning - for example `v1.2.0-rcxx` where xx is 2 digit number starting from 01. So the release candidates for `v1.2.0` would be `v1.2.0-rc01`, `v1.2.0-rc02` etc.

RELEASE.md
```diff
@@ -0,0 +1,115 @@
+# Release Strategy
+
+## Overview
+
+This document provides a comprehensive guide to our release management process, defining how we coordinate, version, and deploy our Docker-based application. It establishes clear procedures for creating consistent releases while maintaining quality and minimizing coordination overhead.
```
atanmarko, line 5: Would be good to add a note that crates/libraries may follow different versioning system.

RELEASE.md
Freyskeyd: The release PR represents a PR that contains the changelog update + the decision of the codeowner to do the release. The release PR could be on a branch name: `chore/prepare-X.X.X`

RELEASE.md
Freyskeyd: I think that the convention and the general approach for this is to use a number, so `1.2.0-rc.1`, if we're using `1.2.0-rc.01` it makes a difference between the RC number and the other numbers. Is there a particular reason for this suggestion?

RELEASE.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Branch Strategy
+
+Our branching strategy is designed to support parallel development and maintenance workflows while maintaining clear separation between ongoing feature development and stable release maintenance. This approach enables us to deliver new features on the main branch while simultaneously providing critical fixes for previously released versions.
+
+### Main Branch Philosophy
+
+The main branch serves as our primary development artery, representing the latest state of our codebase and the foundation for all future releases. This branch maintains a continuous flow of new features, improvements, and non-breaking changes that have passed through our merge group validation process. 
+
+The main branch follows a strict versioning discipline where the version number in `Cargo.toml` always reflects the next intended minor release in the format `vX.Y.0`. This convention ensures that every commit on main is clearly associated with an upcoming feature release, eliminating ambiguity about which version will contain specific changes.
```
Ekleog-Polygon, line 69: I think this is opposite to what we were discussing? AFAIU we were saying we'd have the version number of the application be the next intended minor release, but all our crates would stay with the previous release number, so that release-plz could Just Work(tm)

Ekleog-Polygon (commented):
LGTM! I'm just with Marko that we probably don't need the release/vX.Y.Z branch(es)

RELEASE.md
Freyskeyd: This part isn't updated based on the recent discussion, because there are still some issues that I need to sort out, but to summarize:

The `application` version will be held by the `crate` that represents the application.
All others crates can rely on `release-plz`, to bump their version. But as suggested by @atanmarko, if we want to have the same version for every crates that belong to the same application, we do not need to have `release-plz` automation.

As an example, if we take `agglayer` application, the main `crate` is `agglayer` crate which will contains the next version `0.4.0`, thus all the crate that belongs to this particular application will have `0.4.0` version.

The only place where we could rely on release-plz automation could be on isolated crates, small group of crates that are not linked to application, or on `interop` repo. 

WDYT?

Ekleog-Polygon: I agree with you that release-plz is not as useful for crates we don't release (if only because we can all do it ourselves), but IMO it's still easier to click "merge" on a PR than to trigger a release workflow on a specific commit. And if we can decide that we bump the applications' version numbers only at the time of releasing a new version, it'd mean we can use release-plz for that simplification.

Do you think we have a real need to bump the version number just after the release? If yes, then as you say we could ignore any non-published crates on release-plz, and use it only for the crates we publish on crates.io (that thus basically need to be handled via release-plz). If no, then I'd suggest just having a group for application crates on release-plz so that they're all on the same version number, and we can benefit from a single process for all our releases.

RELEASE.md
atanmarko: No special reasoning about it, just my taste, this is a nit and I do not feel strongly about it. For example if we release `1.2.0-rc.1`, then `1.2.0-rc.2` and later `1.2.0-rc.11`, it is a bit harder to determine (and observe in any sorted list), which rc is the latest

RELEASE.md
Freyskeyd: I agree with you on the automation stuff, but my point is to really differentiate those 2 paths, between a library release and an application release. The rationale behind bumping the application version just after a release is that it clearly state the beginning and the end of each timeline. It also helps to understand that any changes to a version should be addressed at the right place, on the right branch. It gives a framework that clearly defines where and how we manage our versions and our application release.

Taking this into account, having this split means that we're able to iterate at different pace for the library, using different tools or even different workflows. And honestly, `release-plz` seem good, but we do not have experience with it and I would prefer having a first round of implementation that isn't part of the critical aspect of the release of an application.

What I can suggest is that we start by using `release-plz` where we want to publish things to crate.io, while starting to implement the application release flow defined in this PR. Let's test it, gain experience with `release-plz` overtime and then, if we want to modify anything in this doc, we will. Let's do this iteratively :)

RELEASE.md
Freyskeyd: I think the decision of major release will follow the same path as defined, but in a less regular way. Let's see during the first iterations

RELEASE.md
Freyskeyd: I know that I'm using `git tag --sort=-v:refname` with an alias for that.

PR #861, opened 2025-06-24
chore: bump PP program version

iljakuklic:
Per-emptively bump PP program version before the next release. Version 5.0.0 is skipped as it corresponds to PP program for Agglayer v0.3.5.

* related (and kinda backport of) #859 

## Breaking changes

vkey selector is updated

## PR Checklist:


crates/pessimistic-proof-program/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pessimistic-proof-program"
-version = "4.0.0"
+version = "6.0.0"
```
atanmarko, line 3: Do we bump the pp program version for a new minor release even if there are no changes in that program? @Freyskeyd

Freyskeyd: @atanmarko @iljakuklic maybe we should check the vkey from the 0.3.5 and main

iljakuklic: The idea here is to preemptively version bump so by the next release, the program version would have already been incremented. Let me know if you think it's not a good idea. As far as I know, main has already updated the program, so setting it to 5 would create a selector clash again (between 0.3.5 and main or whatever version main becomes).

hadjiszs (approved):

atanmarko (dismissed):

Freyskeyd:
@iljakuklic can you update this one?

Freyskeyd (approved):

PR #865, opened 2025-06-25
test: take a snapshot of DB storage structure

### LINKED ISSUE #816: Statically enforce storage backwards compatibility
Agglayer node has to be able to correctly load certificates and other pieces of data written by older versions of the software. We currently rely on code reviews to accomplish that which has proven to be rather error prone. Tooling should be introduced to automate this at least in part.

### Suggested approach

Take a snapshot of the internal structure of serialized types as understood by serde. Commit it into the repo. Differences will be way more obvious during the code review, making potential incompatibilities easier to spot.

* [serde_reflection](https://crates.io/crates/serde-reflection) to extract the structure
* [insta](https://crates.io/crates/insta) to take the snapshot

### Possible follow up

The CI could checkout the structure snapshot from the base branch and from the PR branch. Then feed them both into a script checking backwards incompatibilites.

* Enum arms can only be added at the end of the enum
* Adding fields to structs is not allowed

False positives are to some degree expected but it should help with the most common cases.

iljakuklic:
Take a snapshot of the internal structure of data types stored in the persistent database. This will allow us to easier spot changes that break backwards compatibility.

Currently applied to `CerificateHeader` and types it consists of. Most notorious among these are all the error enums.

* fixes #816
* requires agglayer/interop#40

## PR Checklist:


Cargo.toml
```diff
@@ -44,12 +44,12 @@ pessimistic-proof-core = { path = "crates/pessimistic-proof-core" }
...
-agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
-agglayer-primitives = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
-agglayer-tries = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
-unified-bridge = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
+agglayer-bincode = { git = "https://github.com/agglayer/interop.git", branch = "test/storage-encoding-format" }
+agglayer-interop = { git = "https://github.com/agglayer/interop.git", branch = "test/storage-encoding-format" }
+agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", branch = "test/storage-encoding-format" }
+agglayer-primitives = { git = "https://github.com/agglayer/interop.git", branch = "test/storage-encoding-format" }
+agglayer-tries = { git = "https://github.com/agglayer/interop.git", branch = "test/storage-encoding-format" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "test/storage-encoding-format" }
```
iljakuklic: TODO use a tag here

Ekleog-Polygon (dismissed):
LGTM! Hopefully we'll switch to protobuf soon, which should avoid these issues, but for now this should help 😄

crates/agglayer-types/src/lib.rs
```diff
@@ -249,6 +249,7 @@ pub struct CertificateHeader {
 }
 
 #[derive(Debug, thiserror::Error, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename = "agglayer_types::Error")]
```
Freyskeyd, line 252: Does it change how `InError` status are parsed?

Cargo.toml
```diff
@@ -44,12 +44,12 @@ pessimistic-proof-core = { path = "crates/pessimistic-proof-core" }
...
-agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
-agglayer-primitives = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
-agglayer-tries = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
-unified-bridge = { git = "https://github.com/agglayer/interop.git", tag = "v0.7.1" }
+agglayer-bincode = { git = "https://github.com/agglayer/interop.git", branch = "main" }
+agglayer-interop = { git = "https://github.com/agglayer/interop.git", branch = "main" }
+agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", branch = "main" }
+agglayer-primitives = { git = "https://github.com/agglayer/interop.git", branch = "main" }
+agglayer-tries = { git = "https://github.com/agglayer/interop.git", branch = "main" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "main" }
```
atanmarko: Just a reminder to update the tag when new release on `interop` is made.

crates/pessimistic-proof-program/elf/riscv32im-succinct-zkvm-elf
atanmarko, line 1: So this change will take `6.0.0` version from the current main?

crates/agglayer-storage/src/types/certificate/tests/structure.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    tracer.registry()
+}
+
+#[test]
+fn structure_snapshot() {
+    // Take a snapshot of the internal structure of the encoding of Certificate
+    // to avoid accidental changes.
+    match trace() {
+        Ok(registry) => insta::assert_json_snapshot!(&registry),
```
atanmarko, line 46: Just to remember not to `cargo insta review accept` automatically. Maybe some additional comment or warning is needed here, that change in this snapshot may have serious consequences.

iljakuklic: > Just to remember not to `cargo insta review accept` automatically. Maybe some additional comment or warning is needed here, that change in this snapshot may have serious consequences.

Will add the comment although maybe in the snapshot file itself would be more appropriate. Also note that even if you just `insta review accept`, the diff will still show up in review.

crates/agglayer-types/src/lib.rs
iljakuklic: > Does it change how `InError` status are parsed?

No. The data type name does not affect parsing of formats we care about (bincode, json, toml, etc.). The keys are determined by the enum arm name, which stays `CertificateStatus::InError`.

PR #874, opened 2025-07-02
chore: introduce send_to_network_task helper

Ekleog-Polygon:
Handle delayed https://github.com/agglayer/agglayer/pull/819#discussion_r2124732959

Also trace all messages sent to network task, as it could always be useful

### DIFF
--- a/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
@@ -305,10 +305,10 @@ where
 
     async fn send_to_network_task(
         &self,
-        msg: NetworkTaskMessage,
+        message: NetworkTaskMessage,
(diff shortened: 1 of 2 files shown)


crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -314,6 +302,14 @@ where
         self.header.status = status;
         Ok(())
     }
+
+    async fn send_to_network_task(
+        &self,
+        msg: NetworkTaskMessage,
+    ) -> Result<(), CertificateStatusError> {
+        trace!(?msg, "Sending message to network task");
```
iljakuklic: Nit: call the log key "message" rather than "msg"
```suggestion
        trace!(message = ?msg, "Sending message to network task");
```

(or rename the varible)

iljakuklic (dismissed):

PR #877, opened 2025-07-03
feat: add a private gRPC port, and private networks that can only submit through it

Ekleog-Polygon:
Fixes https://github.com/agglayer/protocol-research/issues/155

The code is not as good as I'd have liked it to be, but considering the time constraints I think this should be good enough to merge. Of note is the fact that I haven't added a test for the gRPC ports themselves, because I couldn't find any pre-existing gRPC submit_certificate test to copy from, and writing one from scratch would likely take too much time right now.

### DIFF
--- a/crates/agglayer-config/src/lib.rs
+++ b/crates/agglayer-config/src/lib.rs
@@ -73,7 +73,7 @@ pub struct Config {
     #[serde(default)]
     pub rpc: RpcConfig,
 
-    /// The private networks configuration.
+    /// The proxied networks configuration.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub proxied_networks: Option<ProxiedNetworksConfig>,
 
@@ -197,8 +197,8 @@ impl Config {
         std::net::SocketAddr::from((self.rpc.host, self.rpc.grpc_port))
     }
 
-    /// Get the private gRPC socket address from the configuration.
-    pub fn private_grpc_addr(&self) -> Option<std::net::SocketAddr> {
+    /// Get the proxied gRPC socket address from the configuration.
+    pub fn proxied_grpc_addr(&self) -> Option<std::net::SocketAddr> {
         self.proxied_networks
             .as_ref()
             .map(|pn| std::net::SocketAddr::from((pn.host, pn.grpc_port)))
--- a/crates/agglayer-jsonrpc-api/src/testutils.rs
+++ b/crates/agglayer-jsonrpc-api/src/testutils.rs
@@ -49,7 +49,7 @@ pub struct TestContext {
     pub pending_store: Arc<PendingStore>,
     pub client: jsonrpsee::http_client::HttpClient,
     pub admin_client: jsonrpsee::http_client::HttpClient,
-    pub private_grpc_client: Option<jsonrpsee::http_client::HttpClient>,
+    pub proxied_grpc_client: Option<jsonrpsee::http_client::HttpClient>,
     pub config: Arc<Config>,
     pub certificate_receiver: tokio::sync::mpsc::Receiver<(NetworkId, Height, CertificateId)>,
 }
@@ -75,9 +75,9 @@ impl TestContext {
         let admin_url = format!("http://{}/", raw_rpc.config.admin_rpc_addr());
         let client = HttpClientBuilder::default().build(api_url).unwrap();
         let admin_client = HttpClientBuilder::default().build(admin_url).unwrap();
-        let private_grpc_client = raw_rpc.config.proxied_networks.as_ref().map(|pn| {
-            let private_url = format!("http://{}:{}", pn.host, pn.grpc_port);
-            HttpClientBuilder::default().build(private_url).unwrap()
+        let proxied_grpc_client = raw_rpc.config.proxied_networks.as_ref().map(|pn| {
+            let proxied_url = format!("http://{}:{}", pn.host, pn.grpc_port);
+            HttpClientBuilder::default().build(proxied_url).unwrap()
         });
 
         let listener_api = tokio::net::TcpListener::bind(raw_rpc.config.readrpc_addr())
@@ -102,7 +102,7 @@ impl TestContext {
             pending_store: raw_rpc.pending_store,
             client,
             admin_client,
-            private_grpc_client,
+            proxied_grpc_client,
             config: raw_rpc.config,
             certificate_receiver: raw_rpc.certificate_receiver,
         }
@@ -134,12 +134,12 @@ impl TestContext {
         config.rpc.readrpc_port = addr.port();
         config.rpc.admin_port = admin_addr.port();
         if let Some(proxied_networks) = config.proxied_networks.as_mut() {
-            let private_addr = next_available_addr();
-            proxied_networks.host = match private_addr.ip() {
+            let proxied_addr = next_available_addr();
+            proxied_networks.host = match proxied_addr.ip() {
                 IpAddr::V4(ip) => ip,
                 IpAddr::V6(_) => Ipv4Addr::new(127, 0, 0, 1),
             };
-            proxied_networks.grpc_port = private_addr.port();
+            proxied_networks.grpc_port = proxied_addr.port();
         };
 
         let config = Arc::new(config);
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -276,26 +276,26 @@ impl Node {
         .start()
         .await?;
 
-        // List the public vs. private networks.
+        // List the public vs. proxied networks.
         let (public_networks, proxied_networks) = {
-            let private = config
+            let proxied = config
                 .proxied_networks
                 .as_ref()
                 .map(|n| n.networks.iter().cloned().collect::<HashSet<_>>())
                 .inspect(|pn| {
                     if pn.is_empty() {
                         warn!(
-                            "No private networks configured, but a port was configured. All \
+                            "No proxied networks configured, but a port was configured. All \
                              networks will be considered public."
                         );
                     }
                 });
-            let private2 = private.clone();
+            let proxied2 = proxied.clone();
             (
                 // Is the incoming network public?
-                move |incoming| private.as_ref().is_none_or(|p| !p.contains(&incoming)),
-                // Is the incoming network private?
-                private2.map(|private| move |incoming| private.contains(&incoming)),
+                move |incoming| proxied.as_ref().is_none_or(|p| !p.contains(&incoming)),
+                // Is the incoming network proxied?
+                proxied2.map(|proxied| move |incoming| proxied.contains(&incoming)),
             )
         };
 
@@ -305,12 +305,12 @@ impl Node {
                 .start()
                 .await?;
 
-        let private_grpc_router = match proxied_networks {
+        let proxied_grpc_router = match proxied_networks {
             None => None,
             Some(pn) => Some(
                 agglayer_grpc_api::Server::with_config(config.clone(), rpc_service.clone(), pn)
                     .build()
-                    .inspect_err(|err| error!(?err, "Failed to build private gRPC router"))?,
+                    .inspect_err(|err| error!(?err, "Failed to build proxied gRPC router"))?,
             ),
         };
         let public_grpc_router =
@@ -326,17 +326,17 @@ impl Node {
 
         let readrpc_listener = tokio::net::TcpListener::bind(config.readrpc_addr()).await?;
         let public_grpc_listener = tokio::net::TcpListener::bind(config.public_grpc_addr()).await?;
-        let private_grpc_listener = match config.private_grpc_addr() {
+        let proxied_grpc_listener = match config.proxied_grpc_addr() {
             None => None,
             Some(addr) => Some(tokio::net::TcpListener::bind(addr).await?),
         };
         let admin_listener = tokio::net::TcpListener::bind(config.admin_rpc_addr()).await?;
         info!(on = %config.readrpc_addr(), "ReadRPC listening");
         info!(on = %config.public_grpc_addr(), "Public gRPC listening");
-        if let Some(on) = config.private_grpc_addr() {
-            info!(%on, nets=?config.proxied_networks.as_ref().map(|p| &p.networks), "Private gRPC listening");
+        if let Some(on) = config.proxied_grpc_addr() {
+            info!(%on, nets=?config.proxied_networks.as_ref().map(|p| &p.networks), "Proxied gRPC listening");
         } else {
-            debug!("No private gRPC server configured.");
+            debug!("No proxied gRPC server configured.");
         }
         info!(on = %config.admin_rpc_addr(), "AdminRPC listening");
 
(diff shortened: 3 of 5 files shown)


crates/agglayer-config/src/lib.rs
```diff
@@ -71,6 +73,10 @@ pub struct Config {
     #[serde(default)]
     pub rpc: RpcConfig,
 
+    /// The private networks configuration.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub private_networks: Option<PrivateNetworksConfig>,
```
iljakuklic: Would it make sense to turn this into a vec or a map to have multiple "sets" of private networks with distinct ports and potentially distinct configuration?

crates/agglayer-config/src/private_networks.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[serde(deserialize_with = "deserialize_grpc_port")]
+    pub grpc_port: u16,
+
+    /// The host for the local RPC server for private networks.
+    pub host: Ipv4Addr,
+}
+
+impl PrivateNetworksConfig {
+    /// Create a new `PrivateNetworksConfig` with default values.
+    pub fn for_tests(networks: Vec<NetworkId>) -> Self {
```
iljakuklic: Nit: gate under the "testutils" feature?

crates/agglayer-node/src/node.rs
```diff
@@ -276,17 +276,48 @@ impl Node {
...
+                .private_networks
+                .as_ref()
+                .map(|n| n.networks.iter().cloned().collect::<HashSet<_>>());
+            let private2 = private.clone();
+            (
+                // Is the incoming network public?
+                move |incoming| match &private {
+                    Some(private) => private.contains(&incoming),
+                    None => true,
+                },
```
iljakuklic: Nit: Could [Option::is_none_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or) be used here?

crates/agglayer-node/src/node.rs
```diff
@@ -276,17 +276,48 @@ impl Node {
...
+        let (public_networks, private_networks) = {
+            let private = config
+                .private_networks
+                .as_ref()
+                .map(|n| n.networks.iter().cloned().collect::<HashSet<_>>());
+            let private2 = private.clone();
+            (
+                // Is the incoming network public?
+                move |incoming| match &private {
+                    Some(private) => private.contains(&incoming),
```
iljakuklic: Should this condition be negated?
```suggestion
                // Is the incoming network public?
                move |incoming| match &private {
                    Some(private) => !private.contains(&incoming),
```

crates/agglayer-node/src/node.rs
```diff
@@ -276,17 +276,48 @@ impl Node {
...
-            })?;
+        let private_grpc_router = match private_networks {
+            None => None,
+            Some(pn) => Some(
+                agglayer_grpc_api::Server::with_config(config.clone(), rpc_service.clone(), pn)
+                    .build()
+                    .map_err(|err| {
+                        error!("Failed to build private gRPC router: {}", err);
+                        err
+                    })?,
```
iljakuklic: Nit: use `inspect_err` here?

crates/agglayer-node/src/node.rs
```diff
@@ -276,17 +276,48 @@ impl Node {
...
-        let grpc_router = agglayer_grpc_api::Server::with_config(config.clone(), rpc_service)
-            .build()
-            .map_err(|err| {
-                error!("Failed to build gRPC router: {}", err);
-                err
-            })?;
+        let private_grpc_router = match private_networks {
+            None => None,
+            Some(pn) => Some(
+                agglayer_grpc_api::Server::with_config(config.clone(), rpc_service.clone(), pn)
```
iljakuklic: Warn here if `private_networks` is an empty vec?

crates/agglayer-node/src/node.rs
```diff
@@ -276,17 +276,48 @@ impl Node {
...
+                    })?,
+            ),
+        };
+        let public_grpc_router =
+            agglayer_grpc_api::Server::with_config(config.clone(), rpc_service, public_networks)
+                .build()
+                .map_err(|err| {
+                    error!("Failed to build public gRPC router: {}", err);
+                    err
+                })?;
```
iljakuklic: Nit: inspect_err. Also clippy might complain about the formatter string.

crates/agglayer-jsonrpc-api/src/testutils.rs
```diff
@@ -125,6 +133,14 @@ impl TestContext {
         }
         config.rpc.readrpc_port = addr.port();
         config.rpc.admin_port = admin_addr.port();
+        if let Some(private_networks) = config.private_networks.as_mut() {
+            let private_addr = next_available_addr();
+            private_networks.host = match private_addr.ip() {
+                IpAddr::V4(ip) => ip,
+                IpAddr::V6(_) => Ipv4Addr::new(127, 0, 0, 1),
```
iljakuklic, line 140: This mapping of any ipv6 address to localhost is a bit weird. What's the reasoning here? Could it be an error instead?

crates/agglayer-config/src/lib.rs
Freyskeyd: This config is about the gRPC configuration, right ?
Maybe we could split this into two things:

- `grpc_internal` which uses a `Option<GrpcConfig>`
- `prioxed_networks` which defines the network that are behind a proxy

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
```diff
@@ -54,6 +63,19 @@ where
             None => return Err(tonic::Status::invalid_argument("Missing certificate")),
         };
 
+        if !(self.allowed_networks)(certificate.network_id) {
```
Freyskeyd, line 67: I must say that I don't find this really explicit `self.allowed_networks.contains(certificate.network_id)` could be more clear.

Freyskeyd, line 67: Can we have a comment explaining what this does?

crates/agglayer-node/src/node.rs
```diff
@@ -276,17 +276,48 @@ impl Node {
         .start()
         .await?;
 
+        // List the public vs. private networks.
+        let (public_networks, private_networks) = {
```
Freyskeyd: At some point, we could explore using a tower layer on top of the `CertificateSubmission` service for this

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
iljakuklic: > Can we have a comment explaining what this does?

Yea took me a while to parse too :sweat_smile:

crates/agglayer-config/src/lib.rs
Ekleog-Polygon: That was my first thought too, but then I looked at the current implementation, and allowing multiple network sets would require a few refactorings, that'd be incompatible with the current timeline :smiling_face_with_tear:

crates/agglayer-node/src/node.rs
Ekleog-Polygon: Good catch! I hadn't known of this function

crates/agglayer-node/src/node.rs
Ekleog-Polygon: Good point, that was a copy-paste of the previous code and I hadn't thought more about it :smile:

crates/agglayer-node/src/node.rs
Ekleog-Polygon: Added the warning above, at the place where we still have the list of private networks as a Vec instead of a callback

crates/agglayer-jsonrpc-api/src/testutils.rs
Ekleog-Polygon: No real reasoning, I hesitated with doing the same as non-private addresses (ie. bind on 0.0.0.0 if the returned address is ipv6), but TBH IMO this should require quite a bit of change — in particular, why do we even take host/port as separate, instead of just parsing a SocketAddr? And why do we not allow grpc, readrpc and admin ports listening on different interfaces?

IMO we'll have to refactor this a bit anyway; I'll open an issue about that soon

crates/agglayer-config/src/lib.rs
Ekleog-Polygon: The downside about that would be exactly the, it wouldn't be compatible with the future improvements discussed in https://github.com/agglayer/agglayer/pull/877#discussion_r2183949233 :confused:

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
Ekleog-Polygon: The reason for not using this is, that in one case we need `allowed.contains()` and in the other `!allowed.contains()` — we don't want to list all the networks that could connect to us in the configuration file.

That's the reason why I went with a closure — otherwise, the solution would be to have both `private_networks: HashSet` and `is_private_endpoint: bool`, and have the condition be `if is_private_endpoint == private_networks.contains(certificate.network_id)`.

That'd work too for now, but would also not be compatible with future improvements like https://github.com/agglayer/agglayer/pull/877#discussion_r2183949233 ; if we ever want to have multiple private networks that are each on a separate port so that one private aggsender can't spoof being another private aggsender.

WDYT?

crates/agglayer-node/src/node.rs
Ekleog-Polygon: That's definitely true! We could likely add a new tower layer, maybe with that layer directly on top of `AgglayerService` rather than the individual grpc/json-rpc service; and then we'd wrap differently the main `AgglayerService` depending on which endpoint we send it to.

That said, for the sake of time I'd argue we should land this without the tower service, and see later if we want to spend more effort on this filtering, or on the contrary if we want to discard that feature with future changes.

WDYT?

crates/agglayer-jsonrpc-api/src/testutils.rs
Ekleog-Polygon: Opened [Tech debt: agglayer-node host/port config](https://github.com/agglayer/agglayer/issues/879#top)#879

iljakuklic (approved):
Looks good for now. A tower service and a bit more principled filtering mechanism with multiple "sets" of private networks would be nice at some point.

crates/agglayer-config/src/private_networks.rs
```diff
@@ -0,0 +1,52 @@
...
+        Self {
+            networks,
+            // Invalid host/port, to be overridden by the test context
+            grpc_port: u16::MAX,
+            host: Ipv4Addr::new(255, 255, 255, 255),
+        }
+    }
+}
+
+fn deserialize_grpc_port<'de, D>(deserializer: D) -> Result<u16, D::Error>
```
atanmarko: Isn't there default implementation for `u16` de-serialization that would work out of the box?

crates/agglayer-config/src/private_networks.rs
```diff
@@ -0,0 +1,52 @@
...
+where
+    D: Deserializer<'de>,
+{
+    let port = u16::deserialize(deserializer)?;
+
+    Ok(from_env_or_default("AGGLAYER_PRIVATE_GRPC_PORT", port))
+}
+
+/// Get an environment variable or a default value if it is not set.
+fn from_env_or_default<T: FromStr>(key: &str, default: T) -> T {
```
atanmarko: There is already an implementation of this function in the `crates/agglayer-config/src/rpc.rs`. Move it to some common place and reuse it.

crates/agglayer-config/src/lib.rs
Freyskeyd: Keep in mind that this new gRPC server will be removed as soon as we've signature improvement on the aggkit side, I'm not expecting to have a long-living corner case for network again.

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
atanmarko: @Ekleog-Polygon Could it be:
```
pub trait AllowedNetworks {
    fn contains(&self, network_id: NetworkId) -> bool;
or 
    fn is_allowed(&self, network_id: NetworkId) -> bool;
}
```
and 
```
 AllowedNetworksCb: AllowedNetworks + Send + Sync + 'static,
```
I would prefer it mostly for the reason that call `(self.allowed_networks)(certificate.network_id)` looks a bit odd.

Freyskeyd: Yeah, with the comment that you added it makes more sense.
But don't overthink the fact that we'll allow many private networks, it's not the goal agglayer-node to support customized behaviour in the long run, it aims at providing a standard on which agg chain comply.
If they can't, then it's an adaptation on aggkit side that need to be made.

crates/agglayer-node/src/node.rs
Freyskeyd: Completely agree, if we ever have to add more filtering or checks on our entrypoint, we could explore using tower more often. But yes, in this case, and due to time constraint, I'm fine having this that way :)

crates/agglayer-node/src/node.rs
```diff
@@ -295,25 +325,47 @@ impl Node {
...
+        let private_grpc_listener = match config.private_grpc_addr() {
+            None => None,
+            Some(addr) => Some(tokio::net::TcpListener::bind(addr).await?),
+        };
         let admin_listener = tokio::net::TcpListener::bind(config.admin_rpc_addr()).await?;
         info!(on = %config.readrpc_addr(), "ReadRPC listening");
-        info!(on = %config.grpc_addr(), "gRPC listening");
+        info!(on = %config.public_grpc_addr(), "Public gRPC listening");
+        if let Some(on) = config.private_grpc_addr() {
+            info!(%on, "Private gRPC listening");
```
atanmarko: You can probably print list of private networks here.

crates/agglayer-node/src/node.rs
```diff
@@ -295,25 +325,47 @@ impl Node {
...
 
         let readrpc_server = axum::serve(readrpc_listener, readrpc_router)
             .with_graceful_shutdown(cancellation_token.clone().cancelled_owned());
 
-        let grpc_server = axum::serve(grpc_listener, grpc_router)
+        let public_grpc_server = axum::serve(public_grpc_listener, public_grpc_router)
             .with_graceful_shutdown(cancellation_token.clone().cancelled_owned());
 
+        let private_grpc_server = private_grpc_listener.map(|listener| {
+            // Both options are Some iff the private_networks section is configured.
```
atanmarko:
```suggestion
            // Both parameters are existing if the `private_networks` section is configured.
```

crates/agglayer-config/src/private_networks.rs
Ekleog-Polygon: There is, but it wouldn't allow us to override it with an environment variable — this is a copy-paste of the code we have in the rpc.rs file neighbouring

crates/agglayer-config/src/private_networks.rs
Ekleog-Polygon: Good point, as mentioned right above this was a copy-paste :sweat_smile: Now fixed!

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
Ekleog-Polygon: @atanmarko It would definitely be possible! But it's also quite a bit more boilerplate to define it, especially considering it's code we hope to be temporary

@Freyskeyd Makes sense! I'll consider the conversation as resolved for now, to be reevaluated when we have more clarity on the next steps then :smile:

crates/agglayer-node/src/node.rs
Ekleog-Polygon: :+1: adjusted, I just left `iff` for the standard "if and only if"

atanmarko (dismissed):

iljakuklic (dismissed):

Freyskeyd (approved):
there is a mix between `private` and `proxied` in some place, but overall looks good to me

atanmarko (approved):

iljakuklic (approved):

PR #882, opened 2025-07-07
feat: handle optional extra certificate signer

hadjiszs:
Accept extra certificate signer in the gRPC query metadata.  
The signature is on the same message as the one in the certificate.  
The signer has to be the one specified in the configuration.

CONFIG-CHANGE: Accept one extra signer for a given network id.  
  The configuration variable is `[extra-certificate-signer]`.  
  Set as `1337 = "0xabcdefabcdefabcdefabcdefabcdefabcdefabcd"`  
  to set an expected extra signer for the network id `1337`.


crates/agglayer-config/src/lib.rs
```diff
@@ -14,6 +14,7 @@ use shutdown::ShutdownConfig;
 use url::Url;
 
 pub use self::telemetry::TelemetryConfig;
+use crate::rate_limiting::NetworkId;
```
atanmarko: We already have `NetworkId` defined in the `interop` , could we use that global definition?

crates/agglayer-config/src/lib.rs
```diff
@@ -125,6 +126,12 @@ pub struct Config {
 
     #[serde(default, skip_serializing_if = "crate::is_default")]
     pub grpc: GrpcConfig,
+
+    /// Extra Certificate signer per network.
+    #[serde(default)]
+    #[serde_as(as = "HashMap<DisplayFromStr, _>")]
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub extra_certificate_signer: HashMap<NetworkId, Address>,
```
atanmarko: As discussed, some tests would be needed here, also to demonstrate the users the toml/json format of this config parameter.

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
```diff
@@ -37,6 +41,24 @@ where
...
+                })?;
+
+                let sig = Signature::from_str(sig_str).map_err(|e| {
+                    tonic::Status::invalid_argument(format!("invalid hex in signature: {e}"))
+                })?;
+
+                Some(sig)
+            } else {
+                None
+            }
```
atanmarko: You can also write it this way. It really depends from the taste, feel free to ignore this suggestion.

```suggestion
                    request
                .metadata()
                .get(GRPC_METADATA_NAME_EXTRA_CERTIFICATE_SIGNATURE)
                .map(|value| {
                    value
                        .to_str()
                        .map_err(|e| {
                            tonic::Status::invalid_argument(format!(
                                "invalid signature encoding: {e}"
                            ))
                        })
                        .and_then(|sig_str| {
                            Signature::from_str(sig_str).map_err(|e| {
                                tonic::Status::invalid_argument(format!(
                                    "invalid hex in signature: {e}"
                                ))
                            })
                        })
                })
                .transpose()?
```

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
```diff
@@ -67,3 +76,57 @@ where
...
+        })?;
+
+        let sig = Signature::from_str(sig_str).map_err(|e| {
+            tonic::Status::invalid_argument(format!("invalid hex in signature: {e}"))
+        })?;
+
+        Ok(Some(sig))
+    } else {
+        Ok(None)
+    }
```
atanmarko, line 117: Depending on the taste really (feel free to ignore suggestion), but alternative would be:

```suggestion
    metadata
        .get(GRPC_METADATA_NAME_EXTRA_CERTIFICATE_SIGNATURE)
        .map(|value| {
            value
                .to_str()
                .map_err(|e| {
                    Box::new(tonic::Status::invalid_argument(format!(
                        "invalid signature encoding: {e}"
                    )))
                })
                .and_then(|sig_str| {
                    Signature::from_str(sig_str).map_err(|e| {
                        Box::new(tonic::Status::invalid_argument(format!(
                            "invalid hex in signature: {e}"
                        )))
                    })
                })
        })
        .transpose()
```

crates/agglayer-jsonrpc-api/src/lib.rs
```diff
@@ -204,7 +204,7 @@ where
     }
 
     async fn send_certificate(&self, certificate: Certificate) -> RpcResult<CertificateId> {
-        Ok(self.rpc_service.send_certificate(certificate).await?)
+        Ok(self.rpc_service.send_certificate(certificate, None).await?)
```
atanmarko: Little note about why we don't use extra signatures here would be great.

crates/agglayer-rpc/src/lib.rs
```diff
@@ -327,10 +328,56 @@ where
...
         Ok(())
     }
+
+    /// Verify the extra [`Certificate`] signature.
+    #[instrument(skip(self), level = "debug")]
+    pub(crate) fn verify_extra_cert_signature(
+        &self,
+        certificate: &Certificate,
+        maybe_extra_signer: Option<&H160>,
+        maybe_extra_signature: Option<Signature>,
```
atanmarko: Suggestion (not strong opinion from my side): we haven't used `maybe` on other places where we have options AFAIK, for consistency I wouldn't use them here either. Comment also applies for other `maybe` variables in this PR.

crates/agglayer-jsonrpc-api/src/lib.rs
hadjiszs: yeah 👍

crates/agglayer-rpc/src/lib.rs
hadjiszs: yeah true, I have renamed those

crates/agglayer-config/src/lib.rs
hadjiszs: It requires to implement `FromStr` in interop and doing new tag, however the right interop `NetworkId` is properly used in other crates for this PR, so I moved to u32 only in the config struct same as the already existing `pub proof_signers: HashMap<u32, Address>,` and ` pub full_node_rpcs: HashMap<u32, Url>,`

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
hadjiszs: I feel it requires bit more mental load, but happy to change it if more people prefer the alternative

crates/agglayer-config/src/lib.rs
hadjiszs: added config test with snapshot

crates/agglayer-config/src/lib.rs
```diff
@@ -131,6 +131,12 @@ pub struct Config {
 
     #[serde(default, skip_serializing_if = "crate::is_default")]
     pub grpc: GrpcConfig,
+
+    /// Extra Certificate signer per network.
+    #[serde(default)]
+    #[serde_as(as = "HashMap<DisplayFromStr, _>")]
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub extra_certificate_signer: HashMap<u32, Address>,
```
atanmarko, line 141: One more question, are we sure there would be only one extra certificate signer? Same question for `proof_signers` field.

hadjiszs: Discussed there https://github.com/agglayer/protocol-research/issues/156#issuecomment-3036224148

atanmarko (approved):

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
```diff
@@ -90,3 +98,57 @@ where
         }))
     }
 }
+
+pub(crate) fn get_extra_signature(
+    metadata: &tonic::metadata::MetadataMap,
+) -> Result<Option<Signature>, Box<Status>> {
+    if let Some(value) = metadata.get(GRPC_METADATA_NAME_EXTRA_CERTIFICATE_SIGNATURE) {
```
Ekleog-Polygon, line 105: Do I understand correctly that this is a temporary thing, that we'll be removing as soon as we have a real multisig option? I feel like if we were to keep this long-term we'd want to encode it directly into the Certificate otherwise

crates/agglayer-rpc/src/lib.rs
```diff
@@ -327,10 +328,56 @@ where
 
         Ok(())
     }
+
+    /// Verify the extra [`Certificate`] signature.
+    #[instrument(skip(self), level = "debug")]
```
Ekleog-Polygon: You'll probably want to skip `certificate` here — certificates are pretty huge, and hopefully this is called inside another span that already traces the certificate_id anyway?

crates/agglayer-rpc/src/lib.rs
```diff
@@ -341,14 +388,39 @@ where
...
-                error!(error = %e, hash = hash_string, "Failed to verify the signature of
+                error!(error = %e, hash = hash_string, "Failed to verify the extra signature for the
         certificate {hash}: {e}");
                 CertificateSubmissionError::SignatureError(e)
             })?;
 
+            self.verify_cert_signature(&certificate)
+                .await
+                .map_err(|e| {
+                    error!(error = %e, hash = hash_string, "Failed to verify the signature of
```
Ekleog-Polygon: Nits: here and above
- you don't need to record `hash`, as it's already recorded in the function-level span
- you could name the variable `error` and use `%error` instead of `error = %e`
- errors usually display better as Debug than Display, as Display only shows the latest level of the error, while Debug shows the whole error trace
- if we have the structured data, we don't need to emit it as string in addition to the structured data, as that just makes logs longer

crates/agglayer-types/src/lib.rs
```diff
@@ -670,6 +670,17 @@ impl Certificate {
         }
     }
 
+    /// Retrieve the signer from the provided signature.
+    pub fn signer_from_signature(&self, signature: Signature) -> Result<Address, SignerError> {
+        // TODO: Verify for both commitment versions and return the version
```
Ekleog-Polygon, line 642: Do I understand correctly that currently if we configure an extra signer for a v3 chain, then it'll always reject all certificates? If so, maybe add a comment to that effect on the config struct, so we don't forget it when we write the release notes?

crates/agglayer-config/src/lib.rs
iljakuklic: The name `extra-certificate-signer` seems somewhat unsatisfactory. Not that I have a better idea though.

crates/agglayer-grpc-api/src/certificate_submission_service/mod.rs
hadjiszs: Yeah that's a temporary thing, that was suggested in order to avoid dealing with bothersome Certificate format change which could change once again when we get to the proper solution (zkp or aggregated signature in lieu of the current `certificate.signature`)

crates/agglayer-types/src/lib.rs
hadjiszs: Right now there is no chain which signs the v3 commitment (the aggsender will need to be updated for that)

On our side, we have the migration code to deal with v2 and v3 but only later in the execution path (in the witness generation + in the PP), but on the first signature verification on the client side, we always verify as it was always a v2 commitment, same for the other `certificate.signature`

I'll propagate the comment in the config struct 👍

Ekleog-Polygon: Can we hardcode a reject of signed v3 commitments here somehow? I'm feeling that it'd be better to be sure we're not allowing eg. a v2 signature to by pure chance (or attacker control) allow a different v3 certificate to go through

Ekleog-Polygon (commented):
LGTM! I don't have a better naming for the configuration parameter. Not approving yet because I'm not 100% sure about the commitment signature verification yet

crates/agglayer-types/src/lib.rs
hadjiszs: Ah that's already the case, right below this TODO you can see that the commitment used to retrieve the signer is hardcoded to `V2`, that's actually a copypaste of the `signer()` right below this function which is used today in production

So if the received signature was done on V3, then the public key retrieval would fail.

crates/agglayer-jsonrpc-api/src/lib.rs
```diff
@@ -216,7 +216,16 @@ where
...
+
+        let extra_signature: Option<Signature> = {
+            // NOTE: Extra certificate signature is not supported on the json rpc api
+            None
+        };
+
+        Ok(self
+            .rpc_service
+            .send_certificate(certificate, extra_signature)
+            .await?)
```
Freyskeyd: This could be enough right? As the `send_certificate` expose the type for `extra_signature`
```suggestion
        // NOTE: Extra certificate signature is not supported on the json rpc api
        let extra_signature = None;

        Ok(self
            .rpc_service
            .send_certificate(certificate, extra_signature)
            .await?)
```

crates/agglayer-rpc/src/lib.rs
```diff
@@ -325,10 +326,56 @@ where
 
         Ok(())
     }
+
+    /// Verify the extra [`Certificate`] signature.
+    #[instrument(skip(self, certificate), level = "debug")]
```
Freyskeyd: Do we really want to add the `extra_signer` and `extra_signature` to the trace's fields?

crates/agglayer-rpc/src/lib.rs
```diff
@@ -325,10 +326,56 @@ where
...
+                // Retrieve the signer from the extra signature
+                let retrieved_extra_signer = certificate
+                    .signer_from_signature(extra_signature)
+                    .map_err(SignatureVerificationError::from_signer_error)?
+                    .into_array()
+                    .into();
+
+                // Verify that the signature is performed by the expected authority
+                if retrieved_extra_signer != expected_extra_signer {
+                    return Err(SignatureVerificationError::InvalidExtraSignature {
```
Freyskeyd, line 350: Do we need an error log here, or do we expect to log this at the call site?

crates/agglayer-rpc/src/lib.rs
```diff
@@ -325,10 +326,56 @@ where
...
+            // Extra signature is expected but missing
+            (Some(&expected_signer), None) => {
+                return Err(SignatureVerificationError::MissingExtraSignature {
+                    network_id: certificate.network_id,
+                    expected_signer,
+                });
+            }
+            // Extra signature provided but not required
+            (None, Some(_)) => {
+                warn!("Extra signature provided but not required for this network");
```
Freyskeyd:
```suggestion
                warn!("Unexpected extra signature provided");
```

crates/agglayer-rpc/src/lib.rs
```diff
@@ -339,14 +386,40 @@ where
...
         self.validate_pre_existing_certificate(&certificate).await?;
-        self.verify_cert_signature(&certificate)
-            .await
-            .map_err(|e| {
-                error!(error = %e, hash = hash_string, "Failed to verify the signature of
-        certificate {hash}: {e}");
-                CertificateSubmissionError::SignatureError(e)
+
+        // Verify the involved signatures
+        {
```
Freyskeyd: I don't think you need a block here

crates/agglayer-types/src/lib.rs
Ekleog-Polygon: Right! What I mean is, is there a risk that a malicious actor could take a v2 commitment signature by aggsender, but drop the certificate and instead push a v3 commitment signature that'd have the same signature (because we're hardcoding commitment version to v2 here)?

If so it could allow an attacker to force-upgrade a chain to v3, which'd lead to a DoS

hadjiszs: >and instead push a v3 commitment signature that'd have the same signature

sorry not entirely to get it but if the question is whether a v2 commitment can be equal to a v3 commitment, then no it's not possible, hashes would necessarily be different, they aren't computed with the same number of fields, so even if the common values are equal, then the hashes will be different, and so the signature will be different too

Ekleog-Polygon: Got it, so we're all good. Thank you for the clarification!

crates/agglayer-rpc/src/lib.rs
hadjiszs: I have skipped them 👍

crates/agglayer-jsonrpc-api/src/lib.rs
hadjiszs: yeah, I put this as a block to emphasize the NOTE but I guess inline NOTE is enough

crates/agglayer-rpc/src/lib.rs
hadjiszs: >or do we expect to log this at the call site?

yeah it is logged by the caller

crates/agglayer-rpc/src/lib.rs
```diff
@@ -325,10 +326,59 @@ where
 
         Ok(())
     }
+
+    /// Verify the extra [`Certificate`] signature.
+    #[instrument(
+        skip(self, certificate, extra_signer, extra_signature),
```
Freyskeyd: nit:

```suggestion
        skip_all,
```

Ekleog-Polygon (approved):

Ekleog-Polygon (approved):

PR #889, opened 2025-07-10
chore: clean up certificate task

Ekleog-Polygon:
Handle review comment from https://github.com/agglayer/agglayer/pull/819#discussion_r2111744145

### DIFF
--- a/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
@@ -276,12 +276,14 @@ where
         .await?;
 
         let settlement_tx_hash = settlement_submitted.await.map_err(recv_err)??;
+        #[cfg(feature = "testutils")]
         fail::fail_point!("certificate_task::process_impl::about_to_record_candidate");
         self.header.settlement_tx_hash = Some(settlement_tx_hash);
         self.state_store
             .update_settlement_tx_hash(&certificate_id, settlement_tx_hash)?;
-        self.header.status = CertificateStatus::Candidate; // No set_status: update_settlement_tx_hash already updates the status in
-                                                           // database
+        // No set_status: update_settlement_tx_hash already updates the status in the
+        // database
+        self.header.status = CertificateStatus::Candidate;
         debug!(settlement_tx_hash = ?self.header.settlement_tx_hash, "Submitted certificate for settlement");
 
         self.process_from_candidate().await
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,6 +115,7 @@ dependencies = [
 name = "agglayer-certificate-orchestrator"
 version = "0.1.0"
 dependencies = [
+ "agglayer-certificate-orchestrator",
  "agglayer-clock",
  "agglayer-config",
  "agglayer-contracts",
--- a/crates/agglayer-certificate-orchestrator/Cargo.toml
+++ b/crates/agglayer-certificate-orchestrator/Cargo.toml
@@ -6,13 +6,25 @@ edition.workspace = true
 [lints]
 workspace = true
 
+[features]
+default = []
+testutils = [
+    "dep:agglayer-test-suite",
+    "dep:fail",
+    "dep:mockall",
+    "agglayer-clock/testutils",
+    "agglayer-config/testutils",
+    "agglayer-storage/testutils",
+    "agglayer-types/testutils",
+]
+
 [dependencies]
 async-trait.workspace = true
 anyhow.workspace = true
 arc-swap.workspace = true
 buildstructor.workspace = true
 alloy.workspace = true
-fail.workspace = true
+fail = { workspace = true, optional = true }
 parking_lot.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
@@ -21,9 +33,11 @@ tokio-util.workspace = true
 tracing.workspace = true
 futures-util = "0.3.31"
 reqwest.workspace = true
-agglayer-contracts.workspace = true
 agglayer-clock.workspace = true
+agglayer-config = { workspace = true, optional = true }
+agglayer-contracts.workspace = true
 agglayer-storage.workspace = true
+agglayer-test-suite = { workspace = true, optional = true }
 agglayer-types.workspace = true
 pessimistic-proof.workspace = true
 
@@ -31,11 +45,10 @@ pessimistic-proof.workspace = true
 mockall = { workspace = true, optional = true }
 
 [dev-dependencies]
-agglayer-clock = { workspace = true, features = ["testutils"] }
-agglayer-config = { workspace = true, features = ["testutils"] }
-agglayer-storage = { workspace = true, features = ["testutils"] }
-agglayer-test-suite = { path = "../agglayer-test-suite" }
-agglayer-types = { workspace = true, features = ["testutils"] }
+agglayer-certificate-orchestrator = { workspace = true, features = [
+    "testutils",
+] }
+
 async-trait.workspace = true
 alloy.workspace = true
 lazy_static.workspace = true
@@ -44,7 +57,3 @@ rstest.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 test-log.workspace = true
-
-[features]
-default = []
-testutils = ["dep:mockall"]
--- a/crates/agglayer-certificate-orchestrator/src/network_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/network_task.rs
@@ -385,6 +385,7 @@ where
                         settlement_submitted_notifier
                             .send(result)
                             .map_err(|_| Error::InternalError("Certificate notification channel closed".into()))?;
+                        #[cfg(feature = "testutils")]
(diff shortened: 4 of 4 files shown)


crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -156,156 +144,203 @@ where
...
+                height,
+                certificate_id,
+            })
+            .await
+            .map_err(send_err)?;
+
+        self.process_from_proven().await
+    }
+
+    async fn recompute_state(&self) -> Result<(), CertificateStatusError> {
```
iljakuklic, line 166: Nit: The `recompute_state` method is placed between `process_from_pending` and `provess_from_proven`. Could it be moved somewhere out of the way so the `process_form_*` methods are next to each other in the execution order?

Ekleog-Polygon: Here I had placed it in the same order as when everything was linear: the `recompute_state` function is actually an alternative, currently it's either `process_from_pending` or `recompute_state`, and then the other ones.

Would you like it better with `recompute_state` eg. between `process_impl` and `process_from_pending`, even though it wouldn't be the first step of the processing?

(I really dislike the need for `recompute_state`, not even counting the fact that it's technically wrong, but until I get some time to seriously get to the storage changes we'll have to deal with it :smiling_face_with_tear: )

iljakuklic: > Here I had placed it in the same order as when everything was linear: the `recompute_state` function is actually an alternative, currently it's either `process_from_pending` or `recompute_state`, and then the other ones.
> 
> Would you like it better with `recompute_state` eg. between `process_impl` and `process_from_pending`, even though it wouldn't be the first step of the processing?

Yea I'd slightly prefer the "clean" scenario to be readable top to bottom. The `recompute_state` method could also be inserted before `process_from_candidate` so its relative positioning with respect to the `process_from_*` methods is already ambiguous.

> (I really dislike the need for `recompute_state`, not even counting the fact that it's technically wrong, but until I get some time to seriously get to the storage changes we'll have to deal with it 🥲 )

Agreed. I'd say it's one more reason to place it aside somewhere.

This is by no means a showstopper, just a mild style preference.

Ekleog-Polygon: Got it! I just moved it up — didn't understand what you meant by `process_from_candidate` (it's being used just before `process_from_proven` too), so I went with this one :smile:

iljakuklic (approved):
+1 Looks good.

Some general points:

1. This interferes with #874. When one of the PRs is merged, the other one will need to be updated.

2. Would be nice to abstract the pattern of sending a request to another task and waiting for a reply on an oneshot channel so it looks a bit more like an async function call. That is for another day.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -156,156 +144,203 @@ where
             self.pending_store.remove_generated_proof(&certificate_id)?;
         }
 
-        // First, prove the certificate (or recompute just the new state if it's already
-        // proven)
-        if self.header.status < CertificateStatus::Proven {
```
polykeith, line 159: Just a small comment here -- changing this comparison to be strictly a match statement is absolutely the right move, since me being a newcomer here, this would immediately raise the question "what certificate status are 'less than' `Proven`?", and obviously the ideal here is to have code that is self-explanatory, instead of assuming the reader has all the context.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -154,146 +142,193 @@ where
...
-            // Retrieve local network state
-            trace!("Retrieving local network state");
-            let (response, state) = oneshot::channel();
-            self.send_to_network_task(NetworkTaskMessage::GetLocalNetworkStateBeforeHeight {
-                height,
-                response,
-            })
-            .await?;
-            let state = state.await.map_err(recv_err)??;
+        match &self.header.status {
```
polykeith, line 145: This can be done later (or maybe it's already done?), but it would be really helpful to have unit tests on all `process_from_*` methods, so that we assert that these functions would _always_ be progressing the certificate state (or not if it's desired).

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -154,146 +142,193 @@ where
...
+            settlement_submitted_notifier,
+        })
+        .await?;
+
+        let settlement_tx_hash = settlement_submitted.await.map_err(recv_err)??;
+        fail::fail_point!("certificate_task::process_impl::about_to_record_candidate");
+        self.header.settlement_tx_hash = Some(settlement_tx_hash);
+        self.state_store
+            .update_settlement_tx_hash(&certificate_id, settlement_tx_hash)?;
+        self.header.status = CertificateStatus::Candidate; // No set_status: update_settlement_tx_hash already updates the status in
```
atanmarko: Nit: format this comment better, put it above this line.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -154,146 +142,193 @@ where
...
+        );
+        self.send_to_network_task(NetworkTaskMessage::CertificateSettled {
+            height,
+            certificate_id,
+            settled_certificate,
+        })
+        .await?;
+
         if self.header.status != CertificateStatus::Settled {
             return Err(CertificateStatusError::InternalError(
```
atanmarko, line 335: Should we log here error or at least warning?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -154,146 +142,193 @@ where
...
+        let (settlement_submitted_notifier, settlement_submitted) = oneshot::channel();
+        self.send_to_network_task(NetworkTaskMessage::CertificateReadyForSettlement {
+            height,
+            certificate_id,
+            settlement_submitted_notifier,
+        })
+        .await?;
+
+        let settlement_tx_hash = settlement_submitted.await.map_err(recv_err)??;
+        fail::fail_point!("certificate_task::process_impl::about_to_record_candidate");
```
atanmarko, line 280: Could you put this under `testutils` feature, mark with some comment - not great that we are mixing test code with production code.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -154,146 +142,193 @@ where
...
-            let state = state.await.map_err(recv_err)??;
+        match &self.header.status {
+            CertificateStatus::Pending => self.process_from_pending().await,
+            CertificateStatus::Proven => {
+                self.recompute_state().await?;
+                self.process_from_proven().await
+            }
+            CertificateStatus::Candidate => {
+                self.recompute_state().await?;
+                self.process_from_candidate().await
```
atanmarko, line 153: Do we log properly errors that happen here on the level above?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Totally agree! I actually have a todo on my list that is "fuzz the certificate_task", but I have higher-prio stuff on my plate for now :smiling_face_with_tear: 

I just converted it to an actual issue for future reference: [Write fuzzer for the CertificateTask](https://github.com/agglayer/agglayer/issues/892#top)#892

If you're interested in working on it, I'd be happy to pair with you! The last thing I shared in knowledge&learn was actually a fuzzing 101 "class"; I can also find you the link of the recording — don't know how familiar you're with fuzzing yet.

This being said, I think the other option I had suggested on Slack would be more interesting for you as an entry point :smile:

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: All errors are already logged in the `process` call above :smile: 

I'll probably be able to refactor this to use `eyre` as we had discussed and make everything even better some time, but once I'm done with the review comments on the certificate task refactoring, I'll probably start designing storage, so no promises yet

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Good idea! I had copied previous fail points, but `fail` should definitely be a testutils-gated dependency :smile: 

(Well, even better would be if there was a feature inside `fail` itself to disable it like I did in [`reord`](https://crates.io/crates/reord); but I'm too lazy to go and try to push that upstream :sweat_smile: )

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: Yup, everything is being logged in `process`, which calls `process_impl` exactly with the goal of logging errors at a centralized place :smile:

iljakuklic (approved):

PR #897, opened 2025-07-15
feat: backport extra certificate signer

hadjiszs:
Accept extra certificate signer in the gRPC query metadata.  
The signature is on the same message as the one in the certificate.  
The signer has to be the one specified in the configuration.

CONFIG-CHANGE: Accept one extra signer for a given network id.  
  The configuration variable is `[extra-certificate-signer]`.  
  Set as `1337 = "0xabcdefabcdefabcdefabcdefabcdefabcdefabcd"`  
  to set an expected extra signer for the network id `1337`.


iljakuklic (approved):

crates/agglayer-rpc/src/error.rs
```diff
@@ -72,4 +71,25 @@ pub enum SignatureVerificationError<Rpc: Middleware> {
     /// SP1-based Aggchain proof not yet supported.
     #[error("SP1-based Aggchain proof not yet supported")]
     SP1AggchainProofUnsupported,
+
+    /// Signature is missing.
+    #[error("signature not provided")]
+    SignatureMissing,
```
Ekleog-Polygon, line 77: Looks to me like this variant is not actually used anywhere?

atanmarko (approved):

crates/agglayer-rpc/src/error.rs
hadjiszs: I have updated the code with #920 which uses this one normally 👍

atanmarko (approved):

PR #899, opened 2025-07-15
perf: reduce in-memory certificate status size by 10x, improving bincode parsing perf by 17%

Ekleog-Polygon:
Another review comment from the certificate task refactor. The struct size was computed by my IDE, here is the speed delta on the newly-added benchmark:

```
default cert header     time:   [280.21 ns 280.75 ns 281.41 ns]
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe

  default cert header     time:   [231.87 ns 233.42 ns 235.67 ns]
                        change: [-17.389% -16.787% -16.017%] (p = 0.00 < 0.05)
                        Performance has improved.
```


crates/agglayer-storage/src/types/certificate/tests/status.rs
```diff
@@ -8,7 +8,7 @@ type Cse = CertificateStatusError;
 type Pve = ProofVerificationError;
 
 fn err(error: CertificateStatusError) -> CertificateStatus {
-    CertificateStatus::InError { error }
+    CertificateStatus::InError { error: Box::new(error) }
```
iljakuklic: Introduce a helper constructor method on `CertificateStatus` for this?

iljakuklic (dismissed):

iljakuklic (approved):

crates/agglayer-types/src/lib.rs
```diff
@@ -450,6 +448,14 @@ impl fmt::Display for CertificateStatus {
     }
 }
 
+impl CertificateStatus {
+    pub fn error(err: CertificateStatusError) -> Self {
```
polykeith: Would it be even faster if we `#[inline]` this function?

iljakuklic: > Would it be even faster if we `#[inline]` this function?

Probably. But since it's constructing an error, one would hope it will tend to be on the cold path :slightly_smiling_face:.

Ekleog-Polygon: I'd say probably no actually, because AFAIR rust already inlines pretty freely across a workspace. Also, even if it did get things faster, it'd only be building an in-error CertificateStatus manually. Bincode parsing would not actually get faster, because it wouldn't use this function anyway; and in-memory size would not decrease either.

Finally, inlining on cold paths risks making hot paths slower, due to the added icache thrashing.

All this together, I'd lean towards not adding `#[inline]` here :smile:

iljakuklic (approved):

PR #909, opened 2025-07-21
perf: Optimize NetworkState serialization for zero-copy SP1 I/O

sai-deng:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Additions and Changes

Please explain in detail, if applicable, which changes you performed in the PR and link them to issues. (If not applicable you can remove this whole section)

## Breaking changes

Please explain in detail which changes are breaking changes in the PR. (If not applicable you can remove this whole section)

## PR Checklist:

### DIFF
--- a/crates/pessimistic-proof-core/src/local_state/mod.rs
+++ b/crates/pessimistic-proof-core/src/local_state/mod.rs
@@ -350,9 +350,14 @@ mod tests {
         let unaligned = &bytes[1..];
         assert!(NetworkState::from_bytes_zero_copy(unaligned).is_err());
 
-        // Test wrong size
+        // Test wrong size (too small)
         let too_small = &bytes[..bytes.len() - 1];
         assert!(NetworkState::from_bytes_zero_copy(too_small).is_err());
+
+        // Test wrong size (too large)
+        let mut too_large = bytes.clone();
+        too_large.push(0);
+        assert!(NetworkState::from_bytes_zero_copy(&too_large).is_err());
 
         // Test empty data
         assert!(NetworkState::from_bytes_zero_copy(&[]).is_err());


crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -15,6 +15,97 @@ use crate::{
...
+
+        // Reconstruct the nullifier tree
+        let nullifier_tree = NullifierTree::<Keccak256Hasher> {
+            root: agglayer_primitives::Digest::from(self.nullifier_tree_root),
+            empty_hash_at_height: self
+                .nullifier_empty_hash_at_height
+                .map(|h| agglayer_primitives::Digest::from(h)),
+        };
+
+        NetworkState {
```
4l0n50, line 70: I think this this ends up allocating a new network state, no?

crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -15,6 +15,97 @@ use crate::{
...
+        std::mem::size_of::<Self>()
+    }
+
+    /// Check if the given byte slice has the correct size for this struct.
+    pub fn check_size(data: &[u8]) -> bool {
+        data.len() == Self::size()
+    }
+
+    /// Safely transmute a byte slice to this struct.
+    /// This is only safe if the data was originally a NetworkStateZeroCopy.
```
4l0n50: What happens then when `data` is a non-deterministic advice?

crates/pessimistic-proof-core/src/local_state/mod.rs
sai-deng: You're right, to_network_state allocates a new NetworkState . But it’s still zero-copy where it counts: from_bytes borrows the input bytes without copying, and Digest::from just wraps them. The allocations are minimal and needed for a usable NetworkState in SP1.

crates/pessimistic-proof-core/src/local_state/mod.rs
sai-deng: I think it works as intended for trusted advice in proofs; adversarial advice just results in unprovable or invalid executions?

4l0n50: I don't see why it could be unprovable, so this may end up being an attack. Maybe we need to make `from_bytes` safe  and argue that the checks are enough for that, even when datea is arbitrary.

crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -15,6 +15,97 @@ use crate::{
...
+    pub unsafe fn from_bytes(data: &[u8]) -> Option<&Self> {
+        let align = std::mem::align_of::<Self>();
+        let ptr = data.as_ptr() as usize;
+        if Self::check_size(data) && (ptr % align == 0) {
```
4l0n50: Do we also need to check that fields are aligned?

crates/pessimistic-proof-core/src/local_state/mod.rs
sai-deng: For both methods, if data is "structurally valid" but semantically wrong (e.g., invalid hashes or leaf_count), it could lead to incorrect proofs if not caught downstream. The main difference is the unsafe pointer cast in the zero-copy method could theoretically cause UB on malformed data, but our size (3140 bytes) and alignment (4 bytes) checks ensure the input matches the fixed layout of primitives/arrays, preventing that—arbitrary advice just returns None safely. In SP1, bad advice makes the proof unprovable anyway, similar to bincode's Err. Think that's sufficient, or add extra validation?

crates/pessimistic-proof-core/src/local_state/mod.rs
sai-deng: The repr(C) layout ensures all fields are properly aligned—u32 needs 4-byte alignment (covered by the struct’s start check), and the hash arrays ([u8; 32]) are 1-byte aligned, so no extra field checks needed?

sai-deng: An alternative approach for deserializing NetworkStateZeroCopy by using sp1_zkvm::io::read::<u32> to read the exit_tree_leaf_count separately and then using sp1_zkvm::io::read_vec to read the remaining fields (all [u8; 32] arrays) as a single vec.

4l0n50: No, you convinced me with your previous answer

crates/pessimistic-proof-core/src/local_state/mod.rs
4l0n50: Ah yes, I now see that the two checks are enough for `NetworkState`. So I think you can remove the `unsafe` from `from_bytes`

crates/pessimistic-proof-core/src/local_state/mod.rs
```diff
@@ -15,6 +15,97 @@ use crate::{
 
 pub mod commitment;
 
+/// Zero-copy representation of NetworkState for safe transmute.
+/// This struct has a stable C-compatible memory layout.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct NetworkStateZeroCopy {
```
4l0n50, line 23: Why do we need `NetworkStateZeroCopy`? Shouldn't be enough to add repr(C) to NetworkState as well as all its fields?

crates/pessimistic-proof-core/src/local_state/mod.rs
sai-deng: I switched to use bytemuck, which added thousands of cycles but still saved 96,000 cycles than the original method. It is safer and eliminates all unsafe code. What do you think?

crates/pessimistic-proof-core/src/local_state/mod.rs
sai-deng: NetworkState contains complex types (LocalExitTree, LocalBalanceTree, NullifierTree) that have methods and logic - they're not just data. These types can't be made Pod (required for zero-copy serialization) because they're not Copy.

PR #918, opened 2025-07-24
fix: clock L1 reconnection adjustments

iljakuklic:
* Pass the config to `max_interval`
* Increase the time before we give up reconnecting L1
* Add diagnostic logs

CONFIG-CHANGE: `l1.max-reconnection-elapsed-time` was replaced with `l1.total-reconnect-timeout`. Added `reconnect-attempt-interval`.


crates/agglayer/tests/snapshots/config__config_display.snap
```diff
@@ -39,7 +39,8 @@ settlement-timeout = "20m"
 chain-id = 1337
 node-url = "http://zkevm-mock-l1-network:8545/"
 ws-node-url = "ws://zkevm-mock-l1-network:8546/"
-max-reconnection-elapsed-time = "10s"
+reconnect-attempt-interval = "10s"
+total-reconnect-timeout = "10m"
```
iljakuklic, line 43: TODO: Describe this in `CONFIG-CHANGE:`.

iljakuklic:
Will leave this PR here for a while for reference but it is likely going to be replaced by:
* #922

crates/agglayer-clock/src/block.rs
```diff
@@ -372,9 +382,16 @@ impl PubSubConnect for WsConnectWithRetries {
                 // progress when the client is disconnected
                 fail::fail_point!("block_clock::PubSubConnect::try_reconnect::add_delay");
```
atanmarko, line 383: Nit: could you put failpoint under some test cfg?

crates/agglayer-config/src/l1.rs
```diff
@@ -57,6 +63,10 @@ fn default_ws_node_url() -> Url {
     "ws://zkevm-mock-l1-network:8546".parse().unwrap()
 }
 
-const fn default_max_reconnection_elapsed_time() -> Duration {
+const fn default_total_reconnect_timeout() -> Duration {
+    Duration::from_secs(10 * 60)
+}
+
+const fn default_reconnect_attempt_interval() -> Duration {
     Duration::from_secs(10)
```
atanmarko, line 71: Nit: I wonder if this should be higher, as we are already using exponential scale starting from default initial 500 ms until this value?

atanmarko (approved):

crates/agglayer-clock/src/block.rs
```diff
@@ -372,9 +382,16 @@ impl PubSubConnect for WsConnectWithRetries {
...
 
-                Ok(self.0.try_reconnect().await?)
+                let handle = self
+                    .connection
+                    .try_reconnect()
+                    .await
+                    .inspect_err(|err| debug!(?err, "Failed to reconnect to the L1 node"))?;
```
atanmarko, line 389: Nit: this should probably be an info or a warning. We would probably like to know if there are issues with the L1 node communication from the standard logs.

iljakuklic: > Nit: this should probably be an info or a warning. We would probably like to know if there are issues with the L1 node communication from the standard logs.

This refers to every attempt to reconnect. When all attempts fail, there is an `error!` to report that just below.

crates/agglayer-config/src/l1.rs
iljakuklic: > Nit: I wonder if this should be higher, as we are already using exponential scale starting from default initial 500 ms until this value?

This is removed in subsequent PR.

PR #920, opened 2025-07-24
feat: update extra signature and adapt commitment versions

hadjiszs:
Follow up of the first iteration done with #882 

- Update the extra signature commitment on the certificate id and the l1 info tree leaf count
- Adapt the verification of the PP signature to accept either a V2 or a V3 commitment
  - Note that the verification on whether the version is the expected one as per what is settled in the L1 is only done at the witness generation step
- Adapt the tests

### DIFF
--- a/crates/agglayer-types/src/certificate/mod.rs
+++ b/crates/agglayer-types/src/certificate/mod.rs
@@ -116,21 +116,90 @@ impl Certificate {
         }
     }
 
-    /// Retrieve the signer from the provided signature.
-    pub fn signer_from_signature(&self, signature: Signature) -> Result<Address, SignerError> {
-        // TODO: Verify for both commitment versions and return the version
-        let version = CommitmentVersion::V2;
-        let commitment = SignatureCommitmentValues::from(self).commitment(version);
+    /// Computes the commitment used to verify the extra signature on the
+    /// agglayer only.
+    /// The commitment is expected to be on the certificate id and the optional
+    /// l1 info tree leaf count.
+    fn extra_signature_commitment(&self) -> Digest {
+        let certificate_id = self.hash();
 
-        signature
-            .recover_address_from_prehash(&B256::new(commitment.0))
-            .map_err(SignerError::Recovery)
+        match self.l1_info_tree_leaf_count {
+            Some(leaf_count) => keccak256_combine([
+                certificate_id.as_digest().as_slice(),
+                leaf_count.to_le_bytes().as_slice(),
+            ]),
+            None => *certificate_id,
+        }
     }
 
+    /// Verify the extra certificate signature.
+    pub fn verify_extra_signature(
+        &self,
+        expected_signer: Address,
+        signature: Signature,
+    ) -> Result<(), SignerError> {
+        let expected_commitment = self.extra_signature_commitment();
+
+        let retrieved_signer = signature
+            .recover_address_from_prehash(&B256::new(expected_commitment.0))
+            .map_err(SignerError::Recovery)?;
+
+        (expected_signer == retrieved_signer)
+            .then_some(())
+            .ok_or(SignerError::InvalidExtraSignature { expected_signer })
+    }
+
+    /// Verify the signature on the PP commitment and returns the [`Address`]
+    /// upon success.
+    pub fn verify_cert_signature(&self, expected_signer: Address) -> Result<(), SignerError> {
+        let pp_commitment_values = SignatureCommitmentValues::from(self);
+
+        let recovered_expected_signer = match &self.aggchain_data {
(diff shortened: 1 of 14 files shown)


crates/agglayer-types/src/certificate/mod.rs
```diff
@@ -116,21 +116,85 @@ impl Certificate {
...
+    /// The commitment is expected to be on the certificate id and the optional
+    /// l1 info tree leaf count.
+    fn extra_signature_commitment(&self) -> Digest {
+        let certificate_id = self.hash();
 
-        signature
-            .recover_address_from_prehash(&B256::new(commitment.0))
-            .map_err(SignerError::Recovery)
+        match self.l1_info_tree_leaf_count {
+            Some(leaf_count) => keccak256_combine([
```
Freyskeyd, line 127: Is having two different content length a potential problem?

crates/agglayer-types/src/certificate/mod.rs
```diff
@@ -116,21 +116,85 @@ impl Certificate {
...
+                certificate_id.as_digest().as_slice(),
+                leaf_count.to_le_bytes().as_slice(),
+            ]),
+            None => *certificate_id,
+        }
+    }
+
+    /// Verify the extra certificate signature.
+    pub fn verify_extra_signature(
```
Freyskeyd, line 136: This could be called a `Certificate signature` and the legacy signature as `commitment signature`, WDYT?

hadjiszs: hm I dont know if it would help differentiating them given that both signatures are on a `commitment`, and both comes with the `certificate` :/

crates/agglayer-types/src/certificate/mod.rs
hadjiszs: I believe that issues could arise if we hash variable length within one hash, but here we have fixed length, just that we have two different branches to define what should be the fixed length content for the hash

crates/agglayer-types/src/certificate/mod.rs
Freyskeyd: What I was suggesting is that one signature is representing the `certificate` signature using the `certificateId` while the other one is representing the `chain commitment` signature.
I don't have a strong opinion about that, feel free to update or not, or even create a ticket if needed :)

hadjiszs: yeah I created a dedicated ticke here:

- https://github.com/agglayer/agglayer/issues/923

I remember that this naming wasn't clear for @iljakuklic either, and it wasn't straightforward to find good alternatives :/ it might get easier to name things here once we'll get closer to the complete solution with aggchain proof or multisig/threshold signature

crates/agglayer-types/src/certificate/mod.rs
```diff
@@ -116,21 +116,84 @@ impl Certificate {
...
+                        .aggchain_proof_commitment(aggchain_params)
+                        .0,
+                );
+                let recovered = signature
+                    .recover_address_from_prehash(&commitment)
+                    .map_err(SignerError::Recovery)?;
+
+                recovered == expected_signer
+            }
+        };
```
atanmarko, line 186: Could you reuse here other cert function `retrieve_signer`? Or adjust `retrieve_signer`/`verify_cert_signature` so that retrieval of the signer is not re-implemented?

hadjiszs: Hm I gave a shot but then it brings confusion it feels:

```suggestion
        let recovered_expected_signer = match &self.aggchain_data {
            // Verify if one of the commitment version is signed.
            // NOTE: The legitimacy of the version is verified during the witness generation,
            // especially in order to forbid version rollback by the chain.
            AggchainData::ECDSA { signature } => [CommitmentVersion::V3, CommitmentVersion::V2]
                .iter()
                .any(|version| match self.retrieve_signer(Some(version)) {
                    Ok(recovered) => recovered == expected_signer,
                    Err(_) => false,
                }),
            AggchainData::Generic {
                signature,
                aggchain_params,
                ..
            } => expected_signer == self.retrieve_signer(CommitmentVersion::V2),
        };

        recovered_expected_signer
            .then_some(())
            .ok_or(SignerError::InvalidPessimisticProofSignature { expected_signer })
    }
```

Especially the `self.retrieve_signer(CommitmentVersion::V2)` in the `Generic` branch, because there is no commitment version for the generic path

Ideally I would go in refactoring of the `CommitmentVersion` enum to have also `Generic` in it, then we could compute all signture commitment with `SignatureCommitmentValues::commitment()`, but it will change the PP code and so the vkey and I think we dont want that for the phase 2

atanmarko (approved):

PR #922, opened 2025-07-25
refactor: adding more logs and metrics to clock

Freyskeyd:
This PR is adding more logs and metrics to the BlockClock

### DIFF
--- a/crates/agglayer-telemetry/src/clock.rs
+++ b/crates/agglayer-telemetry/src/clock.rs
@@ -40,28 +40,10 @@ lazy_static! {
         .with_description("Current health status of the block clock")
         .build();
 
-    /// Gauge for seconds since last block received (for health monitoring)
-    pub static ref SECONDS_SINCE_LAST_BLOCK: Gauge<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
-        .u64_gauge("seconds_since_last_block")
-        .with_description("Seconds since the last block was received")
-        .build();
-
-    /// Gauge for seconds since last epoch event (for health monitoring)
-    pub static ref SECONDS_SINCE_LAST_EPOCH: Gauge<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
-        .u64_gauge("seconds_since_last_epoch")
-        .with_description("Seconds since the last epoch event was sent")
-        .build();
-
     /// Counter for subscription lag events
     pub static ref SUBSCRIPTION_LAG: Counter<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
-        .u64_counter("subscription_lag_total")
+        .u64_counter("blocks_subscription_lag_total")
         .with_description("Total number of subscription lag events")
-        .build();
-
-    /// Gauge for epoch progress (0.0 to 1.0)
-    pub static ref EPOCH_PROGRESS: Gauge<f64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
-        .f64_gauge("epoch_progress")
-        .with_description("Progress through current epoch (0.0 to 1.0)")
         .build();
 
     /// Counter for connection errors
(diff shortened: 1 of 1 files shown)


crates/agglayer-telemetry/src/clock.rs
```diff
@@ -0,0 +1,144 @@
...
+/// Helper function to record subscription lag
+pub fn record_subscription_lag(lagged_count: u64) {
+    SUBSCRIPTION_LAG.add(lagged_count, &[]);
+}
+
+/// Helper function to record clock shutdown
+pub fn record_clock_shutdown() {
+    CONNECTION_STATUS.record(0, &[]);
+    HEALTH_STATUS.record(0.0, &[]); // Unhealthy status
+}
```
iljakuklic: Nit: Should the `record_*` functions be `#[inline]`?

crates/agglayer-telemetry/src/clock.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub static ref SUBSCRIPTION_LAG: Counter<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
+        .u64_counter("subscription_lag_total")
+        .with_description("Total number of subscription lag events")
+        .build();
+
+    /// Gauge for epoch progress (0.0 to 1.0)
+    pub static ref EPOCH_PROGRESS: Gauge<f64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
+        .f64_gauge("epoch_progress")
+        .with_description("Progress through current epoch (0.0 to 1.0)")
+        .build();
```
iljakuklic: I don't think this is used?

crates/agglayer-telemetry/src/clock.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub static ref SECONDS_SINCE_LAST_EPOCH: Gauge<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
+        .u64_gauge("seconds_since_last_epoch")
+        .with_description("Seconds since the last epoch event was sent")
+        .build();
+
+    /// Counter for subscription lag events
+    pub static ref SUBSCRIPTION_LAG: Counter<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
+        .u64_counter("subscription_lag_total")
+        .with_description("Total number of subscription lag events")
+        .build();
```
iljakuklic: Maybe mention in the name this refers to lag to subscription to *blocks*? Just "subscription" can refer to any number of things.

crates/agglayer-telemetry/src/clock.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub static ref SECONDS_SINCE_LAST_BLOCK: Gauge<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
+        .u64_gauge("seconds_since_last_block")
+        .with_description("Seconds since the last block was received")
+        .build();
+
+    /// Gauge for seconds since last epoch event (for health monitoring)
+    pub static ref SECONDS_SINCE_LAST_EPOCH: Gauge<u64> = global::meter(AGGLAYER_CLOCK_OTEL_SCOPE_NAME)
+        .u64_gauge("seconds_since_last_epoch")
+        .with_description("Seconds since the last epoch event was sent")
+        .build();
```
iljakuklic: These look unused too.

crates/agglayer-telemetry/src/clock.rs
```diff
@@ -0,0 +1,144 @@
...
+
+/// Helper function to record current epoch
+pub fn record_current_epoch(epoch_number: u64) {
+    CURRENT_EPOCH.record(epoch_number, &[]);
+}
+
+/// Helper function to record subscription lag
+pub fn record_subscription_lag(lagged_count: u64) {
+    SUBSCRIPTION_LAG.add(lagged_count, &[]);
+}
```
iljakuklic: Is there a way to detect how many blocks lagged on each lag occurrence or should there be a separate metric for that, e.g. `LAG_OCCURRENCES` and `LAG_TOTAL_BLOCKS` or something like that?

iljakuklic (commented):
Some comments

crates/agglayer-telemetry/src/clock.rs
Freyskeyd: I think that's what the gauge is doing, it registers the number of lagged blocks

crates/agglayer-clock/src/block.rs
```diff
@@ -224,34 +265,50 @@ where
                     break;
                 }
                 block_result = Self::recv_block(&mut stream) => {
-                    let block = block_result?;
+                    let block = block_result.map_err(|e| {
+                        error!(error = %e, "Failed to receive block from stream");
```
atanmarko, line 269:
```suggestion
                        error!(error = ?e, "Failed to receive block from stream");
```

crates/agglayer-clock/src/block.rs
```diff
@@ -297,6 +358,9 @@ where
             .fetch_add(1, Ordering::Release)
             .checked_add(1)
         {
+            // Record block processing metrics
+            agglayer_telemetry::clock::CURRENT_BLOCK_HEIGHT.record(current_block, &[]);
```
atanmarko, line 362: Nit: why not using inline function e.g. `record_current_block_height()` like for other metric values?

crates/agglayer-clock/src/block.rs
```diff
@@ -370,28 +443,34 @@ impl PubSubConnect for WsConnectWithRetries {
...
-                    .inspect_err(|err| debug!(?err, "Failed to reconnect to the L1 node"))?;
-
-                Ok(handle)
+                    .inspect(|_| {
+                        info!("Successfully reconnected to L1 WebSocket");
+                        agglayer_telemetry::clock::record_connection_established();
+                    })
+                    .inspect_err(|e| {
+                        warn!(error = %e, "Failed to reconnect to L1 WebSocket");
+                        agglayer_telemetry::clock::record_connection_lost();
```
atanmarko, line 470: Nit: `record_connection_lost` - maybe just a comment/doc note that this function does not always mean that connection was lost after it was present, but inability to establish connection. Maybe `record_no_connection` would be more appropriate name for all scenarious?

crates/agglayer-clock/src/block.rs
iljakuklic: addressed in #930

crates/agglayer-clock/src/block.rs
iljakuklic: addressed in #930

crates/agglayer-clock/src/block.rs
iljakuklic: Hmm, why you suggest this? Not quite sure why it would be better to use `Debug` over `Display`.

PR #924, opened 2025-07-25
fix: add a L1 reconnection timeout

iljakuklic:
CONFIG-CHANGE: Introduced `l1.connect-attempt-timeout` to limit the amount of time the node waits for the L1 websocket connection to be established.


crates/agglayer-clock/src/block.rs
```diff
@@ -457,18 +461,25 @@ impl PubSubConnect for WsConnectWithRetries {
...
-                    })?)
+                let handle =
+                    tokio::time::timeout(Duration::from_secs(3), self.connection.connect())
+                        .await
+                        .map_err(|_| {
+                            let err = Box::new(ConnectionTimeout);
+                            let err = alloy::transports::TransportErrorKind::Custom(err);
+                            err.into()
+                        })
+                        .and_then(|res| res)
```
iljakuklic: Note: this is `Result::flatten` but it is unstable.

crates/agglayer-clock/src/block.rs
```diff
@@ -457,18 +461,25 @@ impl PubSubConnect for WsConnectWithRetries {
...
-                    .inspect(|_| {
-                        info!("Successfully reconnected to L1 WebSocket");
-                        agglayer_telemetry::clock::record_connection_established();
-                    })
-                    .inspect_err(|e| {
-                        warn!(error = %e, "Failed to reconnect to L1 WebSocket");
-                        agglayer_telemetry::clock::record_connection_lost();
-                    })?)
+                let handle =
+                    tokio::time::timeout(Duration::from_secs(3), self.connection.connect())
```
iljakuklic: TODO: make configurable

Issue #904, opened 2025-07-18
Improvements to `agglayer-clock`


iljakuklic:
A bunch of improvements to the epoch clock implementation seem possible.

## Streamlining external interface

The `ClockRef` type exposes a method to query the block height. It is not used anywhere (only internally) but puts constraints on the implementation. The time-based clock has to provide the block height although it does not and does not need to know it. Subsequently, the time-based clock could be simplified accordingly.

## Block clock initialization

There is a oneshot channel that notifies the caller that the intialization of the clock is complete. Explore if the initialization could instead be performed on the current task and only run the regular clock workload on its dedicated task. That should eliminate the need for the channel and make the control flow a bit simpler.

## Use a lighter channel to notify about epoch changes

The epoch changes are currently communicated by a broadcast channel with the capacity of 100. Since the epoch updates are monotonic, it seems sufficient to get the latest value. A much lighter [`watch` channel](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) provides exactly that.

The watch channel does not offer a queue of past values or detection if some values have been skipped. However, it seems the clients are only interested in the latest value for most part. Skipped epochs can be detected by comparing a new value with a previously received value.

## Use a lighter channel to listen for block updates

Alloy uses a broadcast channel to present a stream of blocks, so switching to `notify` is not possible here. It does, however, offer a way to set the channel capacity, which is 16 by default. Even capacity of 1 should be good enough in this case, for similar reasons as described in the previous section.
 
## Related

* Noticed while developing #902 
* One more improvement #375

PR #930, opened 2025-07-29
feat: block clock adjustments

### LINKED ISSUE #919: Clock L1 reconnection failures
The block-based clock seems to have issues with reconnecting to the L1 node after a lost connection.

The reconnection backoff mechanism appears to be misconfigured to give up after 10s (by default), which seems rather short.

iljakuklic:
Multiple changes related to block clock:

* Change L1 reconnection logic to rely on alloy built-in mechanism. This mechanism is currently not configurable, so unfortunately, we cannot change it apart from hacks.
* Remove custom reconnection mechanism with backoff. This interferes with the default alloy mechanism and makes the behaviour more convoluted.
* Introduce a timeout on an attempt to establish an L1 connection.
* Add more logging and metrics.
* Add clock documentation.
* Tests cleanup.

Related:
* fixes #919 
* replaces #918 
* replaces #922
* replaces #924
* replaces #925

CONFIG-CHANGE: Removed `max-reconnection-elapsed-time`. Alloy does not allow us to configure this.

CONFIG-CHANGE: Introduced `connect-attempt-timeout` to limit how long a L1 reconnection attempt takes.


crates/agglayer-clock/src/block.rs
```diff
@@ -219,34 +262,50 @@ where
                     break;
                 }
                 block_result = Self::recv_block(&mut stream) => {
-                    let block = block_result?;
+                    let block = block_result.map_err(|e| {
+                        error!(error = %e, "Failed to receive block from stream");
```
atanmarko: @iljakuklic Continuing discussion from [here](https://github.com/agglayer/agglayer/pull/922#discussion_r2240553249). I think in general it is always useful to show debug printout of the errors in the traces, to display all the information related to this error (e.g. source error chain, or some data missed in the display implementation). In this particular case, this is relevant for `BlockClockError::Transport` source error.

crates/agglayer-clock/src/block.rs
```diff
@@ -152,35 +166,55 @@ where
...
-        debug!("Current L1 Block number: {}", self.latest_seen_block);
         let provider = self.provider.clone();
 
         // Subscribe to the L1 Block stream.
-        let mut stream = provider
-            .subscribe_blocks()
-            .await
-            .map_err(|_| BlockClockError::SubscribeBlocks)?;
+        let mut stream = provider.subscribe_blocks().await.map_err(|e| {
+            error!(error = %e, "Failed to subscribe to L1 block stream");
```
atanmarko: Same question about debug display of errors.

crates/agglayer-clock/src/block.rs
```diff
@@ -219,34 +262,50 @@ where
...
+                        block_hash = %block.hash,
+                        timestamp = block.timestamp,
+                        blocks_to_process = block.number - self.latest_seen_block,
+                        "Received new L1 block"
                     );
 
                     // Overwrite the block number to simulate an overflow
                     // This is used for testing purposes only and doesn't affect the production
                     // code.
                     fail::fail_point!("block_clock::BlockClock::run::overwrite_block_number_on_new_block");
```
atanmarko: Should this go under some test configuration feature?

iljakuklic: According to the `fail` docs:

> Fail points generation by this macro is disabled by default, and can be enabled where relevant with the `failpoints` Cargo feature.

We enable the feature in `dev-dependencies` only. Larger code blocks that only exist to support fail points are already guarded by `#[cfg(test)]`.

atanmarko: Still, I would prefer that it is visually separated from the rest of the code. It is not great to have the test code scattered around the critical production path.

crates/agglayer-clock/src/block.rs
```diff
@@ -90,6 +89,9 @@ impl<P> BlockClock<P> {
     /// Create a new [`BlockClock`] instance based on a genesis block number and
     /// an Epoch duration.
     pub fn new(provider: P, genesis_block: u64, epoch_duration: NonZeroU64) -> Self {
+        // Initialize metrics for clock startup
+        agglayer_telemetry::clock::record_clock_startup();
```
Ekleog-Polygon, line 93: AFAIU the metrics are completely global, yet `BlockClock` is not a static variable.

Can we introduce something like a "clock name", and have that as a label on all the metrics? Otherwise, if we have two block clocks, they'll clobber each other's metrics

crates/agglayer-clock/src/block.rs
```diff
@@ -108,15 +110,27 @@ impl<P> BlockClock<P> {
...
+        connect_attempt_timeout: Duration,
     ) -> Result<Self, BlockClockError> {
-        let ws = WsConnectWithRetries(ws, Some(max_reconnection_elapsed_time));
+        let ws = WsConnectWithRetries {
+            connection,
+            connect_attempt_timeout,
+        };
+        info!(
+            genesis_block = genesis_block,
+            epoch_duration = epoch_duration.get(),
```
Ekleog-Polygon: Nit: `#[tracing::instrument]` and remove the two manually-added fields here

crates/agglayer-clock/src/block.rs
```diff
@@ -152,35 +166,55 @@ where
         start_sender: oneshot::Sender<()>,
         cancellation_token: CancellationToken,
     ) -> Result<(), BlockClockError> {
-        info!("Starting the BlockClock task");
+        info!(
+            genesis_block = self.genesis_block,
+            epoch_duration = self.epoch_duration.get(),
```
Ekleog-Polygon: Considering we seem to also have these quite a few times down below, also have these in a function-wide span, so all traces automatically have them?

crates/agglayer-clock/src/block.rs
```diff
@@ -197,8 +231,17 @@ where
             Ordering::Relaxed,
```
Ekleog-Polygon, line 224: Oh wow this is my first dive into agglayer-clock, and it looks like we vastly overengineered everything, if we're doing atomics for something that is called once every tens of seconds and does network accesses :open_mouth: 

I'm pretty surprised we even need them, intuition seems to indicate we should just be able to remove them all, but I'd need to spend more time on this than it's worth right now I guess.

crates/agglayer-clock/src/block.rs
```diff
@@ -301,7 +367,16 @@ where
                         return Err(BlockClockError::SetEpochNumber(previous, expected));
                     }
                     Ok(epoch_ended) => {
-                        info!("Clock detected the end of the Epoch: {epoch_ended}");
+                        info!(
+                            epoch_number = epoch_ended.as_u64(),
```
Ekleog-Polygon:
```suggestion
                            finished_epoch_number = epoch_ended.as_u64(),
```

crates/agglayer-clock/src/block.rs
```diff
@@ -349,32 +424,46 @@ where
     }
 }
 
-struct WsConnectWithRetries(WsConnect, Option<Duration>);
+struct WsConnectWithRetries {
```
Ekleog-Polygon: ~Did we consider just using [`RetryBackoffLayer`](https://docs.rs/alloy/latest/alloy/transports/layers/struct.RetryBackoffLayer.html)?~

Nit: Rename `WsConnectWithTimeout`, as we're not actually doing any retries

crates/agglayer-telemetry/src/clock.rs
```diff
@@ -0,0 +1,133 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_metrics_initialization() {
+        // Test that metrics can be accessed without panicking
+        CURRENT_BLOCK_HEIGHT.record(100, &[]);
+        CURRENT_EPOCH.record(10, &[]);
+        CONNECTION_STATUS.record(1, &[]);
+        HEALTH_STATUS.record(1.0, &[]);
```
Ekleog-Polygon, line 118: Nit: test the remaining three metrics?

crates/agglayer-telemetry/src/clock.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn test_helper_functions() {
+        // Test helper functions
+        record_clock_startup();
+        record_connection_established();
+        record_connection_failed();
+        record_reconnection_attempt();
+
+        record_current_block_height(1000);
+        record_current_epoch(50);
+        record_subscription_lag(5);
```
Ekleog-Polygon, line 136: Nit: test the eighth function?

Ekleog-Polygon (commented):
LGTM! Happy we're going to see better observability soon :tada: 

I just have a few nits below, feel free to object if you disagree!

crates/agglayer-clock/src/block.rs
iljakuklic: Yea I think a task that shouts into a watch channel on epoch change should do.

Added to the list I collected when working on this:

* https://github.com/agglayer/agglayer/issues/904

crates/agglayer-clock/src/block.rs
iljakuklic: > Still, I would prefer that it is visually separated from the rest of the code. It is not great to have the test code scattered around the critical production path.

Changed. I find it a bit cumbersome but I don't have a very strong preference.

crates/agglayer-clock/src/block.rs
iljakuklic: Hmm not sure how to best address this. There are not supposed to be multiple clocks running within a node. Labeling them seems redundant if the node operates correctly (i.e. has only one clock). But it is still not enforced statically.

Thinking of having a static bool to check if clock has been started and erroring out if it has. However, that may cause trouble in tests where it is expected to have multiple active clock instances due to tests running in parallel.

atanmarko (approved):

crates/agglayer-clock/src/block.rs
```diff
@@ -219,10 +219,11 @@ where
         // Calculate the local Block height based on the current L1 Block number.
         let current_block = self.calculate_block_number(self.latest_seen_block);
 
-        // Overwrite the block number to simulate an overflow
-        // This is used for testing purposes only and doesn't affect the production
-        // code.
-        fail::fail_point!("block_clock::BlockClock::run::overwrite_block_number");
+        #[cfg(test)]
```
Ekleog-Polygon: Do we want these gated more generically on `testutils`?

crates/agglayer-clock/src/block.rs
```diff
@@ -109,6 +109,7 @@ impl<P> BlockClock<P> {
 }
 
 impl BlockClock<BlockProvider> {
```
Ekleog-Polygon, line 111: Now that this is done, you can remove the genesis_block and epoch_duration keys in the `info!` below

crates/agglayer-clock/src/block.rs
```diff
@@ -204,11 +207,7 @@ where
             let header = Self::recv_block(&mut stream).await?;
             self.latest_seen_block = header.number;
 
-            debug!(
-                current_block = self.latest_seen_block,
-                genesis_block = self.genesis_block,
-                "Waiting for genesis block"
-            );
+            debug!("Waiting for genesis block");
```
Ekleog-Polygon, line 206: There's one more genesis_block key just below this block

Ekleog-Polygon (approved):
Awesome, thank you!

Ekleog-Polygon (approved):

Issue #931, opened 2025-07-30
SP1 Execute can block the async threads


Ekleog-Polygon:
The main known culprit is the PP execution during certificate handling, but there may be others we haven't thought of yet

hadjiszs:
Notes from refinement:

- this is for the huge "genesis" certificate to bootstrap existing networks as PP chain
- issue occurred for a certificate of ~50k bridge exits + ~200k imported bridge exits
  - need for more than ~32GB ram
- issue might not be solved with spawn blocking because the runtime is already multithread

questions:
- why is the API blocked?

PR #934, opened 2025-07-31
fix: run expensive computation in a blocking task (#933)

Ekleog-Polygon:
Fixes https://github.com/agglayer/agglayer/issues/931

(cherry picked from commit 15c93ae16b4f88d652a98e9b4ae2d87964eb07e3)

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
@@ -166,12 +166,19 @@ where
         let (pv_sp1_execute, report) = {
             // Do not verify the deferred proof if we are in mock mode
             let deferred_proof_verification = !self.config.mock_verifier;
-            let (pv, report) = self
-                .verifier
-                .execute(ELF, &stdin.clone())
-                .deferred_proof_verification(deferred_proof_verification)
-                .run()
-                .map_err(CertificationError::Sp1ExecuteFailed)?;
+            let (pv, report) = tokio::task::spawn_blocking({
+                let verifier = self.verifier.clone();
+                let stdin = stdin.clone();
+                move || {
+                    verifier
+                        .execute(ELF, &stdin)
+                        .deferred_proof_verification(deferred_proof_verification)
+                        .run()
+                }
+            })
+            .await
+            .map_err(|e| CertificationError::InternalError(e.to_string()))?
+            .map_err(CertificationError::Sp1ExecuteFailed)?;
 
             let pv_sp1_execute: PessimisticProofOutput = PessimisticProofOutput::bincode_options()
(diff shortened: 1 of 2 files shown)


atanmarko (approved):

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -166,12 +166,19 @@ where
...
-            let (pv, report) = self
-                .verifier
-                .execute(ELF, &stdin.clone())
-                .deferred_proof_verification(deferred_proof_verification)
-                .run()
-                .map_err(CertificationError::Sp1ExecuteFailed)?;
+            let (pv, report) = tokio::task::spawn_blocking({
+                let verifier = self.verifier.clone();
+                let stdin = stdin.clone();
+                move || {
```
dwgrantham, line 172: This inner async closure is unnecessary and over-complicates the solution. The `spawn_blocking` call will run the outer closure in a thread in a separate "blocking" thread pool. If the goal is to prevent the verifier from blocking the main thread pool, then `spawn_blocking` is all you need.

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -457,9 +464,14 @@ where
 
         // Perform the native PP execution without the STARK verification in order to
         // cross check the target roots.
-        let (pv, targets_native_execution) =
-            generate_pessimistic_proof(initial_state.clone().into(), &multi_batch_header)
-                .map_err(|source| CertificationError::NativeExecutionFailed { source })?;
+        let (pv, targets_native_execution) = tokio::task::spawn_blocking({
+            let initial_state = initial_state.clone();
+            let multi_batch_header = multi_batch_header.clone();
+            move || generate_pessimistic_proof(initial_state.into(), &multi_batch_header)
```
dwgrantham, line 470: Here too. This is not necessary.

dwgrantham (commented):
@Ekleog-Polygon I think you should remove the inner async closure to keep this code as simple as possible. The `spawn_blocking` and the main closure is all you need to run the long-running verifier in its own thread pool.

dwgrantham:
@Freyskeyd @Ekleog-Polygon I know this is closed but it should be cleaned up.

The reason is that using the blocking thread pool thread to await a future that runs on the main thread pool is 1. unnecessary 2. overly complicated and 3. could cause the blocking thread pool to get exhausted due to contention on the main thread pool. The correct solution is:

```rust
tokio::task::spawn_blocking({
    // run the verifier
})
```

This runs the verifier on the blocking thread pool and keeps the main thread pool from getting exhausted.

PR #939, opened 2025-08-01
ci(docs): deploy doc to cloudflare workers

sebastiendan:
This PR deploys the rust docs to Cloudflare Workers on PR events and shares the new docs URL in the PR.


.github/workflows/doc.yml
```diff
@@ -48,6 +57,111 @@ jobs:
...
+        run: wrangler deploy --env staging
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CF_WORKER_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CF_WORKER_ACCOUNT_ID }}
+
+      - name: Comment PR with deployment URL
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const deploymentComment = `📖 Documentation deployed to: https://pr-${{ github.event.number }}-rust-docs.agglayer.dev`;
```
Freyskeyd: What if we have other repository doing the same ?
We could have the same PR number as they're independent.
Could we have a URL such as `https://${{ github.event.repository.name }}-pr-${{ github.event.number }}-rust-docs.agglayer.dev` ?

sebastiendan: Applied the change 👍

Ekleog-Polygon (dismissed):

Freyskeyd (dismissed):

atanmarko (approved):

.github/workflows/doc.yml
```diff
@@ -1,18 +1,24 @@
...
 
+# Default permissions (individual jobs can override as needed)
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Build documentation
     runs-on: ubuntu-latest-16-cores
+    if: github.event.action != 'closed'
```
polykeith, line 21: If we have to prefix this on practically every job here, why don't we just not trigger any jobs on `closed`, i.e. just take it out from `on.pull_request.types`?

sebastiendan: We have a another job triggered on `closed` events to remove the workers on Cloudflare.

PR #944, opened 2025-08-02
fix: wonky quotes in bash command

defitricks:
the old quotes were breaking things—messed up variable expansion and escaping.
switched outer quotes to double, escaped the inner ones properly, and kept `$time` safe in singles. now it actually works.  

before:  
```bash
'echo "Time is $time"'
```  

after:  
```bash
"echo \"Time is \$time\""
```  

no more weird errors.

### DIFF
--- a/scripts/fuzz.sh
+++ b/scripts/fuzz.sh
@@ -9,8 +9,8 @@ fuzzers=(
     "agglayer-grpc-types/compat::v1::tests::fuzz_round_trip_epoch_configuration"
 )
 
-printf '%s\0' "${fuzzers[@]}" | parallel --null --bar --joblog fuzz.log bash -c '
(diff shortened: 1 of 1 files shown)


atanmarko (commented):
Hey @defitricks , thank you for your contribution! Unfortunately, we are still waiting for approval from our legal department, and until then, we cannot merge PRs from external contributors.

Ekleog-Polygon (changes requested):
I'm not sure I understand, which errors were you getting? The changes you make actually break the system, because you'll be using the crate/target from outside of parallel's bash call, while we want them to be taken from inside parallel's bash call.

And I just checked, it does work on my machine, so I'm curious about what's going wrong.

Freyskeyd:
Closing as PR's author didn't respond for a while

PR #951, opened 2025-08-04
fix: error handling

atanmarko:
Fix init of the tracing subscriber.

Fixes https://github.com/agglayer/security/issues/87

### DIFF
--- a/crates/agglayer-node/src/lib.rs
+++ b/crates/agglayer-node/src/lib.rs
@@ -49,7 +49,7 @@ pub fn main(
     }
 
     // Initialize the logger
-    logging::tracing(&config.log);
+    logging::tracing(&config.log)?;
 
     info!("Starting agglayer node version info: {}", version);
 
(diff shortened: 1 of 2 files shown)


crates/agglayer-node/src/lib.rs
```diff
@@ -49,7 +49,23 @@ pub fn main(
...
+        Ok(()) => {
+            info!("Tracing initialized successfully.");
+        }
+        Err(e)
+            if e.to_string()
+                .contains("trace dispatcher has already been set") =>
+        {
+            // This is a common case in integration tests where the logger is initialized
+            // multiple times. We can safely ignore this error.
+            debug!("Logger already initialized, ignoring error: {e}");
```
Freyskeyd, line 62: would it panic if the logger fails to initialize here?

atanmarko: No, it just continues. Tested on my side.

Issue #855, opened 2025-06-18
Double check that `sp1::execute` on PP doesn't panic if `aggchain-proof` is encoded with invalid content


Freyskeyd:
# Description

hadjiszs:
Note from backlog refinement:

> Note to self: maybe do it by fuzzing the certificate task? That'd give us a reusable fuzzer that could also detect it.

Can be encompassed by the fuzzing on the certificate but which is less of a priority

- #892

Ekleog-Polygon:
Outcome of today's discussion: let's have an `sp1_blocking` and an `sp1_async` helper functions in prover-executor, and just wrap literally all sp1 calls in one of them 😄

PR #957, opened 2025-08-07
fix: use sp1 wrappers everywhere

Ekleog-Polygon:
Fixes https://github.com/agglayer/agglayer/issues/855

Depends on https://github.com/agglayer/provers/pull/267


crates/agglayer-certificate-orchestrator/src/error.rs
```diff
@@ -125,6 +125,9 @@ pub enum CertificationError {
         from_proof: Box<AggchainProofPublicValues>,
         from_witness: Box<AggchainProofPublicValues>,
     },
+
+    #[error(transparent)]
+    Other(eyre::Error),
```
iljakuklic, line 130: Would it make sense to make this error more specific, for example:
```suggestion
    #[error("SP1 crashed")]
    SP1Crash(#[source] eyre::Error),
```

Or maybe "proving system crash" if mentioning SP1 is deemed too specific.

Ekleog-Polygon: Hmm so my thinking here is, we'll hopefully soon be turning this whole type into `eyre::Error`. So I'm using `Other` as a placeholder to say "we haven't migrated all the sub-types yet but it'll come soon; all new sub-types are expected to be Other". Does that make sense to you? If not I'll make a follow-up PR to adjust this!

PR #958, opened 2025-08-07
feat: certificate versioning

iljakuklic:
Versioning mechanism for the `Certificate` structure. Will allow handling of multiple certificate versions with different features. Also makes it possible to evolve the way in which `CertificateId` is calculated.


iljakuklic:
Fairly out of date by now

PR #963, opened 2025-08-11
feat: multisig on PP and agglayer witness generation

### LINKED ISSUE #943: Adapt the PP program and witness generation for the multisig
### Description

In consensus type 1, we should now verify either one aggchain proof, or one multisig, or both.

This ticket encompasses also the changes in the agglayer node for the witness generation.

### Questions

1. Should we merge `consensus_type == 0` within `consensus_type == 1`. See @invocamanman proposal [here](https://github.com/agglayer/protocol-research/issues/169#issuecomment-3140745509)

hadjiszs:
Includes the basic structures and logic for the witness generation + the PP program modifications

Fixes #943


Cargo.toml
```diff
@@ -44,13 +44,21 @@ pessimistic-proof-core = { path = "crates/pessimistic-proof-core" }
...
-agglayer-primitives = "0.9.0"
-agglayer-tries = "0.9.0"
-unified-bridge = "0.9.0"
+agglayer-bincode = { git = "https://github.com/agglayer/interop.git", branch = "rename-commitment-version" }
+agglayer-interop = { git = "https://github.com/agglayer/interop.git", branch = "rename-commitment-version" }
+agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", branch = "rename-commitment-version" }
+agglayer-elf-build = { git = "https://github.com/agglayer/interop.git", branch = "rename-commitment-version" }
+agglayer-primitives = { git = "https://github.com/agglayer/interop.git", branch = "rename-commitment-version" }
+agglayer-tries = { git = "https://github.com/agglayer/interop.git", branch = "rename-commitment-version" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "rename-commitment-version" }
```
hadjiszs: todo: proper tag once released

crates/agglayer-storage/src/types/certificate/tests/structure.rs
```diff
@@ -10,6 +10,7 @@ fn trace() -> serde_reflection::Result<serde_reflection::Registry> {
     dbg!(tracer.trace_value::<Address>(&mut samples, &Address::new([0xad; 20]))?);
     dbg!(tracer.trace_value::<Digest>(&mut samples, &Digest([0xd1; 32]))?);
     dbg!(tracer.trace_value::<NetworkId>(&mut samples, &NetworkId::new(42))?);
+    dbg!(tracer.trace_value::<B256>(&mut samples, &B256::from([0xbd; 32]))?);
```
iljakuklic: I believe this is no longer needed.

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
```diff
@@ -0,0 +1,208 @@
...
+                                leaf_count: declared_leaf,
+                            },
+                        });
+                    }
+                }
+
+                retrieved_root
+            }
+            // Inconsistent declared L1 info tree entry
+            (l1_leaf, l1_info_root) => {
```
polykeith: Nit: not due to your PR, but I find it much easier to review when match patterns are exhaustive, i.e. by the time the match expression evaluates to this arm, we know that `l1_leaf` MUST be `None`, and `l1_info_root` is `Some`. It would also defend us against possible additional match cases that we didn't foresee today, such as adding an additional element in the tuple to match against

I _think_ the following syntax is possible in today's Rust:

```suggestion
            (l1_leaf @ None, l1_info_root @ Some(_)) => {
```

crates/agglayer-types/src/aggchain_data/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn from(value: AggchainData) -> Self {
+        match value {
+            AggchainData::ECDSA { signature } => global::Payload::LegacyEcdsa { signature },
+            AggchainData::Generic {
+                proof,
+                aggchain_params,
+                signature,
+                public_values,
+            } => global::Payload::AggchainProofOnly {
+                signature: *signature.unwrap(), /* this signature must be mandatory, need
```
polykeith: But panicking here seems non-ideal. It basically halts the task/process that the unwrap is in because an irrecoverable fault has happened. If a `Signature` happens to be `None`, is it really such a catastrophic error, or can we just log it and move onto processing other things?

If it's the latter, then we should instead implement `TryFrom` and return a `None` here if the signature happens to be `None`.

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -0,0 +1,88 @@
...
+use pessimistic_proof::core::{self};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::aggchain_data::PayloadWithCtx;
+
+/// Multisig data from the chain.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Payload {
+    signatures: Vec<Signature>,
```
polykeith, line 27: This only gathers the signatories that have signed and approved of the certificate, right?

Moreover, is this actually enough data to encapsulate what a multisig payload should contain? What happens if we add or remove a signatory during a multisig signing process? It sounds like it should also store the ID of the remaining signatories so that the nodes know which signature is missing to get a quorum.

crates/agglayer-types/src/aggchain_data/mod.rs
hadjiszs: yeah you're totally right, my bad I didn't come back on this one

crates/agglayer-types/src/aggchain_data/multisig.rs
hadjiszs: The idea is that you have two things:
- payload from the chain (this `Payload` struct)
- context from the L1 (the `Ctx` struct right below)

for the multisig, we want the chain to provide their signatures, and then we enforce the verification with the multisig committee and the threshold registered on the L1 (values which are fetched from the L1 and so contained in the `Ctx` struct right below)

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
```diff
@@ -0,0 +1,212 @@
...
+                expected: aggchain_vkey.to_hex(),
+                actual: proof_vk_hash.to_hex(),
+            });
+        }
+
+        Ok(AggchainProofCtx {
+            aggchain_vkey: aggchain_proof_payload
+                .aggchain_vkey_from_proof()
+                .vk
+                .hash_u32(),
```
Ekleog-Polygon: Nit: Can we use a temporary variable to avoid recomputing `aggchain_vkey_from_proof` twice?

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
```diff
@@ -0,0 +1,212 @@
...
+        })
+    }
+
+    pub async fn fetch_multisig_ctx(
+        &self,
+        certificate: &Certificate,
+    ) -> Result<MultisigCtx, CertificationError> {
+        let (_multisig_signers, _multisig_threshold): (Vec<Address>, usize) = {
+            // TODO: To fetch from the L1
```
Ekleog-Polygon:
```suggestion
            // TODO: To fetch from the L1: https://github.com/agglayer/agglayer/issues/941
```

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
```diff
@@ -0,0 +1,212 @@
...
+            threshold: _multisig_threshold,
+            prehash: B256::new(
+                SignatureCommitmentValues::from(certificate)
+                    .commitment(SignatureCommitmentVersion::V5)
+                    .0,
+            ),
+        })
+    }
+
+    /// Fetch, verify consistency, and wait the finalization for the l1 info
```
Ekleog-Polygon:
```suggestion
    /// Fetch, verify consistency, and wait for the finalization of the l1 info
```

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -457,12 +347,20 @@ where
 
         // Verify consistency on the aggchain proof public values if provided in the
         // optional context
+        // TODO: to move
```
Ekleog-Polygon: Where?

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
```diff
@@ -0,0 +1,49 @@
...
+use pessimistic_proof::core::{self};
+use serde::{Deserialize, Serialize};
+use sp1_sdk::SP1VerifyingKey;
+use unified_bridge::AggchainProofPublicValues;
+
+use crate::aggchain_data::PayloadWithCtx;
+
+/// Aggchain proof with aggchain params and optional public values for debug
+/// purposes.
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
Ekleog-Polygon: Do we really need (de)serialize here?

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl Payload {
+    pub fn aggchain_vkey_from_proof(&self) -> SP1VerifyingKey {
+        let crate::aggchain_proof::Proof::SP1Stark(sp1_reduce_proof) = &self.proof;
+        sp1_reduce_proof.vkey.clone()
+    }
+}
+
+/// Aggchain proof data from the L1 and enforced by the agglayer.
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
Ekleog-Polygon: And here too; I can't see off the top of my head why we'd need them

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
```diff
@@ -0,0 +1,49 @@
+use agglayer_interop_types::aggchain_proof::Proof;
+use agglayer_primitives::Digest;
+use pessimistic_proof::core::{self};
```
Ekleog-Polygon:
```suggestion
use pessimistic_proof::core;
```

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -0,0 +1,146 @@
...
+//!
+//! 2. current generic, receive one aggchain proof with trusted sequencer
+//!    signature verified only on the agglayer
+//!   - stark proof
+//!   - single ecdsa verified on the agglayer (+ optional pv for debug)
+//!   - aggchain params
+//!
+//! 3. generic + multisig (one or the other, or both)
+//!   - aggchain proof + single ecdsa (trusted sequencer, not multisig) ->
+//!     single signature verified only on the agglayer
```
Ekleog-Polygon, line 15: Hmm… why keep this case separate from the proof+multisig, with one signer and threshold of one? Now we have the infra to verify the signature outside of just agglayer, it feels like it'd be better to not have this case at all?


(Also, reading the code below it seems to not actually be implemented)

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::aggchain_data::{
+    aggchain_proof::{self},
+    multisig::{self, MultisigError},
+    PayloadWithCtx,
+};
+
+/// Represents the data needed from the API/Certificate to verify aggchain
+/// proofs and multisig.
+/// Made separately for now to not have to deal with storage and API design.
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
Ekleog-Polygon: Same here and down below, do we really need (de)serialize for these? Even if we need to serialize to pass into PP, it feels like at most one of the various structs of this name should be required?

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -0,0 +1,146 @@
...
+                    multisig_ctx,
+                    aggchain_proof_ctx,
+                },
+            ) => core::AggchainData::MultisigAndAggchainProof {
+                multisig: PayloadWithCtx(multisig, multisig_ctx)
+                    .try_into()
+                    .map_err(AggchainDataError::InvalidMultisig)?,
+                aggchain_proof: PayloadWithCtx(aggchain_proof, aggchain_proof_ctx).into(),
+            },
+            (payload, ctx) => {
```
Ekleog-Polygon: If `PayloadWithCtx` were a single enum with 4 branches that each contain two fields, instead of a struct with two fields that each are a 4-branch enum, then we wouldn't need this case

crates/agglayer-types/src/aggchain_data/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+    aggchain_proof::{Ctx as AggchainProofCtx, Payload as AggchainProofPayload},
+    global::{
+        AggchainDataError, Ctx as CertificateAggchainDataCtx, Payload as CertificateAggchainData,
+    },
+    multisig::{Ctx as MultisigCtx, MultisigError, Payload as MultisigPayload},
+    PayloadWithCtx as CertificateAggchainDataWithCtx,
+};
+
+/// Represents the payload from the chain, with the context fetched from the L1.
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
Ekleog-Polygon: (Here too, is (de)serialize really required? With our current way of dealing with storage it's scary to modify serializable structs, so if we can not derive it it'd be best IMO)

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use agglayer_primitives::{Address, Digest, Signature, B256};
+use pessimistic_proof::core::{self};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::aggchain_data::PayloadWithCtx;
+
+/// Multisig data from the chain.
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
Ekleog-Polygon: Here again the comment about possibly-unneeded (de)serialize. (Also, I'm quite surprised to have seen at least 3 Payload structs I think, maybe we could deduplicate some of them? And the same for Ctx below.)

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct Payload {
```
Ekleog-Polygon, line 26: Thinking about the payload signing process… Won't there be a circular dependency?
- To sign the certificate, the signers need the certificate id
- To compute the certificate id, they need all the hashed fields, including the payload
- To compute the payload, they need the list of signatures
- But they can compute the signatures only after everyone has signed the certificate.

I'm probably missing something obvious here, but I'm not seeing what :sweat_smile:

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -0,0 +1,88 @@
...
+                    .map_err(|_| MultisigError::InvalidSignature)?;
+                let i = *index.get(&recovered.into_array()).ok_or(
+                    MultisigError::UnknownRecoveredSigner {
+                        recovered,
+                        prehash: prehash.into(),
+                    },
+                )?;
+                Ok::<_, Self::Error>((i, sig))
+            })
+            .collect::<Result<_, _>>()?;
```
Ekleog-Polygon, line 84: This doesn't verify that a single signer has not signed `threshold` times the certificate.

I guess I'll see that check in the PP once I get to it below, but this still feels wrong, especially considering we're already checking the threshold above.

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -0,0 +1,88 @@
...
+                got: signatures.len(),
+                expected: threshold,
+            });
+        }
+
+        let index: HashMap<[u8; 20], usize> = signers
+            .iter()
+            .enumerate()
+            .map(|(i, s)| (s.into_array(), i))
+            .collect();
```
Ekleog-Polygon: Nit: for all practical purposes (with <128 signers), just using `Vec::position` will most likely be much more efficient than converting to HashMap and then indexing: performing 128 cached comparisons is most likely not slower than hashing, and it'd avoid memory allocation, random access, collision management and lots of other HashMap overhead.

If the code were cleaner with HashMap I'd shrug because perf isn't very important, but here it feels like using `Vec::position` would both make the code simpler and faster :sweat_smile:

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -0,0 +1,125 @@
...
+            AggchainHashValues::ConsensusType0 { .. } => 0,
+            AggchainHashValues::ConsensusType1 { .. } => 1,
+        }
+    }
+}
+
+impl AggchainHashValues {
+    /// Returns the commitment on signers in case of no signer.
+    pub fn empty_signers() -> Digest {
+        keccak256_combine([Digest::default()])
```
Ekleog-Polygon, line 35: I think it should be like this?
```suggestion
        keccak256_combine([])
```

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+                let commitment =
+                    SignatureCommitmentValues::new(&constrained_values, None).multisig_commitment();
+
+                multisig
+                    .verify(commitment)
+                    .map_err(ProofError::InvalidMultisig)?;
+            }
+            AggchainData::AggchainProofOnly(aggchain_proof) => {
+                // Panic upon invalid proof.
+                aggchain_proof.verify_aggchain_proof(&constrained_values);
```
Ekleog-Polygon, line 82: Why does `aggchain_proof.verify_aggchain_proof` panic while `multisig.verify` return an error? Would it be hard to make both consistent?

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -0,0 +1,66 @@
...
+                        .map(|recovered| recovered != *expected)
+                })
+                .unwrap_or(true)
+        });
+
+        if has_invalid_signature {
+            return Err(MultisigError::HasInvalidSignature);
+        }
+
+        Ok(())
```
Ekleog-Polygon, line 94: Ok so my aomment above was actually wrong, and this is a probably bug: PP should verify that all the signatures were made by different signers, should it not?

Here, agglayer (or the chain's aggkit) could either give a set of signatures that just refers many times to the same signer, and it'd AFAIU get accepted. Or (haven't checked if some other component wouldn't prevent that) give a set of signers that has the same address multiple times, and it could be accepted too (I'm much less sure about this attack, but checking that here wouldn't cost much)

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -0,0 +1,125 @@
+use agglayer_primitives::{
+    bytes::{BigEndian, ByteOrder as _},
+    keccak::keccak256_combine,
+    Address, Digest,
+};
+
+use crate::aggchain_data::{aggchain_proof::AggchainProof, AggchainData, Vkey};
+
+type ConsensusType = u32;
```
Ekleog-Polygon: Nit: can we make that `struct ConsensusType(u32)`, to stay on the good track of using strong types we started doing a few months ago?

crates/pessimistic-proof-core/src/local_state/commitment.rs
```diff
@@ -4,10 +4,23 @@
...
 
-use crate::{proof::EMPTY_PP_ROOT_V2, ProofError};
+use crate::{
+    aggchain_data::AggchainHashValues,
+    proof::{ConstrainedValues, EMPTY_PP_ROOT_V2},
+    ProofError,
+};
+
+#[derive(Debug, Clone, Copy)]
+pub enum PessimisticRootCommitmentVersion {
```
Ekleog-Polygon, line 21: Nit: could we have comments, here and on the signature version below, that explain what each version corresponds to?

crates/pessimistic-proof-core/src/multi_batch_header.rs
```diff
@@ -3,35 +3,48 @@ use std::collections::BTreeMap;
...
 
 use crate::{
-    aggchain_proof::AggchainData, local_balance_tree::LocalBalancePath,
+    aggchain_data::AggchainData, local_balance_tree::LocalBalancePath,
     nullifier_tree::NullifierPath,
 };
 
 /// Represents the chain state transition for the pessimistic proof.
-#[serde_as]
 #[derive(Clone, Debug, Serialize, Deserialize)]
```
Ekleog-Polygon, line 16: Random thought: do you think protobuf would be more expensive than serde's (de)serialization for the PP? I'm thinking about the idea to desynchronize PP and agglayer versioning to more fearlessly move both forward, and having a protobuf layer of backward-compatibility could help there. If you think that's an idea worth exploring, I can open an issue :)

docs/validity_checks.md
```diff
@@ -0,0 +1,123 @@
...
+
+| #   | Case             | Verified in Agglayer | Verified in PP   | Commitment Version             |
+| --- | ---------------- | -------------------- | ---------------- | ------------------------------ |
+| 1   | Multisig only    | Multisig             | Multisig         | V5                             |
+| 2   | STARK + ECDSA    | STARK + ECDSA        | STARK            | (currently V4, must become V5) |
+| 3   | STARK + Multisig | STARK + Multisig     | STARK + Multisig | V5                             |
+
+Notes:
+
+- Aggchain proof alone isn’t accepted by the Agglayer; it must be accompanied by at least one ECDSA signature from the trusted sequencer (see #[issue to whitelist submitter in case of sp1 issue]).
```
Ekleog-Polygon: Nit: add the link

Ekleog-Polygon (changes requested):
Overall the changes LGTM, thank you! I love the new abstractions, though it looks to me like there's been unneeded duplication of quite a few structs (I stopped counting the number of Payload and Ctx structs after a few, for example :sweat_smile: )

Requesting changes for now, mostly due to the security issue about the same signer being set multiple times in multisigs.

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
hadjiszs: yes great catch 👍 

- duplication on the signers: this has to be enforced at least by the L1, because in the end the hash of the entire ordered list of signers is hashed and put as public input, and then the L1 re-compute this hash on the signers to verify the PP, so should be enforced at least by the L1. But we can totally log error on the agglayer node if we fetch duplicated signers from the L1, and error as well in the PP circuit for in depth prevention.

- duplication on the signatures: yes this one definitely we should prevent, I'll fix this in this PR and add more tests potentially in another PR

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
hadjiszs: for all those default values they are some kind of placeholder, I still need to align the 4 default values with the protocol team because they need to be the same, so it's just a matter of agreeing on the default values (can be anything, just need to be the same values)

crates/agglayer-types/src/aggchain_data/global.rs
hadjiszs: because the deployed chains which uses this case (e.g., katana) doesn't sign the right commitment for the multisig, so for now I cannot merge this case with the case `MultisigAndAggchainProof` otherwise it wont be backward compatible with at least katana: it'd require the aggsender to be upgraded

there are some discussions right now to understand for which thing people want backward compatibility, and then it'd lead us to merge some path or not

exact same reason as to why `legacy_ecdsa` isn't merged yet in the path `multisig_only` with a multisig 1-of-1

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
hadjiszs: that's because `verify_aggchain_proof` does the syscall to `sp1_zkvm::lib::verify::verify_sp1_proof`, which doesnt return any result, it just panic upon failure

crates/agglayer-types/src/aggchain_data/multisig.rs
hadjiszs: yeah same response as here: https://github.com/agglayer/agglayer/pull/963#discussion_r2293161483

crates/agglayer-types/src/aggchain_data/multisig.rs
hadjiszs: no you're right, we need to adapt the certificate id computation as a consequence, so the certificate id wouldnt include the signatures I believe, same as today: the certificate id doesnt include `certificate.aggchain_data.signature`

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
hadjiszs: I wanted to move it to the module `l1_context` where I already moved many fetching from L1, but it might require more refactoring, that's not critical and it could be done in many ways I believe, I think I'll just remove this todo and keep it there for now.

crates/agglayer-types/src/aggchain_data/global.rs
hadjiszs: hm it'd just move this case to the place where I would generate the `PayloadWithCtx` out of the two other enums no?

Ekleog-Polygon: Hmm… maybe? I haven't checked that part of the code, and especially whether or not it already knows which branch of the enum it must be. But regardless, if `PayloadWithCtx` were one big enum, it'd make invalid states impossible to represent in the Rust type system, which is always a good thing. Here, it'd manifest in having the match at the one place you generate it, and not at every place you use it (though I don't think it's being read many times right now).

This being said I definitely won't block this PR on that, so feel free to ignore if you disagree :smile:

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
Ekleog-Polygon: Oh ok so we don't have a choice about it panicking… unless we were to catch_unwind, which probably wouldn't be worth it :smiling_face_with_tear:

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -36,14 +42,31 @@ impl MultiSignature {
         keccak256_combine([buf.as_slice()])
     }
 
+    /// Ensure that the signatures refer to unique and existing signer.
+    fn unique_signers(&self) -> Result<(), MultisigError> {
```
Ekleog-Polygon, line 46: Nit: `-> Result<usize, MultisigError>` where the `usize` is the number of signers. With this we can directly use this function everywhere instead of taking `.signatures.len()` and then remembering to call this function separately :)

Also, would such a change allow us to reuse this function from `agglayer-types/src/aggchain_data/multisig.rs` instead of duplicating the code?

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
```diff
@@ -0,0 +1,48 @@
...
+use pessimistic_proof::core;
+use sp1_sdk::SP1VerifyingKey;
+use unified_bridge::AggchainProofPublicValues;
+
+use crate::aggchain_data::PayloadWithCtx;
+
+/// Aggchain proof with aggchain params and optional public values for debug
+/// purposes.
+#[derive(Clone, Debug)]
+pub struct Payload {
```
atanmarko, line 12: Nit: have we settled with the `Payload` name here? How about `AggchainProofPayload`?

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
```diff
@@ -0,0 +1,48 @@
...
+impl Payload {
+    pub fn aggchain_vkey_from_proof(&self) -> SP1VerifyingKey {
+        let crate::aggchain_proof::Proof::SP1Stark(sp1_reduce_proof) = &self.proof;
+        sp1_reduce_proof.vkey.clone()
+    }
+}
+
+/// Aggchain proof data from the L1 and enforced by the agglayer.
+#[derive(Clone, Debug)]
+pub struct Ctx {
```
atanmarko: Nit: How about `AggchainProofCtx` here?

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -0,0 +1,146 @@
...
+    aggchain_proof::{self},
+    multisig::{self, MultisigError},
+    PayloadWithCtx,
+};
+
+/// Represents the data needed from the API/Certificate to verify aggchain
+/// proofs and multisig.
+/// Made separately for now to not have to deal with storage and API design.
+#[derive(Clone, Debug)]
+pub enum Payload {
```
atanmarko, line 35: Nit: `Payload` is too generic - `AggchainPayload` or something similar would probably help with the code readability. Not to repeat same for the other structures - e.g. `Ctx` would be better as `AggchainCtx`. Especially, having the non compatible structures with different usage but with the same names `Payload`, `Ctx` in different modules for the same crate is not a great idea I think. We have 3 `Payload` types - multisig, aggchain proof and this one enum.

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -0,0 +1,146 @@
...
+                return Err(AggchainDataError::InvalidVariant(format!(
+                    "payload: {payload}, ctx: {ctx}"
+                )))
+            }
+        })
+    }
+}
+
+// For error and debugging
+impl Display for Payload {
```
atanmarko: Nit: I see we have `strum` in sub-dependencies, you may use it to automate this manual display functions:
 ```
#[strum_discriminants(strum(serialize_all = "kebab_case"))]
pub enum Payload {}

impl fmt::Display for Payload {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let disc = EventDiscriminants::from(self);
        f.write_str(disc.as_ref()) // "legacy_ecdsa", "multisig_only", "aggchain_proof_only","multisig_and_aggchain_proof",
    }
}

```
or to manually set names:
```
#[derive(Display)]
pub enum Payload {
   #[strum(to_string = "legacy_ecdsa")]
    LegacyEcdsa {
        /// ECDSA Signature from the trusted sequencer.
        signature: Signature,
    },
```

**Review by atanmarko — COMMENTED** [review]
Did a quick pass-through of this behemont, I have few nits to mention.

File: crates/pessimistic-proof-core/src/aggchain_data/multisig.rs [hunk 33]
```diff
@@ -0,0 +1,157 @@
...
+        for addr in &self.expected_signers {
+            buf.extend_from_slice(addr.as_slice()); // address (20bytes)
+        }
+        keccak256_combine([buf.as_slice()])
+    }
+
+    /// Ensure that the signatures refer to unique and existing signer.
+    fn unique_signers(&self) -> Result<(), MultisigError> {
+        let total = self.expected_signers.len();
+        let mut seen = vec![false; total];
```
iljakuklic, line 48: Would be more efficient to just check the indices are strictly increasing, This would add the requirement that the signatures are sorted by the index. Could be done as a preprocessing step on the host when preparing data to be passed to the PP. Also not sure how L1 would handle that. But in general, it is usually better avoid malleability where practical.

hadjiszs: yeah I hesitated to sort it during witness generation for similar reasons, I was assuming that contiguous access would be better on the prover side, but I didnt want to dive into optimizations right now so I have avoided over constraining the prover inputs for now, to also avoid any unnecessary things which could trigger interrogations for the readers

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
iljakuklic: I think it's fine to use `aggchain_proof::Ctx` from other modules by convention. However, I'd prefer it to be called `Context` rather than `Ctx`.

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -0,0 +1,171 @@
...
+    type Error = MultisigError;
+
+    fn try_into(self) -> Result<core::MultiSignature, Self::Error> {
+        let PayloadWithCtx(Payload { signatures }, multisig) = self;
+
+        let mut seen = vec![false; multisig.signers.len()];
+
+        let indexed_signatures: Vec<(usize, Signature)> = signatures
+            .into_iter()
+            .map(|sig| {
```
iljakuklic, line 74: Is it an external requirements for the signatures to be just plain signatures? It would be easier to require the multisig coordinator to present signatures in the same order as specified in the signers vector, possibly even with the appropriate index into the vector.

crates/pessimistic-proof-core/src/local_state/commitment.rs
```diff
@@ -77,39 +105,94 @@ impl PessimisticRoot {
...
+}
 
+impl SignatureCommitmentValues {
     /// Returns the expected signed commitment for the provided version.
     #[inline]
-    pub fn commitment(&self, version: CommitmentVersion) -> Digest {
-        let imported_bridge_exit_commitment = self.commit_imported_bridge_exits.commitment(version);
-        match version {
-            CommitmentVersion::V2 => keccak256_combine([
+    pub fn commitment(&self, version: SignatureCommitmentVersion) -> B256 {
```
iljakuklic, line 144: Nit: does it still make sense for this to be `#[inline]`?

crates/agglayer-types/src/aggchain_data/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    global::{
+        AggchainDataError, Ctx as CertificateAggchainDataCtx, Payload as CertificateAggchainData,
+    },
+    multisig::{Ctx as MultisigCtx, MultisigError, Payload as MultisigPayload},
+    PayloadWithCtx as CertificateAggchainDataWithCtx,
+};
+
+/// Represents the payload from the chain, with the context fetched from the L1.
+#[derive(Clone, Debug)]
+pub struct PayloadWithCtx<Payload, Context>(pub Payload, pub Context);
```
iljakuklic, line 21: Note: At risk of overengineering, you could enforce the correct context type for any given payload by introducing a trait:

```rust
trait PayloadType {
    type Context;
}
```

And using it like this:

```suggestion
pub struct PayloadWithCtx<Payload: PayloadType>(pub Payload, pub Payload::Context);
```

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
iljakuklic: > Oh ok so we don't have a choice about it panicking… unless we were to catch_unwind, which probably wouldn't be worth it 

This is inside the proof. I guess `catch_unwind` s not even possible there as zkvm program is compiled to abort on panic.

iljakuklic (commented):
Some more nits:

crates/pessimistic-proof-core/src/multi_batch_header.rs
hadjiszs: ah I'm not sure but I think the idea would be to move to rkyv for here, some PRs are opened for that

crates/agglayer-types/src/aggchain_data/multisig.rs
hadjiszs: no external requirements, I think that would add some difficulties to the integration with the aggsender

also not entirely sure that it'd make the code easier on the agglayer-node, I think we'd just have the same number of possible error for example, and instead of indexing ourselves in the agglayer-node, we'd verify that it is properly indexed, so I dont expect the code to be very different, and at the cost of having more constraints on the data that we receive from the chain

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -0,0 +1,146 @@
...
+    MissingSignature,
+}
+
+// Generate the prover inputs from the chain payload and the L1 context.
+impl TryInto<core::AggchainData> for PayloadWithCtx<Payload, Ctx> {
+    type Error = AggchainDataError;
+
+    fn try_into(self) -> Result<core::AggchainData, Self::Error> {
+        let PayloadWithCtx(payload, ctx) = self;
+        Ok(match (payload, ctx) {
```
polykeith: Nit: I find it a bit difficult to reason about closed-over states like this, so instead of `Ok(match foo { pattern => ... })`, I think it makes more sense to do `match foo { pattern => Ok( ... ) }`.

crates/agglayer-types/src/aggchain_data/mod.rs
polykeith: Is the number of `PayloadType`s a closed set? If so, an enum is generally used over a trait, since traits allow for an open set.

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
Ekleog-Polygon: Beyond contiguous access, @iljakuklic's suggestion would allow not actually having the `seen` array at all — you could just check that each index is strictly greater than the previous one, and it'd be enough :smile: 

So just having the constraint that eg. in the submitted certificate the signers must be sorted by increasing signing keys (because signing keys are just u256 so they can be sorted) would likely remove the malleability @iljakuklic mentions :+1:

crates/agglayer-types/src/aggchain_data/multisig.rs
Ekleog-Polygon: FWIW "at the cost of having more constraints on the data that we receive from the chain" is actually an advantage to me: the fewer things the chain can manipulate, the fewer opportunities for attacks it has

crates/pessimistic-proof-core/src/multi_batch_header.rs
Ekleog-Polygon: Oh yes I had forgotten! Got it, then considering the expected perf wins it probably makes most sense to just use rkyv, if it works from a security standpoint :+1:

crates/agglayer-types/src/aggchain_data/mod.rs
iljakuklic: > Is the number of `PayloadType`s a closed set? If so, an enum is generally used over a trait, since traits allow for an open set.

Good point

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
iljakuklic: Just occurred to me: another way of achieving the same thing is to have the witness be a `Vec<Option<Signature>>`. The positions in this vec correspond to the positions in the `signers` vec, and it has to be of the same length, so you can get pairs of signers and corresponding signatures by `zip`-ing them together. Number of `Some(_)` entries is required to be `threshold`. It does not fundamentally change the proposal, just a different way of representing it.

crates/agglayer-types/src/aggchain_data/multisig.rs
hadjiszs: >FWIW "at the cost of having more constraints on the data that we receive from the chain" is actually an advantage to me: the fewer things the chain can manipulate, the fewer opportunities for attacks it has

"more constraints on the data" here refers more to the expected format, it adds redundancy on the data to submit

If we want to abstract a bit to emphasize the differences, I think we can outline like that:

option 1 (the one I went for)

- receive from chain: `a`
- fetch from L1: `b`
- `verify(f(a, b))`

option 2

- receive from chain: `f(a, b')`
- fetch from L1: `b`
  - necessary, we cannot rely on what the chain submits, we need to cross check no matter what with the list registered in the L1
- verify that `f(a, b') == f(a, b)`
- `verify(f(a, b))`

where
- `a` are the signatures
- `b` (resp. `b'`) are the expected signers from the L1 (resp. from the chain)
- `f` being whatever function to generate the layout that we expect for the prover

----

note that the API change which is about to be merged in interop to unblock the aggkit team goes with `option 1` as well

I can add it as topic to discuss in our next Monday meeting if we feel that we want to move to option 2, cc @agglayer/agglayer-developers

hadjiszs: Outcome from team meeting:

I got convinced by @iljakuklic approach to move to the option 2 above. The main argument from @iljakuklic which convinced me is that the introduced redundancy might introduce more possible error cases but that's actually something desirable, to have more explicit and finer granularity on the errors

Following steps:

- Update the API to cover the `option 2`
- Keep the current PR with `option 1` and do the move to `option 2` on following PRs

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
hadjiszs: yeah I'll iterate on this on a following PR, especially given the move to the option 2 from there https://github.com/agglayer/agglayer/pull/963#discussion_r2301037810

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
hadjiszs: will iterate on this part on the following PR where I'll move to the option 2 defined here: https://github.com/agglayer/agglayer/pull/963#discussion_r2301037810

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
hadjiszs: seems we are aligned for `module_name::Context` in this comment: https://github.com/agglayer/agglayer/pull/963#discussion_r2298605509

so I guess for consistency `module_name::Payload` pairs well with that.

Note that outside the crates, they are renamed. For example `aggchain_proof::Payload` is exposed as `AggchainProofPayload` in `crates/agglayer-types/src/aggchain_data/mod.rs`

crates/agglayer-types/src/aggchain_data/global.rs
hadjiszs: replied there: https://github.com/agglayer/agglayer/pull/963#discussion_r2302682004

crates/agglayer-types/src/aggchain_data/multisig.rs
polykeith: In option 2, where exactly does `a` come from?

hadjiszs: @polykeith from our discussion, we'd receive a `map usize => Signature`, where the `usize` corresponds to the signer index in the ordered list registered in the L1.

the `a` corresponds to the signatures, so in option 2 in my example it simply corresponds to the map values received from the chain

notations aren't very precise sorry for that, that was to quickly emphasize the differences between the two approaches, at the protocol design level

crates/agglayer-types/src/aggchain_data/global.rs
hadjiszs: `EventDiscriminants` I was unable to find this type in the `Display` implementation you shared

in the end, I went with deriving their `strum_macros::Display` which allows to not have to implement any display

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -78,69 +77,48 @@ pub enum AggchainDataError {
...
     type Error = AggchainDataError;
 
     fn try_into(self) -> Result<core::AggchainData, Self::Error> {
         let PayloadWithCtx(payload, ctx) = self;
-        Ok(match (payload, ctx) {
-            (Payload::LegacyEcdsa { signature }, Ctx::LegacyEcdsa { signer }) => {
-                core::AggchainData::LegacyEcdsa { signer, signature }
+        match (payload, ctx) {
+            (Payload::LegacyEcdsa { signature }, Context::LegacyEcdsa { signer }) => {
+                Ok(core::AggchainData::LegacyEcdsa { signer, signature })
```
Ekleog-Polygon, line 87: FWIW I'm more used to seeing the version with the Ok outside of the match in Rust code, but :shrug:

Ekleog-Polygon (commented):
Out of curiosity as I wasn't in the meeting last night-JST, what led to the choice of `SignerIndex -> Signature` rather than `Signer -> Signature` for the signature list? I'd have thought the signer to be as good an identifier as the signer index, and to have less risk of getting stale whenever there's a change of signer list on L1

Ekleog-Polygon (dismissed):
After chatting with @hadjiszs I now understand the reason :smile:

atanmarko (dismissed):

polykeith (dismissed):

polykeith (approved):

atanmarko (approved):

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub fn aggchain_vkey_from_proof(&self) -> SP1VerifyingKey {
+        let crate::aggchain_proof::Proof::SP1Stark(sp1_reduce_proof) = &self.proof;
+        sp1_reduce_proof.vkey.clone()
+    }
+}
+
+/// Aggchain proof data from the L1 and enforced by the agglayer.
+#[derive(Clone, Debug)]
+pub struct Context {
+    pub aggchain_vkey: [u32; 8],
```
dwgrantham, line 31: This is a naked type we should consider wrapping in a newtype so that conversions are explicit and managed.

crates/agglayer-types/src/aggchain_data/global.rs
```diff
@@ -0,0 +1,122 @@
...
+    },
+}
+
+#[derive(Clone, Debug, Error, Deserialize, Serialize, Eq, PartialEq)]
+pub enum AggchainDataError {
+    #[error("Invalid variant: {0}")]
+    InvalidVariant(String),
+
+    #[error("Invalid multisig: {0}")]
+    InvalidMultisig(#[source] MultisigError),
```
dwgrantham, line 73: Use `#[from]` instead as it generates `From<MultisigError> for InvalidMultisig` and also implicitly sets the source value as well.

Issue #942, opened 2025-08-01
Store multiple signatures with the Certificate


hadjiszs:
### Description

We want to accept multiple ECDSA signatures coming from the chain to authenticate the Certificate and its content, to be verified by the agglayer and the PP.

### Notes

- A set of signatures coming with the certificate have to be on the exact same commitment (unlike in phase 2 where the commitment was different)
  - the commitment is defined [here](https://github.com/agglayer/protocol-research/issues/169#issuecomment-3136968862)

### Questions

1. What is the maximal number of multisig signers which can be registered on the L1 for a given network? This will tell us the maximal number of signatures that we should accept, cc @invocamanman 
2. How do we want to get the signatures from the chain?
   - grpc metadata field (no api change)
   - additional grpc request parameter, aside of the certificate
   - within the certificate itself (db questions)
3. Can we afford doing a cleanup on the legacy ways to provide the signature (`certificate.signature`, `certificate.aggchain_data.signature`, gRPC metadata field)

Ekleog-Polygon:
This is specifically about how we want to internally store the certificate

PR #978, opened 2025-08-21
feat: support storing multisig data

Ekleog-Polygon:
Fixes https://github.com/agglayer/agglayer/issues/942

AFAICT all the `todo!()` added here are already handled in one of @hadjiszs's two multisig PRs.


crates/agglayer-storage/src/types/certificate/tests/mod.rs
iljakuklic, line 1: Could you add `#[case]` to the aggchain data and certificate tests to cover `Multisig` and `GenericWithMultisig`? In particular to `encoding` and to `encoding_roundtrip_consistent_with_into`.

iljakuklic (approved):
These changes look fine to me in terms of correctness.

crates/agglayer-storage/src/types/certificate/tests/mod.rs
Ekleog-Polygon: Great point! It's now done :smile:

Freyskeyd (approved):

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -214,6 +212,17 @@ pub enum AggchainDataV1<'a> {
         signature: Option<Box<Signature>>,
         public_values: Cow<'a, Box<AggchainProofPublicValues>>,
     },
+
+    Multisig {
+        multisig: Cow<'a, [Option<Signature>]>,
+    },
+
+    GenericWithMultisig {
```
hadjiszs: nit: maybe we want to align the naming with the ones in `AggchainData` defined in `interop`, which name this struct as `MultisigAndAggchainProof`

hadjiszs (approved):

crates/agglayer-storage/src/types/certificate/mod.rs
Ekleog-Polygon: Great point! I just fixed that, hadn't thought of updating the names to the latest ones :smile:

Freyskeyd (approved):

PR #979, opened 2025-08-24
feat: remove extra certificate signer

### LINKED ISSUE #923: Find different terminology for the extra signature verified by the agglayer
The extra certificate signature got introduced in:
- #882 
- #920

the naming is somewhat confusing for some, partially due to the fact that the whole feature is partially implemented, making the abstraction not very great to begin with.

The ticket is to rename the occurrences of `extra certificate signature`, it might be easier once we get closer to the final stage of the multisig/threshold feature.

hadjiszs:
Fixes #923 

CONFIG-CHANGE: Removal of the `[extra-certificate-signer]` config field.

BREAKING-CHANGE: Cannot set an `extra-certificate-signer` anymore,
  feature replaced by the upcoming multisig.

### DIFF
--- a/crates/agglayer-config/src/lib.rs
+++ b/crates/agglayer-config/src/lib.rs
@@ -125,14 +125,6 @@ pub struct Config {
 
     #[serde(default, skip_serializing_if = "crate::is_default")]
     pub grpc: GrpcConfig,
-
-    /// Extra Certificate signer per network.
-    /// Signatures is expected to be performed on the same commitment as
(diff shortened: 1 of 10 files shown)


atanmarko (approved):

hadjiszs:
> LGTM! But IIUC today's meeting, we don't actually want to land this for now?

yeah right, I move it as draft then

hadjiszs:
>yeah right, I move it as draft then

even closing it, will re-open when needed

PR #980, opened 2025-08-24
feat: bring api update for multisig

hadjiszs:
Put several placeholders just to expose the new API.

For now, the requests with multisig are errored just to have the API exposed.


Cargo.toml
```diff
@@ -44,13 +44,21 @@ pessimistic-proof-core = { path = "crates/pessimistic-proof-core" }
...
-agglayer-primitives = "0.9.0"
-agglayer-tries = "0.9.0"
-unified-bridge = "0.9.0"
+agglayer-bincode = { git = "https://github.com/agglayer/interop.git", branch = "multisig-api" }
+agglayer-interop = { git = "https://github.com/agglayer/interop.git", branch = "multisig-api" }
+agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", branch = "multisig-api" }
+agglayer-elf-build = { git = "https://github.com/agglayer/interop.git", branch = "multisig-api" }
+agglayer-primitives = { git = "https://github.com/agglayer/interop.git", branch = "multisig-api" }
+agglayer-tries = { git = "https://github.com/agglayer/interop.git", branch = "multisig-api" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "multisig-api" }
```
Ekleog-Polygon, line 53: Note: must use a released version before merging

crates/agglayer-storage/src/types/certificate/tests/snapshots/agglayer_storage__types__certificate__tests__structure__structure_snapshot.snap
```diff
@@ -604,6 +603,9 @@ snapshot_kind: text
             }
           ]
         }
+      },
+      "14": {
+        "UnsupportedMultisig": "UNIT"
```
Ekleog-Polygon, line 608: Hmm... Considering this seems to be entering our backward compatibility guarantees... I don't feel very good landing this PR, because we don't want to commit to such an error code I think? 🥲

Ekleog-Polygon (commented):
I'm fine with the idea of having the scaffolding ready first, but the error type added to our compat guarantees makes me feel awkward.

Maybe we should just land that with the actual implementation?

hadjiszs:
closing in favor of #984

Issue #941, opened 2025-08-01
Fetch the threshold and the set of signers from the L1 for the given network


hadjiszs:
From the L1 we should get:

- the threshold (integer): it designates the minimal number of signature that should be successfully verified for the given network
- the ordered list of signers addresses

Notes:

- It is important to fetch the entire list of signers and to preserve the same ordering as in the smart contract. Because it impacts the hash on the signers set which will be involved in the public inputs of the PP.

Questions:

- Will it still be from the rollup manager address or from another contract?
- Should we monitor specific L1 smart contract events to ensure finality on modifications of the signers set?

invocamanman:
1. It will be done in the "aggchain" contracts ( e.g same contract we fetch trusted sequencer). This is done because each chain can have their own thresholds and their own signers
2. Yes we should monitor that, maybe similar to now that , if i'm not mistaken, once receive a certificate we fetch some parameters from those smart contracts. Even tho without config versioning, if the signer set changes (or threshold) certificate will get invalidated

PR #984, opened 2025-08-25
feat: add multisig verification at the rpc level

hadjiszs:
Verify the multisig (if any) at the rpc level, when receiving the request.

> [!IMPORTANT]  
> Modifications in `agglayer-storage` are in `todo!()`


crates/agglayer-rpc/Cargo.toml
```diff
@@ -10,6 +10,7 @@ agglayer-storage.workspace = true
 agglayer-config.workspace = true
 agglayer-rate-limiting.workspace = true
 agglayer-types.workspace = true
+pessimistic-proof.workspace = true
```
Freyskeyd: Why do we need this deps here?

hadjiszs: I removed

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -373,11 +373,32 @@ where
...
             .get_trusted_sequencer_address(u32::from(cert.network_id))
             .await?;
 
-        cert.verify_cert_signature(sequencer_address)
-            .map_err(SignatureVerificationError::from_signer_error)
+        let multisig_ctx = MultisigCtx {
+            signers: Default::default(), // TODO: to fetch from L1
+            threshold: 1,                // TODO: to fetch from L1
+            prehash: cert.signature_commitment_values().multisig_commitment(),
+        };
```
Ekleog-Polygon, line 386: Nit: make that into another function that we can implement later (and either add a `todo!()` or open an issue so we don't forget about these `// TODO` comments — we can easily grep for `todo!()` but there's already quite a few `// TODO` comments :smiling_face_with_tear: )

crates/agglayer-rpc/src/lib.rs
```diff
@@ -324,8 +326,27 @@ where
                 SignatureVerificationError::UnableToRetrieveTrustedSequencerAddress(cert.network_id)
             })?;
 
-        cert.verify_cert_signature(sequencer_address)
-            .map_err(SignatureVerificationError::from_signer_error)
+        let multisig_ctx = MultisigCtx {
+            signers: Default::default(), // TODO: to fetch from L1
+            threshold: 1,                // TODO: to fetch from L1
+            prehash: cert.signature_commitment_values().multisig_commitment(),
+        };
```
Ekleog-Polygon, line 333: (and we can reuse the same function here, it'll be a single place to fix later)

crates/agglayer-rpc/src/lib.rs
```diff
@@ -324,8 +326,27 @@ where
...
+                cert.verify_aggchain_proof_signature(sequencer_address, signature)
+            }
+            AggchainData::MultisigOnly(multisig) => {
+                cert.verify_multisig(multisig.into(), multisig_ctx)
+            }
+            AggchainData::MultisigAndAggchainProof { multisig, .. } => {
+                cert.verify_multisig(multisig.into(), multisig_ctx)
+            }
+        }
+        .map_err(SignatureVerificationError::from_signer_error)
```
Ekleog-Polygon, line 349: Actually this whole match is also duplicated between json-rpc and rpc. Why not just keep all that inside `verify_cert_signature` as it was before?

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -27,9 +27,19 @@ pub struct Payload {
...
-impl From<Vec<Signature>> for Payload {
-    fn from(signatures: Vec<Signature>) -> Self {
-        Self { signatures }
+impl From<&[Signature]> for Payload {
+    fn from(signatures: &[Signature]) -> Self {
+        Self {
+            signatures: signatures.to_vec(),
+        }
+    }
+}
```
Ekleog-Polygon: Nit: This will introduce an unneeded allocation and copy whenever there's already a vec. Maybe just keep the two? A `From` implementation doesn't cost much.

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -27,9 +27,19 @@ pub struct Payload {
...
+
+impl From<&agglayer_interop_types::aggchain_proof::MultisigPayload> for Payload {
+    fn from(value: &agglayer_interop_types::aggchain_proof::MultisigPayload) -> Self {
+        Self {
+            signatures: value.0.clone().into_iter().flatten().collect(),
```
Ekleog-Polygon, line 47:
```suggestion
            signatures: value.0.clone(),
```

Nit: Would this not be enough?

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -157,7 +167,7 @@ mod tests {
         let signatures = signatures_from_indices(&signer_indices, &prehash);
 
         let payload_with_ctx = PayloadWithCtx(
-            Payload::from(signatures),
+            Payload::from(signatures.as_slice()),
```
Ekleog-Polygon:
```suggestion
            Payload::from(signatures),
```

Nit: as a consequence of the above suggestion

crates/agglayer-types/src/certificate/mod.rs
```diff
@@ -151,71 +154,83 @@ impl Certificate {
...
+        let signature = signature.as_ref().ok_or(SignerError::Missing)?;
+        let commitment = signature_commitment_values.commitment(SignatureCommitmentVersion::V4); // NOTE: will be upgraded to V5 eventually
+        let recovered = signature
             .recover_address_from_prehash(&commitment)
-            .map_err(SignerError::Recovery)
+            .map_err(SignerError::Recovery)?;
+
+        (recovered == expected_signer)
+            .then_some(())
+            .ok_or(SignerError::InvalidPessimisticProofSignature { expected_signer })
```
Ekleog-Polygon: Personal opinion: to me this is easier to read

```suggestion
        if recovered != expected_signer {
            return Err(SignerError::InvalidPessimisticProofSignature { expected_signer });
        }
        
        Ok(())
```

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
hadjiszs: yeah that's because it'd have been a placeholder method of the l1 rpc traits, and didn't want to bother modifying those traits which are systemic changes in the mock and tests iirc, and normally #941 should come soon so I didnt want to bother now on this

crates/agglayer-rpc/src/lib.rs
hadjiszs: replied there https://github.com/agglayer/agglayer/pull/984#discussion_r2309310229

crates/agglayer-types/src/aggchain_data/multisig.rs
hadjiszs: will try in a follow up PR 👍

crates/agglayer-types/src/aggchain_data/multisig.rs
hadjiszs: hm not enough I think? under the wrappers my intent is to get a `Vec<Signature>` out of a `Vec<Option<Signature>>` basically

crates/agglayer-types/src/certificate/mod.rs
hadjiszs: will do 👍

crates/agglayer-rpc/src/lib.rs
hadjiszs: yeah I tried but I deleted because the semantic of the function felt weird if being as a method `verify_cert_signature` of `Certificate` as before

it would be something like:

```rust
impl Certificate {
   fn verify_cert_signature(&self, trusted_sequencer: Address, multisig_ctx: Option<MultisigContext>) {

    // here some match on the `self.aggchain_data`
    // and ensure that we have `Some(_)` for `multisig_ctx` for the branches of aggchain_data which involves a multisig
   }
}
```

I would have to add and return new errors just in case of inconsistent cases like `None` multisig context while we expect one because of the `certificate.aggchain_data` which involve one multisig etc

Note that in the current function in `agglayer-rpc`, `MultisigCtx` will have to be fetched only in case that it involves a multisig. In particular, in the case of the legacy ecdsa we wont fetch that because no multisig committee will be registered in the L1, hence the Option

So given the succinctness of the match, I felt clearer and simpler to have it as-is in `agglayer-rpc`

Ekleog-Polygon: > Note that in the current function in agglayer-rpc, MultisigCtx will have to be fetched only in case that it involves a multisig. In particular, in the case of the legacy ecdsa we wont fetch that because no multisig committee will be registered in the L1, hence the Option

Ooh, I thought you were planning on unconditionally fetching the `MultisigContext`, like in the current code. Then I understand better the changes, but I'm still sad to see this duplication. Maybe `verify_cert_signature` should just take the `L1Rpc` argument directly, to avoid that duplication?

hadjiszs: it means I need to bring all the trait bounds machinery coming with `L1Rpc` in the `Certificate` implementation blocks no? :/

I dont know if we want to see L1 fetching interactions in `agglayer-types`

hadjiszs: one thing to note is that the function `verify_cert_signature` is duplicated from the beginning prior to this PR in `agglayer-jsonrpc-api`, and is solely involved in the tests, so it makes me wonder whether it shouldnt be the tests to be refactored instead

Ekleog-Polygon: Oh right we're in agglayer-types here. Then one option would be to take in `impl FnOnce() -> MultisigContext` as parameter in `verify_cert_signature`, and have the caller provide it with the L1Rpc machinery :smile:

crates/agglayer-types/src/aggchain_data/multisig.rs
Ekleog-Polygon: Oh right sorry I forgot to check the exact types :sweat_smile:

Ekleog-Polygon: Though as a follow-up nit we could have `value.0.iter().flatten().cloned().collect()` to avoid cloning the `None` parts of the vec :innocent:

crates/agglayer-rpc/src/lib.rs
hadjiszs: will experiment that in following pr 👌

atanmarko (approved):

PR #985, opened 2025-08-25
test: enhance `AggchainData` storage decoding tests

iljakuklic:
In addition to taking snapshots of aggchain data encoding, also decode the bytes back and check the value. This would be better done using fuzzing but this is still an improvement.


crates/agglayer-storage/src/types/certificate/tests/mod.rs
```diff
@@ -260,10 +265,22 @@ fn encoding_roundtrip_consistent_with_into(#[case] orig: impl Into<Certificate>
...
     insta::assert_snapshot!(name, bytes);
+
+    // Also check decoding must produce the same value.
+    let from_bytes: T = bincode::default()
+        .deserialize(bytes.as_ref())
+        .expect("deserialization failed");
+
+    // This should really compare the certificates directly but that requires adding
+    // whole bunch of `Eq` impl to many types.
+    assert_eq!(format!("{from_bytes:?}"), format!("{value:?}"));
```
Ekleog-Polygon, line 283: Hmmm… can we compare the hashes instead here, maybe? The issue with debug repr is that often it's missing quite a bit of details :smiling_face_with_tear:

Ekleog-Polygon (approved):
I'm totally with you that this seems like a prime example for fuzzing!

The main complexity there would likely be that we want to compare behavior with previous versions, which means we'll have to make the crate dev-depend on itself at a previous version.

Then, we can have two differential fuzzers:
- One that takes random bytes, and says "if old version can deserialize it, then new version can also deserialize it with the same outcome
- One that takes an arbitrary value, serializes it and deserializes it with the new version, and asserts it's the same value. And then it tries to serialize it with the old version, and if it works then it must also deserialize with the same value

We could then fuzz against multiple old versions relatively easily once this is done, and compare the current behavior with all previous versions. The main issue would likely be technically-backward-incompatible changes that we made while knowing what we were doing (eg. NetworkId u32::MAX), which would need manual whitelisting.

crates/agglayer-storage/src/types/certificate/tests/mod.rs
iljakuklic: > Hmmm… can we compare the hashes instead here, maybe? The issue with debug repr is that often it's missing quite a bit of details 🥲

Yea this is in a number of places in this file already. For hashes, we need to find a serialization format that would be hashed, but serialization is (in part) what is being tested so any serialization based on `serde` should be avoided. A proper `Eq` would be better but there is an issue as per the comment. Will try  to put `Eq` on the serialization structs, using hash for things that we don't have control over, and see how that goes.

iljakuklic: For now, I just filed an issue for this:
* https://github.com/agglayer/agglayer/issues/989

PR #991, opened 2025-08-28
feat: move all dynamic error types to eyre

Ekleog-Polygon:
Fixes https://github.com/agglayer/security/issues/80

BREAKING-CHANGE: Error types have been migrated to eyre in the API. No expected user-facing change unless the binary enables color-eyre.

### DIFF
--- a/crates/agglayer-grpc-types/src/compat/v1/snapshots/agglayer_grpc_types__compat__v1__tests__bad_data__debug.snap
+++ b/crates/agglayer-grpc-types/src/compat/v1/snapshots/agglayer_grpc_types__compat__v1__tests__bad_data__debug.snap
@@ -1,5 +1,8 @@
 ---
 source: crates/agglayer-grpc-types/src/compat/v1/tests.rs
-expression: "format!(\"{:?}\", anyhow::Error::from(error))"
+expression: "format!(\"{:?}\", eyre::Error::from(error))"
 ---
 invalid value
+
+Location:
+    crates/agglayer-grpc-types/src/compat/v1/tests.rs:21:25
--- a/crates/agglayer-node/src/logging.rs
+++ b/crates/agglayer-node/src/logging.rs
@@ -1,8 +1,7 @@
 use agglayer_config::log::LogFormat;
-use anyhow::Result;
 use tracing_subscriber::{prelude::*, util::SubscriberInitExt, EnvFilter};
 
-pub(crate) fn tracing(config: &agglayer_config::Log) -> Result<()> {
+pub(crate) fn tracing(config: &agglayer_config::Log) -> eyre::Result<()> {
     // TODO: Support multiple outputs.
     let writer = config.outputs.first().cloned().unwrap_or_default();
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,9 +48,9 @@ dependencies = [
  "agglayer-prover",
  "agglayer-prover-config",
  "agglayer-storage",
- "anyhow",
  "assert_cmd",
  "clap",
+ "color-eyre",
  "dirs",
  "dotenvy",
  "eyre",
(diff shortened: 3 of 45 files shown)


atanmarko (approved):

crates/agglayer-grpc-types/src/compat/v1/snapshots/agglayer_grpc_types__compat__v1__tests__bad_data__debug.snap
```diff
@@ -1,5 +1,8 @@
 ---
 source: crates/agglayer-grpc-types/src/compat/v1/tests.rs
-expression: "format!(\"{:?}\", anyhow::Error::from(error))"
+expression: "format!(\"{:?}\", eyre::Error::from(error))"
 ---
 invalid value
+
+Location:
+    crates/agglayer-grpc-types/src/compat/v1/tests.rs:21:25
```
Freyskeyd: Is it expected to have that as part of the gRPC error at the end?

crates/agglayer-node/src/logging.rs
```diff
@@ -1,8 +1,7 @@
 use agglayer_config::log::LogFormat;
-use anyhow::Result;
 use tracing_subscriber::{prelude::*, util::SubscriberInitExt, EnvFilter};
 
-pub(crate) fn tracing(config: &agglayer_config::Log) -> Result<()> {
+pub(crate) fn tracing(config: &agglayer_config::Log) -> eyre::Result<()> {
```
Freyskeyd, line 4: Is there a particular reason to not just replace `use anyhow::Result;` by `use eyre::Result;` ?
It could have make less code changes right?

crates/agglayer-grpc-types/src/compat/v1/snapshots/agglayer_grpc_types__compat__v1__tests__bad_data__debug.snap
Ekleog-Polygon: I didn't do it on purpose, but when `cargo insta` sent me that diff, I was like
> :shrug: that'd probably help us for debugging whenever aggkit will send us the error message

Do you expect it to be a problem? (Actually I'm just now thinking that it'd probably be a problem for the snapshot tests themselves, because the location would change; so I'll need to find some way of censoring the location in the snapshot test)

crates/agglayer-node/src/logging.rs
Ekleog-Polygon: It would, that mostly comes down to a matter of taste — by not importing `eyre::Result` I'm making it obvious what the error type is.

My personal opinion is that importing another crate's `Result` alias is often an issue, because the reader doesn't actually know which other crate it comes from.

Do you have a strong preference for importing `eyre::Result`? If yes I can definitely revert that change!

crates/agglayer-grpc-types/src/compat/v1/snapshots/agglayer_grpc_types__compat__v1__tests__bad_data__debug.snap
Freyskeyd: I think that you need to implement a custom handler for eyre to achieve that.

crates/agglayer-node/src/logging.rs
Freyskeyd: No, I don't have strong preference nor opinion on this :)

crates/agglayer-grpc-types/src/compat/v1/snapshots/agglayer_grpc_types__compat__v1__tests__bad_data__debug.snap
Ekleog-Polygon: I ended up simply censoring away the location part of the error messages in the tests — it seemed easier than a custom eyre handler, and as it's only for tests we don't care that much about stability :smile:

atanmarko (dismissed):

atanmarko (approved):

Freyskeyd (approved):

PR #994, opened 2025-08-28
feat: use hardcoded default metadata value

### LINKED ISSUE #992: Use hardcoded constant value for the Certificate metadata field
We remove the `metadata` field from the agglayer API.

The `certificate.metadata` will still be present in the `Certificate` payload in the storage, and in the `certificate_id` calculation.

As a consequence, we should consider the same constant value on both sides: aggsender and agglayer.

hadjiszs:
- Use hardcoded default metadata value (zeros) if none is provided from the API.
  - The API keep accepting metadata value for backward compatibility
- `CertificateId` calculation enforces the default value.
  - Not breaking because the deployed aggsenders do not depend on the certificate id calculation.

Fixes #992


Freyskeyd (dismissed):

crates/agglayer-types/src/certificate/mod.rs
```diff
@@ -79,7 +79,7 @@ impl Certificate {
             self.new_local_exit_root.as_ref(),
             commit_bridge_exits.as_slice(),
             commit_imported_bridge_exits.as_slice(),
-            self.metadata.0.as_slice(),
+            Metadata::DEFAULT.as_slice(),
```
Ekleog-Polygon: Hmm… I think this will break hash computation (and thus certificate id) for all old certificates, won't it?

We could add an `if` here to solve this, but I thought we wanted to see RPC enforce that any incoming certificate has empty metadata instead?

(Which'd also mean we can't keep `Digest::ZERO` as the default value, because it's certainly not the result of hashing empty metadata.)

hadjiszs: >but I thought we wanted to see RPC enforce that any incoming certificate has empty metadata instead?

this would be breaking because we have the requirement to be backward compatible with already deployed aggsenders, so it should still be doable to declare a metadata field, and we should expose it accordingly

also one thing is that the already deployed aggsenders are agnostic from the certificate id calculation, as it isnt included in the message they sign

rather, the idea is to enforce a default value if the metadata isn't provided (which should be the case with the upcoming version of aggsender)

>I think this will break hash computation (and thus certificate id) for all old certificates

yeah it is true that it'd break the lookup of already stored certificate :/ actually just writing the default metadata value in the storage if none is received at the RPC level, without changing the certificate id calculation should be enough in the end, do you agree @agglayer/agglayer-developers  ?

atanmarko (dismissed):

Freyskeyd (approved):

crates/agglayer-types/src/certificate/mod.rs
Ekleog-Polygon: That's basically what I'm suggesting; but in order to do that we'll need to have the default metadata value be the hash of empty metadata, and not a simple all-0 hash 😄

hadjiszs:
> I think there's one main thing missing here: AFAIR, we don't want any aggsender on the new multisig scheme to be able to submit a certificate with actual metadata, or that could cause all the certificate duplication issues we're trying to fix with the removal.
> 
> So I think we should add some check that ensures that if we're on a network with the multisig scheme, then its metadata must be `Metadata::default`. And hopefully we can remove that check as soon as we're done upgrading old aggsenders.
> 
> Does that make sense to you?

as discussed I'll bring this extra check in another PR later today

hadjiszs:
>as discussed I'll bring this extra check in another PR later today

- https://github.com/agglayer/agglayer/pull/1003

PR #997, opened 2025-08-28
chore: use version dependencies for interop crates

polykeith:
This PR simply removes the usage of git dependencies on the interop cratest and instead uses v0.10.0 for interop crates.


Ekleog-Polygon (changes requested):
See https://github.com/agglayer/agglayer/pull/991 for the changes I think we'd need for tests to pass.

polykeith:
Closing in favor of #991.

Issue #995, opened 2025-08-28
Update the PP for the latest alignments with L1 for multisig


hadjiszs:
- accept V5 signature to migrate to PP root V3
- merge the case `AggchainProofOnly` into the case `MultisigAndAggchainProof`
  - verify the signature coming in this case within the PP
  - expect a V5 commitment for this signature instead of a V4 to verify this single signature as a multisig 1-of-1
- get rid of consensus type 0
- aligns the commitments involved in the aggchain hash with test vector generated by the contracts team (defined [there](https://github.com/agglayer/specs/blob/b16acc09c3550946122f11fc9f51efb266e5e8c9/specs/agglayer_consensus_types.md))
- aligns the default values involved in the aggchain hash (defined [there](https://github.com/agglayer/specs/blob/b16acc09c3550946122f11fc9f51efb266e5e8c9/specs/agglayer_consensus_types.md))

PR #999, opened 2025-08-29
feat: update the PP to align with the smart contracts for multisig

hadjiszs:
- accept V5 signature to migrate to PP root V3
- merge the case `AggchainProofOnly` into the case `MultisigAndAggchainProof`
  - verify the signature coming in this case within the PP
  - expect a V5 commitment for this signature instead of a V4 to verify this single signature as a multisig 1-of-1
- get rid of consensus type 0
- aligns the commitments involved in the aggchain hash with test vector generated by the contracts team (defined [there](https://github.com/agglayer/specs/blob/b16acc09c3550946122f11fc9f51efb266e5e8c9/specs/agglayer_consensus_types.md))
- aligns the default values involved in the aggchain hash (defined [there](https://github.com/agglayer/specs/blob/b16acc09c3550946122f11fc9f51efb266e5e8c9/specs/agglayer_consensus_types.md))

Fixes https://github.com/agglayer/agglayer/issues/995


crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -1,85 +1,67 @@
...
+    pub const EMPTY_MULTISIG_HASH: Digest = Digest(hex!(
+        "290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563"
+    ));
 
-    /// Returns the empty vkey used in case of no aggchain proof
-    pub fn empty_sp1_vkey() -> Vkey {
-        Vkey::default()
-    }
+    /// Value if no aggchain vkey.
+    pub const EMPTY_AGGCHAIN_VKEY: Digest = Digest::ZERO;
```
atanmarko:
```suggestion
    pub const EMPTY_AGGCHAIN_VKEY_HASH: Digest = Digest::ZERO;
```

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -30,16 +30,18 @@ pub enum MultisigError {
...
-    pub fn signers_commit(&self) -> Digest {
-        // addresses does not need to be padded in 32bytes because
-        // addresses won't be decoded in solidity, just encoded from the contract
-        // storage.
-        let mut buf = Vec::with_capacity(self.expected_signers.len());
-        buf.extend_from_slice(self.threshold.to_be_bytes().as_slice());
-        for addr in &self.expected_signers {
-            buf.extend_from_slice(addr.as_slice()); // address (20bytes)
+    pub fn multisig_hash(&self) -> Digest {
+        let mut buf = Vec::with_capacity(32 + 20 * self.expected_signers.len());
```
atanmarko: Would be nice to define some constants for those magic numbers.

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
atanmarko: Just to be consistent - on one place we refer to it as `aggchain_vkey_hash`, on another as aggchain_vkey

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -1,85 +1,67 @@
...
-        signers_commit: Option<Digest>,
-        threshold: Option<usize>,
+        multisig_hash: Digest,
     },
 }
 
 impl From<&AggchainHashValues> for ConsensusType {
     fn from(value: &AggchainHashValues) -> Self {
         match value {
-            AggchainHashValues::ConsensusType0 { .. } => Self(0),
```
Ekleog-Polygon, line 26: Are we fine with the backward-compatibility implications of killing this consensus type?

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -1,85 +1,67 @@
...
 }
 
 impl AggchainHashValues {
-    /// Returns the commitment on signers in case of no signer.
-    pub fn empty_signers() -> Digest {
-        keccak256_combine([Digest::default()])
-    }
+    /// Value if no multisig.
+    pub const EMPTY_MULTISIG_HASH: Digest = Digest(hex!(
+        "290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563"
```
Ekleog-Polygon, line 32: Can we have a test that ensures this is the value of the expected hash contents? And maybe a comment here pointing to said test, so readers don't just see a magic value with no explanation

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -1,85 +1,67 @@
...
+    pub fn hash(&self) -> Digest {
         let consensus_type: u32 = ConsensusType::from(self).0;
 
         match self {
-            AggchainHashValues::ConsensusType0 { signer } => {
-                keccak256_combine([&consensus_type.to_be_bytes(), signer.as_slice()])
-            }
             AggchainHashValues::ConsensusType1 {
-                aggchain_vkey,
+                aggchain_vkey: aggchain_vkey_u32,
```
Ekleog-Polygon, line 47: Random thought for the future, not now: we're having this kind of endianness writes at way too many places. We should probably introduce a wrapper type around vkey hashes so that we can easily get them in u32, u8, etc. formats without manual fiddling every time

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -89,23 +71,20 @@ impl AggchainHashValues {
 impl From<&AggchainData> for AggchainHashValues {
     fn from(value: &AggchainData) -> Self {
         match value {
-            AggchainData::LegacyEcdsa { signer, .. } => {
-                AggchainHashValues::ConsensusType0 { signer: *signer }
-            }
-            AggchainData::MultisigOnly(multi_signature) => AggchainHashValues::ConsensusType1 {
+            AggchainData::LegacyEcdsa { signer, signature } => AggchainHashValues::ConsensusType1 {
```
Ekleog-Polygon, line 75: Do I understand correctly that this change will require a coordinated update of PP, agglayer and maybe contracts to adjust the consensus type there?

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
```diff
@@ -89,23 +71,20 @@ impl AggchainHashValues {
...
-                signers_commit: Some(multi_signature.signers_commit()),
-                threshold: Some(multi_signature.threshold),
+                multisig_hash: MultiSignature {
+                    signatures: vec![(0, *signature)],
+                    expected_signers: vec![*signer],
+                    threshold: 1,
+                }
+                .multisig_hash(),
             },
-            AggchainData::AggchainProofOnly(AggchainProof {
```
Ekleog-Polygon, line 101: Same question about backward-compat implications here: what will removing this variant entail for the following stakeholders?
- Aggsenders who could still be sending aggchain-proof-only data (I think that's fine thanks to the conversion layer)
- Storage that could still have stored aggchain-proof-only certificates in one of the databases (I don't see any changes to agglayer-storage in this PR so I'm a bit worried, though I'm also surprised it even compiles at all)
- Contracts that could still have some chains configured in aggchain-proof-only mode

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
```diff
@@ -43,8 +42,6 @@ pub enum AggchainData {
     },
     /// Multisig only
     MultisigOnly(MultiSignature),
-    /// Aggchain proof only (stark proof)
-    AggchainProofOnly(AggchainProof),
```
Ekleog-Polygon, line 47: (for my comment above, this is the most scary part to me, among others because this struct is (de)serialize and we don't have the protobuf abstraction layer yet)

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
```diff
@@ -110,32 +103,24 @@ impl AggchainData {
...
-                    signature_commitment.commitment(SignatureCommitmentVersion::V2),
-                    signature,
-                )?
-            {
-                PessimisticRootCommitmentVersion::V2
-            } else {
-                return Err(ProofError::InvalidSignature);
-            }
+        let target_pp_root_version = if is_signed_with_version(SignatureCommitmentVersion::V3)?
+            || is_signed_with_version(SignatureCommitmentVersion::V5)?
```
Ekleog-Polygon, line 117: Just to confirm: there's currently no aggsender sending certificates with commitment version V4, as they would be broken by this change otherwise. Right? If I'm correct, can we just name the new one V4 instead of V5, to avoid skipping a number in our code?

crates/pessimistic-proof-test-suite/tests/pp_public_values.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        let computed = match inputs {
+            Inputs::Multisig(MultisigInputs { threshold, signers }) => MultiSignature {
+                signatures: vec![], // not needed
+                expected_signers: signers,
+                threshold: threshold as usize,
+            }
+            .multisig_hash(),
+            Inputs::Aggchain(AggchainInputs {
+                consensus_type: _, // not needed
```
Ekleog-Polygon: For this one, we should probably `assert_eq!(consensus_type, 1)`, because the test framework would be broken otherwise?

crates/pessimistic-proof-test-suite/tests/pp_public_values.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    for entry in entries {
+        let TestVector {
+            inputs,
+            expected_output,
+        } = entry;
+
+        let computed = match inputs {
+            Inputs::Multisig(MultisigInputs { threshold, signers }) => MultiSignature {
+                signatures: vec![], // not needed
```
Ekleog-Polygon: This `// not needed` comments feels wrong. Why is it part of the test files if it's actually not needed? Can we just remove it?

crates/pessimistic-proof-test-suite/tests/pp_public_values.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        assert_eq!(*expect, *computed);
+    }
+}
+
+// Convert [u8;32] to [u32;8] in big endian
+fn u8x32_to_u32x8_be(b: [u8; 32]) -> [u32; 8] {
+    core::array::from_fn(|i| {
+        u32::from_be_bytes([b[i * 4], b[i * 4 + 1], b[i * 4 + 2], b[i * 4 + 3]])
+    })
```
Ekleog-Polygon, line 52: Wasn't there a helper function for exactly this, that we were using in other parts of the codebase?

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s00_be000.snap
```diff
@@ -3,45 +3,45 @@ source: crates/pessimistic-proof-test-suite/tests/cycle-tracker.rs
...
-     4911 or
-     4789 and
-     4001 sltu
-     2743 auipc
-     1764 sub
-     1726 mul
-     1622 ecall
-     1592 mulhu
-     1101 jal
+gas: 2182636
```
Ekleog-Polygon, line 6: That's a 5-6% increase in total gas cost, feels like quite a lot. Do you know where it's coming from? (It seems to go up to 10% increase at some places below)

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s01_be020.snap
```diff
@@ -3,45 +3,45 @@ source: crates/pessimistic-proof-test-suite/tests/cycle-tracker.rs
...
-     37329 xor
-     32788 bltu
-     23046 jalr
-     17788 and
-     13594 or
-     11811 sll
-     11537 auipc
-     10252 srl
-      8012 sltu
+gas: 4052013
```
Ekleog-Polygon, line 6: This one seems to be the worst, at a 22% increase in gas cost

Ekleog-Polygon (commented):
I have a few questions regarding backward-compatibility listed below, because I'm missing some context on whether such changes are fine or not.
Also, I was quite surprised by the performance impact of the changes. Maybe something went wrong in the changes, that resulted in such drastic performance losses?

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
hadjiszs: yes, this change is the result of discussion + decision with the protocol team and their contracts

- https://github.com/agglayer/agglayer/issues/995
- https://github.com/agglayer/protocol-team-kanban/issues/686#issuecomment-3223641009

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
hadjiszs: ah I added one test for that already in this PR, it is at the end of `crates/pessimistic-proof-core/src/aggchain_data/multisig.rs`

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
hadjiszs: yes, in the smart contracts, the protocol team wants to upgrade the consensus type of all networks at the same time, so even the legacy ecdsa will now be expected to put a an aggchain hash with the shape of a consensus type 1

> will require a coordinated update of [...], agglayer

However, the agglayer node side of things is agnostic to this specific change, it is just for the shape of the aggchain hash as part of the PP public inputs

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
hadjiszs: ### Before

- `AggchainProofOnly` comes with a single signature `V4`
- This signature is verified only in the agglayer node side, not in the PP


### After

- `AggchainProofOnly` comes with a single signature `V5` (aggsender will be upgraded for katana)
- This signature has to be verified in the agglayer node AND in the PP (unlike before)
- As a consequence of the two points above: `AggchainProofOnly` becomes nothing more different than a `MultisigAngAggchainProof` where the multisig is a 1-of-1

### Notes

- this `AggchainProofOnly` path is used only by one network: katana, hence being ok to require the upgrade of their aggsender to sign a V4 instead of a V5
- the protocol team will register their trusted sequencer address as a multisig committee, so this PR is a follow up of those discussions/decisions

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
hadjiszs: related reply: https://github.com/agglayer/agglayer/pull/999#discussion_r2315463402

there is no required change in the API, however we have the possibility to optimize the layout as we mentioned when we were discussing the proto: we could remove the signature in the `AggchainProof` message in the proto, and then require the aggsender to send directly a `MultisigAndAggchainProof` with their multisig 1-of1.

But this change is optional, the proto as defined today does the job, I will open a ticket though to keep track of this possible optimization

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
hadjiszs: V4 is used only for katana, deployed today in mainnet, for the signature which comes with the aggchain proof in the path `AggchainProofOnly`.

This signature will be upgraded to V5 in the upcoming release (mainly via the changes from this PR), so we wont have any V4 anymore. However, I wont be in favor of removing this V4 definition, I think erasing (+ overwriting) part of the story can be very confusing. Typically here, for many of my comments I would have had to explain that the katana network has to upgrade from its V4 commitment to a new V4 :/ so better stay with simply a V4 and a V5 I think

crates/pessimistic-proof-test-suite/tests/pp_public_values.rs
hadjiszs: not part of the test file, it's part of the struct that I want to test cover, I cannot remove it but I can hide it maybe, I can define an helper method to generate a `pessimistic_proof_core::MultiSignature` only in test profile

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s01_be020.snap
hadjiszs: There are some impactful changes in terms of perf for the path considered in the cycle tracker (legacy ecdsa):

- Instead of verifying against signature V3 to migrate to pp root V3. Verify against V3 and V5
  - require the hashing for the signature commitment V5
  - require the additional signature verification for this V5

- Compute an aggchain hash consensus type 1, instead of a consensus type 0
   - more expensive, the hash is heavier: https://github.com/agglayer/agglayer/blob/8a2346fd1d95cb5abc35e403afebc1249795dca3/crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs#L56-L85

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s00_be000.snap
hadjiszs: replied there: https://github.com/agglayer/agglayer/pull/999#discussion_r2315536509

crates/pessimistic-proof-core/src/aggchain_data/mod.rs
hadjiszs: > I will open a ticket though to keep track of this possible optimization

- https://github.com/agglayer/interop/issues/100

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
hadjiszs: yes, @iljakuklic made that [here](https://github.com/agglayer/provers/blob/e6a6dc9af825ec9c3a086f920c69e898f8ce3e9c/crates/aggkit-prover-types/src/vkey_hash.rs#L8-L9) but it is in the `provers` repo, it'd need to be promoted to the `interop` repo and make a new release then

I have created a dedicated issue for that: https://github.com/agglayer/interop/issues/101

crates/pessimistic-proof-test-suite/tests/pp_public_values.rs
hadjiszs: replied there: https://github.com/agglayer/agglayer/pull/999#discussion_r2316487367

crates/pessimistic-proof-test-suite/tests/pp_public_values.rs
hadjiszs: I have removed it from the test vector in the end as it isn't involved at all (fyi the test vector was generated by the contracts team, this field was surely informational)

crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs
hadjiszs: I moved the test at the end of the aggchain hash module

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s01_be020.snap
Ekleog-Polygon: Do you know if it'd be easy to generate a certificate close to the one in this test, but with a v5 commitment? This way we could erase the issue of double-hashing, and see how much PP perf really gets hit by the heavier hash.

I'm a bit scared of the 20% slowdown right now, because we've been making big projects like the zero-copy deserialization for AFAIR roughly that amount of expected speedup :smiling_face_with_tear:

Ekleog-Polygon (approved):
Thank you for your answers! I'm not yet 100% clear on all the exact details of how backward-compatibility will be handled, but I'm now reasonably confident about it :smile:

crates/pessimistic-proof-test-suite/tests/snapshots/cycle_tracker__s01_be020.snap
hadjiszs: I'll give a shot in a dedicated PR, I'll need to first wrap up the fetch from the L1, but will try to get this benchmark as direct follow up

>I'm a bit scared of the 20% slowdown right now, because we've been making big projects like the zero-copy deserialization for AFAIR roughly that amount of expected speedup 🥲

my feeling is that I assume the benchmarks on zero copy to have been performed with the legacy ecdsa, so without the aggchain hash computed as consensus type 1. So I feel that this 20% increase is already there in mainnet for all chains which needs to compute the consensus type 1

but to be sure I'll try to update the cycle tracker to make this visible

PR #1003, opened 2025-09-01
feat: forbid providing a metadata if has multisig

hadjiszs:
If a multisig is provided, we forbid to provide a metadata


crates/agglayer-grpc-types/src/compat/v1/certificate.rs
```diff
@@ -7,7 +7,24 @@ impl TryFrom<v1::Certificate> for Certificate {
     type Error = Error;
 
     fn try_from(value: v1::Certificate) -> Result<Self, Self::Error> {
-        Ok(Certificate {
+        let aggchain_data: AggchainData = required_field!(value, aggchain_data);
+
+        // whether it involves a multisig
+        let has_multisig = matches!(
+            aggchain_data,
+            AggchainData::Generic { .. } // 1-of-1
```
Ekleog-Polygon, line 15: Hmm… not sure whether we do want this one too?

Ekleog-Polygon (approved):

crates/agglayer-grpc-types/src/compat/v1/certificate.rs
hadjiszs: yeah it takes the path of a multisig even within the PP since #999

Ekleog-Polygon: Hmm… won't that risk breaking old aggsenders then? I'll review #999 tomorrow-JST, but it feels to me like that could be problematic in practice?

hadjiszs: the `Generic` path is only used for katana, and the idea is to update the commitment for this signature to V5, so that its signature can be verified in the PP as a multisig 1-of-1. The protocol team made this change in the contracts to consider the trusted sequencer for katana as a multisig committee 1-of-1

Ekleog-Polygon: SGTM! Happy to hear that we can simplify our code, then :smile:

Ekleog-Polygon (approved):

atanmarko (approved):

PR #1009, opened 2025-09-02
build: adding protobuf publish

Freyskeyd:
This PR add the publication of the protobuf files to buf registry


atanmarko (approved):
I am not really an expert for this topic, so trust you on this one ;)

.github/workflows/proto-publish.yml
```diff
@@ -0,0 +1,35 @@
...
+  contents: read
+
+jobs:
+  publish:
+    name: Publish protobuf to BSR
+    uses: agglayer/reusable-workflows/.github/workflows/proto-publish.yml@main
+    permissions:
+      contents: read
+    with:
+      ref: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.ref || github.ref }}
```
polykeith, line 34: As I read from [the docs](https://github.com/agglayer/reusable-workflows/blob/a0dcef96b55592ae101c093c30f9483a72933072/.github/workflows/proto-publish.yml#L6), without `use-git-metadata` nor `labels`, the workflow will not publish anything, and I don't seem to see any defined here... unless `ref` already defines them somehow?

Ekleog-Polygon: Right, it's interesting, but looking at [how the code is currently written](https://github.com/agglayer/reusable-workflows/blob/main/.github/workflows/proto-publish.yml#L106) I'm seeing labels being taken from `input` and not from the result of the `extract-labels` job?

Ekleog-Polygon (approved):
I don't think this will work as per @polykeith's comment (and mine below), but CI is a magic beast where it's often better to just land the thing and see what happens, with basically no risk if it's wrong. So I'm approving still :smile:

.github/workflows/proto-publish.yml
Freyskeyd: For `agglayer-node` we're relying only on the `git-metadata` that is automatically defined in the reusable action.
I updated the reusable action to either use the input or the extracted labels

hadjiszs (approved):

PR #1010, opened 2025-09-02
chore: add epoch store reader to agglayer rpc

atanmarko:
Add epoch store reader to `agglayer-rpc`. Needed for get network state implementation.


crates/agglayer-storage/src/stores/epochs.rs
```diff
@@ -80,4 +81,12 @@ where
     PendingStore: PendingCertificateReader + PendingCertificateWriter,
     StateStore: StateWriter + MetadataWriter + StateReader,
 {
+    fn get_certificate(
+        &self,
+        epoch_number: EpochNumber,
+        index: CertificateIndex,
+    ) -> Result<Option<Certificate>, Error> {
+        let per_epoch_store = self.open(epoch_number)?;
```
Freyskeyd: Maybe this should be open using `readonly` to avoid any race condition?

crates/agglayer-rpc/src/lib.rs
```diff
@@ -212,10 +217,14 @@ where
             latest_epoch_with_settlement: None,
         })
     }
+
+    pub fn get_current_epoch(&self) -> Option<agglayer_types::EpochNumber> {
```
Freyskeyd: Is it really needed?

atanmarko: Not really. Just wanted to use something from the epoch store to prevent unused warnings. I'll remove it.

crates/agglayer-storage/src/stores/epochs.rs
atanmarko: I'll look into this tomorrow morning.

Ekleog-Polygon (commented):
LGTM! I don't have any comments beyond @Freyskeyd 's point on using readonly; but I do think that we really ought to use readonly here

Freyskeyd:
> LGTM! I don't have any comments beyond @Freyskeyd 's point on using readonly; but I do think that we really ought to use readonly here

The issue could be that we're trying to read the current epoch database which can't be open twice, leading to an error.

crates/agglayer-storage/src/stores/epochs.rs
Freyskeyd: You can use something like this in the `storage/db`, to initialize the `epoch` store:

```rust
pub fn open_cf_read_only(path: &Path, cfs: Vec<ColumnFamilyDescriptor>) -> Result<DB, DBError> {
    let mut options = Options::default();
    options.create_if_missing(true);
    options.create_missing_column_families(true);

    Ok(DB {
        rocksdb: rocksdb::DB::open_cf_descriptors_read_only(&options, path, cfs, false)?,
        default_write_options: WriteOptions::default(),
    })
}
```

atanmarko:
> > 
> 
> The issue could be that we're trying to read the current epoch database which can't be open twice, leading to an error.

So what do you suggest as a workaround?

crates/agglayer-storage/src/stores/per_epoch/mod.rs
```diff
@@ -56,16 +56,61 @@ impl<PendingStore, StateStore> PerEpochStore<PendingStore, StateStore> {
...
+        let db = Arc::new(DB::open_cf_readonly(&path, epochs_db_cf_definitions())?);
+        
+        Self::try_open_with_db(
+            db,
+            epoch_number,
+            pending_store,
+            state_store,
+            None, // No start checkpoint for readonly
+            BackupClient::noop(), // No backup needed for readonly access
+            true, // readonly mode
```
Freyskeyd, line 99: nit: Maybe we can introduce an `Enum` for more clarity?

atanmarko: I'll add enum to the next PR in the chain.

PR #1011, opened 2025-09-02
feat: implement agglayer rpc get network state

atanmarko:


crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +271,206 @@ where
...
+
+        let latest_pending_certificate = self
+            .get_latest_pending_certificate_header(network_id)
+            .inspect_err(|error| {
+                warn!(?error, "Failed to get latest pending certificate");
+            })
+            .unwrap_or_default();
+
+        // Determine network type from the latest available certificate
+        let network_type = match self.get_latest_available_certificate_for_network(network_id) {
```
Ekleog-Polygon, line 398: Should we not take the network type from the latest settled, or at least the latest proven certificate for network? AFAICT this certificate has not yet undergone much checks yet, so we could be answering an unexpected network type if a type-changing certificate was sent but not yet validated yet.

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +271,206 @@ where
...
+                    }
+                    agglayer_types::aggchain_proof::AggchainData::MultisigOnly { .. } => {
+                        Ok(NetworkType::MultisigOnly)
+                    }
+                    agglayer_types::aggchain_proof::AggchainData::MultisigAndAggchainProof {
+                        ..
+                    } => Ok(NetworkType::MultisigAndAggchainProof),
+                }
+            }
+            Ok(None) => Err(GetNetworkStateError::UnknownNetworkType { network_id }),
```
Ekleog-Polygon, line 416: Should this not be `UnknownNetwork`? In this case we literally have no information on the network AFAICT

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +271,206 @@ where
...
+            Err(error) => {
+                error!(?error, "Unable to determine network type");
+                Err(GetNetworkStateError::UnknownNetworkType { network_id })
```
Ekleog-Polygon: And this feels like a perfect use case for an `GetNetworkStateError::Other(#[source] eyre::Error)` variant to the enum, as AFAICT this is an internal error

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +271,206 @@ where
...
+        let settled_pp_root = latest_settled_certificate.as_ref().and_then(|cert| {
+            // Get the proof for the settled certificate
+            self.get_proof(cert.certificate_id)
+                .inspect_err(|error| {
+                    error!(
+                        ?error,
+                        "get network status: failed to get proof for settled certificate"
+                    );
+                })
+                .ok()
```
Ekleog-Polygon: Should this error not result in a failing RPC call, rather than just an RPC call with no "settled pp root" result? Returning none here, afaict, would mean to the caller that there was no pp root settled yet

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +271,206 @@ where
...
+                            sp1_proof.public_values.as_slice(),
+                        )
+                        .inspect_err(|error| {
+                            error!(
+                                ?error,
+                                "get network status: failed to deserialize pessimistic proof \
+                                 output"
+                            );
+                        })
+                        .ok()
```
Ekleog-Polygon: Same here

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +271,206 @@ where
...
+        let settled_let_leaf_count = self
+            .state
+            .read_local_network_state(network_id)
+            .inspect_err(|error| {
+                error!(
+                    ?error,
+                    "Failed to get local network state for network {network_id}"
+                );
+            })
+            .ok()
```
Ekleog-Polygon: And here

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +271,206 @@ where
...
+
+        let latest_pending_status = latest_pending_certificate
+            .as_ref()
+            .map(|cert| cert.status.clone());
+
+        // Get pending certificate error if exists
+        let latest_pending_error = latest_pending_certificate
+            .as_ref()
+            .and_then(|cert| match &cert.status {
+                agglayer_types::CertificateStatus::InError { error } => Some(*error.clone()),
```
Ekleog-Polygon:
```suggestion
                agglayer_types::CertificateStatus::InError { error } => Some((*error).clone()),
```
Clone the value inside the box directly, instead of cloning the whole box and then copying out. Might be optimized away by the compiler but :shrug:

crates/agglayer-storage/src/stores/per_epoch/tests.rs
```diff
@@ -324,3 +324,61 @@ fn adding_certificate_and_restart() {
...
+    assert!(
+        store
+            .add_certificate(certificate.hash(), agglayer_types::ExecutionMode::Default)
+            .is_ok(),
+        "Failed to add certificate to epoch store"
+    );
+
+    // Verify that we can retrieve the proof at index 0
+    let proof = store.get_proof_at_index(CertificateIndex::ZERO).unwrap();
+    assert!(proof.is_some(), "Should retrieve a proof");
```
Ekleog-Polygon, line 379: Can we easily assert for equality here, so that we also validate our (de)serialization implementation automatically?

proto/agglayer/node/v1/node_state.proto
```diff
@@ -54,20 +54,14 @@ enum GetNetworkStateErrorKind {
...
 
-  // Invalid network local state.
-  GET_NETWORK_STATE_ERROR_KIND_NETWORK_LOCAL_STATE_ERROR = 4;
+  // Invalid network state.
+  GET_NETWORK_STATE_ERROR_KIND_NETWORK_STATE_ERROR = 2;
 
   // Could not determine network type
-  GET_NETWORK_STATE_ERROR_KIND_UNKNOWN_NETWORK_TYPE = 5;
+  GET_NETWORK_STATE_ERROR_KIND_UNKNOWN_NETWORK_TYPE = 3;
 }
```
Ekleog-Polygon: Just to confirm, did these ever reach production? This will break any other users depending on it, but AFAICT it should be fine, because they never got into production yet. Right?

Ekleog-Polygon (changes requested):
LGTM, thank you! Requesting changes only due to the comments about silently returning `None` instead of having a proper error — all the rest are just nits that we can live with or handle in follow-up PRs :smile:

crates/agglayer-rpc/src/lib.rs
atanmarko: Ok, I'll remove pending certificate. Proven or settled.

crates/agglayer-rpc/src/lib.rs
atanmarko: We just have no certificate - maybe we know about the network but certificate has not yet been received (we have just started a new chain). So `UnknowNetworkType` is appropriate here I think.

crates/agglayer-rpc/src/lib.rs
atanmarko: I am against making so many custom error types for this function in proto definition - they are internal details of the agglayer node. I am logging the error, but I am not sure that we should return this details as part of the get_network_status grpc api.

crates/agglayer-rpc/src/lib.rs
atanmarko: Good question. I am not sure. You may have not settled pp root because simply you do not have that happened yet - tThat still does not mean that you should not return other data. Hence the approach that most of the return values are optional.

crates/agglayer-rpc/src/lib.rs
atanmarko: This is part of the approach that you return only the data you have - I don't think that we should error on any missing piece. What do you think @Freyskeyd @hadjiszs ?

crates/agglayer-storage/src/stores/per_epoch/tests.rs
atanmarko: `SP1ProofWithPublicValues` does not implement equality.

proto/agglayer/node/v1/node_state.proto
atanmarko: @Ekleog-Polygon Not they have not - this is new API.

crates/agglayer-storage/src/storage/mod.rs
```diff
@@ -28,6 +28,9 @@ pub enum DBError {
 
     #[error("Trying to access an unknown ColumnFamily")]
     ColumnFamilyNotFound,
+
+    #[error("Database is opened in read-only mode")]
```
Ekleog-Polygon: Nit: was (for some reason github doesn't want me to send a suggestion :shrug: )

crates/agglayer-storage/src/stores/per_epoch/mod.rs
```diff
@@ -56,16 +56,61 @@ impl<PendingStore, StateStore> PerEpochStore<PendingStore, StateStore> {
...
+    pub fn try_open_readonly(
+        config: Arc<agglayer_config::Config>,
+        epoch_number: EpochNumber,
+        pending_store: Arc<PendingStore>,
+        state_store: Arc<StateStore>,
+    ) -> Result<Self, Error> {
+        let path = config
+            .storage
+            .epochs_db_path
+            .join(format!("{epoch_number}"));
```
Ekleog-Polygon: Nit: this is repeated twice, could be a function like `config.storage.epoch_db_path(epoch_number)`

crates/agglayer-storage/src/stores/per_epoch/mod.rs
```diff
@@ -87,40 +132,52 @@ impl<PendingStore, StateStore> PerEpochStore<PendingStore, StateStore> {
...
-                        return Err(Error::Unexpected(
-                            "Start checkpoint doesn't match the expected one, using the one from \
-                             the DB"
-                                .to_string(),
-                        ))?;
-                    } else {
-                        checkpoint
+            if readonly {
+                // For readonly access, we just use the existing checkpoint
+                checkpoint
```
Ekleog-Polygon: Nit: the code would be cleaner here with a `break` to avoid the indent-to-the-right for the whole `match` block. Something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=eea3a2b1cdb090fabfa875f67cb06db0

crates/agglayer-rpc/src/lib.rs
```diff
@@ -383,55 +411,101 @@ where
...
-            });
+        let settled_let_leaf_count =
+            match self
+                .state
+                .read_local_network_state(network_id)
+                .map(|local_network_state| {
+                    // We return the leaf count of the latest local exit tree
+                    local_network_state.map(|v| v.exit_tree.leaf_count as u64)
+                }) {
+                Ok(leaf_count) => leaf_count,
```
Ekleog-Polygon: Nit: the code in the `map` could just be in the `Ok` branch, it'd make the code simpler

crates/agglayer-rpc/src/lib.rs
```diff
@@ -149,22 +149,46 @@ where
...
+            .inspect_err(|e| error!("Failed to get latest settled certificate: {e}"))?
+            .map(|(_, SettledCertificate(id, height, _, _))| (id, height));
+
+        let certificate_id = [
+            proven_certificate_id_and_height,
+            settled_certificate_id_and_height,
+        ]
+        .into_iter()
+        .flatten()
+        .max_by(|x, y| x.1.cmp(&y.1))
```
Ekleog-Polygon: Is it actually possible that the latest settled certificate would be higher than the latest proven certificate? If not, maybe we can simplify code here?

Ekleog-Polygon (dismissed):
LGTM! I'm just sad due to the sheer amount of code required for error handling :smiling_face_with_tear:

crates/agglayer-storage/src/stores/per_epoch/mod.rs
atanmarko: It is quite rare to see this break construct in Rust code, so may confuse potential reader. It also reminds me a bit to the infamous `goto` command from the BASIC language I used on Commodore 64. Hence would vote against this change :)

crates/agglayer-rpc/src/lib.rs
atanmarko: I have reused existing `get_latest_known_certificate_header` implementation here, so I guess this is out of scope for this PR. But anyway I would prefer to be on the safe side.

crates/agglayer-rpc/src/error.rs
```diff
@@ -108,8 +108,49 @@ impl SignatureVerificationError {
...
+    },
+
+    #[error(
+        "Invalid certificate retrieval - certificate id and hash mismatch - expected \
+         certificate_id: {expected} got: {got}"
+    )]
+    CertificateIdHashMismatch {
+        expected: CertificateId,
+        got: CertificateId,
+    },
```
hadjiszs:
```suggestion
    #[error(
        "Mismatch on the certificate id. expected: {expected}, re-computed from the certificate in DB: {got}"
    )]
    CertificateIdMismatch {
        expected: CertificateId,
        got: CertificateId,
    },
```

bringing difference between the certificate id and its hash as of today feels confusing to me

crates/agglayer-rpc/src/error.rs
```diff
@@ -108,8 +108,49 @@ impl SignatureVerificationError {
...
+#[derive(Debug, thiserror::Error)]
+pub enum GetLatestCertificateError {
+    #[error(transparent)]
+    Storage(#[from] StorageError),
+
+    #[error("Failed to get latest known certificate header for network {network_id}")]
+    UnknownLatestCertificateHeader {
+        network_id: NetworkId,
+        #[source]
+        source: Box<CertificateRetrievalError>,
```
hadjiszs:
```suggestion
        #[source]
        source: Box<CertificateRetrievalError>,
```

if the field is named `source` then no need `#[source]` no? also is it printed in the error message?

crates/agglayer-rpc/src/error.rs
```diff
@@ -108,8 +108,49 @@ impl SignatureVerificationError {
...
+    NotFound { certificate_id: CertificateId },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum GetLatestCertificateError {
+    #[error(transparent)]
+    Storage(#[from] StorageError),
+
+    #[error("Failed to get latest known certificate header for network {network_id}")]
+    UnknownLatestCertificateHeader {
```
hadjiszs: Enum name suggests that it is unknown, while the error message doesn't and just say there is failure to get the latest known (which doesnt mean it's unknown no?)

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +296,274 @@ where
...
+                    );
+                    return Err(GetNetworkStateError::InternalError {
+                        network_id,
+                        source: error.into(),
+                    });
+                }
+            }
+        } else {
+            None
+        };
```
hadjiszs, line 477: you have access to `self.l1_rpc_provider` here so instead of all of that you could call `get_prev_pessimistic_root` which will fetch the latest settled pp root from L1

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +296,274 @@ where
...
+                    source: eyre::eyre!("Inconsistent network state error"),
+                });
+            }
+            Err(error) => {
+                error!(
+                    ?error,
+                    "get network status: failed to read local network state for network \
+                     {network_id}"
+                );
+                None
```
hadjiszs: shouldnt it be an error here?

we want `None` only in the case of `Ok(None)` in the first branch no?

---

maybe the 3 branches `Err(_)` can be a single branch where you print the error as-is and return the `GetNetworkStateError::InternalError`?

crates/agglayer-rpc/Cargo.toml
```diff
@@ -13,14 +13,17 @@ agglayer-rate-limiting.workspace = true
...
 
 
 alloy.workspace = true
+eyre.workspace = true
 futures.workspace = true
 serde.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
 tracing.workspace = true
+log = "0.4.27"
```
hadjiszs: I dont see where it is used

crates/agglayer-storage/src/storage/mod.rs
```diff
@@ -29,7 +29,7 @@ pub enum DBError {
     #[error("Trying to access an unknown ColumnFamily")]
     ColumnFamilyNotFound,
 
-    #[error("Database is opened in read-only mode")]
+    #[error("Database was opened in read-only mode")]
```
hadjiszs, line 32: in datadog people might wonder how is this an error, maybe let's precise that we got a write attempt

```suggestion
    #[error("Cannot write in database opened in read-only mode")]
```

crates/agglayer-rpc/src/lib.rs
hadjiszs: I think there is a `max_by_key` to do something like this

```suggestion
        .max_by_key(|x| x.1)
```

proto/agglayer/node/v1/node_state.proto
```diff
@@ -54,20 +54,17 @@ enum GetNetworkStateErrorKind {
...
 
-  // Missing latest settled certificate.
-  GET_NETWORK_STATE_ERROR_KIND_MISSING_LATEST_SETTLED_CERTIFICATE = 1;
-
-  // Missing latest pending certificate.
-  GET_NETWORK_STATE_ERROR_KIND_MISSING_LATEST_PENDING_CERTIFICATE = 2;
-
   // Invalid network ID/network not found.
-  GET_NETWORK_STATE_ERROR_KIND_INVALID_NETWORK_ID = 3;
+  GET_NETWORK_STATE_ERROR_KIND_INVALID_NETWORK_ID = 1;
```
iljakuklic, line 58: Should this be done in a more backwards compatible manner? I.e:
* mark the kinds 1 and 2 (and other removed kinds) as reserved
* maintain the same numeric mapping for `GET_NETWORK_STATE_ERROR_KIND_INVALID_NETWORK_ID` and `GET_NETWORK_STATE_ERROR_KIND_UNKNOWN_NETWORK_TYPE`
* allocate a fresh number for each newly introduced kind

crates/agglayer-rpc/src/lib.rs
```diff
@@ -144,6 +148,104 @@ where
...
+            .map(|(_, SettledCertificate(id, height, _, _))| (id, height));
+
+        let certificate_id = [
+            proven_certificate_id_and_height,
+            settled_certificate_id_and_height,
+        ]
+        .into_iter()
+        .flatten()
+        .max_by(|x, y| x.1.cmp(&y.1))
+        .map(|v| v.0);
```
iljakuklic: Style nit: Since there are only two things to compare, `std::cmp::max_by_key` could be used here, also exploiting that `None` < `Some(_)`, roughly like:

```rust
        let certificate_id = std::cmp::max_by_key(
            proven_certificate_id_and_height,
            settled_certificate_id_and_height,
            |v| v.map(|(_, ht)| ht),
        )
        .map(|v| v.0);
```
Additinally, plain `std::cmp::max` could be used if you change the tuple to have the height as the first component.

crates/agglayer-rpc/src/lib.rs
```diff
@@ -144,6 +148,104 @@ where
...
+            None => Ok(None),
+            Some(certificate_id) => self.fetch_certificate_header(certificate_id).map(Some),
+        }
+        .map_err(|error| {
+            error!(?error, "Failed to get latest known certificate header");
+            GetLatestCertificateError::UnknownLatestCertificateHeader {
+                network_id,
+                source: Box::new(error),
+            }
+        })?;
```
iljakuklic: Nit: could the error handling be moved into the `Some(_)` branch? Seems it could make the code a bit more readable to have the error processing closer to where the error is generated.

crates/agglayer-rpc/src/lib.rs
```diff
@@ -144,6 +148,104 @@ where
...
+                // If not found in pending store, try to get from epoch store.
+                if let (Some(epoch_number), Some(certificate_index)) =
+                    (epoch_number, certificate_index)
+                {
+                    match self
+                        .epochs_store
+                        .get_certificate(epoch_number, certificate_index)
+                    {
+                        Ok(Some(certificate)) => {
+                            debug!("Found certificate {} in epoch store", certificate_id);
```
iljakuklic: Nit:
```suggestion
                            debug!("Found certificate {certificate_id} in epoch store");
```

Also the `warn!` below

crates/agglayer-rpc/src/lib.rs
```diff
@@ -194,27 +296,274 @@ where
...
+                    }
+                }
+            }
+            Err(error) => {
+                error!(
+                    ?error,
+                    "Failed to get proof for certificate {certificate_id} from pending store",
+                );
+                Err(ProofRetrievalError::Storage(error))
+            }
```
iljakuklic: Nit: This could be right after `get_proof` in a `map_err`, immediately dealt with using the `?` operator to get the error out of the way and continue the happy path code without the extra indentation.

There are multiple occurrences of this pattern in this file.

proto/agglayer/node/v1/node_state.proto
Freyskeyd: I'm not sure we're having backward compatibility concern as this feature is still in development

iljakuklic: > I'm not sure we're having backward compatibility concern as this feature is still in development

Ok, I was deceived by the fact it modifies an already merged proto definition.

crates/agglayer-rpc/src/error.rs
atanmarko: Yeah I think so, but was not sure :)

crates/agglayer-rpc/src/lib.rs
atanmarko: Yes I know about that alternative. I chose to go to the database as faster local solution as value is already available locally.

crates/agglayer-rpc/src/lib.rs
atanmarko: @hadjiszs Having no data is not a breaking error for get network status. E.g. `NoCertificate`, `NoCertificateHeader`, `NoProof` are expected when network has just started.  I have tried to isolate and return error in `agglayer_storage::Error` cases where some serious inconsistencies  are met and invariants are broken. `agglayer_storage::Error` has many possible errors , I am not sure which one can actually happen with the `read_local_network_state`.

crates/agglayer-rpc/Cargo.toml
atanmarko: Yeah it was automatically added by the IDE

crates/agglayer-rpc/src/lib.rs
atanmarko: We will revamp this part soon with the introduction of the new network state column family, as discussed today in the meeting. Until then I will leave it as it is.

crates/agglayer-rpc/src/lib.rs
atanmarko: EDIT: As discussed today in a meeting, this part will be revamped very soon with the introduction of the network state column family. Until then I will leave it as it is.

crates/agglayer-rpc/src/lib.rs
atanmarko: Yeah, good one.

crates/agglayer-rpc/src/lib.rs
atanmarko: Resolved

crates/agglayer-storage/src/stores/per_epoch/mod.rs
Ekleog-Polygon: Right, it's still a pretty new feature of the language. I personally find it helps quite a lot with rightward drift, while still being easy enough to understand (it can only break from an enclosing block), but I'll definitely not block this PR on that :smile:

Ekleog-Polygon (dismissed):

crates/agglayer-rpc/src/lib.rs
hadjiszs: >Having no data is not a breaking error for get network status. 

Yea I understand, but here we're not in this case: the read of local network state returns Ok(None) if there is no state for this network (that's the first branch of your match)

Here we're in a branch Err, so it means there is really a failure at reading the state (as opposed to not finding a state), so in this case i dont think we want None

iljakuklic (dismissed):

crates/agglayer-rpc/src/lib.rs
atanmarko: Ok, I'll return an error.

Freyskeyd (approved):

Issue #996, opened 2025-08-28
Update the witness generation to consider the indices from aggsender


hadjiszs:
## What we have

Multisig witness generation got introduced in
- #963 

In it, we assume that the chain sends a raw and unordered list of signatures.
Out of them, we recover the addresses for each of them and simply check if it leads to an acceptable threshold of registered address from the L1.

> [!IMPORTANT] 
> The witness generation is already functional as-is, but we would like to update it to have more granularity on the errors.

## What we want

We now receive indexed signatures from the chain: `signatures: Vec<Option<Signature>>`, where `signatures[i]` corresponds to the optional signature made by the signer `i` as per the indexes registered in the L1.

## To do

- adapt the witness generation to leverage this new data structure instead of inferring everything
- adapt the error to bring more granularity (we now have the ability to explicit the indices for which the signatures are unrecoverable)

PR #1014, opened 2025-09-04
feat: use the new format for signature list for the proof

Ekleog-Polygon:
Fixes https://github.com/agglayer/agglayer/issues/996

AFAICT no breaking change despite the storage modification, because the previous version had never been deployed to production.

### DIFF
--- a/crates/agglayer-types/src/aggchain_data/multisig.rs
+++ b/crates/agglayer-types/src/aggchain_data/multisig.rs
@@ -16,25 +16,22 @@ pub enum MultisigError {
 
     #[error("Unknown signer or invalid prehash. prehash: {prehash:?}, recovered: {recovered}")]
     UnknownRecoveredSigner { recovered: Address, prehash: Digest },
-
-    #[error("At least one signer comes duplicated. signer: {signer}.")]
-    HasDuplicateSigner { signer: Address },
 }
 
 /// Multisig data from the chain.
 #[derive(Clone, Debug)]
 pub struct Payload {
-    signatures: Vec<Signature>,
+    signatures: Vec<Option<Signature>>,
 }
 
-impl From<Vec<Signature>> for Payload {
-    fn from(signatures: Vec<Signature>) -> Self {
+impl From<Vec<Option<Signature>>> for Payload {
+    fn from(signatures: Vec<Option<Signature>>) -> Self {
         Self { signatures }
     }
 }
 
-impl From<&[Signature]> for Payload {
-    fn from(signatures: &[Signature]) -> Self {
+impl From<&[Option<Signature>]> for Payload {
+    fn from(signatures: &[Option<Signature>]) -> Self {
         Self {
             signatures: signatures.to_vec(),
         }
@@ -44,7 +41,7 @@ impl From<&[Signature]> for Payload {
 impl From<&agglayer_interop_types::aggchain_proof::MultisigPayload> for Payload {
     fn from(value: &agglayer_interop_types::aggchain_proof::MultisigPayload) -> Self {
         Self {
-            signatures: value.0.clone().into_iter().flatten().collect(),
+            signatures: value.0.clone(),
         }
     }
 }
@@ -60,22 +57,6 @@ pub struct Ctx {
     pub prehash: B256,
 }
 
-impl Ctx {
-    /// Returns the index of the signer from the provided signature.
-    fn signer_from_signature(&self, signature: &Signature) -> Result<usize, MultisigError> {
-        let recovered = signature
-            .recover_address_from_prehash(&self.prehash)
-            .map_err(|_| MultisigError::InvalidSignature)?;
-
-        self.signers.iter().position(|s| s == &recovered).ok_or(
-            MultisigError::UnknownRecoveredSigner {
-                recovered,
-                prehash: self.prehash.into(),
-            },
-        )
-    }
-}
-
 // Generate the prover inputs from the chain payload and the L1 context.
 impl TryInto<core::MultiSignature> for PayloadWithCtx<Payload, Ctx> {
     type Error = MultisigError;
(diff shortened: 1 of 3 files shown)


crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -42,55 +39,36 @@ impl MultiSignature {
...
-                .unwrap_or(true)
-        });
-
-        if has_invalid_signature {
-            return Err(MultisigError::HasInvalidSignature);
+        if self.signatures.len() > self.expected_signers.len() {
+            return Err(MultisigError::OutOfBoundSignerIndex {
+                idx: self.signatures.len() - 1,
+                total: self.expected_signers.len(),
+            });
```
hadjiszs: maybe having fields both referring to indexes or number of signers:

```suggestion
            return Err(MultisigError::OutOfBoundSignerIndex {
                got: self.signatures.len() - 1,
                max: self.expected_signers.len() - 1,
            });
```

or

```suggestion
            return Err(MultisigError::OutOfBoundSigner {
                got: self.signatures.len(),
                max: self.expected_signers.len(),
            });
```

crates/agglayer-types/src/aggchain_data/multisig.rs
```diff
@@ -60,129 +57,17 @@ pub struct Ctx {
...
-                Ok::<_, MultisigError>((idx, sig))
-            })
-            .collect::<Result<_, _>>()?;
-
-        if indexed_signatures.len() < multisig.threshold {
-            return Err(MultisigError::UnderThreshold {
-                got: indexed_signatures.len(),
-                expected: multisig.threshold,
-            });
-        }
```
hadjiszs: ah then we need to update here: https://github.com/agglayer/agglayer/blob/b82c3da4a7d1473e426b8159436834bdca6326c3/crates/agglayer-types/src/certificate/mod.rs#L214-L218

to call the `core::MultiSignature::verify_multisig`

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -18,11 +18,8 @@ pub enum MultisigError {
     #[error("Multisig is under the required threshold. got: {got}, expected: {expected}")]
     UnderThreshold { got: usize, expected: usize },
 
-    #[error("At least one signature is invalid or from an unregistered signer.")]
-    HasInvalidSignature,
-
-    #[error("At least one signer comes duplicated. signer: {signer}.")]
-    HasDuplicateSigner { signer: Address },
+    #[error("Signature {idx} is invalid or from an unregistered signer.")]
```
hadjiszs:
```suggestion
    #[error("Signer #{idx} has an invalid signature.")]
```

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -42,55 +39,36 @@ impl MultiSignature {
...
+            let Some(signature) = signature else {
+                continue; // No signature is a valid signature
+            };
+            let Ok(recovered) = signature.recover_address_from_prehash(&commitment) else {
+                return Err(MultisigError::HasInvalidSignature { idx }); // Failed to recover from prehash
+            };
+            let expected = self.expected_signers[idx];
+            if recovered != expected {
+                return Err(MultisigError::HasInvalidSignature { idx }); // Signer doesn't match
+            }
```
hadjiszs: can be useful for troubleshooting to have two different errors here

```suggestion
            let recovered = signature.recover_address_from_prehash(&commitment).map_err(|e| MultisigError::Recovery { idx, source: e })?;
            
            let expected = self.expected_signers[idx];
            if recovered != expected {
                return Err(MultisigError::HasInvalidSignature {
                     idx,
                     recovered,
                     expected
                }); // Signer doesn't match
            }
```

- possible reasons for `MultisigError::Recovery`:
   - ill-formed signature (not necessarily just wrong signer or wrong message, just ill-formed ecdsa)

- possible reasons for `HasInvalidSignature`
   - wrong signer or wrong commitment

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
Ekleog-Polygon: I tried improving on the various error messages, lmk what you think :)

Ekleog-Polygon:
I took advantage of the changes to remove the duplicate MultisigError struct — it made more sense to just remove it than to copy-paste the modifications on both sides to me, but lmk if there was a specific reason for the duplication and I'll revert :smile:

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -15,17 +15,17 @@ pub struct MultiSignature {
 
 #[derive(Clone, Debug, Error, Serialize, Deserialize, PartialEq, Eq)]
 pub enum MultisigError {
+    #[error("Too many signers provided. got: {num}, committee size: {max}.")]
+    TooManySigners { num: usize, max: usize },
+
     #[error("Multisig is under the required threshold. got: {got}, expected: {expected}")]
     UnderThreshold { got: usize, expected: usize },
 
-    #[error("At least one signature is invalid or from an unregistered signer.")]
```
Freyskeyd, line 21: Could this be breaking in regard of the storage?

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
```diff
@@ -48,55 +48,36 @@ impl MultiSignature {
...
+        for (idx, signature) in self.signatures.iter().enumerate() {
+            let Some(signature) = signature else {
+                continue; // No signature is a valid signature
+            };
+            let recovered = signature
+                .recover_address_from_prehash(&commitment)
+                .map_err(|_| MultisigError::InvalidSignature { idx })?;
+            let expected = self.expected_signers[idx];
+            if recovered != expected {
+                return Err(MultisigError::InvalidSigner { idx });
```
hadjiszs:
```suggestion
                return Err(MultisigError::InvalidSigner { idx, expected, commitment });
```

by rephrasing the error message

```diff
- #[error("Signature claimed to be from signer {idx} is invalid.")]
+ #[error("Signer #{idx} ({expected}) provided a wrong signature. expected signed message: {commitment}")]
```

because if the `recovered != signer`, it might be due to two reasons: either the message isn't the right one, or the signer is indeed different, we have no way to distinguish those two cases unfortunately (so will be valuable to expose the prehash/commitment, alongside the index and signer address that we consider there)

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
Ekleog-Polygon: It would be if multisig was ever deployed in production, but AFAIU what @hadjiszs suggested, if was actually never, so we're still fine to make these changes 😄

hadjiszs: yeah multisig related so never deployed yet 👍

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
Ekleog-Polygon: Hmm… is it possible for the message to not be the right one, but for the signature to still `recover_from_prehash(wrong_commitment)` to a valid other signer key? I'd expect that to be impossible without access to the other signer private key?

I've tried rewording in a way that's as "consensual as possible, and scheduled a 1:1 for today so we can close this question :smile:

atanmarko (approved):

crates/pessimistic-proof-core/src/aggchain_data/multisig.rs
hadjiszs: >Hmm… is it possible for the message to not be the right one, but for the signature to still recover_from_prehash(wrong_commitment) to a valid other signer key? I'd expect that to be impossible without access to the other signer private key?

discussed f2f, tldr: yeah if the recover succeed, no way to distinguish the two cases: wrong commitment or wrong signer

the new error message lgtm 👌

hadjiszs (approved):
thank you for the back and forth 👍

PR #1016, opened 2025-09-04
feat: fetch the multisig context from the L1

### LINKED ISSUE #941: Fetch the threshold and the set of signers from the L1 for the given network

hadjiszs:
The fetch occurs twice:

- At the rpc level upon receiving one certificate, we need to verify its multisig prior to inserting the Certificate in the storage
- At the witness generation, we need the context to generate the prover inputs

Fixes #941

### DIFF
--- a/crates/agglayer-contracts/src/aggchain.rs
+++ b/crates/agglayer-contracts/src/aggchain.rs
@@ -88,12 +88,7 @@ where
             .getAggchainSigners()
             .call()
             .await
-            .map(|alloy_vec| {
-                alloy_vec
-                    .iter()
-                    .map(|addr| Address::from_alloy(*addr))
-                    .collect()
-            })
+            .map(|alloy_vec| alloy_vec.into_iter().map(Address::from_alloy).collect())
             .map_err(L1RpcError::MultisigSignersFetchFailed)?;
 
         let threshold: usize = {


crates/agglayer-contracts/src/aggchain.rs
```diff
@@ -74,4 +79,37 @@ where
...
+        &self,
+        rollup_address: Address,
+    ) -> Result<(Vec<Address>, usize), L1RpcError> {
+        let signers = AggchainBase::new(rollup_address.into(), self.rpc.clone())
+            .getAggchainSigners()
+            .call()
+            .await
+            .map(|alloy_vec| {
+                alloy_vec
+                    .iter()
```
iljakuklic: Nit:
```suggestion
                alloy_vec
                    .into_iter()
```

crates/agglayer-contracts/src/contracts/AggchainBase.json
Freyskeyd, line 1: Is it a contract from another branch that the one define in the `script/make/Makefile.contracts.toml` ?

Freyskeyd (approved):

PR #1019, opened 2025-09-05
feat: add settled claim

atanmarko:
Add last settled claim for get network state

### DIFF
--- a/crates/agglayer-rpc/src/lib.rs
+++ b/crates/agglayer-rpc/src/lib.rs
@@ -570,16 +570,19 @@ where
             .as_ref()
             .and_then(|cert| cert.epoch_number.map(|num| num.as_u64()));
 
-        // TODO: implement settled claim retrieval
-        let settled_claim = self
-            .get_last_settled_claim(network_id, settled_height.unwrap_or(Height::from(0)))
-            .map_err(|error| {
-                error!(?error, "Failed to get last settled claim");
-                GetNetworkStateError::InternalError {
-                    network_id,
-                    source: error.into(),
-                }
-            })?;
+        // Get the last settled claim if we have a settled height
+        let settled_claim = if let Some(height) = settled_height {
+            self.get_last_settled_claim(network_id, height)
+                .map_err(|error| {
+                    error!(?error, "Failed to get last settled claim");
+                    GetNetworkStateError::InternalError {
+                        network_id,
+                        source: error.into(),
+                    }
+                })?
+        } else {
+            None
+        };
 
         Ok(NetworkState {
             network_status,
--- a/crates/agglayer-rpc/src/lib.rs
+++ b/crates/agglayer-rpc/src/lib.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use agglayer_config::{epoch::BlockClockConfig, Config, Epoch};
 use agglayer_contracts::{L1TransactionFetcher, RollupContract};
+use agglayer_primitives::Hashable;
 use agglayer_rate_limiting as rate_limiting;
 use agglayer_storage::{
     columns::latest_settled_certificate_per_network::SettledCertificate,
@@ -23,8 +24,8 @@ pub use self::error::{
     CertificateRetrievalError, CertificateSubmissionError, GetNetworkStateError,
 };
 use crate::{
-    error::{GetLatestCertificateError, ProofRetrievalError},
-    network_state::{NetworkState, NetworkType},
+    error::{GetLatestCertificateError, GetLatestSettledClaimError, ProofRetrievalError},
+    network_state::{NetworkState, NetworkType, SettledClaim},
 };
 
 pub mod error;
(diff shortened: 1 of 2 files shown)


crates/agglayer-rpc/src/lib.rs
```diff
@@ -347,6 +348,57 @@ where
...
+        // Iterate from the given height down to 0
+        for current_height in (0..=height.as_u64()).rev().map(Height::from) {
+            // Fetch certificate header for the current height
+            let header_opt = self
+                .state
+                .get_certificate_header_by_cursor(network_id, current_height)?;
+
+            let header = match header_opt {
+                Some(h) => h,
+                None => continue, // No certificate at this height, continue to next lower height
```
Freyskeyd: If we don't have the latest expected height, it means that we've a problem right?

crates/agglayer-rpc/src/lib.rs
```diff
@@ -347,6 +348,57 @@ where
...
+            };
+
+            // Only proceed if both epoch_number and certificate_index are present
+            let (epoch_number, certificate_index) =
+                match (header.epoch_number, header.certificate_index) {
+                    (Some(epoch), Some(idx)) => (epoch, idx),
+                    _ => continue,
```
Freyskeyd: This should return an error as we're iterating over settled certificates only

crates/agglayer-rpc/src/lib.rs
```diff
@@ -347,6 +348,57 @@ where
...
+
+            // Fetch the certificate from the epoch store
+            let certificate_opt = self
+                .epochs_store
+                .get_certificate(epoch_number, certificate_index)
+                .map_err(GetLatestSettledClaimError::from)?;
+
+            let certificate = match certificate_opt {
+                Some(cert) => cert,
+                None => continue,
```
Freyskeyd: Same here, should we return an error?

crates/agglayer-rpc/src/lib.rs
atanmarko: Yeah. But should the get network status API decide about those inconsistencies and return error?

crates/agglayer-rpc/src/lib.rs
atanmarko: Yeah, I was thinking about that. In all those cases, I'll return and error inconsistent state.

crates/agglayer-rpc/src/lib.rs
Freyskeyd: I don't think it's the responsibility of the API, nor the rpc crate to deal with storage inconsistency.
The RPC expect to walk through the list of settled certificate, a settled certificate has a `header` which contains `epoch_number` and `index`. If one of those things are not present, it means that there is a problem, that should be logged, but the query can't continue.

crates/agglayer-rpc/src/lib.rs
```diff
@@ -347,6 +348,88 @@ where
         }
     }
 
+    pub fn get_last_settled_claim(
```
Ekleog-Polygon: Nit: naming is inconsistent, between `get_last_settled_claim` and `GetLatestSettledClaimError`

crates/agglayer-rpc/src/lib.rs
```diff
@@ -347,6 +348,88 @@ where
...
+                    _ => {
+                        // Missing epoch information, return an error indicating inconsistent state
+                        error!(
+                            "Missing epoch information in certificate header for network \
+                             {network_id} at height {current_height}, inconsistent state"
+                        );
+                        return Err(GetLatestSettledClaimError::InconsistentState {
+                            network_id,
+                            height: settled_height,
+                        });
```
Ekleog-Polygon, line 392: Nit: all this is triplicated, we could have a `GetLatestSettledClaimError::inconsistent_and_log()` error that abstracts this away, or we could have an inner function that does this only once for all possible failure modes

crates/agglayer-rpc/src/lib.rs
```diff
@@ -347,6 +348,88 @@ where
...
+            if let Some(last_imported_exit) = certificate.imported_bridge_exits.last() {
+                let global_index_hash = last_imported_exit.global_index.hash();
+                let bridge_exit_hash = last_imported_exit.bridge_exit.hash();
+
+                return Ok(Some(SettledClaim {
+                    global_index: global_index_hash,
+                    bridge_exit_hash,
+                }));
+            }
+        }
```
Ekleog-Polygon, line 429: Very small nit: add a comment saying that we'll keep looping only if there's no imported bridge exit — there are many returns throughout the function, though maybe just making it shorter with the error management nit above could be enough to make it more readable :)

crates/agglayer-rpc/src/lib.rs
```diff
@@ -347,6 +348,88 @@ where
...
+
+        // No imported bridge exits found in any certificate from height down to 0
```
Ekleog-Polygon:
```suggestion
        // No imported bridge exits found in any certificate from `settled_height` down to 0
```

Ekleog-Polygon (dismissed):
LGTM! Just have a few nits below :smile:

crates/agglayer-rpc/src/lib.rs
atanmarko: @Ekleog-Polygon This is quite temporary implementation until we switch to alternative data collection, so does not make sense to invest time in polishing it.

Freyskeyd (approved):

Ekleog-Polygon (dismissed):

Ekleog-Polygon (dismissed):

Ekleog-Polygon (approved):

PR #1029, opened 2025-09-08
build: adding git cliff file

Freyskeyd:
This PR is adding git-cliff config


cliff.toml
```diff
@@ -0,0 +1,129 @@
...
+- [Brief description of new features]
+
+## ✅ Testing & Validation
+<!-- TEST_PLACEHOLDER_START -->
+<!-- TEST_PLACEHOLDER_END -->
+
+{% if breaking_commits %}
+  ## ⚠️ Breaking Changes
+  {%- for commit in breaking_commits %}
+    {%- for line in commit.footers %}
```
polykeith, line 39: It looks like we can also use the `breaking_description` attribute in the `commit`, as referenced here: https://git-cliff.org/docs/templating/context. Maybe that's a cleaner alternative?

Freyskeyd: Yes we could, I created a [ticket](https://github.com/agglayer/agglayer/issues/1031) to not block this PR

PR #1032, opened 2025-09-09
feat: adding proto storage network state

Freyskeyd:


crates/agglayer-types/src/network_info.rs
```diff
@@ -69,3 +68,23 @@ pub struct NetworkInfo {
     /// The epoch number of the latest settlement.
     pub latest_epoch_with_settlement: Option<u64>,
 }
+
+impl NetworkInfo {
+    pub fn with_network_id(network_id: NetworkId) -> Self {
```
iljakuklic: The `with_` prefix usually suggest some sort of builder method. Although `Vec::with_capacity` is a counterexample, I still find `from_` to be more natural here.

```suggestion
    pub fn from_network_id(network_id: NetworkId) -> Self {
```

proto/agglayer/storage/v0/network_info.proto
```diff
@@ -0,0 +1,86 @@
...
+  bytes global_index = 1;
+  /// Hash of the claimed imported bridge exit.
+  bytes bridge_exit_hash = 2;
+}
+
+// The height of the latest pending certificate.
+message LatestPendingHeight {
+  // The actual certificate height that is in pending
+  uint64 latest_pending_height = 1;
+}
```
iljakuklic: This could be changed to something like `LatestPendingCertificateInfo`. It can initially hold just the `height` field but the suggestion is to make it easy to extend it with other bits of information in the future as needed.

proto/agglayer/storage/v0/network_info.proto
```diff
@@ -0,0 +1,86 @@
...
+  NETWORK_TYPE_MULTISIG_ONLY = 3;
+  /// Multisig and aggchain proof network type.
+  NETWORK_TYPE_MULTISIG_AND_AGGCHAIN_PROOF = 4;
+}
+
+// Represents the current settled height for a given network.
+message SettledHeight {
+  // The actual certificate height settled
+  uint64 height = 1;
+}
```
iljakuklic: This is not used. I guess it's supposed to be included in the `SettledCertificate` message.

proto/agglayer/storage/v0/network_info.proto
```diff
@@ -0,0 +1,86 @@
...
+
+// Represents the current settled certificate for a given network.
+message SettledCertificate {
+  // The ID of the latest settled certificate.
+  SettledCertificateId certificate_id = 1;
+  // The pessimistic proof root of the latest settled certificate.
+  SettledPessimisticProofRoot new_pp_root = 2;
+  // The leaf count of the latest settled local exit tree.
+  SettledLocalExitTreeLeafCount let_leaf_count = 3;
```
iljakuklic, line 27: Hmm I am not sure if the extra wrapper messages (`SettledCertificateId` etc.) bring much value. This is already a message so if any of these need to be changed in the future, it could be done by deprecating or introducing new fields here directly. The extra indirection slightly increases code complexity and encoded size.

iljakuklic (commented):
Some initial comments, mostly regarding the protobuf definitions.

proto/agglayer/storage/v0/network_info.proto
Freyskeyd: Yes, but with that you're just changing the inner type, not the whole type of `SettledCertificate`.
It means that if you're changing the certificate_id layout 10 times, you end up with a really large `SettledCertificate` representation, while the other changes are scoped to a dedicated type representation.

crates/agglayer-jsonrpc-api/src/lib.rs
```diff
@@ -267,9 +266,9 @@ where
         Ok(header)
     }
 
-    async fn get_network_info(&self, network_id: NetworkId) -> RpcResult<NetworkInfo> {
+    async fn get_network_info(&self, network_id: NetworkId) -> RpcResult<Option<NetworkInfo>> {
         let state = self.rpc_service.get_network_info(network_id)?;
-        Ok(state)
+        Ok(Some(state))
```
Ekleog-Polygon: I'm curious, why make it return an `Option`? AFAIU it should either return something, or an `Err` if it failed to find anything? (because this is an RPC call)

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -25,6 +36,7 @@ pub const CERTIFICATE_PER_NETWORK_CF: &str = "certificate_per_network_cf";
 pub const NULLIFIER_TREE_PER_NETWORK_CF: &str = "nullifier_tree_per_network_cf";
 pub const BALANCE_TREE_PER_NETWORK_CF: &str = "balance_tree_per_network_cf";
 pub const LOCAL_EXIT_TREE_PER_NETWORK_CF: &str = "local_exit_tree_per_network_cf";
+pub const NETWORK_STATE_CF: &str = "network_state_cf";
```
Ekleog-Polygon: Hmm why do you need this? AFAIU NetworkInfo lives in `METADATA_CF` as the code is currently written, and we're not changing the code paths for network state?

crates/agglayer-storage/src/storage/cf_definitions/state.rs
```diff
@@ -1,13 +1,14 @@
...
-pub const CFS: [&str; 7] = [
+pub const CFS: [&str; 8] = [
     crate::columns::CERTIFICATE_HEADER_CF,
     crate::columns::CERTIFICATE_PER_NETWORK_CF,
     crate::columns::LATEST_SETTLED_CERTIFICATE_PER_NETWORK_CF,
     crate::columns::METADATA_CF,
     crate::columns::LOCAL_EXIT_TREE_PER_NETWORK_CF,
     crate::columns::BALANCE_TREE_PER_NETWORK_CF,
     crate::columns::NULLIFIER_TREE_PER_NETWORK_CF,
+    crate::columns::NETWORK_STATE_CF,
```
Ekleog-Polygon: Same here, AFAIU we don't need a new CF (and if we needed one it'd be `NETWORK_INFO_CF`)

crates/agglayer-storage/src/storage/mod.rs
```diff
@@ -96,6 +96,36 @@ impl DB {
             .map_or(Ok(None), |v| v.map(Some))
     }
 
+    pub fn atomic_multi_get<C: ColumnSchema>(
```
Ekleog-Polygon, line 109: Just to check, did you (or someone else) actually investigate and confirm that `multi_get` really is not atomic by default? I only had what ChatGPT told me, and the docs were unclear; but I didn't check much further. Maybe `multi_get` already takes a snapshot internally, in which case this would not be required?

crates/agglayer-storage/src/stores/interfaces/reader/network_info_reader.rs
```diff
@@ -0,0 +1,14 @@
...
+
+pub trait NetworkInfoReader: Send + Sync {
+    fn get_network_info(&self, network_id: NetworkId) -> Result<Option<NetworkInfo>, Error>;
+
+    fn get_latest_pending_height(&self, network_id: NetworkId) -> Result<Option<Height>, Error>;
+
+    fn get_latest_settled_certificate_id(
+        &self,
+        network_id: NetworkId,
+    ) -> Result<Option<CertificateId>, Error>;
```
Ekleog-Polygon, line 13: AFAIR these are currently unused, can we remove them?

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -591,3 +596,209 @@ impl MetadataReader for StateStore {
...
+                            network_info::v0::network_info_value::Value::SettledClaim(claim,),
+                            agglayer_types::SettledClaim {
+                                global_index: Digest::try_from(&*claim.global_index).map_err(
+                                    |_| {
+                                        Error::Unexpected(
+                                            "Unable to deserialize GlobalIndex into a Digest"
+                                                .to_string(),
+                                        )
+                                    }
+                                )?,
```
Ekleog-Polygon: This pattern seems repeated quite often, it might deserve a macro too?

crates/agglayer-storage/src/types/network_info/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+        let mut buf = Vec::with_capacity(expected_length);
+        <Value as prost::Message>::encode(self, &mut buf)?;
+
+        Ok(buf)
+    }
+
+    fn decode(buf: &[u8]) -> Result<Self, crate::columns::CodecError> {
+        <Value as prost::Message>::decode(buf).map_err(Into::into)
+    }
+}
```
Ekleog-Polygon, line 51: Nit: it'd be great to have a similar macro as the one @iljakuklic recently did for the `serde` implementation

crates/agglayer-types/src/network_info.rs
```diff
@@ -22,14 +20,15 @@ pub enum NetworkStatus {
 // The aggchain type of network
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub enum NetworkType {
```
Ekleog-Polygon, line 25: Will this not break the (de)serialize API?

crates/agglayer-types/src/network_info.rs
```diff
@@ -69,3 +68,23 @@ pub struct NetworkInfo {
     /// The epoch number of the latest settlement.
     pub latest_epoch_with_settlement: Option<u64>,
 }
+
+impl NetworkInfo {
+    pub fn from_network_id(network_id: NetworkId) -> Self {
+        Self {
+            network_status: NetworkStatus::Active,
```
Ekleog-Polygon: `NetworkStatus::Active` sounds weird. Should we have an `Unspecified` variant, and use it here? After all `network_id` might as well not exist yet.

scripts/make/Makefile.proto.toml
```diff
@@ -1,6 +1,27 @@
 [tasks.generate-proto]
+dependencies = [
+    "install-buf",
+    "install-protoc-gen-prost",
+    "install-protoc-gen-prost-serde",
+    "install-protoc-gen-prost-crate",
+    "install-protoc-gen-tonic",
```
Ekleog-Polygon, line 7:
```suggestion
```
I think having these as deps of the sub-tasks is enough?

crates/agglayer-storage/src/storage/mod.rs
Freyskeyd: I didn't find any proof that `multi_get` is atomic or is using snapshot internally.

crates/agglayer-storage/src/types/network_info/mod.rs
Freyskeyd: Created a ticket for this [here](https://github.com/agglayer/agglayer/issues/1034)

crates/agglayer-types/src/network_info.rs
Freyskeyd: this is not deploy anywhere so I guess its fine

crates/agglayer-rpc/src/lib.rs
```diff
@@ -634,6 +629,31 @@ where
...
+where
+    PendingStore: Send + Sync + 'static,
+    StateStore: NetworkInfoReader + 'static,
+    DebugStore: Send + Sync + 'static,
+    L1Rpc: Send + Sync + 'static,
+    EpochsStore: EpochStoreReader + 'static,
+{
+    /// Assemble the current state of the specified network from
+    /// the data in various sources.
+    pub fn get_network_state(
```
iljakuklic:
```suggestion
    pub fn get_network_info(
```

buf.storage.gen.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+plugins:
+  - local: protoc-gen-prost
+    out: crates/agglayer-storage/src/types/generated
+    opt:
+      - bytes=.
+      - compile_well_known_types
+
+  - local: protoc-gen-prost-serde
+    out: crates/agglayer-storage/src/types/generated
```
iljakuklic: Do we want/need serde impls on the protobuf types?

(Removing may require #1033 first)

crates/agglayer-storage/src/storage/mod.rs
```diff
@@ -96,6 +96,36 @@ impl DB {
...
+            .collect();
+
+        let results: Result<Vec<Option<_>>, _> = snapshot
+            .multi_get_cf(keys?)
+            .into_iter()
+            .map(|r| r.map_err(DBError::from))
+            .collect();
+
+        results?
+            .into_iter()
```
iljakuklic: Style nit (have not checked it works):
```rust
        let results = snapshot
            .multi_get_cf(keys?)
            .into_iter()
            .map(|r| r.map_err(DBError::from))
            .collect::<Result<Vec<Option<_>>, _>>()?;

        results
            .into_iter()
```

proto/agglayer/storage/v0/network_info.proto
```diff
@@ -0,0 +1,82 @@
...
+  bytes bridge_exit_hash = 2;
+}
+
+// The latest pending certificate.
+message LatestPendingCertificateInfo {
+  // The actual certificate height that is in pending
+  uint64 height = 1;
+  // The ID of the latest pending certificate.
+  bytes id = 2;
+}
```
iljakuklic, line 108: There is an apparent inconsistency. The `SettledCertificate` message only contains other message types. However, the likes of `SettledClaim` or `LatestPendingCertificateInfo` contain fields with plain types directly.

Any specific reason for that?

* related discussion https://github.com/agglayer/agglayer/pull/1032#discussion_r2336470140

crates/agglayer-storage/src/types/network_info/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::{columns::Codec, types::network_info::v0::NetworkType};
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Key {
+    pub(crate) network_id: u32,
+    pub(crate) kind: KeyKind,
+}
+
+impl Key {
+    pub(crate) fn all_keys_for_network(network_id: agglayer_types::NetworkId) -> Vec<Self> {
```
iljakuklic: How about returning an iterator?
```suggestion
    pub(crate) fn all_keys_for_network(network_id: agglayer_types::NetworkId) -> impl ExactSizeIterator<Item = Self> {
```

crates/agglayer-storage/src/types/network_info/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+            Key {
+                network_id: network_id.to_u32(),
+                kind: KeyKind::LatestPendingHeight,
+            },
+        ]
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub enum KeyKind {
```
iljakuklic: Nit: might be a bit more modular to add a method to iterate over all `KeyKind`s (I believe the `strum` package can even derive it), and use it in the implementation of `Key::all_keys_for_network`.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -591,3 +596,209 @@ impl MetadataReader for StateStore {
...
+                value: Some($pattern),
+            }) => Ok(Some($return_value)),
+            None => Ok(None),
+            _ => Err(Error::Unexpected($err_msg.to_string())),
+        }
+    };
+}
+impl NetworkInfoReader for StateStore {
+    fn get_network_info(&self, network_id: NetworkId) -> Result<Option<NetworkInfo>, Error> {
+        let mut state = NetworkInfo::from_network_id(network_id);
```
iljakuklic:
```suggestion
        let mut info = NetworkInfo::from_network_id(network_id);
```

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -591,3 +596,209 @@ impl MetadataReader for StateStore {
...
+        self.db
+            .atomic_multi_get::<NetworkInfoColumn>(keys.clone())?
+            .into_iter()
+            .zip(keys)
+            .try_for_each(|(maybe_value, network_info::Key { kind, .. })| {
+                match kind {
+                    network_info::KeyKind::LatestSettledClaim => {
+                        state.settled_claim = expected_type_or_fail!(
+                            maybe_value,
+                            network_info::v0::network_info_value::Value::SettledClaim(claim,),
```
iljakuklic: Nit: weird looking extra comma:

```suggestion
                            network_info::v0::network_info_value::Value::SettledClaim(claim),
```

I'm wondering if the scope of the macro could be reduced somehow? Whatever is inside of a macro is not subject to the processing by the formatter etc.

crates/agglayer-jsonrpc-api/src/lib.rs
```diff
@@ -82,7 +81,7 @@ trait Agglayer {
     ) -> RpcResult<Option<CertificateHeader>>;
 
     #[method(name = "getNetworkInfo")]
-    async fn get_network_info(&self, network_id: NetworkId) -> RpcResult<NetworkInfo>;
+    async fn get_network_info(&self, network_id: NetworkId) -> RpcResult<Option<NetworkInfo>>;
```
atanmarko: Does it make sense to not being able to return any info about the network, without an error happened? Not having any info because the network just started could be an error "no data about the network yet".

crates/agglayer-storage/src/stores/interfaces/reader/network_info_reader.rs
atanmarko: I don't think it hurts to have it here.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -358,10 +369,7 @@ impl StateStore {
         Ok(())
     }
 
-    fn read_local_exit_tree(
-        &self,
-        network_id: NetworkId,
-    ) -> Result<Option<LocalExitTree>, Error> {
+    fn read_local_exit_tree(&self, network_id: NetworkId) -> Result<Option<LocalExitTree>, Error> {
```
atanmarko, line 362: Nit: do you have some different formatting settings?

crates/agglayer-types/src/network_info.rs
```diff
@@ -69,3 +68,23 @@ pub struct NetworkInfo {
     /// The epoch number of the latest settlement.
     pub latest_epoch_with_settlement: Option<u64>,
 }
+
+impl NetworkInfo {
+    pub fn from_network_id(network_id: NetworkId) -> Self {
```
atanmarko: Nit: implement `Default` and `new`

crates/agglayer-types/src/network_info.rs
atanmarko: Yes, `Unspecified` makes sense here.

atanmarko (approved):
Approved with some nits.

buf.storage.gen.yaml
Freyskeyd: ~~it doesn't seem to be that straightforward, will merge this PR with `serde` and iterate after that with [this](https://github.com/agglayer/agglayer/issues/1044) ticket~~

crates/agglayer-storage/src/stores/state/mod.rs
Freyskeyd: I don't think so, this update respects the `rustfmt` config that we're using tho

crates/agglayer-rpc/src/lib.rs
```diff
@@ -434,198 +437,202 @@ where
...
-                        network_id,
-                        source: error.into(),
+        if network_info.network_type == NetworkType::Unspecified {
+            // Determine network type from the latest available certificate
+            network_info.network_type =
+                match self.get_latest_available_certificate_for_network(network_id) {
+                    Ok(Some(certificate)) => {
+                        // Determine network type based on aggchain_data variant
+                        match certificate.aggchain_data {
+                    agglayer_types::aggchain_proof::AggchainData::ECDSA { .. } => {
```
Ekleog-Polygon: Hmm the indent looks wrong here? Also, it might make sense to have that match be a helper function IMO

crates/agglayer-storage/src/types/network_info/mod.rs
```diff
@@ -41,17 +46,21 @@ pub enum KeyKind {
...
 
-impl Codec for Key {}
-
 pub type Value = super::generated::agglayer::storage::v0::NetworkInfoValue;
 
 impl Codec for Value {
-    fn encode(&self) -> Result<Vec<u8>, crate::columns::CodecError> {
-        let expected_length = self.encoded_len();
-        let mut buf = Vec::with_capacity(expected_length);
+    fn encode_into<W: io::Write>(&self, mut writer: W) -> Result<(), crate::columns::CodecError> {
```
Ekleog-Polygon: Can we take `&mut bytes::BufMut` here instead of `io::Write`? It'd allow us to skip one allocation for free, I think, assuming this will still mostly be used to write into a `Vec`, because `Vec` implements `BufMut` — this being said I don't see that function being called anywhere yet so I may be wrong

crates/agglayer-storage/src/types/network_info/mod.rs
```diff
@@ -2,50 +2,58 @@ use std::io;
...
 
 #[derive(Clone, Serialize, Deserialize)]
 pub struct Key {
     pub(crate) network_id: u32,
-    pub(crate) kind: KeyKind,
+    #[serde(
+        serialize_with = "serialize_kind",
+        deserialize_with = "deserialize_kind"
+    )]
+    pub(crate) kind: ValueDiscriminants,
```
Ekleog-Polygon, line 19: IIUC you can use [`#[strum_discriminants(derive(serde::Deserialize, serde::Serialize))]`](http://docs.rs/strum_macros/latest/strum_macros/derive.EnumDiscriminants.html) to avoid the need for this code :smile:

Ekleog-Polygon (approved):
LGTM! I just have a few nits down below :smile:

crates/agglayer-storage/src/types/network_info/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+
+pub type Value = super::generated::agglayer::storage::v0::NetworkInfoValue;
+
+impl Codec for Value {
+    fn encode_into<W: io::Write>(&self, mut writer: W) -> Result<(), crate::columns::CodecError> {
+        let len = self.encoded_len();
+
+        let mut buf = BytesMut::new();
+        buf.reserve(len);
+        <Value as prost::Message>::encode(self, &mut buf)?;
```
iljakuklic, line 39: This could be factored out into the `encode` trait method, instead relying on the default impl (which creates an extra temporary).

crates/agglayer-storage/src/types/network_info/mod.rs
iljakuklic: Hmm, `bincode` can write into a `io::Write` but protobuf favours `BufMut`. Seems the intersection between the two would be to just take a `&mut Vec<u8>`. Or require `W: io::Write + BufMut` which is satisfied by `&mut Vec<u8>` and `&mut [u8]` and maybe handful of other types. However it seems a bit awkward to basically define interface requiremets on basis of the union of implementation requirements.

iljakuklic (dismissed):

crates/agglayer-storage/src/types/network_info/mod.rs
Freyskeyd: @iljakuklic @Ekleog-Polygon I created a ticket for this [here](https://github.com/agglayer/agglayer/issues/1049)

crates/agglayer-storage/src/types/network_info/mod.rs
Freyskeyd: That's what I did at first, but as this is defined in the buf yaml I wasn't sure about the visibility, I'll try to refactor

crates/agglayer-rpc/src/lib.rs
Freyskeyd: Mmmh I don't see the issue with the indent, I will refactor this to make it more clear

iljakuklic (approved):

polykeith (approved):

PR #1033, opened 2025-09-10
chore: remove `serde` requirement from storage codec

iljakuklic:
With the gradual move towards protobuf in storage, the requirement for all storage types to implement the `serde` traits is unnecessary and a bit limiting. This removes the `serde` supertraits from `agglayer_storage::columns::Codec`. Consequently, we do not have convenient default implementation for the methods so a macro is introduced for it.

### DIFF
--- a/crates/agglayer-storage/src/columns/mod.rs
+++ b/crates/agglayer-storage/src/columns/mod.rs
@@ -1,3 +1,5 @@
+pub use std::io;
+
 use agglayer_types::bincode;
 
 #[derive(Debug, thiserror::Error)]
@@ -50,7 +52,14 @@ pub const PROOF_PER_CERTIFICATE_CF: &str = "proof_per_certificate_cf";
 pub const DEBUG_CERTIFICATES_CF: &str = "debug_certificates";
 
 pub trait Codec: Sized {
-    fn encode(&self) -> Result<Vec<u8>, CodecError>;
+    #[inline]
+    fn encode(&self) -> Result<Vec<u8>, CodecError> {
+        let mut buffer = Vec::new();
+        self.encode_into(&mut buffer)?;
+        Ok(buffer)
+    }
+
+    fn encode_into<W: io::Write>(&self, writer: W) -> Result<(), CodecError>;
 
     fn decode(buf: &[u8]) -> Result<Self, CodecError>;
 }
@@ -59,8 +68,11 @@ macro_rules! impl_codec_using_bincode_for {
     ($($type:ty),* $(,)?) => {
         $(
             impl $crate::columns::Codec for $type {
-                fn encode(&self) -> Result<Vec<u8>, $crate::columns::CodecError> {
-                    Ok($crate::columns::bincode_codec().serialize(self)?)
+                fn encode_into<W: $crate::columns::io::Write>(
+                    &self,
+                    writer: W,
+                ) -> Result<(), $crate::columns::CodecError> {
+                    Ok($crate::columns::bincode_codec().serialize_into(writer, self)?)
                 }
 
                 fn decode(buf: &[u8]) -> Result<Self, $crate::columns::CodecError> {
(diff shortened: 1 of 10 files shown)


crates/agglayer-storage/src/lib.rs
```diff
@@ -1,10 +1,27 @@
+macro_rules! impl_codec_using_bincode_for {
```
iljakuklic: I am not mega thrilled this is at the top of the crate root file. However, I find the declarative macro scoping rules somewhat confusing and this is what I managed to get working. Suggestions on how to move it in a submodule (either in `columns` or a new one) are welcome.

Ekleog-Polygon: AFAIR this `pub use` solution is the correct one, though I haven't tried it in your exact example 😄
https://stackoverflow.com/questions/26731243/how-do-i-use-a-macro-across-module-files

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -50,14 +49,10 @@ pub const PROOF_PER_CERTIFICATE_CF: &str = "proof_per_certificate_cf";
 // debug CFs
 pub const DEBUG_CERTIFICATES_CF: &str = "debug_certificates";
 
-pub trait Codec: Sized + Serialize + DeserializeOwned {
-    fn encode(&self) -> Result<Vec<u8>, CodecError> {
-        Ok(bincode_codec().serialize(self)?)
-    }
+pub trait Codec: Sized {
+    fn encode(&self) -> Result<Vec<u8>, CodecError>;
```
Ekleog-Polygon: Nit: If we're breaking compatibility here anyway, could we have `fn encode(&self, buf: &mut Vec<u8>)` instead, that appends the encoding to the passed-in buffer? This could allow us to reuse allocations, and various other slight performance wins — not worth breaking compat for, but we're doing it already here :smile:

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -50,14 +49,10 @@ pub const PROOF_PER_CERTIFICATE_CF: &str = "proof_per_certificate_cf";
...
 
-    fn decode(buf: &[u8]) -> Result<Self, CodecError> {
-        Ok(bincode_codec().deserialize(buf)?)
-    }
+    fn decode(buf: &[u8]) -> Result<Self, CodecError>;
```
Ekleog-Polygon, line 64: I'd like to suggest also returning the number of bytes actually read from the buffer here, but it seems like while `bincode::decode_from_slice` does return it, the serde variant doesn't, so maybe it's not worth thinking too much about for now.

crates/agglayer-storage/src/lib.rs
```diff
@@ -1,10 +1,27 @@
...
+}
+
 // Physical storage
 #[rustfmt::skip]
 pub mod storage;
 // Logical store
 #[rustfmt::skip]
 pub mod stores;
 
+#[macro_use]
```
polykeith, line 8: Nit: remove this if not declaring the macro inside of `columns`.

iljakuklic: > Nit: remove this if not declaring the macro inside of `columns`.

Just moved the macro there.

crates/agglayer-storage/src/columns/mod.rs
iljakuklic: Added `encode_into` which takes a writer. Does that look reasonable to you?

crates/agglayer-storage/src/columns/mod.rs
iljakuklic: The semantics here are that there should not be any leftover bytes in the buffer, so it should be `buf.len()` basically. Plus to avoid the additional burden on the implementations to track bytes, I'd prefer not to add it.

PR #1041, opened 2025-09-16
fix: use `DB::default_write_options` consistently

iljakuklic:
* The options were not previously passed to the batch write method.
* Also a small refactor to get the column family easier.
* Fixes agglayer/security#104.


atanmarko (approved):
LGTM

Ekleog-Polygon (approved):

PR #1047, opened 2025-09-19
refactor: use vkey hashing utilities from agglayer-primitives

polykeith:
Replace the ad-hoc implementation of the vkey hashing operations and utilities with the one located in agglayer-primitives.

Fixies agglayer/interop#101

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
@@ -110,12 +110,14 @@ where
             .context("Failed to hash SP1 vkey")
             .map_err(CertificationError::Other)?;
 
-        let proof_vk_hash = agglayer_contracts::aggchain::AggchainVkeyHash::new(vkey_hash_bytes);
+        let vkey_digest = Digest::from(vkey_hash_bytes);
+
+        let proof_vk_hash = agglayer_contracts::aggchain::VKeyHash::from(vkey_digest);
 
         if aggchain_vkey != proof_vk_hash {
             return Err(CertificationError::AggchainProofVkeyMismatch {
-                expected: aggchain_vkey.to_hex(),
-                actual: proof_vk_hash.to_hex(),
+                expected: aggchain_vkey.to_bytes().to_string(),
+                actual: proof_vk_hash.to_bytes().to_string(),
             });
         }
 
--- a/crates/agglayer-contracts/src/aggchain.rs
+++ b/crates/agglayer-contracts/src/aggchain.rs
@@ -1,21 +1,10 @@
+pub use agglayer_primitives::vkey_hash::VKeyHash;
+
 use agglayer_primitives::{Address, U256};
 use alloy::primitives::Bytes;
 use tracing::error;
 
 use crate::{contracts::AggchainBase, L1RpcClient, L1RpcError};
-
-#[derive(PartialEq, Eq)]
-pub struct AggchainVkeyHash([u8; 32]);
-
-impl AggchainVkeyHash {
-    pub fn new(vkey: [u8; 32]) -> Self {
-        Self(vkey)
-    }
-
-    pub fn to_hex(&self) -> String {
-        hex::encode(self.0)
-    }
-}
 
 #[async_trait::async_trait]
 pub trait AggchainContract {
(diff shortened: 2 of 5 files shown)


atanmarko (dismissed):

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
```diff
@@ -110,12 +110,14 @@ where
...
+        let vkey_digest = Digest::from(vkey_hash_bytes);
+
+        let proof_vk_hash = agglayer_contracts::aggchain::VKeyHash::from(vkey_digest);
 
         if aggchain_vkey != proof_vk_hash {
             return Err(CertificationError::AggchainProofVkeyMismatch {
-                expected: aggchain_vkey.to_hex(),
-                actual: proof_vk_hash.to_hex(),
+                expected: aggchain_vkey.to_bytes().to_string(),
+                actual: proof_vk_hash.to_bytes().to_string(),
```
iljakuklic, line 120: Note: maybe `LowerHex` and `UpperHex` impls could be added to `VKeyHash`.

crates/agglayer-contracts/src/aggchain.rs
```diff
@@ -55,7 +43,7 @@ where
             .getAggchainVKey(alloy::primitives::FixedBytes(aggchain_selector))
             .call()
             .await
-            .map(|arg0| AggchainVkeyHash::new(arg0.0))
+            .map(|arg0| VKeyHash::from(arg0))
```
iljakuklic: Clippy suggests:
```suggestion
            .map(VKeyHash::from)
```

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
```diff
@@ -110,12 +110,14 @@ where
...
-        let proof_vk_hash = agglayer_contracts::aggchain::AggchainVkeyHash::new(vkey_hash_bytes);
+        let vkey_digest = Digest::from(vkey_hash_bytes);
+
+        let proof_vk_hash = agglayer_contracts::aggchain::VKeyHash::from(vkey_digest);
 
         if aggchain_vkey != proof_vk_hash {
             return Err(CertificationError::AggchainProofVkeyMismatch {
-                expected: aggchain_vkey.to_hex(),
-                actual: proof_vk_hash.to_hex(),
+                expected: aggchain_vkey.to_bytes().to_string(),
```
iljakuklic, line 119: I am not sure if this is exactly the same with respect to lowercase vs uppercase hex and the presence of the `0x` prefix. Hopefully it does not matter much since it is just a string inside of an error.

iljakuklic (approved):

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
polykeith: I did research on the `Display` implementation on the `B256` type (alias for FixedBytes<32>), and it looks like this:

```rust
impl<const N: usize> fmt::Display for FixedBytes<N> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        // If the alternate flag is NOT set, we write the full hex.
        if N <= 4 || !f.alternate() {
            return self.fmt_hex::<false>(f, true);
        }

        // If the alternate flag is set, we use middle-out compression.
        const SEP_LEN: usize = '…'.len_utf8();
        let mut buf = [0; 2 + 4 + SEP_LEN + 4];
        buf[0] = b'0';
        buf[1] = b'x';
        hex::encode_to_slice(&self.0[0..2], &mut buf[2..6]).unwrap();
        '…'.encode_utf8(&mut buf[6..]);
        hex::encode_to_slice(&self.0[N - 2..N], &mut buf[6 + SEP_LEN..]).unwrap();

        // SAFETY: always valid UTF-8
        f.write_str(unsafe { str::from_utf8_unchecked(&buf) })
    }
}
```

Internally it still uses `hex` to encode the bytes, so I think we're fine here.

crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
polykeith: Can do in a later release of `agglayer-primitives`.

Ekleog-Polygon (approved):

atanmarko (approved):

PR #1053, opened 2025-09-19
feat: update contracts

atanmarko:
Resolves https://github.com/agglayer/agglayer/issues/1046

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/lib.rs
+++ b/crates/agglayer-aggregator-notifier/src/lib.rs
@@ -2,7 +2,6 @@
 const ELF: &[u8] = pessimistic_proof::ELF;
 
 mod certifier;
-mod proof;
 mod settlement_client;
 
 pub use certifier::CertifierClient;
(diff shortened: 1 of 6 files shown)


crates/agglayer-contracts/src/contracts.rs
```diff
@@ -1,23 +1,35 @@
...
+        #[derive(Debug, Eq, PartialEq)]
+        AgglayerGateway,
+        "src/contracts/AggLayerGateway.json"
+    );
+}
+
+// Re-export the contracts
+pub use aggchain_base::AggchainBase;
+// Use the IAggchainSigners from one of the modules to avoid duplication
+pub use aggchain_base::IAggchainSigners;
```
Ekleog-Polygon, line 31: Would feel better if there were no IAggchainSigners duplication here :smiling_face_with_tear: I'm fearing some day we'll have issues due to the two not referring to the same type.

I guess there's no way we could unify the JSONs so they use the same IAggchainSigners? And we probably can't just not re-export it either?

Ekleog-Polygon (approved):
LGTM! If everything still builds fine I'm happy with it, we can iterate on my comment whenever an issue actually arises, as it'll certainly be a compile-time issue :smile:

crates/agglayer-contracts/src/contracts.rs
atanmarko: Yeah, this is generated from the ABI compiled by the smart contracts project, we are just using it.

PR #1059, opened 2025-09-22
chore: update provers dependency

atanmarko:


Ekleog-Polygon (approved):

crates/pessimistic-proof-core/src/aggchain_data/aggchain_proof.rs
```diff
@@ -18,7 +18,7 @@ pub struct AggchainProof {
 impl AggchainProof {
     /// Verifies the next proof in the sp1 buffer.
     pub fn verify_aggchain_proof(&self, constrained_values: &ConstrainedValues) {
-        let aggchain_proof_public_values = AggchainProofPublicValues {
+        let _aggchain_proof_public_values = AggchainProofPublicValues {
```
iljakuklic, line 21: Would it make sense to only construct the `aggchain_proof_public_values` object if running on the zkvm?

atanmarko: @iljakuklic Tried it, then whole function, imports need to be in the ` #[cfg(target_os = "zkvm")]`, so decided to go this way eventually.

Issue #1058, opened 2025-09-22
Migrate latest_*_certificate_per_network column families to the new network_info column


Ekleog-Polygon:
1. Copy the contents to network_info at the startup of the node
2. (Might want to first copy the database then open one RO and one RW before doing the copy, so the RO db can serve as implicit backup)
3. Kill the column families after the copy is completed
4. Reroute all reads and writes directly to network_info
5. Write documentation on how to perform the migration from an operational point of view (backup the database before / delete the automatic database backup after?)

PR #1060, opened 2025-09-23
feat: add required fields to network info storage

Ekleog-Polygon:
Add the fields required for [Migrate latest_*_certificate_per_network column families to the new network_info column](https://github.com/agglayer/agglayer/issues/1058#top)#1058 to the network info column.

Next pr is going to be writing the actual migration and rerouting the calls.


proto/agglayer/storage/v0/network_info.proto
```diff
@@ -8,69 +8,86 @@ message NetworkInfoValue {
...
+  LocalExitRoot ler = 4;
+
+  // The height of the latest settled certificate.
+  Height height = 5;
+
+  // The epoch number in which the latest settled certificate was settled.
+  EpochNumber epoch = 6;
+
+  // The index of the certificate in the epoch.
+  CertificateIndexInEpoch index_in_epoch = 7;
```
Freyskeyd, line 47: This is a duplicate of what the `CertificateHeader` contains, are we not afraid of having those info duplicated?

Ekleog-Polygon: It's already present in the `latest_settled_certificate_per_network` CF, so I kind of mindlessly added it here too.
I'm fine with removing that information! I'm not entirely sure why it's here (apart from avoiding a second request to the certificate header CF), and I mostly added it here to avoid too many big changes at once, and because removing fields from proto is usually not a big issue :smile:

This being said… arguably, all the information here should belong in `CertificateId` for tracability anyway, esp. PPRoot. And eg. `height` and `network_id` are already duplicated between the two CFs.

WDYT? My initial guess would be it's best to just duplicate everything until the day we decide to duplicate nothing; but it's not at all a strong opinion :smile:

atanmarko:
@Ekleog-Polygon Is this one still relevant?

Ekleog-Polygon:
I think it is relevant long-term for our protobuf migrations — not a P0, but something we'll want to do at some point

hadjiszs:
let's reopen if/when needed

PR #1062, opened 2025-09-23
refactor: introduce a type for specifying ports in config

iljakuklic:
More ports are coming with the introduction of TLS. This will reduce the repetitive boilerplate somewhat.


Cargo.toml
```diff
@@ -88,6 +88,7 @@ criterion = "0.5.1"
 derive_more = "2.0"
 dirs = "5.0"
 dotenvy = "0.15.7"
+educe = { version = "0.6", features = ["Debug", "Clone", "Eq", "Ord"] }
```
atanmarko, line 91: What was the main reason to add this new dependency, that we were not able to do in a standard way?

iljakuklic: > What was the main reason to add this new dependency, that we were not able to do in a standard way?

Unfortunately, the built-in `derive` adds a bunch of spurious trait bounds on the phantom type. An alternative would be to implement the traits manually.

iljakuklic: > What was the main reason to add this new dependency, that we were not able to do in a standard way?

Also, we already have this dependency in `interop` and `provers` workspaces.

atanmarko (dismissed):

crates/agglayer-config/src/port.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub const fn as_u16(&self) -> u16 {
+        self.port
+    }
+
+    pub fn from_env_var(env_var: &str) -> Option<Self> {
+        env::var(env_var)
+            .inspect_err(|err| match err {
+                env::VarError::NotPresent => (),
+                env::VarError::NotUnicode(_) => warn!("Contents of ${env_var} not Unicode"),
+            })
```
polykeith, line 39:
```suggestion
            .inspect_err(|err| if let env::VarError::NotUnicode(_) = err {
                warn!("Contents of ${env_var} not Unicode"),
            })
```

crates/agglayer-config/src/port.rs
```diff
@@ -0,0 +1,81 @@
+use std::env;
+
+use tracing::warn;
+
+/// Default port for given service.
+pub trait PortDefaults {
```
polykeith, line 6: I'm curious, why not an enum containing all the variants for the services instead of declaring a trait?

iljakuklic: > I'm curious, why not an enum containing all the variants for the services instead of declaring a trait?

I want to have the `Default` impl that fills in the correct default port, so the relevant fields can be `#[serde(default)]`ed. So these need to be separate types. An enum value would still have to be passed in as a const generic or something like that. Also this way, it could be easily factored out into `interop` to be usable elsewhere without knowing all the services upfront.

crates/agglayer-config/src/port.rs
iljakuklic: Since there are not that many enum arms, I prefer to be explicit about the fact that all cases have been covered.

polykeith (dismissed):

crates/agglayer-config/src/port.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    pub const fn as_u16(&self) -> u16 {
+        self.port
+    }
+
+    pub fn from_env_var(env_var: &str) -> Option<Self> {
+        env::var(env_var)
+            .inspect_err(|err| match err {
+                env::VarError::NotPresent => (),
+                env::VarError::NotUnicode(_) => warn!("Contents of ${env_var} not Unicode"),
```
Ekleog-Polygon:
```suggestion
                env::VarError::NotUnicode(_) => warn!("Contents of ${env_var} are not Unicode"),
```

crates/agglayer-config/src/port.rs
```diff
@@ -0,0 +1,81 @@
...
+            })
+            .ok()?
+            .parse()
+            .inspect_err(|_| warn!("Variable ${env_var} not a valid port number"))
```
Ekleog-Polygon:
```suggestion
            .inspect_err(|_| warn!("Variable ${env_var} is not a valid port number"))
```

Ekleog-Polygon (approved):
Very happy to see these things cleaned up :tada:

polykeith (approved):

PR #1063, opened 2025-09-24
chore: replace pessimistic proof `TODO`s with issues

iljakuklic:
Introduce Github issues in place of TODOs so we can track them better.

* [ ] Introduce/locate issues covering the TODOs.
* [ ] List them below.
* [ ] Remove the TODOs from the code.

Related:
* Fixes agglayer/security#182
* Fixes agglayer/security#89

Issues replacing the TODOs:
* NetworkId bit width discussed in agglayer/protocol-research#128
* Merkle proof verification batching #587


crates/pessimistic-proof-test-suite/tests/snapshots/vkey_selector__vkey_snapshot.snap
```diff
@@ -5,5 +5,5 @@ snapshot_kind: text
 ---
 # If this test fails, it means the PP vkey has changed.
 # When that happens, consider updating the selector by bumping the PP version.
-| PP_VKEY          | 0x006d77897e2eb9f59938982739f46f93febf1c04aaa7d4f1539d1bc02785afc8 |
+| PP_VKEY          | 0x00d8a70d1445498036f934082b837c5d00ee35b809f1d3010a859a6461b4dbe1 |
```
iljakuklic, line 8: Aaargh, why does removing bloody comments change the vkey? 😕 🤷 Very irritating.

Probably `panic!` et al. generate a message with a line number that ends up in the binary.

crates/pessimistic-proof-core/src/nullifier_tree.rs
```diff
@@ -7,12 +7,8 @@ use unified_bridge::{GlobalIndex, NetworkId};
 use crate::ProofError;
 
 // 32 bits for the network id and 32 bits for the LET index
-// TODO: consider using less than 32 bits for the network id - unlikely that
-// we'll have 4 billion chains :)
 pub const NULLIFIER_TREE_DEPTH: usize = 64;
```
polykeith, line 10: `usize` doesn't necessarily mean 32 bits -- it's dependent on the target architecture. I would consider changing this to a specific numbered unsigned integer. However, that also doesn't seem ideal since this constant is primarily used in determining the tree depth, which correctly accepts a `usize`... so perhaps the original TODO issue is moot.

iljakuklic: > `usize` doesn't necessarily mean 32 bits

The `NetworkId` type uses `u32`. The bits of network id are used to determine the path down the binary tree (the top 32 levels). The idea is to reduce the number of bits in `NetworkId` so the tree is shallower. That would improve proving efficiency as there would be fewer Merkle tree levels to check. Specifying the total depth as a `usize` seems appropriate to me.

atanmarko (approved):

Ekleog-Polygon (approved):

PR #1077, opened 2025-09-30
fix: update witness_generation to accept a tx hash

Freyskeyd:
updates witness_generation to accept optional prev_pessimistic_root_before_tx parameter and modifies get_prev_pessimistic_root to use transaction-specific block context for more accurate state queries.

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
@@ -353,7 +353,12 @@ where
             .await
             .map_err(|_| CertificationError::LastPessimisticRootNotFound(network_id))?;
 
-        debug!("Prev PP root from L1: {}", Digest(prev_pessimistic_root));
+        debug!(
+            from_specific_block = certificate_tx_hash.is_some(),
+            "Prev PP root from L1: {}",
+            Digest(prev_pessimistic_root)
+        );
+
         let declared_l1_info_root = certificate
             .l1_info_root()
             .map_err(|source| CertificationError::Types { source })?;


crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -325,6 +326,7 @@ where
         &self,
         certificate: &Certificate,
         state: &mut LocalNetworkStateData,
+        prev_pessimistic_root_before_tx: Option<Digest>,
```
atanmarko: Nit: a bit ambiguous argument name, some doc would be nice.

crates/agglayer-aggregator-notifier/src/packer/mod.rs
```diff
@@ -479,6 +479,8 @@ where
         |_| { Ok(None) }
     );
 
+    fail::fail_point!("notifier::packer::settle_certificate::receipt_future_ended::kill_node");
```
atanmarko, line 482: Note: I am not a fan of all those fail points. I would put them behind some `#[cfg(any(test, feature = "testutils"))]`

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
Freyskeyd: I updated the argument name

crates/agglayer-aggregator-notifier/src/packer/mod.rs
Freyskeyd: This macro is already guarded by features, no need to add one more.
You can read more about it [here](https://docs.rs/fail/latest/fail/)

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -344,10 +346,14 @@ where
...
             .l1_rpc
-            .get_prev_pessimistic_root(network_id.to_u32())
+            .get_prev_pessimistic_root(
+                network_id.to_u32(),
+                certificate_tx_hash.map(|digest| digest.0.into()),
+            )
             .await
             .map_err(|_| CertificationError::LastPessimisticRootNotFound(network_id))?;
 
+        debug!("Prev PP root from L1: {}", Digest(prev_pessimistic_root));
```
hadjiszs: nit: can be useful to log whether we fetched from specific or latest block, ideally at least just the Option `certificate_tx_hash` with the certificate id

atanmarko (approved):

PR #1082, opened 2025-10-03
fix: make dry-run sendTx not broadcasting tx

Freyskeyd:


crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -248,35 +251,37 @@ where
     /// constructs a [`FunctionCall`] that can be used to create a dry-run
     /// or send a transaction.
```
atanmarko, line 254: Adjust the outdated comment.

atanmarko (dismissed):

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -248,35 +251,37 @@ where
...
+            sequencer_address.into(),
+            signed_tx
+                .tx
+                .zkp
+                .proof
+                .to_fixed_bytes()
+                .map(|value| value.into()),
+        );
+
+        T::into_result(builder).await
```
iljakuklic: Hmm I am not sure introducing the trait for this is worth it. Is there a problem just returning the builder from this function and follow up with a call to `call` or `send` wherever it's used?

iljakuklic (dismissed):

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
Freyskeyd: It wasn't really working with SolCallBuilder due to lifetime issues, but if you wan to give it a shot @iljakuklic feel free :)

iljakuklic: Here is one possibility. The lifetimes are a bit tricky indeed so it still looks somewhat weird. Let me know what you think.
* #1084

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
iljakuklic: I think the comment is no longer outdated with my changes?

iljakuklic (approved):

atanmarko (approved):

Issue #1083, opened 2025-10-03
L1 RPC losing a settlement tx blocks future progress


Ekleog-Polygon:
If L1 RPC loses a settlement transaction, for now we're blocking the chain until manual intervention happens to resubmit the settlement tx.

We should instead automatically resubmit the same settlement tx with the same nonce. This can never risk running the tx twice, and will also allow us to recover from such cases.

This being said, further consideration must be given to cross-network interactions. For example, if:
1. We submit settlement tx A for network 1
2. We submit settlement tx B for network 2
3. RPC loses settlement tx A

Then we can't resend settlement tx A with the same nonce, because B did land and thus imposed nonce requirements.

TBH I'm not sure to which point we can automate recovering from these errors, but probably at least "try resending with same nonce" would catch a good part of the issues.

PR #1087, opened 2025-10-03
feat: allow removing the settlement tx hash from admin api

Ekleog-Polygon:
Manual work around for https://github.com/agglayer/agglayer/issues/1083

### DIFF
--- a/crates/agglayer-storage/src/stores/state/mod.rs
+++ b/crates/agglayer-storage/src/stores/state/mod.rs
@@ -113,6 +113,52 @@ impl StateWriter for StateStore {
         Ok(())
     }
 
+    fn remove_settlement_tx_hash(
+        &self,
+        certificate_id: &CertificateId,
+    ) -> Result<(), Error> {
+        // TODO: make lockguard for certificate_id
+        let certificate_header = self.db.get::<CertificateHeaderColumn>(certificate_id)?;
+
+        if let Some(mut certificate_header) = certificate_header {
+            if certificate_header.settlement_tx_hash.is_none() {
+                return Err(Error::UnprocessedAction(
+                    "Tried to remove settlement tx hash for a certificate that does not have a \
+                     settlement tx hash"
+                        .to_string(),
+                ));
+            }
+
+            if certificate_header.status == CertificateStatus::Settled {
+                return Err(Error::UnprocessedAction(
+                    "Tried to remove settlement tx hash for a certificate that is already settled"
+                        .to_string(),
+                ));
+            }
+
+            certificate_header.settlement_tx_hash = None;
+            certificate_header.status = CertificateStatus::Proven;
+
+            self.db
+                .put::<CertificateHeaderColumn>(certificate_id, &certificate_header)?;
+
+            if let Err(error) = self.backup_client.backup(BackupRequest { epoch_db: None }) {
+                warn!(
+                    hash = certificate_id.to_string(),
+                    "Unable to trigger backup for the state database: {}", error
+                );
+            }
+        } else {
+            info!(
+                hash = %certificate_id,
+                "Certificate header not found for certificate_id: {}",
+                certificate_id
+            )
+        }
+
+        Ok(())
+    }
+
     fn assign_certificate_to_epoch(
         &self,
         certificate_id: &CertificateId,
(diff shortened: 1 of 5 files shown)


crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -113,6 +113,52 @@ impl StateWriter for StateStore {
...
+
+            if certificate_header.status == CertificateStatus::Settled {
+                return Err(Error::UnprocessedAction(
+                    "Tried to remove settlement tx hash for a certificate that is already settled"
+                        .to_string(),
+                ));
+            }
+
+            certificate_header.settlement_tx_hash = None;
+            certificate_header.status = CertificateStatus::Proven;
```
iljakuklic, line 140: Seems like there is a fair number of places that need to carefully keep status and other fields (such as `settlement_tx_hash`) in sync. Filed issue #1088 for this.

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
```diff
@@ -327,3 +327,62 @@ async fn pending_certificate_in_error_force_set_status() {
     assert!(res.settlement_tx_hash.is_some());
     assert_eq!(res.status, CertificateStatus::Candidate);
 }
+
+#[test_log::test(tokio::test)]
+async fn pending_certificate_in_error_with_settlement_tx_hash_force_set_status() {
```
iljakuklic, line 332: Also a test for attempting remove tx hash from settled certificate would be nice

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -113,6 +113,52 @@ impl StateWriter for StateStore {
         Ok(())
     }
 
+    fn remove_settlement_tx_hash(
```
iljakuklic, line 116: This also reverts the status to `Proven`. Maybe the name should somehow reflect it?

iljakuklic (approved):
Looks good. Some nits:

crates/agglayer-storage/src/stores/state/mod.rs
Ekleog-Polygon: FWIW it's something that is (or should be) in the storage requirements issue I sent on slack a bit ago. Can you link it there maybe? 🙏

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
Ekleog-Polygon: Do you mean, to confirm it fails as expected?

crates/agglayer-storage/src/stores/state/mod.rs
Ekleog-Polygon: I actually hesitated on whether or not to reset to Proven and how to name the function. TBH I quite dislike the current way we have of having so many state invariants that aren't enforced by encoding.

Thinking more about it, maybe I should really just not touch the cert status there? The more I think about the more I'm thinking this function should really just remove the settlement tx hash.

WDYT?

crates/agglayer-storage/src/stores/state/mod.rs
Ekleog-Polygon: Just did it :)

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
iljakuklic: > Do you mean, to confirm it fails as expected?

Yep.

atanmarko (approved):

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
Ekleog-Polygon: Going to do this in a follow-up PR

crates/agglayer-storage/src/stores/state/mod.rs
Ekleog-Polygon: Going to do this in a follow-up PR

iljakuklic: > Thinking more about it, maybe I should really just not touch the cert status there? The more I think about the more I'm thinking this function should really just remove the settlement tx hash.
> 
> WDYT?

I'm not quite sure what all the implicit invariants are. This would probably introduce a new state where certificate is a `Candidate` but settlement tx hash is `None`, which was not possible (or intended to be possible) before. I am not sure how to feel about this, hopefully will get cleaned up during storage redesign.

PR #1089, opened 2025-10-03
fix: handle review comments from remove_settlement_tx_hash addition

Ekleog-Polygon:


crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -113,6 +113,51 @@ impl StateWriter for StateStore {
...
+            }
+
+            if certificate_header.status == CertificateStatus::Settled {
+                return Err(Error::UnprocessedAction(
+                    "Tried to remove settlement tx hash for a certificate that is already settled"
+                        .to_string(),
+                ));
+            }
+
+            certificate_header.settlement_tx_hash = None;
```
iljakuklic, line 140: Adding a conversation about this from the [previous PR](https://github.com/agglayer/agglayer/pull/1087). The context is whether the status should also be reverted to `Pending` here or not.

> > @Ekleog-Polygon:
> > Thinking more about it, maybe I should really just not touch the cert status there? The more I think about the more I'm thinking this function should really just remove the settlement tx hash.
> > WDYT?
> 
> @iljakuklic:
> I'm not quite sure what all the implicit invariants are. This would probably introduce a new state where certificate is a `Candidate` but settlement tx hash is `None`, which AFAIK was not possible (or intended to be possible) before. I am not sure how to feel about this, hopefully will get cleaned up during storage redesign.

In addition to that, it is now confirmed that the version with the status reversal appears to work.

Just to be clear I don't have a definite conclusion.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -113,6 +113,51 @@ impl StateWriter for StateStore {
...
+
+            self.db
+                .put::<CertificateHeaderColumn>(certificate_id, &certificate_header)?;
+
+            if let Err(error) = self.backup_client.backup(BackupRequest { epoch_db: None }) {
```
iljakuklic, line 145: Hmm. Should the backup be done *before* the modification?

crates/agglayer-storage/src/stores/state/mod.rs
Ekleog-Polygon: I don't either, but can say that status reversal or not here should have no impact on the fix — we're reversing the status either way just after calling the settlement tx hash removal, so right now the only user of this function doesn't depend on this specific behavior

crates/agglayer-storage/src/stores/state/mod.rs
Ekleog-Polygon: This is a copy-paste of the other functions above; I'd guess we're always backing up after each changes right now :sweat_smile: 

In a way it makes sense to me too: if we take one backup after each change, then our backup policy can identify more easily whether enough time has passed to make a new backup or not, rather than having to wait for the next change to know whether we should backup the last change.

atanmarko: We also do a backup after change in `update_settlement_tx_hash`, so I guess not?

atanmarko (approved):
LGTM

iljakuklic (approved):

atanmarko (approved):

PR #1091, opened 2025-10-06
feat: encrypt RPC connections with TLS

iljakuklic:
These are backward compatible changes, the node behaviour is not changed by default. The cleartext ports are still open. There is an option to open ports with the encrypted version provided the server certificate and key are set up in the config file.

CONFIG-CHANGE:
  * Added `tls.certificate` and `tls.key` to specify paths to server TLS certificate files.
  * Added `rpc.admin-tls-port`, `rpc.readrpc-tls-port`, and `rpc.grpc-tls-port` to specify encrypted RPC ports.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -27,7 +27,7 @@ jobs:
       checks: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       - uses: rustsec/audit-check@v2.0.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,7 +18,7 @@ jobs:
     name: Unit & Linter coverage
     runs-on: ubuntu-latest-16-cores
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
 
--- a/.github/workflows/cycle-tracker.yml
+++ b/.github/workflows/cycle-tracker.yml
@@ -23,7 +23,7 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -21,7 +21,7 @@ jobs:
     if: github.event.action != 'closed'
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -82,7 +82,7 @@ jobs:
       issues: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Download documentation artifact
         uses: actions/download-artifact@v5
@@ -149,7 +149,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -36,7 +36,7 @@ jobs:
       tags: ${{ steps.meta.outputs.tags }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
--- a/.github/workflows/mq-checking-commit.yml
+++ b/.github/workflows/mq-checking-commit.yml
@@ -17,7 +17,7 @@ jobs:
     if: ${{ github.event_name == 'merge_group' }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - uses: wagoid/commitlint-github-action@v6
--- a/.github/workflows/proto.yml
+++ b/.github/workflows/proto.yml
@@ -19,7 +19,7 @@ jobs:
   buf:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
           ref: ${{ github.head_ref || github.event.pull_request.head.ref || github.ref }}
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@nightly
@@ -40,7 +40,7 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -66,7 +66,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@nightly
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
           echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
 
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Docker meta
         id: meta
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,14 +26,14 @@ jobs:
     runs-on: ubuntu-latest-16-cores
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       - uses: ./.github/actions/unit-test
 
   isolated-feature-checks:
     name: "Isolated feature checks"
     runs-on: ubuntu-latest-16-cores
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
         with:
@@ -49,7 +49,7 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Load image
         run: |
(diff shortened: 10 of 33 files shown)


atanmarko:
@iljakuklic Fix the conflicts.

crates/agglayer-node/src/node.rs
```diff
@@ -326,6 +353,49 @@ impl Node {
         Ok(node)
     }
 
+    fn serve_rpc(
+        name: &str,
+        handles: &mut Vec<axum_server::Handle>,
+        addr_plain: SocketAddr,
+        addr_tls: SocketAddr,
```
atanmarko, line 373: Should both of this endpoints be optional, and handle plain, tls or both endpoints depending of the passed arguments.

tests/integrations/src/agglayer_setup.rs
```diff
@@ -123,26 +136,49 @@ pub async fn start_agglayer(
...
     let client = loop {
         if max_attempts == 0 {
-            panic!("Failed to connect to the server");
+            panic!("Failed to connect to the node RPC server {url}");
         }
         interval.tick().await;
-        if let Ok(client) = WsClientBuilder::default().build(&url).await {
+
+        let client_res = WsClientBuilder::default()
+            .with_custom_cert_store({
```
atanmarko, line 209: Could we return both or one of clients (plain and tls) depending on the configuration from `start_agglayer` test function?

crates/agglayer-node/src/node.rs
iljakuklic: The tls part is already optional, will be turned off if server certificates are not configured. The aim of this PR is to integrate TLS support into the node, and make it testable. By default, the node should behave as before.

Various other potentially breaking changes are planned as a follow-up. That includes:

* Changes to the config file format
* Changing the defaults (e.g. plaintext only listen on localhost)
* Configuration granularity (should the certificates be specified on per-endpoint basis or should they be shared)
* Determining whether TLS should be mandatory for some services. Will quite possibly be required for services that also require authentication.
* etc.

crates/agglayer-node/src/lib.rs
```diff
@@ -69,6 +64,16 @@ pub fn main(
 
     info!("Starting agglayer node version info: {}", version);
 
+    // Initialize rustls crypto provider.
+    crypto::CryptoProvider::install_default(crypto::aws_lc_rs::default_provider())
```
Ekleog-Polygon, line 70: Not sure I understand, AFAIU this is already the default provider in rustls?

crates/agglayer-node/src/node.rs
```diff
@@ -300,21 +300,42 @@ impl Node {
             .merge(health_router)
             .merge(json_rpc_router);
 
-        let readrpc_listener = tokio::net::TcpListener::bind(config.readrpc_addr()).await?;
-        let public_grpc_listener = tokio::net::TcpListener::bind(config.public_grpc_addr()).await?;
-        let admin_listener = tokio::net::TcpListener::bind(config.admin_rpc_addr()).await?;
-        info!(on = %config.readrpc_addr(), "ReadRPC listening");
-        info!(on = %config.public_grpc_addr(), "Public gRPC listening");
         info!(on = %config.admin_rpc_addr(), "AdminRPC listening");
```
Ekleog-Polygon: Why leave just this info log? AFAIU we actually start listening only down below?

crates/agglayer-node/src/node.rs
```diff
@@ -324,6 +345,12 @@ impl Node {
                 _ = cancellation_token.cancelled() => {
                     debug!("Node RPC shutdown requested.");
                 }
+            };
+
```
Ekleog-Polygon, line 351: Nit: log the shutdown, eg. if one of the server tasks completes we wouldn't see anything otherwise

crates/agglayer-node/src/node.rs
```diff
@@ -335,6 +362,49 @@ impl Node {
...
+                Some(tls_server_fut) => tls_server_fut.await,
+                None => std::future::pending().await,
+            }
+        };
+
+        async move {
+            tokio::select! {
+                _ = plain_server_fut => {},
+                _ = tls_server_fut => {},
+            }
```
Ekleog-Polygon, line 408: Nit: I think you could use an `if` precondition to disable the branch without the `std::future::pending` hack? Maybe you'd also need `OptionFuture` to avoid too much boilerplate then.

tests/dev-tls-certs/build.rs
```diff
@@ -0,0 +1,57 @@
+use std::{env, fs, path::Path};
+
+use rcgen::{CertificateParams, ExtendedKeyUsagePurpose as EKU, KeyUsagePurpose as KU, SanType};
+
+fn main() {
+    // Step 1: Generate CA certificate.
+    let ca_keys = rcgen::KeyPair::generate().expect("CA key generation failed");
```
Ekleog-Polygon, line 7: IMO we could have used a hardcoded snake oil certificate here and it'be fine; but now that you've written the code let's just use it :smile:

tests/dev-tls-certs/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+//! Gives access to pre-generated TLS certificates for testing.
+
+/// Path to the CA certificate.
+pub const CA_CERT_PATH: &str = concat!(env!("OUT_DIR"), "/ca.cert.pem");
+
+/// Path to the server certificate.
+pub const SERVER_CERT_PATH: &str = concat!(env!("OUT_DIR"), "/server.cert.pem");
+
+/// Path to the server private key.
+pub const SERVER_KEY_PATH: &str = concat!(env!("OUT_DIR"), "/server.key.pem");
```
Ekleog-Polygon, line 10: Nit: this is technically wrong, because the path is on the dev machine, which will break eg. when if we ever run tests on a different machine than we're building them (which could happen for example in CI).

This being said, it should probably be fine :shrug: Best would be to `include_bytes` here but then we can't just provide it as path, and we'd need a new function to write the bytes to a `tempfile` and return the relevant `PathBuf`.

Feel free to ignore this nit :smile:

tests/integrations/Cargo.toml
```diff
@@ -26,9 +26,12 @@ pin-project.workspace = true
 rand.workspace = true
 reqwest = "0.12.23"
 rstest.workspace = true
+rustls.workspace = true
+rustls-pemfile = "2.2"
```
Ekleog-Polygon: Nit: can we move this dep to the workspace?

tests/integrations/Cargo.toml
```diff
@@ -53,5 +56,4 @@ agglayer-types.workspace = true
 agglayer-prover-config.workspace = true
 pessimistic-proof-test-suite.workspace = true
 pessimistic-proof.workspace = true
-
-test-log.workspace = true
+dev-tls-certs = { path = "../dev-tls-certs" }
```
Ekleog-Polygon: And this one too

tests/integrations/src/agglayer_setup.rs
```diff
@@ -102,12 +103,24 @@ pub async fn start_agglayer(
         }],
     });
 
+    config.tls = Some(agglayer_config::TlsConfig {
+        certificate: dev_tls_certs::SERVER_CERT_PATH.into(),
+        key: dev_tls_certs::SERVER_KEY_PATH.into(),
+    });
+
+    assert!(config.tls.is_some());
```
Ekleog-Polygon: AFAICT this is useless, as we set it just above. A comment explaining why might do a better job at saying "don't remove this from tests"

tests/integrations/src/l1_setup.rs
```diff
@@ -22,7 +22,7 @@ impl L1Docker {
                 &format!("{ws_port}:8546"),
                 "-d",
                 "--name",
-                &name,
+                &format!("{name}_10"),
```
Ekleog-Polygon: Hmm… why is this necessary? (FWIW I've hit issues with integration tests that required me to manually remove the previous container before retrying; if this is the reason then removing the container before running might be a better choice)

tests/dev-tls-certs/build.rs
iljakuklic: > IMO we could have used a hardcoded snake oil certificate here and it'be fine; but now that you've written the code let's just use it 😄

Tried that initially but was getting annoying emails from github about certificates in the repo.

crates/agglayer-node/src/node.rs
iljakuklic: > Nit: I think you could use an `if` precondition to disable the branch without the `std::future::pending` hack? Maybe you'd also need `OptionFuture` to avoid too much boilerplate then.

I am not quite sure how would that work. Observe that both futures need to be in the select arm, so the other one is not cancelled when one makes progress. I could still disable the branch if TLS is not configured but I don't see much of a point in doing that.

crates/agglayer-node/src/lib.rs
iljakuklic: > Not sure I understand, AFAIU this is already the default provider in rustls?

The crypto operations panic without this initialization line.

tests/integrations/src/agglayer_setup.rs
iljakuklic: > Could we return both or one of clients (plain and tls) depending on the configuration from `start_agglayer` test function?

Added the ability to switch between plaintext / tls to the testing environment setup.

tests/integrations/src/agglayer_setup.rs
```diff
@@ -53,11 +53,54 @@ pub async fn start_l1() -> L1Docker {
...
+    ) -> (oneshot::Receiver<()>, WsClient, CancellationToken) {
+        start_agglayer(config_path, l1, self.config, self.token, self.use_tls).await
+    }
+
+    pub async fn setup_network(
+        self,
+        tmp_dir: &Path,
+    ) -> (oneshot::Receiver<()>, L1Docker, WsClient) {
+        let l1 = start_l1().await;
+        let (receiver, client, _token) = self.start_agglayer(tmp_dir, &l1).await;
```
Ekleog-Polygon, line 92: Nit: this feels weird to me, intuitively a "setup" function should set up without starting, but a "start" function could both setup and start.

So I'd argue for renaming "setup_network" into "start_network" here. WDYT?

(Though I see it was already named this way before, so this is definitely not a blocker)

crates/agglayer-node/src/node.rs
Ekleog-Polygon: :+1: Probably not worth investigating further then :smile:

crates/agglayer-config/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition.workspace = true
 [dependencies]
 dirs.workspace = true
 alloy-primitives.workspace = true
+axum-server.workspace = true
```
Freyskeyd, line 9: why is it necessary?

crates/agglayer-node/src/node.rs
```diff
@@ -328,6 +347,14 @@ impl Node {
...
                     debug!("Node RPC shutdown requested.");
                 }
+            };
+
+            info!("Shutting down RPC services");
+
+            // Signal to all RPCs to shut down gracefully.
+            for handle in rpc_service_handles {
+                // Note: shutdown timeout is already applied on runtime level.
+                handle.graceful_shutdown(None);
```
Freyskeyd, line 357: are you sure that it will properly handle a graceful shutdown that way?

Cargo.toml
```diff
@@ -46,6 +50,7 @@ agglayer-test-suite = { path = "crates/agglayer-test-suite" }
 pessimistic-proof = { path = "crates/pessimistic-proof" }
 pessimistic-proof-core = { path = "crates/pessimistic-proof-core" }
 pessimistic-proof-test-suite = { path = "crates/pessimistic-proof-test-suite" }
+dev-tls-certs = { path = "tests/dev-tls-certs" }
```
Freyskeyd, line 53: Maybe in alphabetic order can be better?

tests/integrations/src/agglayer_setup.rs
atanmarko: I agree, `start_network` makes more sense.

crates/agglayer-config/Cargo.toml
iljakuklic: Unfortunately, I did not find a way for normal `axum` to do TLS.

Issue #825, opened 2025-05-28
[Design] Health & readiness checks low-hanging fruit


Ekleog-Polygon:
Following a chat with @tnobayashi, it appeared that we could easily improve the benefits of health and readiness checks by serving health and readiness endpoints on a separate port instead of the same port as ReadRPC. This would allow us to listen on this port at the very startup of the node, without waiting for all the initialization work to be done. In turn, this would allow our infra to actually make use of our health endpoint, as currently the health endpoint is disabled there due to slow startup.

In addition to that, I'm thinking that we should have a single tokio runtime that'd run everything, or at least our RPCs and health endpoints. This way, if a tokio runtime gets starved, then the health endpoint will be starved too and it'll automatically mark the service as unhealthy. I'd even argue for metrics to be in the same runtime, if only to reduce the number of threads and simplify our code. In particular, it'd avoid having to think of on which runtime do things run, and in particular whether they'll automatically affect the health endpoint or not.

Future work would be doable by having the health endpoint automatically check our prometheus metrics, ping our various services or similar; but all that would not be as low-hanging fruit as what I'm listing here.

Freyskeyd:
Defining this as `Design`

PR #1097, opened 2025-10-07
docs: add design document for status health checks

polykeith:
Part of #825.

Adds a design document to detail what kinds of data will be collected in order to determine whether the node status is healthy or not.

### DIFF
--- a/docs/status_checks.md
+++ b/docs/status_checks.md
@@ -3,21 +3,48 @@
 It is often useful to know about the state of the node, especially after it has been started.
 However, information about the state of the node is quite opaque aside from telemetry.
 
-This document lists the kinds of data that we would like to collect and display as a status health check of the node.
+This document lists the signals that we would like to collect and display as a status health check of the node:
 
-* Age of the certificate
-* Frequency and duration of RPC/smart contract call failures with L1
-  - If L1 takes more than 3 minutes of RPC to respond, something is wrong
-* Any failure in submitting a certificate and settling it on L1
-* Any storage call returning an error
-  - If too many errors happen, maybe restarting the node is a good idea and would fix
-* Clock not making progress
-  - If the epoch has not moved in twice of the expected amount of time, something is wrong
-* Multiple errors in trying to retrieve the signing key from GCP
-* No new RPC calls to the node within an hour (an epoch)
-  - Most likely a problem as we do expect traffic to settle funds across blockchains
-* Certificate status invariant violation (e.g. settled certs appearing in PendingStore, pending certs appearing in EpochStore, etc.)
-  - Most probably cannot perform this check cheaply, but nice to have
-* Rate limiter not working
-  - Would cause us to spend too much tx fees
-* Taking too long to perform a read or write (something like 10 seconds) to RocksDB
+1. Age of the certificate
+2. Frequency and duration of RPC/smart contract call failures with L1
+3. Any failure in submitting a certificate and settling it on L1
+4. Any storage call returning an error
+5. Clock not making progress
+6. Multiple errors in trying to retrieve the signing key from GCP
+7. No new RPC calls to the node within an hour (an epoch)
+8. Certificate status invariant violation (e.g. settled certs appearing in PendingStore, pending certs appearing in EpochStore, etc.)
+9. Rate limiter not working
+10. Taking too long to perform a read or write to RocksDB
+
+## Heuristics in determining healthiness
+
+Each signal should also be paired with a criteria in determining how it can be considered healthy or not.
+
+1. A certificate of an age > 1 day is problematic.
+2. Something is wrong if L1 takes more than 3 minutes of RPC to respond.
+3. Certificate submission and settlement should never emit any errors.
+4. Storage call returning errors is unusual but may be fixed by a node restart.
+5. If epoch has not progressed in twice the expected amount of time (i.e. more than 2 hours if each epoch is about an hour), it hints at an unhealthy status.
+6. Signing key retrieval from GCP emitting an error once or twice is tolerated, but more than that hints at a latent issue.
+7. Not having new RPC calls to the node within an epoch is highly unusual and should be flagged.
+8. Any certificate status invariant violation should be reported and is highly unusual.
+9. Rate limiter throwing errors and not working as expected would cause us to spend a lot on tx fees on L1.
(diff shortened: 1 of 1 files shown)


docs/status_checks.md
```diff
@@ -0,0 +1,50 @@
...
+9. Rate limiter not working
+10. Taking too long to perform a read or write to RocksDB
+
+## Heuristics in determining healthiness
+
+Each signal should also be paired with a criteria in determining how it can be considered healthy or not.
+
+1. A certificate of an age > 1 day is problematic.
+2. Something is wrong if L1 takes more than 3 minutes of RPC to respond.
+3. Certificate submission and settlement should never emit any errors.
```
iljakuklic, line 25: Do you mean the settlement tx is deemed invalid by L1? Might be good to clarify, as I don't think this point is meant to also cover RPC issues which are addressed in point (2).

docs/status_checks.md
```diff
@@ -0,0 +1,50 @@
...
+4. Storage call returning errors is unusual but may be fixed by a node restart.
```
iljakuklic, line 26: Just to point out, we have some storage errors that are pretty serious, e.g. failing to decode stored data, which may happen after an upgrade if the decoding is not handles in a backwards compatible manner. These errors do not tear down the node but indicate a bug in the software and would not be fixed by a node restart.

Another class of storage issues has to do with IO stuff. Some would not be fixed by a restart either, e.g. full disk.

docs/status_checks.md
```diff
@@ -0,0 +1,50 @@
...
+5. If epoch has not progressed in twice the expected amount of time (i.e. more than 2 hours if each epoch is about an hour), it hints at an unhealthy status.
+6. Signing key retrieval from GCP emitting an error once or twice is tolerated, but more than that hints at a latent issue.
+7. Not having new RPC calls to the node within an epoch is highly unusual and should be flagged.
```
iljakuklic, line 29: Would this be too noisy in certain testnet settings, where things might get quiet if testing is not actively going on?

docs/status_checks.md
atanmarko: We may emit some errors and then recover - so this should probably include more detailed analysis for particular scenarios.

docs/status_checks.md
atanmarko: Yeah, would probably update this point.

docs/status_checks.md
polykeith: Right, settlement tx should simply Just Work, barring RPC and network issues. Any other errors that may arise from this situation is something that certainly requires some sort of investigation to see what's wrong.

docs/status_checks.md
polykeith: In that case, it sounds like decoding errors are serious enough to warrant an elevated warning level, possibly to critical depending on what it can't decode as it can hamper the normal operation of the node.

docs/status_checks.md
polykeith: As I recall with my discussion with @Ekleog-Polygon, Katana will always submit a tx at least once to the agglayer node, however I'm not sure if this is the case for testnet. Is this true?

docs/status_checks.md
```diff
@@ -0,0 +1,50 @@
+# Status, health and readiness checks
+
+It is often useful to know about the state of the node, especially after it has been started.
+However, information about the state of the node is quite opaque aside from telemetry.
+
+This document lists the signals that we would like to collect and display as a status health check of the node:
+
+1. Age of the certificate
```
Ekleog-Polygon, line 8:
```suggestion
1. Age of the latest certificates in each stage
```

docs/status_checks.md
```diff
@@ -0,0 +1,50 @@
...
+7. No new RPC calls to the node within an hour (an epoch)
+8. Certificate status invariant violation (e.g. settled certs appearing in PendingStore, pending certs appearing in EpochStore, etc.)
+9. Rate limiter not working
+10. Taking too long to perform a read or write to RocksDB
+
+## Heuristics in determining healthiness
+
+Each signal should also be paired with a criteria in determining how it can be considered healthy or not.
+
+1. A certificate of an age > 1 day is problematic.
```
Ekleog-Polygon, line 23:
```suggestion
1. A non-settled certificate of an age > 1 day is problematic.
```

docs/status_checks.md
Ekleog-Polygon: Right, ISTR that Katana in particular (and more generally FEPv2 networks) should always try to submit at least one tx per epoch. So maybe having no RPC over 2 consecutive epochs should be considered bad? One epoch might just be bad luck, if the RPC calls came just before/after said epoch.

docs/status_checks.md
```diff
@@ -0,0 +1,50 @@
...
+7. Not having new RPC calls to the node within an epoch is highly unusual and should be flagged.
+8. Any certificate status invariant violation should be reported and is highly unusual.
+9. Rate limiter throwing errors and not working as expected would cause us to spend a lot on tx fees on L1.
+10. Taking around 10 seconds or more to write to RocksDB is unusual.
+
+## API
+
+In addition, there are a few API setups in which we communicate unhealthiness back to the client who requests for this information:
+
+1. Have multiple endpoints that is specialized for a particular service, and if the signals for that service is unhealthy, we return the corresponding HTTP status code (503 Service Unavailable or 500 Internal server error). Healthy services would simply just return a 200 OK. We may return a JSON as well if desired.
```
Ekleog-Polygon, line 38: Does this suggest a middleware that'd intercept incoming requests and reply 503/500 instead of passing it through to the actual service? If so I'm not sure that's a good idea; we should still try performing what people ask of us to the best of our ability, as the whole system is designed to retry automatically if something doesn't make progress (starting with aggsender's auto-resending of transactions)

docs/status_checks.md
```diff
@@ -0,0 +1,50 @@
...
+
+2. One dedicated endpoint containing a JSON that contains fields corresponding to the service's status and the signals/telemtry used to determine the healthiness of the checkpoint.
```
Ekleog-Polygon, line 40: We'll also want some heuristic to return the true/false reply expected by infra, so that they can auto-restart the node if the health is bad for X consecutive minutes.

Issue #1092, opened 2025-10-07
Recover from the very long settlement tx inclusion times


atanmarko:
Count with the possible situation that the waiting for the pending transaction would fail, but eventually (e.g. after 1 hour) could get included in the l1 block.

Related https://github.com/agglayer/agglayer/issues/1083

atanmarko:
Resolved in #1099

PR #1099, opened 2025-10-07
fix: pending tx handling

atanmarko:
Fixes: https://github.com/agglayer/agglayer/issues/1083 and https://github.com/agglayer/agglayer/issues/1092

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
+++ b/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
@@ -1,8 +1,6 @@
 use std::{sync::Arc, time::Duration};
 
-use agglayer_certificate_orchestrator::{
-    Error, Error::PendingTransactionTimeout, SettlementClient,
-};
+use agglayer_certificate_orchestrator::{Error, SettlementClient};
 use agglayer_config::outbound::OutboundRpcSettleConfig;
 use agglayer_contracts::{rollup::VerifierType, L1TransactionFetcher, RollupContract, Settler};
 use agglayer_storage::stores::{
@@ -347,7 +345,7 @@ where
                         ?timeout,
                         "Timeout while watching the pending settlement transaction"
                     );
-                    PendingTransactionTimeout {
+                    Error::PendingTransactionTimeout {
                         certificate_id,
                         error: format!(
                             "Timeout while watching the pending settlement transaction {:?}, \
@@ -403,7 +401,7 @@ where
                         ?timeout,
                         "Timeout while waiting for the pending settlement transaction"
                     );
-                    Err(PendingTransactionTimeout {
+                    Err(Error::PendingTransactionTimeout {
                         certificate_id,
                         error: format!(
                             "Settlement pending transaction timeout after {:?}, error: {}",
@@ -581,7 +579,7 @@ where
                 // The inner field derefs to the transaction type which implements the
                 // Transaction trait.
                 use alloy::consensus::Transaction as _;
-                (*tx.inner).nonce()
+                tx.inner.nonce()
             }
             None => {
                 warn!("Settlement tx not found on L1 for tx: {}", tx_hash);
--- a/crates/agglayer-certificate-orchestrator/src/network_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/network_task.rs
@@ -415,23 +415,23 @@ where
                         // If the error in the sending transaction happened for whatever reason,
                         // check if maybe the certificate has been settled through some other previous transaction.
                         if let Err(_err) = &result {
-                            for previos_tx_hash in previous_tx_hashes {
-                                match self.settlement_client.fetch_settlement_receipt_status(previos_tx_hash).await {
+                            for previous_tx_hash in previous_tx_hashes {
+                                match self.settlement_client.fetch_settlement_receipt_status(previous_tx_hash).await {
                                     Ok(true) => {
                                         // Transaction is mined, but we haven't known that, return it for further processing.
                                         info!(%certificate_id,
-                                            "Certificate for new height: {height} has been settled on L1 through previous transaction {previos_tx_hash}");
-                                        result = Ok((previos_tx_hash, None));
+                                            "Certificate for new height: {height} has been settled on L1 through previous transaction {previous_tx_hash}");
+                                        result = Ok((previous_tx_hash, None));
                                         break;
                                     }
                                     Ok(false) => {
                                         // Transaction is not mined yet, will retry later.
                                         debug!(%certificate_id,
-                                            "Certificate for new height: {height} is still pending settlement on L1 through previous transaction {previos_tx_hash}");
+                                            "Certificate for new height: {height} is still pending settlement on L1 through previous transaction {previous_tx_hash}");
                                     }
                                     Err(err) => {
                                         debug!(%certificate_id,
-                                            "Error checking receipt status for previous settlement tx {previos_tx_hash}: {err}");
+                                            "Error checking receipt status for previous settlement tx {previous_tx_hash}: {err}");
                                     }
                                 }
                             }
@@ -575,26 +575,6 @@ where
                 nullifier_root: state_commitment.nullifier_root.into(),
             };
         pp_commitment_values.compute_pp_root(version)
-    }
-
-    #[allow(dead_code)]
-    fn is_pending_pessimistic_root(
-        &self,
-        settled_pp_root: Digest,
-        height: u64,
-        commitment: StateCommitment,
-    ) -> bool {
-        let computed_v2 = self.pending_pessimistic_root(
-            height.into(),
-            PessimisticRootCommitmentVersion::V2,
-            &commitment,
-        );
-        let computed_v3 = self.pending_pessimistic_root(
-            height.into(),
-            PessimisticRootCommitmentVersion::V3,
-            &commitment,
-        );
-        settled_pp_root == computed_v2 || settled_pp_root == computed_v3
     }
 
     /// Fetches the latest pessimistic root and transaction hash if both are


crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -64,6 +66,7 @@ where
     async fn submit_certificate_settlement(
         &self,
         certificate_id: CertificateId,
+        nonce: Option<u64>,
```
Freyskeyd, line 67: Couldn't we refactor the `verify_pessimistic_trusted_aggregator` to return the TransactionRequest and thus dealing with configuration outside ? So that we don't propagate this nonce everywhere

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -372,29 +385,118 @@ where
...
+                                }
+                            }
+                            Err(err) => Err(err),
+                        };
+
+                        // If the error in the sending transaction happened for whatever reason,
+                        // check if maybe the certificate has been settled through some other previous transaction.
+                        if let Err(_err) = &result {
+                            for previos_tx_hash in previous_tx_hashes {
+                                match self.settlement_client.get_settlement_receipt_status(previos_tx_hash).await {
```
Freyskeyd: why requesting two time the same payload (request above), couldn't we have it once?

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -397,6 +459,144 @@ where
...
+        network_id: agglayer_types::NetworkId,
+    ) -> Result<(Option<[u8; 32]>, Option<SettlementTxHash>), Error> {
+        use agglayer_contracts::contracts::PolygonRollupManager::VerifyPessimisticStateTransition;
+        use alloy::{providers::Provider, sol_types::SolEvent};
+
+        // Create a filter for the latest VerifyPessimisticStateTransition event for
+        // this network_id Using from_block Latest ensures we only get recent
+        // events
+        let rollup_address = self.l1_rpc.get_rollup_manager_address();
+        // TODO: Set Latest instead of Earliest after testing
```
hadjiszs, line 480: I see that's still set to earliest

crates/agglayer-certificate-orchestrator/src/certifier.rs
```diff
@@ -21,6 +21,7 @@ pub struct CertifierOutput {
     pub height: Height,
     pub new_state: LocalNetworkStateData,
     pub network: NetworkId,
+    pub pp_root: Digest,
```
hadjiszs:
```suggestion
    pub new_pp_root: Digest,
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -27,6 +33,9 @@ pub struct CertificateTask<StateStore, PendingStore, CertifierClient> {
     pending_store: Arc<PendingStore>,
     certifier_client: Arc<CertifierClient>,
     cancellation_token: CancellationToken,
+    pp_root: Option<Digest>,
```
hadjiszs:
```suggestion
    new_pp_root: Option<Digest>,
```

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
```diff
@@ -31,4 +32,26 @@ pub trait SettlementClient: Unpin + Send + Sync + 'static {
...
+    async fn get_settlement_nonce(
+        &self,
+        settlement_tx_hash: SettlementTxHash,
+    ) -> Result<Option<u64>, Error>;
+
+    /// Returns the receipt status for a settlement tx.
+    async fn get_settlement_receipt_status(
+        &self,
+        settlement_tx_hash: SettlementTxHash,
+    ) -> Result<bool, Error>;
```
iljakuklic, line 56: Nit: the `get_` prefix is [not very idiomatic](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) in Rust.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -397,6 +459,144 @@ where
...
+            .map_err(|e| {
+                Error::L1CommunicationError(agglayer_contracts::L1RpcError::FailedToQueryEvents(
+                    e.to_string(),
+                ))
+            })?;
+
+        // Get the most recent event (last in the list) and extract its new pessimistic
+        // root.
+        let result = events.iter().rev().find_map(|log| {
+            info!("Found VerifyPessimisticStateTransition event: {:?}", log);
```
hadjiszs: do we still want this log and should it be info?

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: @hadjiszs I have opened a ticket for that earlier today https://github.com/agglayer/agglayer/issues/1104

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -298,8 +333,15 @@ where
...
 
+        if self.pp_root.is_none() {
+            return Err(CertificateStatusError::InternalError(
+                "CertificateTask::process_from_candidate called without a pp_root".into(),
+            ));
+        }
+
         let height = self.header.height;
         let certificate_id = self.header.certificate_id;
+        let pp_root = self.pp_root.unwrap();
```
hadjiszs:
```suggestion
        let height = self.header.height;
        let certificate_id = self.header.certificate_id;
        let pp_root = self.pp_root.ok_or(CertificateStatusError::InternalError(
                "CertificateTask::process_from_candidate called without a pp_root".into(),
            )
        })?;
```

short term at least to not have naked unwrap, but then for later makes me wonder if we want an option in self there

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: Definitely not as info

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -264,21 +278,42 @@ where
...
             settlement_submitted_notifier,
         })
         .await?;
 
-        let settlement_tx_hash = settlement_submitted.await.map_err(recv_err)??;
+        let (settlement_tx_hash, nonce) = settlement_submitted.await.map_err(recv_err)??;
+        if self.previous_tx_hashes.contains(&settlement_tx_hash) {
+            warn!("Resubmitted the same settlement transaction hash {settlement_tx_hash}");
+        } else {
+            self.previous_tx_hashes.insert(settlement_tx_hash);
```
iljakuklic: Nit: [`HashSet::insert`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert) already returns a bool indicating whether the new value was inserted, no need to check with `contains` separately.

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -372,29 +385,118 @@ where
...
+                            .await;
+
+                        let result = match result {
+                            Ok((epoch, index)) => {
+                                // Certificate has been settled.
+                                CertificateSettlementResult::Settled(epoch, index)
+                            }
+                            Err(Error::PendingTransactionTimeout { settlement_tx_hash, .. }) => {
+                                match self.settlement_client.fetch_settlement_receipt_status(settlement_tx_hash).await {
+                                    Ok(true) => {
```
iljakuklic, line 463: This match with log reporting could be factored into a `report_settlement_state` or something like that (or maybe a function that does fetch and reports to logs) as to not interrupt the main flow of the logic that much. It is also duplicated where previous txs are checked.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -316,10 +356,44 @@ where
...
+                return Err(error);
+            }
+            CertificateSettlementResult::TimeoutError => {
+                // Retry the settlement transaction
+                info!(
+                    "Retrying the settlement transaction after a timeout for certificate \
+                     {certificate_id}"
+                );
+                self.set_status(CertificateStatus::Proven)?;
+                return Box::pin(self.process_from_proven()).await;
```
iljakuklic, line 378: Hmm why is Box::pin needed here? Would not expect that.

Edit: Ah I see, I guess it has something to do with recursion.

crates/agglayer-certificate-orchestrator/src/network_task.rs
atanmarko: Ok, will leave this for follow up fix.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: Yeas, async calls recursion.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -27,6 +33,9 @@ pub struct CertificateTask<StateStore, PendingStore, CertifierClient> {
     pending_store: Arc<PendingStore>,
     certifier_client: Arc<CertifierClient>,
     cancellation_token: CancellationToken,
+    new_pp_root: Option<Digest>,
+    nonce: Option<u64>,
+    previous_tx_hashes: HashSet<SettlementTxHash>,
```
iljakuklic, line 38: I'm wondering if it would be more appropriate to have these as arguments to the various `process_from_*` functions.

For example, in `process_for_pending`, the nonce is always `None`, so it does not need to have that argument at all. In later stages, it is expected to be known, so the argument could be `Digest` (no `Option`). It would allow for tracking which bits of information are (not-)needed where, with the assistance of the type system. Similar reasoning could be applied to `new_pp_root` and to lesser extent also to `previous_tx_hashes`.

This is out of time budget for this PR, so treat as a suggestion for a follow up.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -329,9 +337,38 @@ where
...
+                if let PendingTransactionError::TxWatcher(alloy::providers::WatchTxError::Timeout) =
+                    error
+                {
+                    error!(
+                        %settlement_tx_hash,
+                        ?error,
+                        ?timeout,
+                        "Timeout while watching the pending settlement transaction"
+                    );
+                    PendingTransactionTimeout {
```
Freyskeyd: Why does the enum variant imported that way? It feels odd, I had to go to the top to see that in fact it is: `Error::PendingTransactionTimeout`

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -397,6 +459,143 @@ where
...
+                Error::L1CommunicationError(
+                    agglayer_contracts::L1RpcError::TransactionReceiptNotFound(e.to_string()),
+                )
+            })? {
+            Some(tx) => {
+                // Extract nonce from the inner transaction envelope.
+                // The inner field derefs to the transaction type which implements the
+                // Transaction trait.
+                use alloy::consensus::Transaction as _;
+                (*tx.inner).nonce()
```
Freyskeyd: You can just `tx.nonce()` here I guess

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Freyskeyd, line 378: How do we prevent this from loop forever?

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -372,29 +385,118 @@ where
...
+                                }
+                            }
+                            Err(err) => Err(err),
+                        };
+
+                        // If the error in the sending transaction happened for whatever reason,
+                        // check if maybe the certificate has been settled through some other previous transaction.
+                        if let Err(_err) = &result {
+                            for previos_tx_hash in previous_tx_hashes {
+                                match self.settlement_client.fetch_settlement_receipt_status(previos_tx_hash).await {
```
Freyskeyd: why requesting two time the same payload (request above), couldn't we have it once?

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -372,29 +385,118 @@ where
...
+                                    }
+                                }
+                            }
+                            Err(err) => Err(err),
+                        };
+
+                        // If the error in the sending transaction happened for whatever reason,
+                        // check if maybe the certificate has been settled through some other previous transaction.
+                        if let Err(_err) = &result {
+                            for previos_tx_hash in previous_tx_hashes {
```
Freyskeyd:
```suggestion
                            for previous_tx_hash in previous_tx_hashes {
```

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -454,4 +558,63 @@ where
...
+
+                ler_leaf_count: state_commitment.ler_leaf_count,
+                balance_root: state_commitment.balance_root.into(),
+                nullifier_root: state_commitment.nullifier_root.into(),
+            };
+        pp_commitment_values.compute_pp_root(version)
+    }
+
+    #[allow(dead_code)]
+    fn is_pending_pessimistic_root(
```
Freyskeyd: If it's dead code, maybe we can remove this?

crates/agglayer-contracts/src/settler.rs
```diff
@@ -104,7 +110,26 @@ where
...
+
+            if let Some(nonce) = nonce {
+                debug!(
+                    "Nonce provided, increasing max_fee_per_gas \
+                     {DEFAULT_GAS_PRICE_INCREASE_MULTIPLIER} times"
+                );
+                // If nonce is provided, increase the max fee by 10% to avoid
+                // transaction getting stuck due to nonce gaps.
+                adjusted.max_fee_per_gas *= DEFAULT_GAS_PRICE_INCREASE_MULTIPLIER;
+                adjusted.max_priority_fee_per_gas *= DEFAULT_GAS_PRICE_INCREASE_MULTIPLIER;
```
Freyskeyd, line 123: Are we sure of this behavior if we have multiple retries?

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
```diff
@@ -349,7 +351,7 @@ async fn pending_certificate_in_error_with_settlement_tx_hash_force_set_status()
     let fake_settlement_tx_hash = SettlementTxHash::from(Digest::from([1; 32]));
     context
         .state_store
-        .update_settlement_tx_hash(&certificate_id, fake_settlement_tx_hash)
+        .update_settlement_tx_hash(&certificate_id, fake_settlement_tx_hash, false)
```
Freyskeyd, line 354: maybe having a dedicated method such as `overwrite_settlement_tx_hash` would have prevented this extract field everywhere ?

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: We can - but I am not sure if it would be better that way.

crates/agglayer-certificate-orchestrator/src/network_task.rs
atanmarko: @Freyskeyd Occasionally there were errors on the call to `submit_certificate_settlement` and it fails immediately. So this is just a fail-safe. 
Settlement nonce is retrieved form the get tx status `eth_getTransactionByHash` and receipt `eth_getTransactionReceipt` is available only when the transaction is mined. I have put both to the same function but I had to separate them.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: We don't - transaction is repeated with the same nonce and increased gas price - eventually it will get mined to the l1. If it is repeated because l1 is not working or some infrastructure issue, we get stuck (but we should anyway not progress in that case).

crates/agglayer-certificate-orchestrator/src/network_task.rs
atanmarko: It is not the same - one is `eth_getTransactionByHash` and other `eth_getTransactionReceipt`. First one is available when you send tx, the second one when the transaction is mined to the block.

crates/agglayer-certificate-orchestrator/src/network_task.rs
atanmarko: Yes, forgot to clean it up.

crates/agglayer-contracts/src/settler.rs
atanmarko: Good question. Here we would recalculate the gas price from the network and double it. If the network gas prices have increased, this should be fine. If they are not, previous transaction should be mined. But we should actually keep the previous gas price and number of retries and increase from the latest.

atanmarko: Here is the ticket to resolve this more robustly https://github.com/agglayer/agglayer/issues/1106

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: Added ticket https://github.com/agglayer/agglayer/issues/1107

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: Added ticket https://github.com/agglayer/agglayer/issues/1107 to review this topic

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
atanmarko: Ticket https://github.com/agglayer/agglayer/issues/1108

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -397,6 +457,143 @@ where
...
+        let result = events.iter().rev().find_map(|log| {
+            let latest_pp_root =
+                VerifyPessimisticStateTransition::decode_log(&log.clone().into()).ok();
+            let tx_hash = log.transaction_hash.map(Digest::from);
+            match (
+                latest_pp_root.map(|val| <[u8; 32]>::from(val.newPessimisticRoot)),
+                tx_hash,
+            ) {
+                (Some(pp_root), Some(tx_hash)) => Some((pp_root, SettlementTxHash::new(tx_hash))),
+                _ => None,
```
iljakuklic, line 513: Would it be useful to add logs to flag up the `(None, Some(_))` and `(Some(_), None)` cases?

atanmarko: I'll do it in the next PR.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Freyskeyd: What if the tx never gets mined?
My question is, do we need a `max_retries` ?

atanmarko: We have number of retries from the previous tx transactions. We can put and end after 5 transactions?

atanmarko: Lets handle this in the next PR

PR #1101, opened 2025-10-08
feat: gas price configuration

iljakuklic:
This adds config keys for multiplying and clamping the default gas price
estimation outcome.

CONFIG-CHANGE: Added the following config keys:
  * `rpc.settle.gas-price.multiplier` for scaling the Alloy estimate.
    It is a decimal number, 1.0 means no scaling.
  * `rpc.settle.gas-price.floor` to set the min overall gas price.
    This is a string number (since serde does not do `u128`) with units,
    e.g. "10gwei", "7.3eth", etc. The unit is required to avoid ambiguity.
  * `rpc.settle.gas-price.ceiling` to set the max overall gas price.
    Same format as floor.

### DIFF
--- a/crates/agglayer-contracts/src/settler.rs
+++ b/crates/agglayer-contracts/src/settler.rs
@@ -40,7 +40,6 @@ where
             // Fall back to hex representation of revert data
             return Some(format!("0x{}", hex::encode(revert_data)));
         }
-
         // For non-revert errors, return the debug representation
         Some(format!("{error:?}"))
     }


crates/agglayer-contracts/src/settler.rs
```diff
@@ -98,6 +101,33 @@ where
             tx_call = tx_call.gas(adjusted_gas);
         }
 
+        {
+            // Apply gas price multiplier and floor/ceiling constraints
+            let estimate = self.rpc.estimate_eip1559_fees().await?;
+            let adjust = |fee: u128| -> u128 {
+                let fee = Decimal::from(fee).saturating_mul(self.gas_price_params.multiplier);
+                fee.to_u128().unwrap_or(u128::MAX)
```
atanmarko: We probably don't want to give millions as fee by default if it unwraps (e.g. negative value in the config so wrongly parsed)?  Maybe to put some meaningful default. You clamp it but default ceiling is also MAX :)

crates/agglayer-config/src/outbound.rs
```diff
@@ -97,6 +143,24 @@ const fn default_settlement_timeout() -> Duration {
...
+    Decimal::ONE
+}
+
+fn same_as_default_gas_price_multiplier(v: &Decimal) -> bool {
+    *v == default_gas_price_multiplier()
+}
+
+/// Default gas price ceiling for the transaction.
+const fn default_gas_price_ceiling() -> u128 {
+    u128::MAX
```
atanmarko: Put some meaningful value (e.g. 5 * max gas price in the last 2 years)

crates/agglayer-contracts/src/settler.rs
iljakuklic: Hmm I am not a fan of operations that are not well behaved. Placing a different default here does nothing if the calculated amount is just below `u128::MAX`. Think of it as saturating mul, followed by saturating conversion.

Putting negative multiplier into config should be rejected at config level. I took a shortcut to use the decimal type for arithmetic but it should really be something that does fixed point multiplication at runtime (just like the gas estimate adjustment) but accepts decimal amount in config. Will fix that.

crates/agglayer-config/src/outbound.rs
iljakuklic: Changed the default max and added a warning log if the ceiling is hit.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -243,6 +243,10 @@ where
             .wait_for_transaction_receipt(settlement_tx_hash, certificate_id)
             .await?;
 
+        if !receipt.inner.tx_type().is_eip1559() {
+            warn!(tx = %settlement_tx_hash, "Settlement tx is not eip1559.");
```
Freyskeyd, line 247: why do we need to log this one?

crates/agglayer-config/src/outbound.rs
```diff
@@ -63,6 +69,41 @@ impl Default for OutboundRpcSettleConfig {
...
+/// Gas price configuration for settlement transactions.
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct GasPriceConfig {
+    /// Gas price multiplier for the transaction.
+    /// The gas price is calculated as follows:
+    /// `gas_price = estimate_gas_price * multiplier`
+    #[serde(default, skip_serializing_if = "crate::is_default")]
+    pub multiplier: Multiplier,
```
Freyskeyd, line 86: Should we remove the old `gas_multiplier_factor` ?

crates/agglayer-config/src/outbound.rs
```diff
@@ -97,6 +138,12 @@ const fn default_settlement_timeout() -> Duration {
     Duration::from_secs(20 * 60)
 }
 
+/// Default gas price ceiling for the transaction.
+const fn default_gas_price_ceiling() -> u128 {
+    // 400 gwei
+    400_000_000_000_u128
```
Freyskeyd: Where does this value come from?

crates/agglayer-contracts/src/lib.rs
```diff
@@ -20,6 +20,27 @@ pub use aggchain::AggchainContract;
...
+    /// Minimum gas price floor (in wei) for transactions.
+    pub floor: u128,
+    /// Maximum gas price ceiling (in wei) for transactions.
+    pub ceiling: u128,
+}
+
+impl Default for GasPriceParams {
+    fn default() -> Self {
+        GasPriceParams {
+            multiplier_per_1000: 1000, // 1.0 scaled by 1000
```
Freyskeyd, line 37: by default we increase by 1000 the gas?

iljakuklic: > by default we increase by 1000 the gas?

The multiplier is scaled by 1000 to get three decimal places. So 1000 here means multiplier of 1.0, which does nothing.

crates/agglayer-config/src/outbound.rs
iljakuklic: > Where does this value come from?

Rougly 5x the max gas over the past two years, @atanmarko suggested that as a heuristic.

crates/agglayer-config/src/outbound.rs
iljakuklic: > Should we remove the old `gas_multiplier_factor` ?

I think it makes sense to keep it, as these are separate parameters to the transaction. One specifies the gas limit, the other max price per gas. Price tends to be much more volatile than gas usage. Maybe we should review how these interact though I would consider that to be more of a matter of tuning these params.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
iljakuklic: I expect the transaction to be `eip1559`, otherwise the configuration may not be correctly reflected in the gas amounts. Conceptually, this should be an assert but it does not check some sort of fatal condition worth tearing the node down over.

crates/agglayer-config/src/outbound.rs
Freyskeyd: @praetoriansentry does it sound good to you?

iljakuklic: There are basically three options here:

a. Default to unlimited. I had that originally. Basically, if the config is not present, it does not do anything.
b. Use some arbitrary default to put guardrails around the gas price if not configured (current state).
c. Do not have a default, the user has to specify this amount always. Will be annoying since the config structs that contain this value, directly or indirectly, will have to accept an extra param. Old configs will fail.

I don't have a strong preference between (a) and (b), would rather avoid (c).

praetoriansentry: Makes sense. 400gwei would be very expensive and i think having some kind of ceiling makes sense to avoid paying tons of ETH if the gas prices spike for some reason. As long as this is configurable, I think the control is very valuable.

crates/agglayer-config/src/with/eth_amount.rs
```diff
@@ -0,0 +1,105 @@
...
+        let (number_part, unit) = amount_str.split_at(unit_pos);
+        let number_part = number_part.trim_end();
+
+        // Parse using alloy's parse_units
+        let parsed = parse_units(number_part, unit).map_err(Error::UnitParsing)?;
+
+        // Convert ParseUnits to u128
+        match parsed {
+            ParseUnits::U256(value) => u128::try_from(value).map_err(|_| Error::Overflow),
+            ParseUnits::I256(_) => Err(Error::NegativeAmount),
```
hadjiszs, line 50: considering that the tests pass, I guess positive values aren't parsed as I256 but is it always guaranteed? also what about zero value (if this case havent already been caught earlier in the execution flow)

crates/agglayer-contracts/src/settler.rs
```diff
@@ -98,6 +101,110 @@ where
             tx_call = tx_call.gas(adjusted_gas);
         }
 
+        let tx_call = {
```
hadjiszs, line 104: I have difficulty to see the difference between what is adjusted there, and what is adjusted right above, both code block mention about gas adjustment and estimate

maybe bit more comments to understand what is tuned and where?
also do we want to log a bit more things for the new adjusted values?

crates/agglayer-config/src/outbound.rs
atanmarko: We have put ceiling eventually to 100gwei if it is not set in configuration. But it should be always set in configuration.

crates/agglayer-contracts/src/settler.rs
atanmarko: @hadjiszs First one sets the gas limit, other gas price.

atanmarko: I have added comments.

crates/agglayer-config/src/with/eth_amount.rs
atanmarko: @hadjiszs Even if they are parsed as type I256, it has big enough unsigned limit to cover return value of u128.

atanmarko: @hadjiszs From [here](https://docs.rs/alloy-primitives/latest/src/alloy_primitives/utils/units.rs.html#243-290) it seems to be guarantied that you get U256

hadjiszs: >Even if they are parsed as type I256, it has big enough unsigned limit to cover return value of u128.

my concern was if there is a case where one positive value get parsed as I256 and lead wrongly to `Error::NegativeAmount`, but yeah no issue there the code you linked show indeed that positive values will always be U256

PR #1109, opened 2025-10-10
fix: nonce handling

atanmarko:
Resolves https://github.com/agglayer/agglayer/issues/1106

### DIFF
--- a/crates/agglayer-contracts/src/settler.rs
+++ b/crates/agglayer-contracts/src/settler.rs
@@ -214,7 +214,7 @@ mod test {
     use super::{adjust_gas_price_estimate, GasPriceParams};
 
     #[rstest::rstest]
-    fn test_adjust_gas_estimate_respects_floor_and_ceiling(
+    fn test_adjust_gas_price_estimate_respects_floor_and_ceiling(
         #[values(500, 1000, 1500, 2000)] multiplier_per_1000: u64,
         #[values(10_000_000, 50_000_000)] floor: u128,
         #[values(100_000_000, 200_000_000)] ceiling: u128,
@@ -258,7 +258,7 @@ mod test {
     }
 
     #[rstest::rstest]
-    fn test_adjust_gas_estimate_with_retries(
+    fn test_adjust_gas_price_estimate_with_retries(
         #[values(1, 2, 3, 5)] number_of_retries: u64,
         #[values(1000)] multiplier_per_1000: u64,
         #[values(10_000_000)] floor: u128,
@@ -326,7 +326,7 @@ mod test {
     }
 
     #[test]
-    fn test_adjust_gas_estimate_with_zero_retries() {
+    fn test_adjust_gas_price_estimate_with_zero_retries() {
         let estimate = Eip1559Estimation {
             max_fee_per_gas: 50_000_000,
             max_priority_fee_per_gas: 25_000_000,
@@ -353,7 +353,7 @@ mod test {
     }
 
     #[test]
-    fn test_adjust_gas_estimate_retries_hit_ceiling() {
+    fn test_adjust_gas_price_estimate_retries_hit_ceiling() {
         let estimate = Eip1559Estimation {
             max_fee_per_gas: 50_000_000,
             max_priority_fee_per_gas: 25_000_000,


crates/agglayer-contracts/src/settler.rs
```diff
@@ -140,7 +144,11 @@ where
     }
 }
 
-fn adjust_gas_estimate(estimate: &Eip1559Estimation, params: &GasPriceParams) -> Eip1559Estimation {
+fn adjust_gas_price_estimate(
+    estimate: &Eip1559Estimation,
+    params: &GasPriceParams,
+    number_of_retries: Option<u64>,
```
iljakuklic, line 150: Hmm I'm not convinced this function, which basically is a simple mathematical operation on the estimate, should involve domain concepts such as retries. Would be IMHO be cleaner to pre-process the multiplier param and use this function as is.

crates/agglayer-contracts/src/settler.rs
```diff
@@ -154,6 +162,13 @@ fn adjust_gas_estimate(estimate: &Eip1559Estimation, params: &GasPriceParams) ->
     };
 
     let mut max_fee_per_gas = adjust(estimate.max_fee_per_gas).max(*floor);
+    if let Some(retries) = number_of_retries {
+        if retries > 0 {
```
iljakuklic, line 166: Could `number_of_retries` be just a `u64`? I see `None` and `Some(0)` are treated the same. It also makes sense conceptually... 0 means no retries.

crates/agglayer-contracts/src/settler.rs
```diff
@@ -154,6 +162,13 @@ fn adjust_gas_estimate(estimate: &Eip1559Estimation, params: &GasPriceParams) ->
...
+            max_fee_per_gas = max_fee_per_gas
+                .saturating_mul(retries as u128 * DEFAULT_GAS_PRICE_RETRY_INCREASE_FACTOR)
+                .saturating_div(100);
```
iljakuklic, line 169: I am not sure this does what is intended. It is basically (ignoring saturation and rounding and factor scaling issues):

```rust
let new_gas = if retries > 0 {
    gas * retries * INCREASE_FACTOR
} else {
    gas
}
```

I.e. for increase factor of 1.2, the gas is multiplied by 1.0 initially, then upon retries by 1.2, 2.4, 3.6, etc.

I think either a linear increase (giving 1.0, 1.2, 1.4, 1.6, etc):

```rust
gas * ((INCREASE_FACTOR - 1.0) * retries + 1.0)
```

or exponential increase (1.0, 1.2, 1.44, 1.728):

```rust
gas * INCREASE_FACTOR.pow(retries)
```

would make more sense.

crates/agglayer-contracts/src/settler.rs
atanmarko: @iljakuklic Replacement tx would fail if it doesn't have 10% gas price bigger than the previous. As you are taking the current network estimate, it is a problem. The right solution would be to keep the previous tx gas fee prices, but not sure if it is a too big change now.

crates/agglayer-contracts/src/settler.rs
atanmarko: It can - I just though it is safer like this, to communicate that there are no retries with the Option. retries == 0 is worth the assert check, so just a paranoid check here.

crates/agglayer-contracts/src/settler.rs
atanmarko: > I.e. for increase factor of 1.2, the gas is multiplied by 1.0 initially, then upon retries by 1.2, 2.4, 3.6, etc.

This is what I have wanted (to be safe that the replacement transactions have bigger gas, having in mind that we are doing unrelated estimate again from that point in the network).
To have linear or exponential, we should keep the previous gas prices.

iljakuklic: > > I.e. for increase factor of 1.2, the gas is multiplied by 1.0 initially, then upon retries by 1.2, 2.4, 3.6, etc.
> 
> This is what I have wanted (to be safe that the replacement transactions have bigger gas, having in mind that we are doing unrelated estimate again from that point in the network). To have linear or exponential, we should keep the previous gas prices.

I just found the initial step to be weird. First the price jumps by 20% of the original estimate, and subsequently it jumps by 120%.

PR #1110, opened 2025-10-10
fix: move to readonly db store for get_proof

Freyskeyd:


crates/agglayer-storage/src/stores/epochs.rs
```diff
@@ -49,7 +49,10 @@ where
     StateStore: StateWriter + StateReader + MetadataWriter,
 {
     type PerEpochStore = PerEpochStore<PendingStore, StateStore>;
-    fn open(&self, epoch_number: EpochNumber) -> Result<PerEpochStore<PendingStore, StateStore>, Error> {
+    fn open(
```
hadjiszs, line 52: maybe we can remove the writer trait bounds in the `where` clause?

PR #1111, opened 2025-10-10
fix: repeated tx fee issue

atanmarko:
Resolves: #1106

### DIFF
--- a/crates/agglayer-contracts/src/settler.rs
+++ b/crates/agglayer-contracts/src/settler.rs
@@ -151,7 +151,6 @@ where
                             .unwrap_or(estimate.max_priority_fee_per_gas)
                             .max(self.gas_price_params.floor)
                             .min(self.gas_price_params.ceiling),
-
                     };
                     debug!(
                         "Nonce provided: {nonce_info:?}, increasing  previous max_fee_per_gas and \


crates/agglayer-contracts/src/settler.rs
```diff
@@ -108,32 +110,48 @@ where
...
+                if let Some((nonce, previous_max_fee_per_gas, previous_max_priority_fee_per_gas)) =
+                    nonce_info
+                {
+                    // This is repeated transaction, increase the previous max_fee_per_gas and
+                    // max_priority_fee_per_gas by a factor
+                    // If previous_max_priority_fee_per_gas is None, set it to estimated.
+                    let adjust: Eip1559Estimation = Eip1559Estimation {
+                        max_fee_per_gas: previous_max_fee_per_gas
+                            .saturating_mul(DEFAULT_GAS_PRICE_REPEAT_TX_INCREASE_FACTOR)
+                            .saturating_div(100),
```
iljakuklic: Somewhat offtopic: I am very confused by [`u128::saturating_div`](https://doc.rust-lang.org/std/primitive.u128.html#method.saturating_div). When does it actually saturate? The only case I can think of is to cap the result at `u128::MAX` if the denominator is 0 but the linked docs say it panics in that case 🤷.

crates/agglayer-contracts/src/settler.rs
```diff
@@ -108,32 +110,48 @@ where
...
+                    adjust
+                } else {
+                    // Adjust the gas fees based on the configuration.
+                    let adjust = adjust_gas_estimate(&estimate, &self.gas_price_params);
+                    debug!(
+                        "Calculated adjusted gas estimation for rollup_id: {rollup_id}: \
+                         max_priority_fee_per_gas: {}, max_fee_per_gas: {}. Original estimate: \
+                         {:?}",
+                        adjust.max_priority_fee_per_gas, adjust.max_fee_per_gas, estimate
+                    );
```
iljakuklic: A log like this one is already inside `adjust_gas_estimate`, is it not?

crates/agglayer-contracts/src/settler.rs
```diff
@@ -108,32 +110,48 @@ where
...
+                            })
+                            .unwrap_or(estimate.max_priority_fee_per_gas),
+                    };
+                    debug!(
+                        "Nonce provided: {nonce_info:?}, increasing  previous max_fee_per_gas and \
+                         max_priority_fee_per_gas to {adjust:?} for rollup_id: {rollup_id}"
+                    );
+                    // Set the nonce for the transaction
+                    tx_call = tx_call.nonce(nonce);
+                    adjust
```
iljakuklic, line 161: This should probably also be constrained by the gas price floor and ceiling from the config. Maybe that part could be factored out of `adjust_gas_estimate` into a separate function, as it also includes some useful logs.

crates/agglayer-contracts/src/settler.rs
```diff
@@ -108,32 +110,48 @@ where
...
+            let adjusted_fees =
+                if let Some((nonce, previous_max_fee_per_gas, previous_max_priority_fee_per_gas)) =
+                    nonce_info
+                {
+                    // This is repeated transaction, increase the previous max_fee_per_gas and
+                    // max_priority_fee_per_gas by a factor
+                    // If previous_max_priority_fee_per_gas is None, set it to estimated.
+                    let adjust: Eip1559Estimation = Eip1559Estimation {
+                        max_fee_per_gas: previous_max_fee_per_gas
+                            .saturating_mul(DEFAULT_GAS_PRICE_REPEAT_TX_INCREASE_FACTOR)
```
iljakuklic: Since this increases the gas exponentially with every attempt, I wonder under what conditions is another attempt triggered? If we give up too early on an attempt, the gas price might escalate very quickly.

crates/agglayer-contracts/src/settler.rs
atanmarko: Yeah - just thought to be consistent in calculation 🤷

crates/agglayer-contracts/src/settler.rs
atanmarko: Yeah, this can be removed.

crates/agglayer-contracts/src/settler.rs
atanmarko: We set the timeout in the config. Currently it is 1 hour on bali.

iljakuklic: > We set the timeout in the config. Currently it is 1 hour on bali.

OK. Let's just make sure the price cap is applied, just in case.

crates/agglayer-contracts/src/settler.rs
atanmarko: @iljakuklic This is repeated transaction, so flor was already fixed in the first try. Enough to limit the ceiling.

crates/agglayer-contracts/src/settler.rs
atanmarko: Done

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -275,24 +275,35 @@ where
             )));
         }
 
+        if self.previous_tx_hashes.len() > 5 {
```
iljakuklic: Could be nice to make this configurable eventually, or at least a named constant for now.

crates/agglayer-contracts/src/settler.rs
```diff
@@ -120,17 +120,36 @@ where
...
-                        max_fee_per_gas: previous_max_fee_per_gas
-                            .saturating_mul(DEFAULT_GAS_PRICE_REPEAT_TX_INCREASE_FACTOR)
-                            .saturating_div(100)
-                            .min(self.gas_price_params.ceiling),
+                        max_fee_per_gas: {
+                            let mut new_max_fee_per_gas = previous_max_fee_per_gas
+                                .saturating_mul(DEFAULT_GAS_PRICE_REPEAT_TX_INCREASE_FACTOR)
+                                .div_ceil(100)
+                                .min(self.gas_price_params.ceiling)
+                                .max(self.gas_price_params.floor);
```
iljakuklic: In general I prefer to apply floor and then ceiling, so the ceiling takes precedence in case of misconfiguration (ceiling < floor). I have a WIP branch to raise an error in this case instead but may still be better to have the implementation robust too.

EDIT: The PR is here:
* #1112

crates/agglayer-contracts/src/settler.rs
```diff
@@ -120,17 +120,36 @@ where
...
+                            // In the corner case that the previous fee is the same as the new fee
+                            // due to rounding, multiply it by 2 to
+                            // ensure progress
+                            if new_max_fee_per_gas == previous_max_fee_per_gas {
+                                new_max_fee_per_gas *= 2;
+                            }
```
iljakuklic, line 135: Since this is applied after the ceiling, it may raise the price up to just under 2x the configured ceiling. The priority gas price calculation below does not have this issue.

atanmarko: Will push a fix

crates/agglayer-contracts/src/settler.rs
```diff
@@ -108,32 +110,62 @@ where
...
+                                // the new fee due to rounding,
+                                // multiply it by 2 to ensure progress
+                                if new_max_priority_fee_per_gas == previous {
+                                    new_max_priority_fee_per_gas *= 2;
+                                }
+                                new_max_priority_fee_per_gas
+                            })
+                            .unwrap_or(estimate.max_priority_fee_per_gas)
+                            .min(self.gas_price_params.ceiling)
+                            .max(self.gas_price_params.floor),
```
iljakuklic: Here too, swap the floor and ceiling.

atanmarko: Will do, this would be really extreme case where in config floor is bigger than ceiling.

crates/agglayer-aggregator-notifier/src/certifier/tests.rs
```diff
@@ -276,7 +276,7 @@ mockall::mock! {
             new_pessimistic_root: [u8; 32],
             proof: Bytes,
             custom_chain_data: Bytes,
-            nonce: Option<u64>,
+            nonce: Option<(u64, u128, Option<u128>)>
```
polykeith, line 279: Why don't we just accept a `NonceInfo` instead of passing in a tuple of its fields?

atanmarko: @polykeith I didn't want to introduce dependency from `agglayer-certificate-orchestartor` `serttlement_client` to `agglayer-contract` `L1RpcClient` (which doesn't have any non-alloy dependencies). Dependency goes in the opposite direction.
But I agree it would be nice to move `NonceInfo` to  `agglayer_primitives` or similar.

PR #1112, opened 2025-10-10
chore: gas price config follow up

iljakuklic:
Miscellaneous clean ups, handling subtle corner cases, improved testing
and other assorted nice-to-haves.

### DIFF
--- a/crates/agglayer-contracts/src/lib.rs
+++ b/crates/agglayer-contracts/src/lib.rs
@@ -129,7 +129,7 @@ pub enum L1RpcError {
     ReorgDetected(u64),
     #[error("Cannot get the block hash for the block number {0}")]
     BlockHashNotFound(u64),
-    #[error("Unable to fetch transaction receipt for {tx_hash}: {source}")]
+    #[error("Unable to fetch transaction receipt for {tx_hash}")]
     UnableToFetchTransactionReceipt {
         tx_hash: String,
         #[source]
--- a/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
@@ -156,7 +156,8 @@ where
         info!(
             %prev_pp_root,
             %new_pp_root,
-            "Successfully generated the witness for the PP for the Certificate {certificate_id}"
+            %certificate_id,
+            "Successfully generated the witness for the PP for certificate",
         );
 
         let network_state = pessimistic_proof::NetworkState::from(initial_state);
--- a/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
+++ b/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
@@ -214,11 +214,7 @@ where
                     .unwrap_or_else(|| error.to_string());
                 let error_message = error.to_string();
 
-                error!(
-                    error_message,
-                    error_decoded = error_decoded,
-                    "Failed to settle certificate"
-                );
+                error!(error_message, error_decoded, "Failed to settle certificate");
 
                 return Err(Error::SettlementError {
                     certificate_id,
@@ -333,9 +329,10 @@ where
             .mul_f64(self.config.max_retries as f64);
 
         debug!(
-            "Waiting for transaction receipt with timeout of {timeout:?}, max_retries: {} and \
-             retry_interval: {:?}",
-            self.config.max_retries, self.config.retry_interval
+            ?timeout,
+            max_retries = self.config.max_retries,
+            retry_interval = ?self.config.retry_interval,
+            "Waiting for transaction receipt",
         );
 
         for attempt in 0..=self.config.max_retries {
@@ -363,7 +360,9 @@ where
                         for confirmation_attempt in attempt..=self.config.max_retries {
                             match self.l1_rpc.get_provider().get_block_number().await {
                                 Ok(current_block) => {
-                                    let confirmations = current_block.saturating_sub(receipt_block);
+                                    let confirmations = current_block
+                                        .saturating_sub(receipt_block)
+                                        .saturating_add(1);
                                     if confirmations >= self.config.confirmations as u64 {
                                         info!(
                                             confirmations,
@@ -432,6 +431,7 @@ where
                             %settlement_tx_hash,
                             next_attempt = attempt + 1,
                             max_retries = self.config.max_retries,
+                            retry_interval = ?self.config.retry_interval,
                             "Transaction receipt not found yet, retrying after {:?}",
                             self.config.retry_interval
                         );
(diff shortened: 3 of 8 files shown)


crates/agglayer-config/src/multiplier.rs
```diff
@@ -5,37 +5,67 @@
 #[derive(
```
Ekleog-Polygon, line 5: Can we remove the [*.snap.new gitignore](https://github.com/agglayer/agglayer/commit/aca1782b04bc4d39a6b09cdf057fcaa4c1c12098#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947) here maybe?

crates/agglayer-config/src/multiplier.rs
```diff
@@ -5,37 +5,67 @@
...
+    pub fn try_from_f64_check(x: f64) -> Result<Self, FromF64Error> {
+        // We first get the rounded conversion, check the delta against the original
+        // value and fail if there is too much precision loss, indicating there were
+        // too many decimals in the original floating point number.
+        let r = Self::try_from_f64_round(x)?;
+        let delta = r.as_u64_per_1000() as f64 - Self::scale_f64(x);
+
+        // We still allow some tolerance to account for the fact that floating point
+        // cannot represent base-10 decimals (such as 1.2) exactly.
+        (delta.abs() <= Self::FROM_F64_TOLERANCE)
```
Ekleog-Polygon, line 46: FWIW, we could probably just ensure that the provided f64 is <= 2^53 / SCALE ; considering [the exactly-representable decimals](https://stackoverflow.com/questions/3793838/which-is-the-first-integer-that-an-ieee-754-float-is-incapable-of-representing-e). This being said, your new option works fine for me too :)

crates/agglayer-config/src/multiplier.rs
iljakuklic: > Can we remove the [*.snap.new gitignore](https://github.com/agglayer/agglayer/commit/aca1782b04bc4d39a6b09cdf057fcaa4c1c12098#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947) here maybe?

Will do.

crates/agglayer-config/src/multiplier.rs
iljakuklic: > FWIW, we could probably just ensure that the provided f64 is <= 2^53 / SCALE ; considering [the exactly-representable decimals](https://stackoverflow.com/questions/3793838/which-is-the-first-integer-that-an-ieee-754-float-is-incapable-of-representing-e). This being said, your new option works fine for me too :)

If I understand your point correctly, I think the proposed check is already in the `try_from_f64_lossy` below. Basically, there are two checks to do with conversion from float:

**Check 1:** Very large floats may be out of range for the `u64` representation. That check is in `try_from_f64_lossy` and is applied to `try_from_f64_strict` too, since it delegates to it eventually.

**Check 2:** The scaled `u64` representation only supports 3 decimal places. However `f64` supports way more fractional precision especially around zero. So if the user enters a multiplier such as `1.0004`, the `_lossy` conversion trims it to `1.0` which may result in unexpected behaviour. This check is there to reject the likes of `1.0004`. There is still a tolerance to account for the fact that e.g. `1.1` may end up being something like `1.10000000053269(...garbage)` in f64. I did not bother working out the precise bounds for what the tolerance should be so I just chose `1e-6` semi-arbitrarily.

iljakuklic: Having said that, I'm pretty sure the implementation of the checks could be cleaned up / simplified.

atanmarko (approved):
LGTM

crates/agglayer-config/src/multiplier.rs
Ekleog-Polygon: I can confirm that you already implemented this check in the new version! Sorry, my comment was about the previous PR, and I had skimmed over this PR when writing the comment :sweat_smile:

crates/agglayer-config/src/multiplier.rs
```diff
@@ -45,16 +78,127 @@ impl Default for Multiplier {
     }
 }
 
-impl TryFrom<f32> for Multiplier {
-    type Error = FromF32Error;
+impl std::fmt::Display for Multiplier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // A quick, dirty and inefficient implementation that does not use f64
+        // to print the decimal number since it could result in imprecision.
```
Ekleog-Polygon, line 84: FWIW, we could use https://crates.io/crates/fixed_decimal as a backend, but maybe it'd be overengineering

crates/agglayer-config/src/multiplier.rs
```diff
@@ -45,16 +78,127 @@ impl Default for Multiplier {
...
+        let result = Multiplier::try_from_f64_strict(input).unwrap();
+        assert_eq!(result, Multiplier::from_u64_per_1000(expected));
+    }
+
+    #[rstest]
+    #[case(-1.0)]
+    #[case(-0.001)]
+    #[case(-100.0)]
+    #[case(1.001 * u64::MAX as f64)]
+    fn try_from_f64_strict_out_of_range(#[case] input: f64) {
```
Ekleog-Polygon: Nit: could add a case here with 2^53 ; as that should be out of the safe integer range of ieee754

crates/agglayer-config/src/multiplier.rs
```diff
@@ -45,16 +78,127 @@ impl Default for Multiplier {
...
+    fn try_from_f64_lossy_valid_values(#[case] input: f64, #[case] expected: u64) {
+        let result = Multiplier::try_from_f64_lossy(input).unwrap();
+        assert_eq!(result, Multiplier::from_u64_per_1000(expected));
+    }
+
+    #[rstest]
+    #[case(-1.0)]
+    #[case(-0.001)]
+    #[case(-100.0)]
+    fn try_from_f64_lossy_out_of_range(#[case] input: f64) {
```
Ekleog-Polygon: Nit: Same here

crates/agglayer-contracts/src/lib.rs
```diff
@@ -27,16 +27,43 @@ pub use settler::Settler;
...
+}
+
+impl GasPriceParams {
+    /// Create new gas price parameters.
+    ///
+    /// Returns an error if ceiling < floor.
+    pub fn new(
+        multiplier_per_1000: u64,
+        floor: u128,
+        ceiling: u128,
```
Ekleog-Polygon: Random thought: taking two parameters of the same type makes it relatively likely that we'll inadvertently mix them up, more than before where the name was enforced by the struct initialization syntax. OTOH, making this a builder pattern would likely be overkill. So… let's leave it like this I guess :shrug:

Ekleog-Polygon (dismissed):
LGTM! Feel free to ignore any or all of my nits below :smile:

crates/agglayer-config/src/multiplier.rs
iljakuklic: > FWIW, we could use https://crates.io/crates/fixed_decimal as a backend, but maybe it'd be overengineering

Yea, that occurred to me too after having implemented it :laughing:

crates/agglayer-contracts/src/lib.rs
iljakuklic: > Random thought: taking two parameters of the same type makes it relatively likely that we'll inadvertently mix them up, more than before where the name was enforced by the struct initialization syntax. OTOH, making this a builder pattern would likely be overkill. So… let's leave it like this I guess 🤷

Could change the floor and ceiling to `RangeInclusive<u128>`, should be obvious enough which one is which with the `floor..=ceiling` syntax.

PR #1121, opened 2025-10-13
fix: additional check for previous tx

atanmarko:
Additional pp_root check

### DIFF
--- a/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
@@ -194,7 +194,7 @@ where
         let mut state = state.await.map_err(recv_err)??;
 
         debug!("Recomputing new state for already-proven certificate");
-        // settlement_tx_hash_missing_on_l1 is true if the settlement tx hash in
+        // `settlement_tx_hash_missing_on_l1` is `true` if the settlement tx hash in
         // certificate header is not found on L1.
         let settlement_tx_hash_missing_on_l1: bool =
             if let Some(previous_tx_hash) = self.header.settlement_tx_hash {


crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -435,7 +434,8 @@ where
                         // Transaction not yet included in a block, continue retrying
                         if attempt <= self.config.max_retries {
                             // Log at 25%, 50%, 75% progress milestones
-                            let progress_percent = ((attempt + 1) * 100) / self.config.max_retries;
+                            let progress_percent =
+                                ((attempt + 1) * 100) / self.config.max_retries.max(1);
```
iljakuklic: The check below matches against the exact percentage. That works well when `max_retries` is a multiple of 4. But if max retries is for example 5, the calculation above will return `[20, 40, 60, 80, 100]` so the only time it is logged is actually at the end.

I guess what you want is to check whether a percentage point has been crossed. You can do that by checking the percentage range between `100 * attempt / max_retries` and `100 * (attempt + 1) / retries` contains any of `[25, 50, 75, 100]`.

Since `max_retries` is non-negative, you can also multiply the inequalities by `max_retries` to get rid of the division and the associated non-zero check fix.

A rough sketch:
```rust
let limit = self.config.max_retries;
let cur_progress = (100 * attempt + 1) ..= (100 * (attempt + 1)); // Note the parens
let do_log = [25, 50, 75, 100].iter().any(|p| cur_progress.contains(limit * p));
```

A further refinement is to use N equal intervals instead of percentiles and determine which interval does the current vs next attempt fall into:
```rust
const N: usize = 4; // Report every 25% = split the progress into 4 parts
let max_attempts = self.config.max_retries.max(1);
let curr = N * attempt / max_attempts;
let next = N * (attempt + 1) / max_attempts;
let do_log = curr != next;
```

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -549,6 +572,31 @@ where
...
+                                    "Settlement tx {settlement_tx_hash} is still pending");
+                                Ok(false)
+
+                            }
+                            Err(err) => {
+                                debug!(%certificate_id,
+                                    "Error checking receipt status for settlement tx {settlement_tx_hash}: {err}");
+                                Err(err)
+                            }
+                        };
```
iljakuklic: Style nitpick: separate information retrieval from logging:


```suggestion
                        let mined = self.settlement_client.fetch_settlement_receipt_status(settlement_tx_hash).await;
                        match &mined {
                            Ok(true) => {
                                info!(%certificate_id,
                                    "Settlement tx {settlement_tx_hash} has been mined");
                            }
                            Ok(false) => {
                                debug!(%certificate_id,
                                    "Settlement tx {settlement_tx_hash} is still pending");
                            }
                            Err(err) => {
                                debug!(%certificate_id,
                                    "Error checking receipt status for settlement tx {settlement_tx_hash}: {err}");
                            }
                        };
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -335,6 +392,11 @@ where
             "Submitted certificate for settlement"
         );
 
+        if height.as_u64() >= 1 && self.previous_tx_hashes.len() >= 3 {
+            println!(">>>>>>>>>>>>>>> Turn me off now <<<<<<<<<<<<<<<");
+            tokio::time::sleep(Duration::from_secs(300)).await;
+        }
```
iljakuklic: If this proves a useful debugging tool, maybe it could be turned into a fail point. If the binary is compiled with fail points support enabled, the fail points can be controlled using an environment variable. You can make them crash, return error, sleep for some period, etc. Can be also made to fail only after a certain number of calls. Seems could be useful here.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: Could you push the fix?

crates/agglayer-certificate-orchestrator/src/network_task.rs
atanmarko: @iljakuklic Push the fix please

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: I intend to write a dedicated integration test for this family of issues

atanmarko: Here is the actual ticket https://github.com/agglayer/agglayer/issues/1122

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -97,6 +98,21 @@ pub enum NetworkTaskMessage {
...
+        settlement_tx_hash: SettlementTxHash,
+        // Notifier to send back the result of whether the tx has been mined or not.
+        tx_mined_notifier: oneshot::Sender<Result<bool, Error>>,
+    },
+
+    /// Check if settlement tx has been mined.
+    FetchLatestContractPPRoot {
+        // Notifier to send back the result of the latest pp root from L1.
+        contract_pp_root_notifier:
+            oneshot::Sender<Result<Option<(Digest, SettlementTxHash)>, Error>>,
```
iljakuklic, line 114: Nice type 😆

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -549,6 +579,35 @@ where
...
+                            .send(mined)
+                            .map_err(|_|Error::InternalError("Certificate notification channel closed".into()))?;
+                        continue;
+                    }
+                    Some(NetworkTaskMessage::FetchLatestContractPPRoot { contract_pp_root_notifier }) => {
+                        // Fetch the latest pp root from L1
+                        let latest_pp_root = self.fetch_latest_pp_root_from_l1().await;
+                        contract_pp_root_notifier
+                            .send(latest_pp_root)
+                            .map_err(|_| Error::InternalError("Certificate notification channel closed".into()))?;
```
iljakuklic, line 610: Possible follow up: consider if it makes sense to perform this in certificate task, rather than doing the network task roundtrip.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -193,11 +193,167 @@ where
...
+                debug!(
+                    "Settlement tx {previos_tx_hash} existence on L1: \
+                     {result_is_settlement_tx_mined:?}"
+                );
+                match result_is_settlement_tx_mined {
+                    Ok(true) => false,  // We have fetched the receipt, tx status 1, tx exist on L1
+                    Ok(false) => false, // Tx found on l1, but with status 0 (reverted)
+                    Err(error) => {
+                        if error.to_string().contains("No transaction receipt found") {
+                            true
```
iljakuklic, line 220: Hmm, recovering the error from string seems rather... suboptimal. Not that I have a better suggestion given the current constraints.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -193,11 +193,167 @@ where
         .await?;
         let mut state = state.await.map_err(recv_err)??;
 
-        // Execute the witness generation to retrieve the new local network state
         debug!("Recomputing new state for already-proven certificate");
+        // Check if cert `settlement_tx_hash` exist on the l1
+        let settlement_tx_hash_missing_on_l1: bool =
```
Freyskeyd, line 199: I'm not sure to understand with the comment above, does this bool represent the fact that the transaction exists or not?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -193,11 +193,167 @@ where
...
+            if let Some(previos_tx_hash) = self.header.settlement_tx_hash {
```
Freyskeyd:
```suggestion
            if let Some(previous_tx_hash) = self.header.settlement_tx_hash {
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -193,11 +193,167 @@ where
...
+            let recomputed_from_contract: Option<Digest> = match result_latest_contract_pp_root {
+                Ok(Some((contract_pp_root, contract_settlement_tx_hash))) => {
+                    // Try to recompute the state with the latest tx from contract.
+                    match self
+                        .certifier_client
+                        .witness_generation(
+                            &self.certificate,
+                            &mut state,
+                            Some(contract_settlement_tx_hash.into()),
+                        )
```
iljakuklic, line 262: I am a bit concerned by the double execution of `witness_generation`. The state (passed in as `&mut state`) gets updated by the first execution, so it seems to me the second one gets the updated state while it should get the original one. This may happen to work in the trivial case where there is no state update, i.e. no (imported)bridge exits. But does it work in general?

(Nothing that can't be solved with a bit of cloning.)

crates/agglayer-certificate-orchestrator/src/network_task.rs
atanmarko: Well, network task has the settlement client - so all network related tasks are performed there. So would have have to refactor significantly how we handle things in that case.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: Yeah, here is the follow up ticket - https://github.com/agglayer/agglayer/issues/1123

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: Hm, good point. Let me check.

atanmarko: Fixed in latest commit.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -193,11 +193,167 @@ where
...
+                        "Cert settlement tx is missing from the l1, but failed to fetch latest pp \
+                         root from contract: {error}"
+                    )));
+                }
+            };
+
+            if recomputed_from_contract.is_none() {
+                // Tx not found on L1, and pp root from contract not matching,
+                // clean tx from cert header and move back to Pending
+                let previos_tx_hash = self.header.settlement_tx_hash;
```
Freyskeyd:
```suggestion
                let previous_tx_hash = self.header.settlement_tx_hash;
```

crates/agglayer-certificate-orchestrator/src/error.rs
```diff
@@ -27,8 +27,10 @@ pub enum CertificationError {
     )]
     TrustedSequencerNotFound(NetworkId),
 
-    #[error("Failed to retrieve the last pessimistic root for network {0}")]
-    LastPessimisticRootNotFound(NetworkId),
+    #[error(
+        "Failed to retrieve the last pessimistic root for network {0} for settlement tx hash {1:?}"
+    )]
+    LastPessimisticRootNotFound(NetworkId, Option<Digest>),
```
Freyskeyd, line 33: can it break the storage that already exists? cc @iljakuklic

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -193,11 +193,167 @@ where
...
+                );
+                self.header.settlement_tx_hash = None;
+                if let Err(error) = self.state_store.remove_settlement_tx_hash(&certificate_id) {
+                    error!(
+                        ?error,
+                        "Failed to remove tx_hash {previos_tx_hash:?} from database"
+                    );
+                };
+
+                self.header.status = CertificateStatus::Pending;
```
Freyskeyd: Why not just go back to `Proven` ? `pending` certs can be replaced by the RPC

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: Right, will fix the comment.

crates/agglayer-certificate-orchestrator/src/error.rs
atanmarko: Not this one, only `CertificateStatusError` afaik? cc @iljakuklic

iljakuklic: I think it's fine. `CertificateStatusError::LastPessimisticRootNotFound` is not modified. This is `CertificationError::LastPessimisticRootNotFound` which is not stored.

crates/agglayer-certificate-orchestrator/src/network_task.rs
iljakuklic: The thinking was that this does not appear to require any specific data only available to network task (I mean in terms of state, the settlement client itself is required) but I might have missed something.

atanmarko: Ok, lets have a follow up discussion.

PR #1124, opened 2025-10-15
fix: refactor fetch transaction receipt

### LINKED ISSUE #1123: Refactor `fetch_transaction_receipt` to return Option<TransactionReceipt>
Return `Option<TransactionReceipt>`, not `L1RpcError::TransactionNotYetMined`

atanmarko:
Fixes #1123

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
+++ b/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
@@ -428,20 +428,13 @@ where
                 Ok(None) => {
                     // Transaction not yet included in a block, continue retrying
                     if attempt <= self.config.max_retries {
-                        const N: usize = 4; // Split progress into 4 equal stages.
-                        let max_attempts = self.config.max_retries.max(1);
-                        let curr_stage = N * attempt / max_attempts;
-                        let next_stage = N * (attempt + 1) / max_attempts;
-                        // Report if we cross from one stage to the next.
-                        if curr_stage != next_stage {
-                            debug!(
-                                %settlement_tx_hash,
-                                next_attempt = attempt + 1,
-                                max_retries = self.config.max_retries,
-                                "Transaction receipt not found yet, retrying after {:?}",
-                                self.config.retry_interval
-                            );
-                        }
+                        debug!(
+                            %settlement_tx_hash,
+                            next_attempt = attempt + 1,
+                            max_retries = self.config.max_retries,
+                            "Transaction receipt not found yet, retrying after {:?}",
+                            self.config.retry_interval
+                        );
                         tokio::time::sleep(self.config.retry_interval).await;
                         continue;
                     } else {


crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -196,42 +196,38 @@ where
...
+            );
+            match result_is_settlement_tx_mined {
+                Ok(Some(_)) => false, // We have fetched the receipt, tx status 1, tx exist
+                Ok(None) => true,     // Tx not found on L1
+                Err(error) => {
+                    // Some error happened while checking the tx receipt on L1
+                    warn!(
+                        "Failed to check settlement tx {previous_tx_hash} existence on L1: {error}"
+                    );
+                    false
```
Freyskeyd: You can replace all of this by something like that:

```suggestion
result_is_settlement_tx_mined
    .as_ref()
    .map(Option::is_none)
    .inspect_err(|error| warn!(
        "Failed to check settlement tx {previous_tx_hash} existence on L1: {error}"
    )})
    .unwrap_or(false);
```

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
```diff
@@ -53,7 +53,7 @@ pub trait SettlementClient: Unpin + Send + Sync + 'static {
     async fn fetch_settlement_receipt_status(
         &self,
         settlement_tx_hash: SettlementTxHash,
-    ) -> Result<bool, Error>;
+    ) -> Result<Option<bool>, Error>;
```
iljakuklic: Would be even more readable to replace this with a three-armed enum. Something like `enum TxReceiptStatus { TxSuccessful, TxFailed, NotFound }`, and have helper methods like `.is_receipt_found()`, etc.

The tradeoff is that you lose the immediate ability to do `.ok_or(err)?` but that could be resolved by a conversion function like `.status() -> Option<bool>`.

This would make all the `if..else` cascades way more readable imho, turning them into a `match`.

iljakuklic (approved):

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -425,59 +425,53 @@ where
...
-                        } else {
-                            // Max retries reached
-                            error!(
+                Ok(None) => {
+                    // Transaction not yet included in a block, continue retrying
+                    if attempt <= self.config.max_retries {
+                        const N: usize = 4; // Split progress into 4 equal stages.
+                        let max_attempts = self.config.max_retries.max(1);
+                        let curr_stage = N * attempt / max_attempts;
+                        let next_stage = N * (attempt + 1) / max_attempts;
```
polykeith: Can we use an enum with numeric discriminants to encapsulate the concept of stages? Or wait, did you actually mean "chunks of progress" here rather than stages that are non-fungible?

iljakuklic: > Or wait, did you actually mean "chunks of progress" here rather than stages that are non-fungible?

Correct. The "stages" don't have any distinct meaning.

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
polykeith: That's what I thought as well, and we could simply just implement methods on the enum for syntactic sugar. Could be something that can be included into `agglayer-primitives`.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: Hm, I find the first implementation a bit more clear about what is going on.

iljakuklic: > Hm, I find the first implementation a bit more clear about what is going on.

A sort of middle ground would be to factor out the warn printout:
```suggestion
            result_is_settlement_tx_mined.inspect_err(|error| {
                warn!(?error, "Failed to check settlement tx {previous_tx_hash} existence on L1");
            });
            match result_is_settlement_tx_mined {
                Ok(Some(_)) => false, // We have fetched the receipt, tx status 1, tx exist
                Ok(None) => true,     // Tx not found on L1
                Err(error) => false,  // Some error happened while checking the tx receipt on L1
```

Or even
```suggestion
            result_is_settlement_tx_mined.inspect_err(|error| {
                // Some error happened while checking the tx receipt on L1
                warn!(?error, "Failed to check settlement tx {previous_tx_hash} existence on L1");
            });
            result_is_settlement_tx_mined.is_ok_and(|receipt| receipt.is_none())
```

But I don't have a strong opinion about this. The original version is ok too imho.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
iljakuklic: @Ekleog-Polygon suggests to just drop the progress bar logic and log every attempt, which seems sensible to me.

crates/agglayer-certificate-orchestrator/src/settlement_client.rs
atanmarko: Done

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -425,59 +425,46 @@ where
...
-                        }
+                Ok(None) => {
+                    // Transaction not yet included in a block, continue retrying
+                    if attempt <= self.config.max_retries {
+                        debug!(
+                            %settlement_tx_hash,
+                            next_attempt = attempt + 1,
+                            max_retries = self.config.max_retries,
+                            "Transaction receipt not found yet, retrying after {:?}",
+                            self.config.retry_interval
```
Ekleog-Polygon, line 436: Nit: use the key-value format for retry_interval

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -196,42 +196,39 @@ where
...
+            .await?;
+
+            let result_is_settlement_tx_mined = response_is_settlement_tx_mined
+                .await
+                .map_err(recv_err)?
+                .inspect_err(|error| {
+                    // Some error happened while checking the tx receipt on L1
+                    warn!(
+                        ?error,
+                        "Failed to check settlement tx {previous_tx_hash} existence on L1"
```
Ekleog-Polygon, line 218: Nit: same here, use key-value format for previous_tx_hash

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -104,7 +104,7 @@ pub enum NetworkTaskMessage {
         certificate_id: CertificateId,
         settlement_tx_hash: SettlementTxHash,
         // Notifier to send back the result of whether the tx has been mined or not.
-        tx_mined_notifier: oneshot::Sender<Result<bool, Error>>,
+        tx_mined_notifier: oneshot::Sender<Result<crate::TxReceiptStatus, Error>>,
```
Ekleog-Polygon, line 107: Nit: `use crate::TxReceiptStatus;`

crates/agglayer-rpc/src/lib.rs
```diff
@@ -716,6 +716,23 @@ where
                                     pre_existing_certificate_id, error
                                 );
 
+                                CertificateSubmissionError::UnableToReplacePendingCertificate {
+                                    reason: error.to_string(),
+                                    height: certificate.height,
+                                    network_id: certificate.network_id,
+                                    stored_certificate_id: pre_existing_certificate_id,
+                                    replacement_certificate_id: new_certificate_id,
+                                    source: Some(error),
```
Ekleog-Polygon, line 725: Nit: Can we avoid this big copy-paste, maybe by having it be a closure that takes `error` as parameter and calling it with a hardcoded parameter in `ok_or_else`?

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: I'll fix it in the follow up.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
atanmarko: Fix in the follow up.

crates/agglayer-certificate-orchestrator/src/network_task.rs
atanmarko: Fix in the follow up.

crates/agglayer-rpc/src/lib.rs
atanmarko: @Ekleog-Polygon Could you elaborate a bit more?

Ekleog-Polygon: The code in `.map_err` and `.ok_or_else` is AFAICT exactly the same apart from the `let error` line in `.ok_or_else`. We could have a single closure that we pass in at both places :)

Issue #1120, opened 2025-10-13
Query event `VerifyPessimisticStateTransition` more optimally


atanmarko:
Use other tag, not current `BlockNumberOrTag::Earliest`.

PR #1125, opened 2025-10-15
fix: event handling

atanmarko:
Fixes https://github.com/agglayer/agglayer/issues/1120

### DIFF
--- a/crates/agglayer-contracts/src/lib.rs
+++ b/crates/agglayer-contracts/src/lib.rs
@@ -1,6 +1,9 @@
 //! Agglayer smart-contract bindings.
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
 
 use agglayer_primitives::U256;
 use alloy::{
@@ -10,7 +13,6 @@ use alloy::{
     rpc::types::{Filter, TransactionReceipt},
     signers::k256::elliptic_curve::ff::derive::bitvec::macros::internal::funty::Fundamental,
 };
-use tokio::sync::RwLock;
 use tracing::{debug, error};
 
 pub mod aggchain;
@@ -139,6 +141,8 @@ pub enum L1RpcError {
     TransactionReceiptFailedOnL1(TxHash),
     #[error("Failed to get the events: {0}")]
     FailedToQueryEvents(String),
+    #[error("L1 info roots cache lock poisoned")]
+    CacheLockPoisoned,
 }
 
 impl<RpcProvider> L1RpcClient<RpcProvider>
--- a/crates/agglayer-contracts/src/rollup.rs
+++ b/crates/agglayer-contracts/src/rollup.rs
@@ -66,7 +66,10 @@ where
     async fn get_l1_info_root(&self, l1_leaf_count: u32) -> Result<[u8; 32], L1RpcError> {
         // Check if we already have this l1_info_root cached
         {
-            let cache = self.l1_info_roots.read().await;
+            let cache = self
+                .l1_info_roots
+                .read()
+                .map_err(|_| L1RpcError::CacheLockPoisoned)?;
             if let Some(&cached_root) = cache.get(&l1_leaf_count) {
                 trace!(
                     "Retrieved cached L1 info root for leaf count {}: {}",
@@ -203,7 +206,10 @@ where
 
         // Cache the retrieved l1_info_root for future use
         {
-            let mut cache = self.l1_info_roots.write().await;
+            let mut cache = self
+                .l1_info_roots
+                .write()
+                .map_err(|_| L1RpcError::CacheLockPoisoned)?;
             cache.insert(l1_leaf_count, l1_info_root);
         }
 


crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -545,27 +546,45 @@ where
...
+                .event_signature(VerifyPessimisticStateTransition::SIGNATURE_HASH)
+                .topic1(U256::from(network_id.to_u32()))
+                .from_block(BlockNumberOrTag::Number(start_block))
+                .to_block(BlockNumberOrTag::Number(end_block));
+
+            // Fetch the logs through from the network.
+            events = self
+                .l1_rpc
+                .get_provider()
+                .get_logs(&filter)
```
polykeith, line 577: I'm guessing that this filter specifically looks for the settlement transaction in the logs, right? I'm worried that if none exist, then this loop would continue forever until the genesis block of Ethereum, since it's looping while `end_block > 0`.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -153,6 +160,7 @@ where
             default_l1_info_tree_entry,
             gas_multiplier_factor,
             gas_price_params,
+            l1_info_roots: Arc::new(Mutex::new(HashMap::new())),
```
polykeith: Does it have to be a `Mutex`, or can it be a `RwLock` instead? i.e. allowing multiple readers but no writers, or 1 writer and no other readers/writers.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: True. But it is impossible that there is no such event, it happens on every cert settlement. If it is about new contract and network for test, it will finish in the first loop.

crates/agglayer-contracts/src/lib.rs
atanmarko: Sure, I'll change it to RwLock.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: In the end, this event is queried only in the edge circumstances, where settlement tx has been lost or not settled after a long timeout. It is not queried in happy path. So even if it loops until 0, that is fine.

iljakuklic: Can the filter detect the point of the rollup manager deployment so we can just stop iterating there?

crates/agglayer-contracts/src/lib.rs
```diff
@@ -1,14 +1,16 @@
...
 
 use agglayer_primitives::U256;
 use alloy::{
     eips::BlockNumberOrTag,
     primitives::{Address, FixedBytes, TxHash, B256},
     providers::Provider,
     rpc::types::{Filter, TransactionReceipt},
+    signers::k256::elliptic_curve::ff::derive::bitvec::macros::internal::funty::Fundamental,
 };
+use tokio::sync::RwLock;
```
iljakuklic: Would this work with a std `RwLock`? As far as I can tell, it is not locked across an `await`, and there all sorts of extra complexities (and also perf overhead) associated with async locks.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -171,20 +179,39 @@ where
...
-                .from_block(BlockNumberOrTag::Earliest);
-
-            // Get logs from the contract
-            let logs = rpc.get_logs(&filter).await.map_err(|e| {
-                L1RpcInitializationError::InitL1InfoRootMapEventNotFound(e.to_string())
-            })?;
+            // To not hit the provider limit, we start from genesis and restrict search
+            // to the EVENT_FILTER_BLOCK_RANGE blocks range.
+            let mut events = Vec::new();
+            let mut start_block = 0u64;
```
iljakuklic, line 189: Again, could we work out the point of L1InfoTree deployment and start there?

crates/agglayer-contracts/src/rollup.rs
```diff
@@ -64,22 +64,53 @@ where
...
-            .address(self.l1_info_tree)
-            .event_signature(UpdateL1InfoTreeV2::SIGNATURE_HASH)
-            .topic1(U256::from(l1_leaf_count))
-            .from_block(BlockNumberOrTag::Earliest);
-
-        let events = self
+        // To not hit the provider limit, we start from genesis and restrict search
+        // to the EVENT_FILTER_BLOCK_RANGE blocks range.
+        let mut events = Vec::new();
+        let mut start_block = 0u64;
```
iljakuklic, line 91: Same here

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: @iljakuklic I think it would be too complex and out of context for this function. In the highly unlikely case where we have issue with multiple repeated settlement transactions and there are no events in the contract and contract has been deployed on some non test network who has big block number for historical reasons, I would just leave it to iterate (and it easily would without issues), would not make this function more complex to optimize for that unlikely case. If you are concerned about this, I could increase the block range step.

crates/agglayer-contracts/src/lib.rs
atanmarko: Maybe. But with async RW lock we are on the safe side. This function is queried like once every time we call `witness_generation` - we generate a certificate (so every hour on deployed networks), I don't think saving few microseconds is worth the risk and time to experiment.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: Also, to detect the point of the rollup manager deployment, you would have do to binary search across whole blockchain, so big overhead.

crates/agglayer-contracts/src/lib.rs
atanmarko: We could. Feel free to open a follow up ticket for that. I don't think it is worth the effort. In the worst case, this function would call main-net provider with 200 iterations once and cache the result.

crates/agglayer-contracts/src/rollup.rs
atanmarko: Same answer - we could as a followup. Please open the ticket. It would require binary search for when the address of the contract has code.

crates/agglayer-contracts/src/lib.rs
iljakuklic: > Maybe. But with async RW lock we are on the safe side.

Not quite. Using `std::sync::RwLock` across await is a compiler error so there is no safety concern just changing the import. Worst case it fails to compile and switch back. IIRC, official [tokio docs suggest](https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex-and-tokiosyncmutex) using std locks and only switching to tokio locks if needed.

atanmarko: Ok, switching to std RW lock.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
iljakuklic: > I would just leave it to iterate (and it easily would without issues), would not make this function more complex to optimize for that unlikely case.

That's fair

crates/agglayer-contracts/src/rollup.rs
iljakuklic: > Same answer - we could as a followup. Please open the ticket. It would require binary search for when the address of the contract has code.

Yea, I don't know, binary search sounds rather annoying :smile:

atanmarko: We could review this problem in the follow up.

crates/agglayer-contracts/src/lib.rs
iljakuklic: Filed an issue
* #1126

iljakuklic (approved):

crates/agglayer-contracts/src/lib.rs
```diff
@@ -1,13 +1,17 @@
 //! Agglayer smart-contract bindings.
 
-use std::sync::Arc;
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
```
polykeith, line 5: I don't exactly remember whether `std::sync::RwLock` now implements the fairness rule for readers and writers -- previously, only `parking_lot::RwLock` would make sure that the writer won't get starved due to multiple readers constantly acquiring the lock.

atanmarko: We have one writer and one reader. In the current implementation, they would never compete.

iljakuklic: > I don't exactly remember whether `std::sync::RwLock` now implements the fairness rule for readers and writers -- previously, only `parking_lot::RwLock` would make sure that the writer won't get starved due to multiple readers constantly acquiring the lock.

Yea, std or parking_lot both fine in my book. I don't think these locks will be particularly congested.

Edit: IIRC, one extra advantage of parking_lot is that you don't have to deal with the poisoning errors.

PR #1127, opened 2025-10-16
fix: filter event range

### LINKED ISSUE #1126: Tighten the L1 log retrieval block bounds
In certain scenarios, the L1 provider is asked to scan logs. If the searched for item is not found, the stopping point is currently once the whole blockchain has been scanned.

This can be tightened, although the payoff may be dubious, so better to examine on case by case basis.

For example, when the L1 info root is queried. The issue is largely mitigated by caching the results locally. A tighter bound can be achieved by stopping at the point of the `L1InfoTree` deployment. Alternatively, the existing entries in the cache could be used to tighten the bound if they also store block height for given leaf count.

Relevant discussion:

* https://github.com/agglayer/agglayer/pull/1125#discussion_r2434903362
* https://github.com/agglayer/agglayer/pull/1125#discussion_r2435383108
* https://github.com/agglayer/agglayer/pull/1125#discussion_r2435390978

atanmarko:
Fix the log search.
Fixes #1126

### DIFF
--- a/crates/agglayer-config/src/l1.rs
+++ b/crates/agglayer-config/src/l1.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{num::NonZeroU64, time::Duration};
 
 use agglayer_primitives::Address;
 use serde::{Deserialize, Serialize};
@@ -29,7 +29,7 @@ pub struct L1 {
     pub rpc_timeout: Duration,
 
     #[serde(default = "L1::default_event_filter_block_range")]
-    pub event_filter_block_range: u64,
+    pub event_filter_block_range: NonZeroU64,
 }
 
 impl L1 {
@@ -37,8 +37,8 @@ impl L1 {
         Duration::from_secs(45)
     }
 
-    const fn default_event_filter_block_range() -> u64 {
-        10000
+    const fn default_event_filter_block_range() -> NonZeroU64 {
+        NonZeroU64::new(10000).unwrap()
     }
 }
 
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -209,7 +209,7 @@ impl Node {
                         ceiling: gas_config.ceiling,
                     }
                 },
-                config.l1.event_filter_block_range,
+                config.l1.event_filter_block_range.get(),
             )
             .await?,
         );


crates/agglayer-contracts/src/lib.rs
```diff
@@ -77,6 +83,12 @@ pub struct L1RpcClient<RpcProvider> {
     /// Cached UpdateL1InfoTreeV2 first l1_info_root for each leaf count.
     /// Map<leaf_count, l1_info_root>
     l1_info_roots: Arc<RwLock<HashMap<u32, [u8; 32]>>>,
+    /// Number of blocks to query when filtering for events.
+    /// This is to avoid hitting provider limits when querying large block
+    /// ranges or errors like "query returned more than 10000 results".
+    event_filter_block_range: u64,
+    /// PolygonZkEVMGlobalExitRootV2 contract deployment block number.
+    polygon_zkevm_global_exit_root_v2_contract_block: Option<u64>,
```
iljakuklic: Why option? We always have to start *somewhere*, so just plugging in 0 in tests seems fine to me.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -145,35 +157,146 @@ pub enum L1RpcError {
...
+                ))
+            })?;
+
+        if code_at_mid.is_empty() {
+            // Not deployed yet at `mid` -> search higher
+            debug!("No code at block {}, searching higher", mid);
+            lo = mid.saturating_add(1);
+        } else {
+            // Code exists at `mid` -> remember and search lower
+            debug!("Code found at block {}, searching lower", mid);
```
iljakuklic: These two `debug!`s could imho be `trace!`

crates/agglayer-contracts/src/lib.rs
```diff
@@ -145,35 +157,146 @@ pub enum L1RpcError {
...
+
+        debug!(
+            "Binary search iteration {}: checking block {} (range: {} to {})",
+            iterations, mid, lo, hi
+        );
+
+        // Query bytecode at specific block height
+        let code_at_mid = rpc
+            .get_code_at(address)
+            .block_id(BlockId::Number(BlockNumberOrTag::Number(mid)))
```
iljakuklic: Does this work?

```suggestion
            .number(mid)
```

crates/agglayer-contracts/src/lib.rs
```diff
@@ -145,35 +157,146 @@ pub enum L1RpcError {
...
+        }
+        None => {
+            debug!(
+                "Contract deployment block not found for address: {}",
+                address
+            );
+            Ok(None)
+        }
+    }
+}
```
iljakuklic: A bit more concise:

```suggestion
    answer
        .as_ref()  // may or may not work without this
        .map_or_else(
            || debug!(?address, Contract deployment block not found"),
            |block| debug!("Contract deployment found at block: {block}")),
        );

    Ok(answer)
```

crates/agglayer-contracts/src/lib.rs
```diff
@@ -182,11 +305,31 @@ where
...
+        )
+        .await
+        .inspect_err(|error| {
+            error!(
+                ?error,
+                "Failed to find contract {} deployment block",
+                polygon_zkevm_global_exit_root_v2_contract
+            );
+        })
+        .unwrap_or(None);
```
iljakuklic: Have you considered making this a hard error? Seems to me that if the contract address is not found, we won't be able to do much else anyway.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -145,35 +157,146 @@ pub enum L1RpcError {
...
+        "Contract has code at latest block, starting binary search from block 0 to {}",
+        hi
+    );
+
+    // Binary search for the first block where code is present
+    let mut answer: Option<u64> = None;
+    let mut iterations = 0;
+    const MAX_ITERATIONS: u32 = 256; // Safety limit to prevent infinite loops
+
+    while lo <= hi && iterations < MAX_ITERATIONS {
```
iljakuklic: Why keep iterating if `lo == hi`?

Also, I think the `answer` variable can be removed and just return `lo` at the end.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -145,35 +157,146 @@ pub enum L1RpcError {
...
+
+    debug!(
+        "Contract has code at latest block, starting binary search from block 0 to {}",
+        hi
+    );
+
+    // Binary search for the first block where code is present
+    let mut answer: Option<u64> = None;
+    let mut iterations = 0;
+    const MAX_ITERATIONS: u32 = 256; // Safety limit to prevent infinite loops
```
iljakuklic: We should be able to get the exact block within 25 iterations. Even if limited to just 22, it should get us within some 6 or so blocks of the real answer which is probably good enough.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -145,35 +157,146 @@ pub enum L1RpcError {
...
+            })?;
+
+        if code_at_mid.is_empty() {
+            // Not deployed yet at `mid` -> search higher
+            debug!("No code at block {}, searching higher", mid);
+            lo = mid.saturating_add(1);
+        } else {
+            // Code exists at `mid` -> remember and search lower
+            debug!("Code found at block {}, searching lower", mid);
+            answer = Some(mid);
```
iljakuklic: I don't think this is correct if the loop exits early. To get a safe conservative starting point for log scanning, the answer should be set at point where the code does not yet exist, so in the other branch of the if-else to the updated `lo`. Would be good to maintain the invariant that `answer` holds a correct lower bound at all times.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -145,35 +157,146 @@ pub enum L1RpcError {
...
+    debug!("Finding contract deployment block for address: {address}, latest_block {latest_block}");
+
+    let mut hi = latest_block;
+    let mut lo = 0u64;
+
+    // Quick exit: if no code at latest block, contract was never deployed
+    // (Note: this will treat self-destructed contracts as "not found")
+    let latest_code = rpc
+        .get_code_at(address)
+        .block_id(BlockId::Number(BlockNumberOrTag::Number(hi)))
```
iljakuklic: Again, this should be equivalent (not tested):
```suggestion
        .number(hi)
```

iljakuklic (commented):
First pass. Just a bunch of nits.

crates/agglayer-contracts/src/lib.rs
atanmarko: Do you have an example for the scenario where this would not work - where it would not return the exact block where the contract is deployed?

crates/agglayer-contracts/src/lib.rs
atanmarko: Then sometimes it returns the block with no code. And the name of the function is `find_contract_deployment_block_number` so it should return correct block where contract is deployed. We may use this information for other purposes. 
For example if lo<hi and return lo:

```
=== Test Case 1: Deployment at block 5 (out of 0-10) ===
Searching for deployment block (actual: 5)
Initial: lo=0, hi=10
Iter 1: mid=5, code=true, lo=0, hi=10
Iter 2: mid=2, code=false, lo=0, hi=4
Iter 3: mid=3, code=false, lo=3, hi=4
Final: lo=4, hi=4, result=4
Returned block 4 has code: false  ← ❌ NO CODE!
Block before (3) has code: false
```

atanmarko: I have remove this function.

crates/agglayer-contracts/src/lib.rs
atanmarko: I have removed this function, querying from `Earliest`

iljakuklic: Not really, only in rather synthetic scenarios. E.g. when someone decides to limit the iteration count to some smaller number (say 18), where the loop exits prematurely and there still is a gap between lo and hi. The contract is deployed at some point x, where lo <= x <= mid, and mid is returned. When scanning the logs starting from mid, the logs between contract deployment x and mid will be missed.

So more a matter of robustness in terms of parameter tuning and code consistency. The more invariants can you spell out, the easier it is to reason about the code in general. (In fact, I initially assumed it to be true that the answer always contains either `None` or a conservative bound for start of log scanning.) But I might have missed a different corner case...

crates/agglayer-contracts/src/lib.rs
iljakuklic: > Then sometimes it returns the block with no code.

You are right. My thinking was to return a lower bound that is safe to start scanning logs from, even if the loop exits early. See the other answer. It's correct that the function name is no longer accurate in this case.

crates/agglayer-contracts/src/rollup.rs
```diff
@@ -84,36 +86,20 @@ where
...
+        // Get first `UpdateL1InfoTreeV2` event for the given leaf count.
+        // l1 leaf count increases over time for a network, when we filter by
+        // `l1_leaf_count` we would not get provider limit.
+        debug!("Searching for UpdateL1InfoTreeV2 event with leaf count {l1_leaf_count}");
+        let filter = Filter::new()
+            .address(self.global_exit_root_manager_contract)
+            .event_signature(UpdateL1InfoTreeV2::SIGNATURE_HASH)
+            .topic1(U256::from(l1_leaf_count))
+            .from_block(BlockNumberOrTag::Earliest)
+            .to_block(BlockNumberOrTag::Latest);
```
polykeith: So let me see if I understand this correctly -- instead of looping, we simply search from the block that's tagged "earliest" to the block that's tagged "latest"?

atanmarko: Right. If you filter by `l1_leaf_count`, very few such logs  exist so will not hit any limit.

atanmarko: We only need to loop for `VerifyPessimisticStateTransition`

crates/agglayer-contracts/src/lib.rs
```diff
@@ -149,31 +150,35 @@ impl<RpcProvider> L1RpcClient<RpcProvider>
...
         Self {
             rpc,
             inner,
-            l1_info_tree,
+            global_exit_root_manager_contract,
             default_l1_info_tree_entry,
             gas_multiplier_factor,
             gas_price_params,
             l1_info_roots: Arc::new(RwLock::new(HashMap::new())),
+            event_filter_block_range,
```
iljakuklic, line 171: Low priority follow up suggestion: Would be nice to query the provider for max logs per request somehow, instead of fiddling with the value manually. Ideally not by parsing the error message 😄. I have no idea if the API exposes anything like that.

crates/agglayer-contracts/src/lib.rs
```diff
@@ -303,19 +303,22 @@ mod tests {
     impl ContractSetup {
         pub fn new() -> Self {
             Self {
-                rollup_manager: "0x32d33D5137a7cFFb54c5Bf8371172bcEc5f310ff" // bali: 0xe2ef6215adc132df6913c8dd16487abf118d1764
+                rollup_manager: "0xE2EF6215aDc132Df6913C8DD16487aBF118d1764" // Bali
                     .parse()
                     .unwrap(),
-                ger_contract: "0xAd1490c248c5d3CbAE399Fd529b79B42984277DF" // bali: 0x2968d6d736178f8fe7393cc33c87f29d9c287e78
+                ger_contract: "0x2968D6d736178f8FE7393CC33C87f29D9C287e78" // Bali
```
iljakuklic, line 309: Was this done for testing or should it stay?

crates/agglayer-contracts/src/lib.rs
```diff
@@ -405,4 +443,47 @@ mod tests {
...
+            &rpc_url,
+            prover_alloy::DEFAULT_HTTP_RPC_NODE_INITIAL_BACKOFF_MS,
+            prover_alloy::DEFAULT_HTTP_RPC_NODE_BACKOFF_MAX_RETRIES,
+        )
+        .expect("valid alloy provider");
+
+        tracing::info!("Test fetching of the InitL1InfoRootMap for Bali testnet");
+
+        // Use the specified contract addresses for Bali testnet
+        let rollup_manager_address: Address = "0xE2EF6215aDc132Df6913C8DD16487aBF118d1764"
```
iljakuklic: This string is also in the `ContractSetup` constructor above. Maybe these could be turned into named constants. You can use `agglayer_primitives::address!` or `alloy_primitives::address!` macros (depending on which `Address` is used here) to get a `const` address (without invoking `parse`). The same addresses are also used in `settlement_client::tests`.

crates/agglayer-contracts/src/lib.rs
atanmarko: I haven't heard about that api. AFAIK 10k is the limit with most of the providers.

crates/agglayer-contracts/src/lib.rs
atanmarko: This is Bali public GER address, it should stay. The upper one is Cardona.

iljakuklic (approved):

crates/agglayer-config/src/l1.rs
```diff
@@ -17,20 +17,29 @@ pub struct L1 {
...
+
     #[serde(alias = "PolygonZkEVMGlobalExitRootV2Contract")]
     pub polygon_zkevm_global_exit_root_v2_contract: Address,
 
     #[serde(default = "L1::default_rpc_timeout")]
     #[serde(with = "crate::with::HumanDuration")]
     pub rpc_timeout: Duration,
+
+    #[serde(default = "L1::default_event_filter_block_range")]
+    pub event_filter_block_range: u64,
```
Freyskeyd: Can it be a `NonZeroU64` ? So that we're sure it's not zero, and doesn't panic if we `event_filter_block_range - 1` ?

crates/agglayer-contracts/src/rollup.rs
```diff
@@ -84,36 +86,19 @@ where
...
-        }
+        // Get first `UpdateL1InfoTreeV2` event for the given leaf count.
+        // l1 leaf count increases over time for a network, when we filter by
+        // `l1_leaf_count` we would not get provider limit.
+        debug!("Searching for UpdateL1InfoTreeV2 event with leaf count {l1_leaf_count}");
+        let filter = Filter::new()
+            .address(self.global_exit_root_manager_contract)
+            .event_signature(UpdateL1InfoTreeV2::SIGNATURE_HASH)
+            .topic1(U256::from(l1_leaf_count))
+            .from_block(BlockNumberOrTag::Earliest);
```
Freyskeyd, line 97: Earliest may not return anything, right?

atanmarko: I only met such issue if there were more than 10k events. This was never problem on Bali/Cardona or anywhere where have have tested this event (as there is only one or maybe few `UpdateL1InfoTreeV2` events with particular `l1_leaf_count`)

crates/agglayer-contracts/src/lib.rs
```diff
@@ -149,31 +150,35 @@ impl<RpcProvider> L1RpcClient<RpcProvider>
 where
     RpcProvider: alloy::providers::Provider + Clone + 'static,
 {
+    #[allow(clippy::too_many_arguments)]
```
Ekleog-Polygon, line 153: Nit: would be nice to turn into a builder pattern in a follow-up PR

crates/agglayer-contracts/src/lib.rs
Ekleog-Polygon: Can we keep the cardona addresses nearby, commented out, so we don't have to search for it whenever we'll need it next?

atanmarko: @Ekleog-Polygon You can always fetch it here - https://dashboard.agglayer.dev/

PR #1130, opened 2025-10-21
fix: count the first block in which a tx is included as a confirmation

Ekleog-Polygon:


crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -363,7 +363,9 @@ where
                         for confirmation_attempt in attempt..=self.config.max_retries {
                             match self.l1_rpc.get_provider().get_block_number().await {
                                 Ok(current_block) => {
-                                    let confirmations = current_block.saturating_sub(receipt_block);
+                                    let confirmations = current_block
+                                        .saturating_sub(receipt_block)
+                                        .saturating_add(1);
```
polykeith, line 368: This probably will never happen, but if the subtraction resulted in 0, then adding 1 to it doesn't seem to make sense, right?

atanmarko: @Ekleog-Polygon  is right here, block where the transaction is mined to the chain is actually a first confirmation - it does count.

atanmarko (approved):

hadjiszs (approved):

Issue #1122, opened 2025-10-13
Update/implement integration test with various settlement tx fail scenarious


atanmarko:
Review/update tests in https://github.com/agglayer/agglayer/blob/main/tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs

PR #1137, opened 2025-10-24
feat: add more integration tests

atanmarko:
Resolves https://github.com/agglayer/agglayer/issues/1122

### DIFF
--- a/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
@@ -215,7 +215,8 @@ where
                     // Some error happened while checking the tx receipt on L1
                     warn!(
                         ?error,
-                        "Failed to check settlement tx {previous_tx_hash} existence on L1"
+                        settlement_tx_hash = %previous_tx_hash,
+                        "Failed to check settlement tx prior existence on L1",
                     );
                 });
             match result_is_settlement_tx_mined {
@@ -232,8 +233,8 @@ where
 
         if settlement_tx_hash_missing_on_l1 {
             warn!(
-                "Previous settlement tx hash is missing on L1, tx {:?}",
-                self.header.settlement_tx_hash
+                settlement_tx_hash = ?self.header.settlement_tx_hash,
+                "Previous settlement tx hash is missing on L1",
             );
 
             // If the settlement tx is not found on L1, we need to recover.
@@ -262,9 +263,9 @@ where
                         Ok((_, _, recomputed_output)) => {
                             if contract_pp_root == recomputed_output.new_pessimistic_root {
                                 info!(
+                                    %contract_settlement_tx_hash,
                                     "Certificate new pp root matches the latest settled pp root \
-                                     on L1, updating certificate settlement tx hash to \
-                                     {contract_settlement_tx_hash:?}"
+                                     on L1, updating certificate settlement tx hash to the one in contracts"
                                 );
                                 self.header.settlement_tx_hash = Some(contract_settlement_tx_hash);
                                 if let Err(error) = self.state_store.update_settlement_tx_hash(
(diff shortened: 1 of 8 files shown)


crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -497,6 +498,9 @@ where
             "Submitted certificate for settlement"
         );
 
+        #[cfg(feature = "testutils")]
+        testutils::inject_fail_points(&certificate_id, &mut self.header, &self.state_store);
```
Ekleog-Polygon: Nit: can we rename the function eg. `inject_fail_points_after_proving`, so that it's more obvious where it's used?

tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs
```diff
@@ -127,18 +132,25 @@ async fn transaction_without_receipt_status(#[case] state: Forest) {
 #[tokio::test]
 #[timeout(Duration::from_secs(180))]
 #[case::type_0_ecdsa(crate::common::type_0_ecdsa_forest())]
-async fn transaction_fails_due_to_out_of_gas(#[case] state: Forest) {
```
Ekleog-Polygon, line 130: Hmm… IIUC you're removing the "out of gas" integration test. Did it not bring any value?

tests/integrations/tests/certificate_settlement/retries/recovery.rs
```diff
@@ -16,18 +16,16 @@ use rstest::rstest;
 use tokio_util::sync::CancellationToken;
 
 #[rstest]
-// TODO: re-enable this test, once we handle in-flight in-error certificates properly.
-// #[case::settlement_type_0_ecdsa(
-//     &["notifier::packer::settle_certificate::transaction_sent::kill_node"],
-//     crate::common::type_0_ecdsa_forest()
-// )]
```
Ekleog-Polygon, line 23: Why remove this comment? In particular if it's supposed to be fixed now (I actually think so), hopefully just uncommenting it should just work(tm)?

tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs
atanmarko: We are already testing this case with `transaction_with_receipt_status_0_retry` and `retry_on_error` - it was duplicated test.

tests/integrations/tests/certificate_settlement/retries/recovery.rs
atanmarko: @Ekleog-Polygon It is deprecated. That fail point does not exist anymore.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -615,3 +623,32 @@ fn recv_err(_: oneshot::error::RecvError) -> CertificateStatusError {
...
+        fail::eval(
+            "certificate_task::process_impl::invalid_settlement_tx_hash",
+            |_| {
+                // Write an unexistent tx hash to simulate the settlement tx not being found on
+                // L1
+                warn!("FAIL POINT ACTIVE: Injecting invalid settlement tx hash");
+                let unexistent_tx_hash = SettlementTxHash::new(Digest::from([21u8; 32]));
+                header.settlement_tx_hash = Some(unexistent_tx_hash);
+                let _ =
+                    state_store.update_settlement_tx_hash(certificate_id, unexistent_tx_hash, true);
```
hadjiszs: not so familiar with failpoint but isn't the error silent here?
if there is a failure on writing the unexistent tx hash I guess it has an impact on the test relying on this fail eval no?

atanmarko: It does. I'll unwrap here.

iljakuklic (approved):

PR #1138, opened 2025-10-25
feat: a dedicated db column family for tracking the history of settlement tx hashes

iljakuklic:


crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -225,7 +222,7 @@ where
 
         // Get the transaction hash from the pending transaction
         let tx_hash = *pending_tx.tx_hash();
-        info!("Settlement transaction hash: {}", tx_hash);
+        info!(?tx_hash, "Settlement transaction hash");
```
atanmarko: I think we should keep the `tx_hash` also in the message. It is much easier to follow in datadog.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
```diff
@@ -236,14 +233,15 @@ impl<StateStore, PendingStore, PerEpochStore, RollupManagerRpc>
...
+    StateStore: StateWriter,
 {
     #[tracing::instrument(skip(self), fields(%settlement_tx_hash, %certificate_id))]
     async fn wait_for_settlement(
         &self,
         settlement_tx_hash: SettlementTxHash,
         certificate_id: CertificateId,
     ) -> Result<(EpochNumber, CertificateIndex), Error> {
-        info!(%settlement_tx_hash, "Waiting for settlement of tx {settlement_tx_hash}");
+        info!("Waiting for settlement");
```
atanmarko: Why removing settlement tx hash from the message? It was very useful for inspections of the logs.

atanmarko: Maybe you could unify and rename all `settlement_tx_hash` occurrences to `tx_hash` - it would be easier to filter.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -286,11 +303,11 @@ where
                                 Some(contract_settlement_tx_hash.into())
                             } else {
                                 warn!(
-                                    certificate_settlement_tx_hash = ?self.header.settlement_tx_hash,
                                     certificate_pp_root = %recomputed_output.new_pessimistic_root,
                                     %contract_settlement_tx_hash,
                                     %contract_pp_root,
-                                    "Certificate pp root does not match the latest settled pp root on L1 contract, moving certificate back to Proven",
+                                    "Certificate pp root does not match the latest settled pp root \
+                                     on L1 contract, moving certificate back to Proven",
```
atanmarko: AFAIK we are not moving it explicitly back to Proven anymore.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -321,22 +338,28 @@ where
             if recomputed_from_contract.is_none() {
                 // Tx not found on L1, and pp root from contract not matching,
                 // Make the cert InError and wait for aggkit to resubmit it.
-                return Err(CertificateStatusError::SettlementError(format!(
-                    "Settlement tx {:?} not found on L1, moving certificate back to Proven",
-                    self.header.settlement_tx_hash
-                )));
+                return Err(CertificateStatusError::SettlementError(
+                    "Settlement tx not found on L1, moving certificate back to Proven".to_string(),
```
atanmarko:
```suggestion
                    "Settlement tx not found on L1 during recompute state".to_string(),
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -411,22 +434,27 @@ where
...
+            .get_settlement_tx_hashes_for_certificate(certificate_id)?
+            .into_vec();
+
+        if previous_tx_hashes.len() > MAX_TX_RETRY {
             error!(
-                previous_tx_hashes = ?self.previous_tx_hashes,
+                ?previous_tx_hashes,
                 max_retries = MAX_TX_RETRY,
-                "More settlement transactions submitted for the same certificate than allowed retries, something is wrong"
+                "More than 5 different settlement transactions submitted for the same \
```
atanmarko:
```suggestion
                "More than {MAX_TX_RETRY} different settlement transactions submitted for the same \
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -503,69 +525,94 @@ where
...
         let settled_certificate =
             SettledCertificate(certificate_id, height, epoch_number, certificate_index);
+        tracing::trace!(
+            header = ?self
+                .state_store
+                .get_certificate_header(&certificate_id)
+                .map(|h| h.map(|h| h.status)),
+            "Header before settlement info write",
+        );
         self.set_status(CertificateStatus::Settled)?;
```
atanmarko, line 617: Shouldn't we `update_settlement_tx_hash` in the state store with the correct settlement transaction once we know certificate is settled?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -633,7 +700,7 @@ mod testutils {
                 let unexistent_tx_hash = SettlementTxHash::new(Digest::from([21u8; 32]));
                 header.settlement_tx_hash = Some(unexistent_tx_hash);
                 state_store
-                    .update_settlement_tx_hash(certificate_id, unexistent_tx_hash, true)
+                    .update_settlement_tx_hash(certificate_id, unexistent_tx_hash)
```
atanmarko: Is this fail point correct now with this change or needs to be refactored?

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -305,21 +305,23 @@ where
                 }
                 Some(NewCertificate { certificate_id, height, .. }) = self.certificate_stream.recv(), if !self.at_capacity_for_epoch => {
                     info!(
-                        hash = certificate_id.to_string(),
-                        "Received a certificate event for {certificate_id} at height {height}"
+                        %certificate_id, %height,
+                        "Received a certificate event"
```
atanmarko: You could probably add `certificate_id` to the message.

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
```diff
@@ -345,23 +359,32 @@ async fn pending_certificate_in_error_with_settlement_tx_hash_force_set_status()
...
         .state_store
-        .insert_certificate_header(&pending_certificate, CertificateStatus::Pending)
+        .insert_certificate_header(&pending_certificate, initial_status.clone())
         .expect("unable to insert pending certificate header");
 
     let fake_settlement_tx_hash = SettlementTxHash::from(Digest::from([1; 32]));
+
     context
-        .state_store
-        .update_settlement_tx_hash(&certificate_id, fake_settlement_tx_hash, false)
```
atanmarko: Are we managing properly settlement tx hash in the header, in the state store for settled certificate?

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
iljakuklic: > Why removing settlement tx hash from the message? It was very useful for inspections of the logs.

It is still added as a structured log field by the `#[instrument]` annotation on the function. I thought the trend was to remove duplicate information from messages when that information is already provided in the structured format.

Having said that, I am happy to revert these changes as they are not super relevant to the substance of this PR.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
iljakuklic: > Shouldn't we `update_settlement_tx_hash` in the state store with the correct settlement transaction once we know certificate is settled?

It is now done in `add_certificate` storage method. There are some sanity checks in `update_settlement_tx_hash` preventing you from setting it for already settled certificates, which I find to be useful. However, with the checks in place, there is no good place to call the hash update here.

crates/agglayer-jsonrpc-api/src/tests/send_certificate.rs
iljakuklic: > Are we managing properly settlement tx hash in the header, in the state store for settled certificate?

It is now only set when the certificate is determined to be settled on L1, i.e. when the certificate moves into the `Settled` status. The settlement tx hashes of transactions constructed as candidates for settlement transactions are tracked separately. In this test, the certificate does not reach the `Settled` status.

crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
atanmarko: We should for sure add it to the spans, but not remove it from the messages. It is much easier to track logs this way in my experience. Alternative to have 10 `Waiting for settlement`and to open every one to see the details and the related hash is much less productive and harder to debug.

atanmarko: So spans are for search and filtering, messages are to easily track whole log. Both should have enough info from my recent experience.

iljakuklic: > So spans are for search and filtering, messages are to easily track whole log. Both should have enough info from my recent experience.

Ok, re-added the certificate and transaction IDs where appropriate.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
iljakuklic: > Is this fail point correct now with this change or needs to be refactored?

Good observation. I changed it to use `insert_settlement_tx_hash_for_certificate`. It now tests something slightly different: the ability to recover from history if the most recent tx hash does not work. But the new semantics make sense to me.

atanmarko (dismissed):

atanmarko (dismissed):

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -196,47 +193,67 @@ where
...
+                    | Ok(crate::TxReceiptStatus::TxFailed) => {
+                        false // We have fetched the receipt, tx exists on L1
+                    }
+                    Ok(crate::TxReceiptStatus::NotFound) => true, // Tx not found on L1
+                    Err(_error) => false,                         // On error we do nothing
+                };
+                if !missing {
+                    debug!(%previous_tx_hash, "Found previous settlement tx hash");
+                    previous_settlement_tx_hash = Some(*previous_tx_hash);
+                    break;
```
Ekleog-Polygon, line 243: What should we do if there are two previous settlement txes, one failed and one passed? I'm wondering whether this `break` will not mistakenly assume that settlement failed when it actually succeeded with another tx.

That being said, TBH I'm not sure it's worth answering that question this precisely considering we'll likely refactor all of our settlement logic for the next version anyway, and I'm not at all sure our current code handles it any better than this new code, but… :shrug:

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -267,16 +284,16 @@ where
...
                                 );
-                                self.header.settlement_tx_hash = Some(contract_settlement_tx_hash);
-                                if let Err(error) = self.state_store.update_settlement_tx_hash(
-                                    &certificate_id,
-                                    contract_settlement_tx_hash,
-                                    true,
-                                ) {
+                                let insert_result = self
+                                    .pending_store
+                                    .insert_settlement_tx_hash_for_certificate(
```
Ekleog-Polygon, line 289: Drive-by thought: this feels like it should belong in state store and not pending store, because if we lose it we may lose data that's essential to agglayer… but then maybe we can just forget about it for now until we finally [unify pending and state databases](https://github.com/agglayer/protocol-research/issues/172)

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -466,22 +502,15 @@ where
         #[cfg(feature = "testutils")]
         fail::fail_point!("certificate_task::process_impl::about_to_record_candidate");
 
-        self.header.settlement_tx_hash = Some(settlement_tx_hash);
```
Ekleog-Polygon, line 473: Can we do something about `settlement_tx_hash` still being in the `CertificateHeader` struct? AFAIU from my first pass it's basically no longer used anyway, but can we eg. rename it to `_deprecated_settlement_tx_hash` so that it still serde-deserializes fine, without us risking using it?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -503,74 +532,92 @@ where
...
+                let err_message = String::from("No previous settlement tx matches");
+                error!(?processed_tx_hashes, ?certificate_id, "{err_message}");
+                CertificateStatusError::SettlementError(err_message)
+            })?;
+
+            debug!(
+                "Pre check prev txhash {settlement_tx_hash}: {:?}",
+                self.state_store
+                    .get_certificate_header(&certificate_id)
+                    .map(|h| h.map(|h| h.status))
```
Ekleog-Polygon: Nit: add settlement_tx_hash as a key-value tag for easier filtering. Also, I'm not sure I understand why you're retrieving the status of the same cert id at the beginning of each loop cycle?

crates/agglayer-certificate-orchestrator/src/network_task/tests.rs
```diff
@@ -131,12 +155,6 @@ async fn start_from_zero() {
             }))
         });
 
-    state
-        .expect_update_settlement_tx_hash()
```
Ekleog-Polygon, line 135: Nit: The order is confusing, it might be better to alternate between state. and pending. here. Basically, when reading this I was thinking that we had a missing insert_settlement_tx_hash… but it's actually wrong, it was just quite a bit above. Though all this will disappear when we merge pending and state, so it may not be important :shrug:

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -63,6 +60,8 @@ pub const PER_EPOCH_START_CHECKPOINT_CF: &str = "per_epoch_start_checkpoint_cf";
 // Pending related CFs
 pub const PENDING_QUEUE_CF: &str = "pending_queue_cf";
 pub const PROOF_PER_CERTIFICATE_CF: &str = "proof_per_certificate_cf";
+pub const SETTLEMENT_TX_HASHSES_PER_CERTIFICATE_CF: &str =
```
Ekleog-Polygon:
```suggestion
pub const SETTLEMENT_TX_HASHES_PER_CERTIFICATE_CF: &str =
```

proto/agglayer/storage/v0/settlement_tx_record.proto
```diff
@@ -0,0 +1,19 @@
...
+  // Optional to differentiate between empty history and missing history,
+  // which allows for easier field deprecation in the future.
+  TxHashHistory hashes = 1;
+}
+
+// Transaction hash history wrapper.
+message TxHashHistory {
+  // Settlement transaction hashes, uniqued and in order of insertion.
+  // Each hash represents a settlement transaction associated with a certificate.
+  repeated bytes hashes = 1;
```
Ekleog-Polygon: Can we add one layer of indirection here, by storing `SettlementTxHash` rather than plain `bytes`? This way if we ever have to deal with other kinds of settlement tx hashes (eg. txes on another L1), then we will be able to add a variant at the right place.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
iljakuklic: > Can we do something about `settlement_tx_hash` still being in the `CertificateHeader` struct? AFAIU from my first pass it's basically no longer used anyway, but can we eg. rename it to `_deprecated_settlement_tx_hash` so that it still serde-deserializes fine, without us risking using it?

It is used but only for settled certificates.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
iljakuklic: > Drive-by thought: this feels like it should belong in state store and not pending store, because if we lose it we may lose data that's essential to agglayer… but then maybe we can just forget about it for now until we finally [unify pending and state databases](https://github.com/agglayer/protocol-research/issues/172)

Tx hashes for settled certificates are still recorded in the header. Pending store keeps the history of transaction hashes that have been submitted. That includes ones that failed on L1 or where the L1 status is transiently unknown.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
iljakuklic: > What should we do if there are two previous settlement txes, one failed and one passed? I'm wondering whether this `break` will not mistakenly assume that settlement failed when it actually succeeded with another tx.

I believe network task already detects this case and returns the tx id of the transaction that actually settled.

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -424,37 +439,66 @@ where
...
                             Error::internal("Unable to update certificate status")
                         })?;
                 }
-                Operation::SetSettlementTxHash {
-                    from: _,
-                    to: Some(to),
-                } => {
-                    self.state
-                        .update_settlement_tx_hash(
+                Operation::SetSettlementTxHash { from, to } => {
```
Ekleog-Polygon, line 476: I think this needs some more thought. AFAIU, this replaces one of the pending settlement tx hashes, removing `from` and adding `to`.

I guess the first question would be, is this call actually still useful now that we remember all settlement tx hashes? Can we just remove it? I'm thinking most of the use case it served is basically moot now. The only one I can think of would be to just remove the settlement tx hashes, while manually reverting a cert to Proven for example.

And if not, would it make sense to eg. change it to eg. `--edit-settlement-tx-hashes,remove=0xabc,remove=0xdef,add=0x123`, without forcing minus-one plus-one?

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -424,37 +439,66 @@ where
         for operation in operations {
             match operation {
                 Operation::SetStatus { from: _, to } => {
+                    if to == CertificateStatus::Settled {
```
Ekleog-Polygon, line 442: I think we need to design moving a cert to Settled some more, considering that I'm not sure the cert would properly end up in the right epoch db.

This being said, this change does not seem like a regression to me; it's just that we may have to replace it with a simple `return Err("this is not an allowed operation")`. And if not we should probably let the admin pick which settlement tx hash to use; but I guess that's what you mean by your TODO below.

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -424,37 +439,66 @@ where
...
+                            .copied()
+                            .ok_or_else(|| {
+                                error!("Cannot move to settled without settlement tx");
+                                Error::internal("Cannot move to settled without settlement tx")
+                            })?;
+                        self.state
+                            .update_settlement_tx_hash(
+                                &certificate_id,
+                                settlement_tx_hash,
+                                agglayer_storage::stores::UpdateEvenIfAlreadyPresent::Yes,
```
Ekleog-Polygon, line 462: Should this be a yes? We're picking at random here, so it feels like if there's already one we should leave it there?

Ekleog-Polygon (approved):
Overall LGTM! My remaining comments from the last review would basically only apply to code we'll rewrite with the settlement refactor anyway, so the changes I'm not sure about will most likely never reach production.

I think the only remaining concern I have would be the Admin RPC changes; but they don't look like regressions anyway so we can probably land this and iterate on it afterwards :smile:

crates/agglayer-jsonrpc-api/src/admin.rs
iljakuklic: > And if not, would it make sense to eg. change it to eg. `--edit-settlement-tx-hashes,remove=0xabc,remove=0xdef,add=0x123`, without forcing minus-one plus-one?

Yea this would make sense to me. I just did minimal changes that seem sort of right without changing the admin interface.

PR #1141, opened 2025-10-27
feat: moving agglayer-prover to agglayer repo

Freyskeyd:
Previous commit before removing `agglayer-prover` from provers is https://github.com/agglayer/provers/tree/5022963f8b5b8e586e771a8d29113310499f82fc

crates sum:

- `agglayer-prover: 5370654fb37230aba1ff2172c910e912`
- `agglayer-prover-config: 60bab377fc3e3b05f97327925ab75305`
- `agglayer-prover-types: 8ea57124c0a5cf9e6ae29653d794ddf7`
- `proto files: c2e27a41e817afc3fc60d334f7080914`

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/tests.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/tests.rs
@@ -260,7 +260,7 @@ mockall::mock! {
     impl agglayer_contracts::L1TransactionFetcher for L1Rpc {
         type Provider = alloy::providers::RootProvider<Ethereum>;
 
-        async fn fetch_transaction_receipt(&self, tx_hash: FixedBytes<32>) -> Result<TransactionReceipt, L1RpcError>;
+        async fn fetch_transaction_receipt(&self, tx_hash: FixedBytes<32>) -> Result<Option<TransactionReceipt>, L1RpcError>;
 
         fn get_provider(&self) -> &<Self as agglayer_contracts::L1TransactionFetcher>::Provider;
     }
--- a/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
+++ b/crates/agglayer-aggregator-notifier/src/settlement_client/rpc.rs
@@ -1,6 +1,6 @@
 use std::{sync::Arc, time::Duration};
 
-use agglayer_certificate_orchestrator::{Error, NonceInfo, SettlementClient};
+use agglayer_certificate_orchestrator::{Error, NonceInfo, SettlementClient, TxReceiptStatus};
 use agglayer_config::outbound::OutboundRpcSettleConfig;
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -54,12 +54,15 @@ agglayer-tries = "0.11.0"
...
 # Provers
-agglayer-prover = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.1" }
-agglayer-prover-config = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.1" }
-agglayer-prover-types = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.1" }
-prover-alloy = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.1" }
-prover-config = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.1" }
-prover-executor = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.1" }
+agglayer-prover = { path = "crates/agglayer-prover" }
+agglayer-prover-config = { path = "crates/agglayer-prover-config" }
+agglayer-prover-types = { path = "crates/agglayer-prover-types" }
```
Ekleog-Polygon, line 59: Nit: move these three so they get with the rest of the path-dependencies above?

Cargo.toml
```diff
@@ -54,12 +54,15 @@ agglayer-tries = "0.11.0"
...
+prover-alloy = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.2" }
+prover-config = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.2" }
+prover-logger = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.2" }
+prover-executor = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.2" }
+prover-engine = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.2" }
+prover-utils = { git = "https://github.com/agglayer/provers.git", tag = "v1.4.2" }
```
Ekleog-Polygon, line 65: Is it worth releasing 2.0 without agglayer-prover (because it's API breakage) and depending on it here? No strong opinion, just thinking, because I'm noticing 1.4.2 seems to not have the crate removal

atanmarko: We should upgrade to `provers` `v1.6.0` because of Fusaka upgrade

atanmarko: But I don't think that it matters depending on `1.4.2` `prover` in this PR, because those crates have not been moved or changed.

Cargo.toml
atanmarko: cc @Freyskeyd For follow up

Cargo.toml
atanmarko: Lets do upgrade of the provers with new proposer dependencies in the next PR

PR #1144, opened 2025-10-30
feat: add fee adjustments and receipt retrieval fix

atanmarko:
Resolves https://github.com/agglayer/agglayer/issues/1143

### DIFF
--- a/crates/agglayer-jsonrpc-api/src/kernel/mod.rs
+++ b/crates/agglayer-jsonrpc-api/src/kernel/mod.rs
@@ -436,7 +436,7 @@ where
                                     }
                                 }
                                 Err(error) => {
-                                    if confirmation_attempt <= max_retries {
+                                    if confirmation_attempt < max_retries {
                                         warn!(
                                             ?error,
                                             "Failed to get current block number, retrying"
@@ -468,8 +468,7 @@ where
                 }
                 Ok(None) => {
                     // Transaction not yet included in a block, continue retrying
-                    if attempt <= max_retries {
-                        // Report if we cross from one stage to the next.
+                    if attempt < max_retries {
                         debug!(
                             %tx_hash,
                             next_attempt = attempt + 1,


polykeith (dismissed):

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -316,17 +329,33 @@ where
...
+                    "Applying fee adjustments with gas price params: {:?}. Estimated fees: {:?}, \
+                     Adjusted fees: {:?}",
+                    &self.gas_price_params, estimate, adjusted
+                );
+
+                call.max_priority_fee_per_gas(adjusted.max_priority_fee_per_gas)
+                    .max_fee_per_gas(adjusted.max_fee_per_gas)
+                    .send()
+                    .await
+            })
```
atanmarko, line 354: This is only significant change in the PR. 
The rest is just:
* moving function `adjust_gas_estimate` to other module
* cleanup of the logs in the `send_tx` function

Freyskeyd (dismissed):

crates/agglayer-contracts/src/lib.rs
```diff
@@ -285,10 +285,58 @@ where
...
+        max_fee_per_gas,
+        max_priority_fee_per_gas,
+    };
+
+    if &adjusted != estimate {
+        debug!(
+            "Applied gas price adjustment. Estimated {}, {} priority. Adjusted to {}, {} priority.",
+            estimate.max_fee_per_gas,
+            estimate.max_priority_fee_per_gas,
+            max_fee_per_gas,
```
Ekleog-Polygon: Nit: can we have all these values as key/value pairs instead? It'd be both more machine-readable and easier to read for us when we'll be reading logs to debug 😄

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -310,23 +323,39 @@ where
...
             .verify_batches_trusted_aggregator(signed_tx)
-            .and_then(|call| async move { call.send().await })
+            .and_then(|call| async move {
+                let estimate = self.rpc.estimate_eip1559_fees().await?;
+                let adjusted = adjust_gas_estimate(&estimate, &self.gas_price_params);
+
+                debug!(
+                    "Applying fee adjustments with gas price params: {:?}. Estimated fees: {:?}, \
+                     Adjusted fees: {:?}",
+                    &self.gas_price_params, estimate, adjusted
```
Ekleog-Polygon: Nit: same here

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -310,23 +323,39 @@ where
...
             .await
             .map_err(SettlementError::ContractError)?;
 
-        if let Ok(Some(tx)) = self.check_tx_status(*pending_tx.tx_hash()).await {
-            warn!(hash, "Transaction already settled: {tx:?}");
+        if let Ok(Some(tx_receipt)) = self.check_tx_status(*pending_tx.tx_hash()).await {
+            warn!(
+                signed_tx_hash,
+                l1_tx_hash = %tx_receipt.transaction_hash,
+                "L1 transaction already settled: {tx_receipt:?}"
```
Ekleog-Polygon: Can we just have `?tx_receipt` as a key/value log, if the hash you already added just above is not enough?

crates/agglayer-jsonrpc-api/src/service/mod.rs
```diff
@@ -30,14 +30,15 @@ impl<Rpc> AgglayerService<Rpc>
...
-        tracing::Span::current().record("hash", &hash);
+        let signed_tx_hash = format!("{:?}", tx.hash());
+        tracing::Span::current().record("signed_tx_hash", &signed_tx_hash);
 
         info!(
-            hash,
-            "Received transaction {hash} for rollup {}", tx.tx.rollup_id
+            signed_tx = ?tx,
+            "Received signed transaction {signed_tx_hash} for rollup {}",
+            tx.tx.rollup_id
```
Ekleog-Polygon, line 42: Nit: same here, key/value would make the logs more easily searchable 😄
Though I think the span just above already has both rollup id and tx hash? So you could just remove it from the log text and all the info would already be present

crates/agglayer-jsonrpc-api/src/service/mod.rs
```diff
@@ -75,64 +79,72 @@ where
...
+            .await
+            .inspect_err(|error| error!(?error, "Failed to settle transaction on L1"))?;
 
         agglayer_telemetry::SETTLE.add(1, metrics_attrs);
 
-        info!(hash, "Successfully settled transaction {hash}");
+        info!(
+            l1_tx_hash = %receipt.transaction_hash,
+            "Signed transaction {signed_tx_hash}, block number {:?}, gas_used: {}",
+            receipt.block_number,
```
Ekleog-Polygon: Nit: Same here

Ekleog-Polygon (dismissed):
LGTM, thank you! I added a few nits below to improve logging even more, but your PR is already moving things in the right direction, so 🤷😄

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
atanmarko: Removed from code.

crates/agglayer-jsonrpc-api/src/service/mod.rs
atanmarko: Yeah there is already int he span, but I have intentionally left it like that, as span search and regular search in the datadog require different params, so to have one log text notion of this transaction

Freyskeyd (dismissed):

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -350,16 +354,9 @@ where
...
             .map_err(SettlementError::ContractError)?;
 
-        if let Ok(Some(tx_receipt)) = self.check_tx_status(*pending_tx.tx_hash()).await {
-            warn!(
-                signed_tx_hash,
-                l1_tx_hash = %tx_receipt.transaction_hash,
-                "L1 transaction already settled: {tx_receipt:?}"
-            );
-        }
+        debug!(l1_tx_hash = %pending_tx.tx_hash(), "For signed tx {signed_tx_hash} l1 transaction {} sent", pending_tx.tx_hash());
```
Ekleog-Polygon, line 358: Nit: `signed_tx_hash` is missing from kv tags I think

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -370,7 +367,140 @@ where
...
+    /// Wait for transaction receipt with configurable retries and intervals
+    #[allow(dead_code)]
+    async fn wait_for_transaction_receipt(
+        &self,
+        tx_hash: &TxHash,
+    ) -> Result<TransactionReceipt, SettlementError> {
+        let timeout = self
+            .settlement_config
+            .retry_interval
+            .mul_f64(self.settlement_config.max_retries as f64);
```
Ekleog-Polygon: Nit: We should add some leeway for the overhead between retries — maybe one more second, or one more retry interval? Otherwise we'll probably timeout before the end of the last retry.

Also, this variable is actually not used — I originally expected the big for loop to be wrapped in it, but it is only used for logs AFAIU?

In which case just adding a comment above, to that effect, that clearly explains that we know it doesn't have the leeway, would most likely be fine

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -370,7 +367,140 @@ where
...
+        let max_retries = self.settlement_config.max_retries;
+        let retry_interval = self.settlement_config.retry_interval;
+        let required_confirmations = self.settlement_config.confirmations;
+
+        debug!(
+            "Waiting for signed transaction receipt with timeout of {timeout:?}, max_retries: \
+             {max_retries} and retry_interval: {retry_interval:?}",
```
Ekleog-Polygon, line 392: Nit: add more kv tags

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -370,7 +367,140 @@ where
...
+                        );
+                        return Err(SettlementError::Timeout(timeout));
+                    } else {
+                        // No confirmations required, return immediately
+                        return Ok(receipt);
+                    }
+                }
+                Ok(None) => {
+                    // Transaction not yet included in a block, continue retrying
+                    if attempt <= max_retries {
```
Ekleog-Polygon: This should be `<`, and we should just remove the `else` branch to let it fall through after the end of the `for` — it'd be simpler than having the comment about "unreachable code" which is currently wrong :smiling_face_with_tear:

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -370,7 +367,140 @@ where
...
+                        // Report if we cross from one stage to the next.
+                        debug!(
+                            %tx_hash,
+                            next_attempt = attempt + 1,
+                            max_retries,
+                            "L1 transaction receipt not found yet, retrying after {retry_interval:?}",
```
Ekleog-Polygon: Nit: have retry_interval as kv

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -370,7 +367,140 @@ where
...
+                    error!(
+                        ?error,
+                        "Error watching the pending signed transaction settlement"
+                    );
+                    return Err(SettlementError::ProviderError(error));
+                }
+            }
+        }
+
+        // This should not be reached, but added for completeness
```
Ekleog-Polygon: The comment is wrong, we can currently reach here by reaching the end of the loop. If we just set to `<` as written above then indeed it becomes unreachable, but we could also just remove the `else` branch and have it here, it'd avoid code duplication

Ekleog-Polygon (dismissed):
LGTM! There's a technical error in the loop (the two comments not marked as nits), but AFAICT it only affects logging, so maybe it's not a big deal :shrug: 

The rest are definitely only nits, feel free to ignore as usual!

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
atanmarko: @Ekleog-Polygon No it is in the parent span

atanmarko:
```
[agglayer] {"timestamp":"2025-10-31T13:42:05.998483Z","level":"DEBUG","fields":{"message":"For signed tx 0x48424ffd5b2ada17ed991c188a29efd553c1ea258f60f90b4c5f528257d0f169 l1 transaction 0x730d2d3ada9aeb700e86c07b2a7ac72c31f123e6aa27c401ef4bada887a7e21b sent","l1_tx_hash":"0x730d2d3ada9aeb700e86c07b2a7ac72c31f123e6aa27c401ef4bada887a7e21b"},"target":"agglayer_jsonrpc_api::kernel","span":{"name":"settle"},"spans":[{"rollup_id":1,"signed_tx_hash":"0x48424ffd5b2ada17ed991c188a29efd553c1ea258f60f90b4c5f528257d0f169","name":"send_tx"},{"name":"settle"}]}
```

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
atanmarko: @Ekleog-Polygon Yeah, it is only used for the logs.

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
Ekleog-Polygon: Awesome! I hadn't noticed it. TBH I'd advocate for removing it altogether to avoid duplicated info, but it's definitely a matter of taste :shrug: Resolving this conversation :smile:

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
atanmarko: No it should be `attempt <= max_retries`, because first call on 0 is not retry.

Ekleog-Polygon: Hmm I don't think so? This happens between retries, so even if attemps 0 was not a retry, it'll be displayed (and wait) for the last round of the loop even though there'll be no next retry.

Basically AFAIU currently this if condition can never be false, because 0..=max_retries is always <= max_retries

atanmarko: @Ekleog-Polygon `max_retries` means maximum number of times to retry, after the first call. First call is before any wait for retry (so immediately at time 0).

atanmarko: So in `0..=max_retries`, if max_retries is 5, we have first call and 5 retries. Same logic applies for confirmations `attempt <= max_retries `.
max_retries could be 0 in configuration, so first call has to succeed (without any waiting).

Ekleog-Polygon: I totally agree with you about the for loop needing 0..=max_retries!

But the code is like "n+1 times, do try, and if it failed log, sleep and go back to loop condition". This specific if (not the for condition) will thus log and sleep just before the for loop ends because attempt grew too high. IOW, you're logging and sleeping one too many times, where you won't actually retry.

I'm pretty sure you want < here ; and if the code were written the other way around you'd need > 0 instead of <= max_retries. Because the for loop condition already ensures that attempt will never be > max_retries 😅

Anyway, it only happens when a cert fails to settle despite all retries, so it's probably not a big deal anyway 🤷

Ekleog-Polygon: (also there's also the same issue with confirmation_attempts <= max_retries which will always be true too ; I had missed it above, but just like this one it should have little concrete impact 🤷)

atanmarko: Ok, there is one more log and sleep, you are right. Fixing that.

PR #1145, opened 2025-10-31
feat: add new settlement listener task

polykeith:
This PR adds a new task called the settlement listener, which subscribes to event logs from L1 and specifically looks for the `VerifyPessimisticStateTransition` event emitted by the Rollup Manager contract in order to obtain the settling block for the certificate, among other various important data such as the settlement tx hash.


crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -111,19 +116,33 @@ pub struct CertificateOrchestrator<
...
 
-impl<Sc, CertifierClient, PendingStore, EpochsStore, PerEpochStore, StateStore>
+impl<
+        Sc,
+        CertifierClient,
+        PendingStore,
+        EpochsStore,
+        PerEpochStore,
+        StateStore,
+        RollupManagerRpc,
```
Ekleog-Polygon: Nit: I think we usually call this L1Rpc.

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -209,19 +240,23 @@ where
...
             settlement_client,
             certifier_task_builder,
             pending_store.clone(),
             epochs_store,
             current_epoch,
             state_store,
+            l1_rpc,
         )?;
 
+        orchestrator.listener_task.run(cancellation_token).await?;
```
Ekleog-Polygon: This will need to be `spawn_listener_task`, or you'll end up blocking the orchestrator (and other tasks)' startup

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        let rollup_address = self.l1_rpc.get_rollup_manager_address();
+        let filter = Filter::new()
+            .address(rollup_address.into_alloy())
+            .event_signature(VerifyPessimisticStateTransition::SIGNATURE_HASH)
+            .from_block(BlockNumberOrTag::Number(start_block));
+        let mut subscription = self
+            .l1_rpc
+            .get_provider()
+            .subscribe_logs(&filter)
```
Ekleog-Polygon, line 59: I'm wondering if we'll hit the 10k block limit issue here… but considering we're streaming I'd hope not? This'll probably need testing on some large network.

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    fn handle_event(&self, log: Log) -> Result<(), Error> {
+        let event = VerifyPessimisticStateTransition::decode_log(&log.clone().into()).ok();
+        let latest_pp_root = event
+            .as_ref()
+            .map(|val| <[u8; 32]>::from(val.newPessimisticRoot));
+        let tx_hash = log.transaction_hash.map(Digest::from);
+        let network_id = event.as_ref().map(|val| val.rollupID);
```
Ekleog-Polygon: Nit: all these `.map` could be inside the `if let` and avoid the boilerplate. And even better, you could start with `let Some(...) = ... else { return Ok(()) }`, to avoid rightward drift

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        if let (Some(pp_root), Some(tx_hash), Some(network_id), Some(block_number)) =
+            (latest_pp_root, tx_hash, network_id, log.block_number)
+        {
+            debug!(
+                "Retrieved latest VerifyPessimisticStateTransition event for network {} \
+                latest pp_root: {}, tx_hash: {tx_hash}",
+                network_id,
+                Digest(pp_root)
```
Ekleog-Polygon: Nit: use key-value trace logs for easier filtering

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,98 @@
...
+            );
+
+            self.state_store
+                .set_latest_certificate_settling_block(block_number)?;
```
Ekleog-Polygon: I think this method actually should be called from the settlement task, because events will (I think?) appear before the 5 confirmations, which means that they could still be reorg'd away by L1 and then we could miss some.

Instead, if we set this only from the cert settlement task, on next restart we'll start from a block height that at least had 5 confirmations. The risk is that we could end up with duplicate events, but that shouldn't be a big deal.

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -403,6 +403,10 @@ where
                     }
                     Some(NetworkTaskMessage::CertificateReadyForSettlement { settlement_submitted_notifier,
                         nonce_info, previous_tx_hashes, height, new_pp_root, .. }) => {
+
+                        // Associate the pp root with the certificate ID.
+                        self.state_store.set_certificate_id_for_pp_root(&new_pp_root, &certificate_id)?;
```
Ekleog-Polygon: This should be `add_` and not `set_`.

Also, maybe we should set it as soon as we generate the pp root? After all storage is basically free, and this way if the node crashes in the meantime at least in the db we'll have more metadata to understand what happened

crates/agglayer-certificate-orchestrator/src/tests/mocks.rs
```diff
@@ -24,3 +31,53 @@ mock! {
         ) -> Result<(MultiBatchHeader, LocalNetworkState, pessimistic_proof::PessimisticProofOutput), CertificationError>;
     }
 }
+
+mock! {
+    pub L1Rpc {}
```
Ekleog-Polygon, line 37: I think we already have such a mock somewhere; could we reuse it?

crates/agglayer-storage/src/columns/pp_root_to_certificate_id/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use super::{ColumnSchema, PP_ROOT_TO_CERTIFICATE_ID_CF};
+
+/// Column family for the pp root to certificate ID mapping.
+///
+/// ## Column definition
+///
+/// | key        | value           |
+/// | --         | --              |
+/// | `Digest`   | `CertificateId` |
```
Ekleog-Polygon: I think we said we'd need to store the full list of all certificate ids for one PPRoot? This way we can make sure that whenever we settle one, we set the other ones `InError` so that they don't try to settle one by one when it's impossible.

If I'm correct, you'll have to make this into a protobuf value type, like the network info CF

crates/agglayer-storage/src/columns/pp_root_to_certificate_id/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+pub struct PpRootToCertificateIdColumn;
+
+pub type Key = Digest;
```
Ekleog-Polygon: Nit: use the PessimisticRoot type I'm going to PR to interop most likely today, if it lands and can be updated in agglayer by the time you finish your PR

crates/agglayer-storage/src/stores/interfaces/reader.rs
```diff
@@ -71,6 +71,8 @@ pub trait PendingCertificateReader: Send + Sync {
 pub trait MetadataReader: Send + Sync {
     /// Get the latest settled epoch.
     fn get_latest_settled_epoch(&self) -> Result<Option<EpochNumber>, Error>;
+    /// Get the latest certificate settling block.
+    fn get_latest_certificate_settling_block(&self) -> Result<Option<BlockNumber>, Error>;
```
Ekleog-Polygon: Nit: this reads like "get the latest certificate that settles a block". Maybe `get_latest_block_that_settled_any_cert` would be better?

crates/agglayer-storage/src/stores/interfaces/reader.rs
```diff
@@ -99,6 +101,12 @@ pub trait StateReader: Send + Sync {
         &self,
         network_id: NetworkId,
     ) -> Result<Option<LocalNetworkStateData>, Error>;
+
+    /// Get the certificate ID for a given pp root.
+    fn get_certificate_id_for_pp_root(
+        &self,
+        pp_root: &Digest,
+    ) -> Result<Option<CertificateId>, Error>;
```
Ekleog-Polygon: This'll need to become `Result<Vec<CertId>>`

crates/agglayer-storage/src/stores/interfaces/writer.rs
```diff
@@ -34,6 +37,8 @@ pub trait EpochStoreWriter: Send + Sync {
 pub trait MetadataWriter: Send + Sync {
     /// Set the latest settled epoch.
     fn set_latest_settled_epoch(&self, value: EpochNumber) -> Result<(), Error>;
+    /// Set the latest certificate settling block.
+    fn set_latest_certificate_settling_block(&self, value: BlockNumber) -> Result<(), Error>;
```
Ekleog-Polygon: Same nit here as above for the name of the function.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -614,6 +633,23 @@ impl MetadataWriter for StateStore {
             &MetadataValue::LatestSettledEpoch(value),
         )?)
     }
+
+    fn set_latest_certificate_settling_block(&self, value: BlockNumber) -> Result<(), Error> {
+        if let Some(current_latest_certificate_settling_block) = self.get_latest_certificate_settling_block()? {
+            if current_latest_certificate_settling_block >= value {
+                return Err(Error::UnprocessedAction(
+                    "Tried to set a lower value for latest certificate settling block".to_string(),
+                ));
```
Ekleog-Polygon: Hmm… this might deserve a warning and then returning Ok? After all it's not a big deal, and I wouldn't be surprised if it could actually happen.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -614,6 +633,23 @@ impl MetadataWriter for StateStore {
...
+            }
+        }
+
+        self.db.put::<MetadataColumn>(
+            &MetadataKey::LatestCertificateSettlingBlock,
+            &MetadataValue::LatestCertificateSettlingBlock(value),
```
Ekleog-Polygon: Nit: same naming issue, this sounds like the last certificate that settled a block

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -632,4 +668,20 @@ impl MetadataReader for StateStore {
                 })
             })
     }
+
+    fn get_latest_certificate_settling_block(&self) -> Result<Option<BlockNumber>, Error> {
+        self.db
+            .get::<MetadataColumn>(&MetadataKey::LatestCertificateSettlingBlock)
+            .map_err(Into::into)
```
Ekleog-Polygon: If you just use `?` it'll automatically do the err mapping, and will allow you to remove the rightward drift by killing `and_then` and replacing `map_or` with a simple map… or even just matching on `v`, might be even easier to read :smile:

Ekleog-Polygon (commented):
Overall LGTM! A few nits below based on my preliminary review :smile: 

Note to self: when this gets un-draft'd I'll have to review the code as a whole again

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -111,19 +116,25 @@ pub struct CertificateOrchestrator<
...
-impl<Sc, CertifierClient, PendingStore, EpochsStore, PerEpochStore, StateStore>
+impl<Sc, CertifierClient, PendingStore, EpochsStore, PerEpochStore, StateStore, L1Rpc>
     CertificateOrchestrator<
         Sc,
         CertifierClient,
         PendingStore,
         EpochsStore,
         PerEpochStore,
         StateStore,
+        L1Rpc,
```
atanmarko, line 132: We are already using l1 rpc client in the network task and settlement client. Do we need to explicitly specialize this client here just for listener task?

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -111,19 +116,25 @@ pub struct CertificateOrchestrator<
...
     >
 where
     PendingStore: PendingCertificateReader,
+    L1Rpc: RollupContract + L1TransactionFetcher,
```
atanmarko, line 136: Could we define maybe `ListenerClient` type (similar to `Sc` - settlement client)?

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,121 @@
...
+                        error!("Error handling settlement event: {err}");
+                    }
+                }
+            }
+        }
+    }
+
+    fn handle_event(&self, log: Log) -> Result<(), Error> {
+        let Ok(event) = VerifyPessimisticStateTransition::decode_log(&log.clone().into()) else {
+            return Err(Error::InternalError(format!(
```
atanmarko, line 83: Could we maybe define specific error type for event handling instead of using `InternalError`?

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,121 @@
...
+        let Some(block_number) = log.block_number else {
+            return Err(Error::InternalError(format!(
+                "Failed to get block number from log: {log:?}"
+            )));
+        };
+
+        debug!(
+            "Retrieved latest VerifyPessimisticStateTransition event for network {network_id}, \
+             latest pp_root: {pp_root:?}, tx_hash: {tx_hash:?}, block_number: {block_number}",
+        );
```
atanmarko, line 108:
```suggestion
        debug!(
            %network_id, 
            %tx_hash,
            %pp_root,
            %block_number,
            "Retrieved latest VerifyPessimisticStateTransition event for network {network_id}, \
             latest pp_root: {pp_root:?}, tx_hash: {tx_hash:?}, block_number: {block_number}",
        );
```

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        let certificate_ids = self.state_store.get_certificate_ids_for_pp_root(&pp_root)?;
+
+        let Some(settled_certificate_id) = certificate_ids.last() else {
```
atanmarko, line 114: Why does it have to be last? Could it be any in the list?

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,121 @@
...
+            return Err(Error::InternalError(format!(
```
atanmarko, line 115: Could we have a race here - that certificate is not yet in the database and we receive event? Do we need to keep fetched event log in the database to have history (in case of some restart/crash)?

crates/agglayer-certificate-orchestrator/src/tests.rs
```diff
@@ -531,6 +548,7 @@ async fn test_certificate_orchestrator_can_stop() {
         epochs_store,
         Arc::new(current_epoch),
         state_store.clone(),
+        rollup_manager.clone(),
```
atanmarko: Do we need a clone here? Also applies for other places, not to repeat comment.

crates/agglayer-certificate-orchestrator/src/tests/mocks.rs
atanmarko: We already have this mock in `settlement_client/tests.rs`. Could we reuse it from some common place?

Ekleog-Polygon (commented):
Please re-request my review this after @atanmarko's comments are handled :smile:

crates/agglayer-certificate-orchestrator/src/lib.rs
polykeith: We need some way of getting the rollup manager address and the provider to subscribe to events on L1, and inputting the `L1Rpc` type as a generic is the way I can think of doing so.

crates/agglayer-certificate-orchestrator/src/listener_task.rs
polykeith: As discussed with @Ekleog-Polygon, only the last one in the list is useful, as the previous ones are considered obsolete or unused, and they only share the same cert ID with the last one that's used.

crates/agglayer-certificate-orchestrator/src/listener_task.rs
polykeith: I don't feel like that's possible -- an event settling takes a way longer time than putting certificate data in the DB -- unless the node restarts? I'm not aware of any other process that can delay the storage of certificate data into the DB.

crates/agglayer-certificate-orchestrator/src/listener_task.rs
adria0pol: > As discussed with @Ekleog-Polygon, only the last one in the list is useful, as the previous ones are considered obsolete or unused, and they only share the same cert ID with the last one that's used.

@polykeith Maybe it's a good idea to add a comment about this in the code

crates/agglayer-certificate-orchestrator/src/lib.rs
polykeith: I'm not sure I understand -- you want a type alias or a supertrait containing all the subtraits that you want?

crates/agglayer-certificate-orchestrator/src/listener_task.rs
polykeith: There's a conversion to `CertificateStatusError`, should I still use `InternalError` there or should I create a new error variant?

crates/agglayer-certificate-orchestrator/src/tests/mocks.rs
polykeith: We're mocking the same things from 2 different crates: `agglayer-certificate-orchestrator` and `agglayer-aggregator-notifier`, I'm not sure whether we can actually refactor and make the mock shared amongst the 2 crates -- maybe a shared dev dependency on a 3rd crate?

crates/agglayer-certificate-orchestrator/src/lib.rs
atanmarko: Type alias would be enough

crates/agglayer-certificate-orchestrator/src/listener_task.rs
atanmarko: On the second thought, lets stick to  `InternalError`.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -430,6 +444,15 @@ where
...
         let certificate_id = self.header.certificate_id;
+        let new_pp_root = self
+            .new_pp_root
+            .ok_or(CertificateStatusError::InternalError(
+                "CertificateTask::process_from_proven called without a pp_root".into(),
+            ))?;
+
+        // Associate the pp root with the certificate ID.
+        self.state_store
+            .add_certificate_id_for_pp_root(&PessimisticRoot::from(new_pp_root), &certificate_id)?;
```
atanmarko: Do you need this one, as `add_certificate_id_for_pp_root` is already called in `process_from_pending`

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -520,6 +539,10 @@ where
             %settlement_tx_hash,
             "Waiting for certificate settlement to complete"
         );
+
+        // Associate the pp root with the certificate ID.
+        self.state_store
+            .add_certificate_id_for_pp_root(&PessimisticRoot::from(new_pp_root), &certificate_id)?;
```
atanmarko: Same question - is this one needed? `self.new_pp_root` is only changed in `process_from_pending`.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
Ekleog-Polygon: ~~(Copying from Slack)~~

~~Hmmm so I haven't reviewed the PR yet, but my initial intuition would be that we need to associate the cert id with a PP root after certificate execution completes, which would mean not doing so in `process_from_pending`~~

~~Because we don't know whether the PP root claimed in the cert actually matches the real one of the certificate, and we don't want an attacker to claim a fake PP root and get their cert settled-for-agglayer without it actually being correct~~

~~So my initial guess would be that we should remove the assignment in `process_from_pending`, and instead add the assignment exclusively just after we execute the certificate, roughly at the time we send the `CertificateExecuted` message~~

~~Does that make sense to you?~~

Edit: The one in `process_from_pending` is already exactly in the right place here, I replied too quickly. So I'm with you that we should remove the other ones that are only duplicates. Sorry for the noise!

crates/agglayer-storage/src/columns/pp_root_to_certificate_ids/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+/// | `PessimisticRoot` | `Vec<CertificateId>` |
+pub struct PpRootToCertificateIdsColumn;
+
+pub type Key = PessimisticRoot;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Value(pub(crate) Vec<CertificateId>);
+
+crate::columns::impl_codec_using_bincode_for!(Key);
+crate::columns::impl_codec_using_bincode_for!(Value);
```
Ekleog-Polygon: I don't think we want to introduce new CFs where the value is using bincode — can you turn this into a protobuf-encoded CF?

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -370,6 +376,21 @@ impl StateWriter for StateStore {
...
+    fn add_certificate_id_for_pp_root(
+        &self,
+        pp_root: &PessimisticRoot,
+        certificate_id: &CertificateId,
+    ) -> Result<(), Error> {
+        let mut certificate_ids = self.db.get::<PpRootToCertificateIdsColumn>(pp_root)?
+            .map(|v| v.0)
+            .unwrap_or_default();
+        certificate_ids.push(*certificate_id);
+        Ok(self.db.put::<PpRootToCertificateIdsColumn>(
```
Ekleog-Polygon: This should take some kind of mutex, to avoid concurrent calls overwriting each other — I don't think we use a rocksdb TransactionDB so we can't have the read and write be in a single transaction but a simple mutex should be good enough considering we're only modifying this CF here.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -667,4 +717,16 @@ impl MetadataReader for StateStore {
                 })
             })
     }
+
+    fn get_latest_block_that_settled_any_cert(&self) -> Result<Option<BlockNumber>, Error> {
+        match self.db.get::<MetadataColumn>(&MetadataKey::LatestBlockThatSettledAnyCert)? {
+            Some(MetadataValue::LatestBlockThatSettledAnyCert(value)) => Ok(Some(value)),
+            None => Ok(None),
+            _ => Err(Error::Unexpected(
+                "Wrong value type decoded, was expecting LastCertificateSettlingBlock, decoded \
```
Ekleog-Polygon: Nit:
```suggestion
                "Wrong value type decoded, was expecting LatestBlockThatSettledAnyCert, decoded \
```

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -209,19 +224,26 @@ where
...
             pending_store.clone(),
             epochs_store,
             current_epoch,
             state_store,
+            l1_rpc,
         )?;
 
+        tokio::spawn({
+            let listener_task = orchestrator.listener_task.clone();
+            async move { listener_task.run(cancellation_token).await }
```
Ekleog-Polygon, line 244: Nit: We should have some kind of logging here, if the listener task returns with an error.

crates/agglayer-certificate-orchestrator/src/listener_task.rs
```diff
@@ -0,0 +1,131 @@
...
+            )));
+        };
+        let pp_root = PessimisticRoot::from(event.newPessimisticRoot);
+        let Some(tx_hash) = log.transaction_hash.map(Digest::from) else {
+            return Err(Error::InternalError(format!(
+                "Failed to get tx hash from log: {log:?}"
+            )));
+        };
+        // rollupID is the same as topic1, which is the network ID
+        let network_id = event.rollupID;
```
Ekleog-Polygon, line 94: Did you verify that the contracts do have the network id in topic1, and not the rollup id? IIRC they're off-by-one. The best might be to have some kind of integration test to verify that? But if you're sure about that it's good enough for me :smile:

crates/agglayer-certificate-orchestrator/src/listener_task.rs
Ekleog-Polygon: If I'm not mistaken, any certificate would work just as well: the certificate id already includes all of the information inside the certificate, except for aggchain_data, custom_chain_data and l1_info_tree_leaf_count.
Which means that agglayer-node can consider any certificate of this list as settled as it wants and it shouldn't cause any issue.

@hadjiszs Can you confirm this understanding?

This being said, I do think there's an issue in unconditionally setting the latest certificate as settled: we could end up with two certificates that are marked as settled for the same PP root, which would be wrong.

So we'd need some kind of way to ensure that we're only marking the latest cert as settled if none of the certificates listed there are already settled.

Does that make sense to you?

Also for migration reasons, we won't have PP roots recorded when we restart with the upgraded certificate. So we would likely want to first do a release of agglayer-node that only records pp roots -> cert ids ; and only after we're sure all inflight certs do have that record do a second release of agglayer-node that automatically marks certs as settled.

Ekleog-Polygon (changes requested):
Thank you for your updates! The code almost looks good to me, to the exception of the points I listed below :sweat_smile: Please lmk if there's anything unclear!

crates/agglayer-storage/src/stores/state/mod.rs
polykeith: Just to clarify, a mutex on which data structure? The DB?

Issue #1154, opened 2025-11-04
Make nonce reuse issues less prominent


Ekleog-Polygon:
If we restart agglayer-node while there are unsettled txes in the L1 mempool, this will desync agglayer-node nonce accounting whenever they settle. This can be especially bad if some of the unsettled txes have low priority gas and take a long time to be prioritized.

While the low priority gas issue is separate and should be tracked accordingly, we should at least avoid submitting new txes with the same nonce as old txes.

Proper resolution of this issue would involve writing the last nonce used in storage and reloading it from there on startup.

<s>However, as a workaround (until we fully rewrite the settlement logic for aggregation anyway), we can simply switch from CachedNonceManager to SimpleNonceManager. It will not prevent nonce reuse, but at least whenever nonce reuse does happen we'll skip ahead to the latest nonce, instead of moving forward one-by-one and failing one transaction each time.

Once this workaround is done, we should open a follow-up issue to actually save the last used nonce in storage.</s>
(above stricken text already implemented)

PR #1156, opened 2025-11-04
fix: validium tx wait for confirmations

atanmarko:
Port from release branch.
Simple nonce mangaer used by provider.
Related https://github.com/agglayer/agglayer/issues/1154

### DIFF
--- a/crates/agglayer-rpc/src/lib.rs
+++ b/crates/agglayer-rpc/src/lib.rs
@@ -790,7 +790,7 @@ where
 
     /// Verify that the signer of the given [`Certificate`] is the trusted
     /// sequencer for the rollup id it specified.
-    #[instrument(skip(self), level = "debug")]
+    #[instrument(skip(self, cert), fields(certificate_id = %cert.hash()), level = "debug")]
     pub(crate) async fn verify_cert_signature(
         &self,
         cert: &Certificate,


crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -357,32 +357,43 @@ where
...
+                    rate_guard.record(tokio::time::Instant::now());
+                    info!(
+                        block_hash = ?tx_receipt.block_hash,
+                        block_number = ?tx_receipt.block_number,
+                        l1_tx_hash = %tx_receipt.transaction_hash,
+                        "Inspect settle l1 transaction"
+                    )
+                })
+        });
+        receipt.await.map_err(|_| SettlementError::NoReceipt)?
```
Ekleog-Polygon: Nit: AFAICT this is not actually `NoReceipt`, but an internal error in settlement processing? A join handle that returns Err, AFAIK, means there was a panic somewhere in `wait_for_transaction_receipt_static`.

So I'd see that more as a new `InternalError(#[source] eyre::Error)` variant, but… WDYT?

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
```diff
@@ -357,32 +357,43 @@ where
...
     }
 
     /// Wait for transaction receipt with configurable retries and intervals
-    #[instrument(skip(self), level = "debug")]
-    async fn wait_for_transaction_receipt(
-        &self,
+    /// (static version)
+    async fn wait_for_transaction_receipt_static(
```
Ekleog-Polygon: Nit: can we remove this new parenthesis comment and the `_static` suffix? Considering the old function no longer exists I don't see any value in them.

crates/agglayer-node/src/node.rs
```diff
@@ -187,9 +187,10 @@ impl Node {
         let address = signer.address();
         tracing::info!("Signer address: {:?}", address);
 
-        // Create a new L1 RPC provider with signer support
+        // Create a new L1 RPC provider with signer support and SimpleNonceManager
         let wallet = EthereumWallet::from(signer);
         let provider = ProviderBuilder::new()
+            .with_simple_nonce_management()
```
Ekleog-Polygon, line 193: Nit: this should be in another PR, so that our git history does remember that we fixed two bugs (also, feel free to link https://github.com/agglayer/agglayer/issues/1154 in pr description for the rationale)

Ekleog-Polygon (dismissed):
LGTM! Just have a few nits below, I think only the first one deserves a bit more thinking

crates/agglayer-node/src/node.rs
atanmarko: It is already in the commit on release, hence one PR with the same title/commit message here.

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
atanmarko: You are right, lets fix it in the follow up

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
atanmarko: Good point

crates/agglayer-jsonrpc-api/src/kernel/mod.rs
atanmarko: EDIT: I'll fix it here, what is in release stays in release

iljakuklic (approved):

PR #1159, opened 2025-11-06
feat: make admin api able to force edit settlement tx hash

Ekleog-Polygon:
BREAKING-CHANGE: Admin API was significantly updated
  `forceSetCertificateStatus` was replaced with `forceEditCertificate`
  `forceEditCertificate` takes the certificate id, then a string that must be either `"process-now=true"` or `"process-now=false"`, and then up to two operations
  The operations are strings formatted as `"set-status,from=InError,to=Candidate"` or `"set-settlement-tx-hash,from=0x123...789,to=null"` for example.

### DIFF
--- a/crates/agglayer-jsonrpc-api/src/admin.rs
+++ b/crates/agglayer-jsonrpc-api/src/admin.rs
@@ -638,16 +638,16 @@ where
         self.state
             .disable_network(&network_id, agglayer_types::network_info::DisabledBy::Admin)
             .map_err(|error| {
-                error!(?error, "Failed to disable network {}", network_id);
-                Error::internal(format!("Unable to disable network {}", network_id))
+                error!(?error, "Failed to disable network {network_id}");
+                Error::internal(format!("Unable to disable network {network_id}"))
             })
     }
 
     #[instrument(skip(self), level = "debug")]
     async fn enable_network(&self, network_id: NetworkId) -> RpcResult<()> {
         self.state.enable_network(&network_id).map_err(|error| {
-            error!(?error, "Failed to enable network {}", network_id);
-            Error::internal(format!("Unable to enable network {}", network_id))
+            error!(?error, "Failed to enable network {network_id}");
+            Error::internal(format!("Unable to enable network {network_id}"))
         })
     }
 }
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -27,7 +27,7 @@ jobs:
       checks: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       - uses: rustsec/audit-check@v2.0.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,7 +18,7 @@ jobs:
     name: Unit & Linter coverage
     runs-on: ubuntu-latest-16-cores
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
 
--- a/.github/workflows/cycle-tracker.yml
+++ b/.github/workflows/cycle-tracker.yml
@@ -23,7 +23,7 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -21,7 +21,7 @@ jobs:
     if: github.event.action != 'closed'
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -82,7 +82,7 @@ jobs:
       issues: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Download documentation artifact
         uses: actions/download-artifact@v5
@@ -149,7 +149,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -36,7 +36,7 @@ jobs:
       tags: ${{ steps.meta.outputs.tags }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
--- a/.github/workflows/mq-checking-commit.yml
+++ b/.github/workflows/mq-checking-commit.yml
@@ -17,7 +17,7 @@ jobs:
     if: ${{ github.event_name == 'merge_group' }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - uses: wagoid/commitlint-github-action@v6
--- a/.github/workflows/proto.yml
+++ b/.github/workflows/proto.yml
@@ -19,7 +19,7 @@ jobs:
   buf:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
           ref: ${{ github.head_ref || github.event.pull_request.head.ref || github.ref }}
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@nightly
@@ -40,7 +40,7 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -66,7 +66,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@nightly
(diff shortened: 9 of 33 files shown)


crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -247,18 +252,118 @@ where
...
+            %certificate_id,
+            %process_now,
+            ?operation_1,
+            ?operation_2,
+            "(ADMIN) Editing certificate"
         );
+
+        let process_now = match process_now.as_str() {
+            "process now" => true,
+            "do not process now" => false,
```
atanmarko: I don't think using those sentences to describe one argument, in fragile live surgery operation is a good idea.
Whats wrong with `true`/`false` or even "process"/"no-process" (in the linux shell argument spirit)?

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -247,18 +252,118 @@ where
...
+            },
+            SetSettlementTxHash {
+                from: Option<SettlementTxHash>,
+                to: Option<SettlementTxHash>,
+            },
+        }
+
+        impl Operation {
+            fn parse(operation: &str) -> Result<Self, Error> {
+                if operation.starts_with("set status from ") {
```
atanmarko: I would't use sentences with spaces for one argument. It is very error prone. It is counterintuitive for anybody using the Linux/Mac shell. If you insist on using multi words description for one argument, you could use `set-status-from=InError` and `set-status-to=Settled` or `change-status=InError,Settled`. Do sanitary trimming

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -247,18 +252,118 @@ where
...
+                                ))
+                            })
+                        }
+                    }
+                    Ok(Operation::SetStatus {
+                        from: parse_status(from_status)?,
+                        to: parse_status(to_status)?,
+                    })
+                } else if operation.starts_with("set settlement tx hash ") {
+                    let parts = operation["set settlement tx hash from ".len()..]
```
atanmarko: Same problem, we should not use sentence with spaces for one argument name. Why not `change-settlement-tx-hash=0x20d3b836054b95517ddd4f29cd2face634f039639502a5e91c46fd4cc8693562,0xb01a252d84448603c2067369c7374730c387329b08b4984dfb7d84ef1c32457d`? Easier to parse and verify.

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -270,32 +375,84 @@ where
...
-                     not match the existing one ({:?})",
-                    header.settlement_tx_hash
-                )));
+
+        // Check that the current values match the "from" value
+        for operation in operations.iter() {
+            match operation {
+                Operation::SetStatus { from, to: _ } => {
+                    // Ignore the error details when changing from InError to anything else
+                    if &header.status != from
```
atanmarko, line 396: Additional comment would be nice, e.g.:
`// If the certificate’s status isn’t the same as the from status, and we’re not just transitioning between two InError statuses, then return error.`

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: The fact that it's a fragile live surgery is exactly why it's a good idea :smile: Basically, with true/false, you can make a mistake. Here, if you make a mistake (eg. write `"proces now"`), then nothing will happen.

IIRC this was originally suggested by @sebastiendan because the arguments were hard to remember and understand

Ekleog-Polygon: No strong argument against process/no-process that said, it's just that it wouldn't translate as well to the "set status from a to b" for example

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: @sebastiendan which one do you prefer, between the current, the suggestion from @atanmarko, or any other? I'm open to literally any syntax, and don't really care :smile: 
My only comment would be, I'd rather have a single string for the whole set-status operation, that includes both from and to

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: @sebastiendan (and if you have any preferences here too, just let us know :smile: )

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: Uhh but it's not actually about transitioning between two InError statuses? It's just that I'm ignoring the error message in the InError variant. It's fun cause my AI editor also wanted me to write the exact same comment, but I'm relatively sure it's not actually what I wrote :sweat_smile: 

Like, this is literally just "assert that the from status is the expected one, ignoring the error details for InError".

Do you see another wording that would be more clear?

crates/agglayer-jsonrpc-api/src/admin.rs
sebastiendan: My original ask was to replace `... true ...` with `... my-boolean=true ...` so that we always know what arg we're setting

crates/agglayer-jsonrpc-api/src/admin.rs
sebastiendan: I've never seen prose text like that in arguments so I understand Marko's surprise. I like the `set-status-from-...`. This is essentially what I was asking (named parameters).

crates/agglayer-jsonrpc-api/src/admin.rs
sebastiendan: Same comment than above

crates/agglayer-jsonrpc-api/src/admin.rs
atanmarko: Explain that !matches check is because every `InError` status is parsed to "Set to InError by administrator" so that check does not pass. And we can actually switch error types here, from original to "Inerror by administrator"

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -247,18 +252,116 @@ where
...
 
     #[instrument(skip(self), level = "debug")]
-    async fn force_set_certificate_status(
+    async fn force_edit_certificate(
         &self,
         certificate_id: CertificateId,
-        status: CertificateStatus,
-        process_now: bool,
-        remove_settlement_tx_hash: Option<SettlementTxHash>,
+        process_now: String,
```
Freyskeyd: Can you use an enum here ?

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -247,18 +252,116 @@ where
...
+        operation_1: Option<String>,
+        operation_2: Option<String>,
```
Freyskeyd, line 276: Could be replaced by a `Vec<Operation>`

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -247,18 +252,116 @@ where
...
+                to: CertificateStatus,
+            },
+            SetSettlementTxHash {
+                from: Option<SettlementTxHash>,
+                to: Option<SettlementTxHash>,
+            },
+        }
+
+        impl Operation {
+            fn parse(operation: &str) -> Result<Self, Error> {
```
Freyskeyd, line 298: Could we use json instead of doing manual parsing?
Something like:

```rust

#[derive(serde::Serialize, serde::Deserialize)]
pub enum ExecutionMode {
    DryRun,
    Execute
}

#[derive(Debug, serde::Serialize, serde::Deserialize)]
pub enum Operation {
    SetStatus {
        from: CertificateStatus,
        to: CertificateStatus,
    },
    SetSettlementTxHash {
        from: Option<SettlementTxHash>,
        to: Option<SettlementTxHash>,
    },
}
```

That could be used after that with:

```bash
cast rpc admin_forceEditCertificate --params '{"request_type":"dry-run","operations":[{"set-status":{"from":"pending","to":"proven"}}, {"set-settlement-tx-hash":{"to":"0x00...00"}]}'

```

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: Done, thank you for the precisions!

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: I thought a bit about this, and ended up picking something akin to qemu's way of handling arguments: the syntax would be `set-status,from=Candidate,to=Proven` for example.

Hopefully that sounds good to everyone? :smile:

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: Right, but that'd happen only if someone passed `set-status,from=InError,to=InError`, which sounds… pretty unlikely :sweat_smile: 

Anyway, I tried making the comment clearer, hopefully it's good enough now!

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: Unfortunately that'd change the API, because we move from a pseudo-variadic number of arguments to an inner vec.

For cast, that manifests by having to change the syntax from the first to the second:
```
cast forceEditCertificate 0x123 process-now=true set-status,from=Candidate,to=Proven set-settlement-tx-hash,from=0x123,to=0.456
cast forceEditCertificate 0x123 process-now=true '["set-status,from=Candidate,to=Proven","set-settlement-tx-hash,from=0x123,to=0.456"]'
```

Considering you're off, I suggest going to 0.4.3 with this syntax, and we can easily iterate on this with time, as we have no backward-compat requirements on the admin RPC :smile:

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: We could, but admin RPC is made to be used by a human; so I'd think JSON-based arguments would kinda defeat the point here by making everything much harder to use.

So same as above, let's release 0.4.3 with the current syntax and iterate on it if need be :smile: "

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -32,13 +42,17 @@ pub(crate) trait AdminAgglayer {
...
+    #[method(name = "forceEditCertificate")]
+    async fn force_edit_certificate(
         &self,
         certificate_id: CertificateId,
-        status: CertificateStatus,
-        process_now: bool,
-        remove_settlement_tx_hash: Option<SettlementTxHash>,
+        process_now: ProcessNow,
+        operation_1: Option<String>,
+        operation_2: Option<String>,
```
Freyskeyd, line 51: why not accepting a `Vec<String>` ?

Freyskeyd (approved):

crates/agglayer-jsonrpc-api/src/admin.rs
Ekleog-Polygon: Unfortunately that'd change the API, because we move from a pseudo-variadic number of arguments to an inner vec.

For cast, that manifests by having to change the syntax from the first to the second:
```
cast forceEditCertificate 0x123 process-now=true set-status,from=Candidate,to=Proven set-settlement-tx-hash,from=0x123,to=0.456
cast forceEditCertificate 0x123 process-now=true '["set-status,from=Candidate,to=Proven","set-settlement-tx-hash,from=0x123,to=0.456"]'
```

Unless I'm missing something?

PR #1160, opened 2025-11-06
feat: allowing disabling networks

Freyskeyd:
CONFIG-CHANGE: Short description of the configuration changes.
  More details can be added in the description.
  Link to configuration file example if applicable.
  Link to migration guide if applicable.
  Multiple configuration changes can be listed.

BREAKING-CHANGE: Short description of the breaking changes.
  More details can be added in the description.
  Multiple lines can be used, with **markdown** formatting if desired.
  Multiple breaking changes can be listed.

### DIFF
--- a/crates/agglayer-rpc/src/lib.rs
+++ b/crates/agglayer-rpc/src/lib.rs
@@ -646,6 +646,10 @@ where
             })?;
 
         match network_info.latest_pending_status {
+            _ if network_is_disabled => {
+                // If the network is disabled in storage, mark it as disabled
+                network_info.network_status = NetworkStatus::Disabled;
+            }
             None => {
                 // No pending certificate means the network status is unknown
                 network_info.network_status = NetworkStatus::Unknown;
@@ -653,10 +657,6 @@ where
             Some(CertificateStatus::InError { .. }) => {
                 // Network is in error if the latest pending certificate is in error
                 network_info.network_status = NetworkStatus::Error;
-            }
-            _ if network_is_disabled => {
-                // If the network is disabled in storage, mark it as disabled
-                network_info.network_status = NetworkStatus::Disabled;
             }
             _ => {
                 // Otherwise, the network is active


buf.storage.gen.yaml
```diff
@@ -8,7 +8,7 @@ plugins:
     out: crates/agglayer-storage/src/types/generated
     opt:
       - bytes=.
-      - compile_well_known_types
+      # - compile_well_known_types
```
Ekleog-Polygon: Do you know why this change is required?

crates/agglayer-rpc/src/lib.rs
```diff
@@ -645,6 +645,23 @@ where
             }
         }
 
+        if self
+            .state
+            .is_network_disabled(&network_id)
+            .map_err(|error| {
+                error!(
+                    ?error,
+                    "Failed to check if network {network_id} is disabled in storage"
```
Ekleog-Polygon: Nit: add `network_id` to fields here… though it might be even better to just instrument the whole function so all logs automatically have it

crates/agglayer-storage/src/types/disabled_network/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+
+        let mut buf = BytesMut::new();
+        buf.reserve(len);
+        <Value as prost::Message>::encode(self, &mut buf)?;
+
+        Ok(buf.to_vec())
+    }
+
+    fn encode_into<W: io::Write>(&self, _writer: W) -> Result<(), crate::columns::CodecError> {
+        unimplemented!()
```
Ekleog-Polygon: Uhh this sounds wrong, isn't there a default impl that just delegates to `encode`?

Also IIUC this is not our first prost column in rocksdb, can we have some macro to implement Codec with less boilerplate?

proto/agglayer/storage/v0/disabled_network.proto
```diff
@@ -0,0 +1,23 @@
...
+}
+
+// Enum representing who disabled the network.
+enum DisabledBy {
+  // Default value, should not be used.
+  DISABLED_BY_UNSPECIFIED = 0;
+  // The network was disabled by an admin.
+  DISABLED_BY_ADMIN = 1;
+  // The network was disabled by the system.
+  DISABLED_BY_SYSTEM = 2;
```
Ekleog-Polygon: AFAICT we can never be `DISABLED_BY_SYSTEM` currently, would it make sense to remove the variant?

As the next PR will probably be a way for admin to disable the network.

Ekleog-Polygon (dismissed):
Mostly LGTM for the current changes :smile:  Requesting changes due to the `unimplemented!°

Also, I'd guess we should also disable the network on RPC, rejecting new certificates… unless the get_active_networks fixes would be enough to do that?

crates/agglayer-storage/src/types/disabled_network/mod.rs
Freyskeyd: That's already what we're doing for `NetworkInfo` as `encode_into` isn't compatible with how prost works currently. And basically, `encode` is calling `encode_into` in the default implementation.
@iljakuklic if you have any idea, feel free

Ekleog-Polygon: We can simply call °encode° and then write the resulting Vec to the °io::Write° impl, can't we? Sure it'll be one "spurious" allocation, but it should be reasonably easy without having to hack around prost, and avoid the panic at least :smile:

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -514,10 +511,19 @@ impl StateReader for StateStore {
     /// small. This function is only called once when the node starts.
     /// Benchmark: `last_certificate_bench.rs`
     fn get_active_networks(&self) -> Result<Vec<NetworkId>, Error> {
+        let disabled_networks = self
+            .db
+            .keys::<crate::columns::disabled_networks::DisabledNetworksColumn>()?
+            .filter_map(|v| v.ok())
```
atanmarko, line 548: Should we error early here if there is some issue with reading disabled networks column? We don't want to consider it active if it is disabled by purpose.

crates/agglayer-rpc/src/lib.rs
```diff
@@ -645,6 +645,23 @@ where
             }
         }
 
+        if self
+            .state
+            .is_network_disabled(&network_id)
```
atanmarko: Could we check the `match network_info.latest_pending_status` to learn the network status in the else branch of this if condition (if network is not disabled)?

proto/agglayer/storage/v0/disabled_network.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package agglayer.storage.v0;
+
+import "google/protobuf/timestamp.proto";
+
```
atanmarko, line 6: Once disabled always disabled? Do we need a logic/state log to disable/enable network multiple times?

buf.storage.gen.yaml
Freyskeyd: Yes, sorry I'll have to remove this comment.

Basically, defining `compile_well_known_types` will make protoc generate all well_known types instead of using those from `prost_types` which is more straightforward.

crates/agglayer-rpc/src/lib.rs
Freyskeyd: what would be the purpose of that?

crates/agglayer-storage/src/stores/state/mod.rs
Freyskeyd: I'll just match and add an error log in case it is happening

Freyskeyd: In fact, we're just reading the keys, and it shouldn't be something problematic as the key will never change

proto/agglayer/storage/v0/disabled_network.proto
Freyskeyd: I added some more store method to cover that

crates/agglayer-rpc/src/lib.rs
atanmarko: Better code organization and logic flow? If network is disabled, why check `latest_pending_status`?

atanmarko: Like this:
```
        if self
            .state
            .is_network_disabled(&network_id)
            .map_err(|error| {
                error!(
                    ?error,
                    "Failed to check if network {network_id} is disabled in storage"
                );
                GetNetworkInfoError::InternalError {
                    network_id,
                    source: error.into(),
                }
            })?
        {
            network_info.network_status = NetworkStatus::Disabled;
        } else {
            match network_info.latest_pending_status {
                None => {
                    // No pending certificate means the network status is unknown
                    network_info.network_status = NetworkStatus::Unknown;
                }
                Some(CertificateStatus::InError { .. }) => {
                    // Network is in error if the latest pending certificate is in error
                    network_info.network_status = NetworkStatus::Error;
                }
                _ => {
                    // Otherwise, the network is active
                    network_info.network_status = NetworkStatus::Active;
                }
            }
        }
```

Freyskeyd: I'll move the fetch of the disabled network and match on it on a branch of the pending_status so that we can have the proper `NetworkStatus`

iljakuklic (approved):

crates/agglayer-rpc/src/lib.rs
```diff
@@ -640,29 +654,16 @@ where
                 // Network is in error if the latest pending certificate is in error
                 network_info.network_status = NetworkStatus::Error;
             }
+            _ if network_is_disabled => {
+                // If the network is disabled in storage, mark it as disabled
+                network_info.network_status = NetworkStatus::Disabled;
+            }
```
atanmarko: Should this be first item in the match  - if network is disabled we don't care if it is in error or no pending certificate?

atanmarko: Actually it is also misleading, we would return Unknown or Error status despite network being disabled.

PR #1164, opened 2025-11-06
feat: trace interactions with L1

### LINKED ISSUE #1153: Trace all interactions with L1
We should build our alloy provider with tracing enabled. This means in particular:
1. Switching to https://docs.rs/alloy/0.14.0/alloy/providers/struct.ProviderBuilder.html#method.on_client
2. Build the client with a layer https://docs.rs/alloy/0.14.0/alloy/rpc/client/struct.ClientBuilder.html#method.layer
3. Put a tracing layer there https://docs.rs/tower-http/latest/tower_http/trace/struct.TraceLayer.html

With this, we should get proper logs that allow us to fully understand exactly how we're interacting with L1.

iljakuklic:
The output is rather unsatisfying but it works. Could be improved in subsequent iterations.

Note: did not figure out how to do this in the HTTP level so it works on the jsonrpc level.

* Fixes #1153

### DIFF
--- a/crates/agglayer-node/Cargo.toml
+++ b/crates/agglayer-node/Cargo.toml
@@ -8,7 +8,7 @@ license.workspace = true
 workspace = true
 
 [dependencies]
-alloy = { workspace = true, features = ["full", "json-rpc"] }
+alloy.workspace = true
 axum = { workspace = true, features = ["tokio", "http1", "http2"] }
 arc-swap.workspace = true
 buildstructor.workspace = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ sp1-prover = "=5.0.1"
 sp1-zkvm = { version = "=5.0.0", default-features = false }
 
 # Core dependencies
-alloy = { version = "0.14.0", features = ["full"] }
+alloy = { version = "0.14.0", features = ["full", "json-rpc"] }
 alloy-primitives = { version = "1.3", features = ["serde", "k256"] }
 alloy-signer-local = { version = "0.14.0", features = ["keystore", "mnemonic"] }
 anyhow = "1.0"


crates/agglayer-node/src/l1_tracing.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        let span = tracing::debug_span!("L1 interaction");
+        let _span_guard = span.enter();
```
Ekleog-Polygon: Nit: this could be `#[instrument(skip_all, debug, name="L1 interaction")]` or similar I think?

It'd be even better if we could generate a random ID and put it in the span so that we could associate requests with responses, but :shrug: for the backport this'll definitely do :smile:

crates/agglayer-node/src/l1_tracing.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl<S> tower::Service<Request> for L1TraceService<S>
+where
+    S: tower::Service<Request, Response = Response> + Clone + Send + 'static,
+    S::Future: Send + 'static,
+    S::Error: std::fmt::Debug,
+{
+    type Response = Response;
+    type Error = S::Error;
+    type Future = futures::future::BoxFuture<'static, Result<Response, Self::Error>>;
```
Ekleog-Polygon, line 36: Nit: I think we have associated type impl trait now? Though I've been yearning for them for so long that I never remember whether they're actually here or I'm just dreaming of them :rofl:  https://doc.rust-lang.org/reference/types/impl-trait.html#r-type.impl-trait.return-in-trait

crates/agglayer-node/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license.workspace = true
 workspace = true
 
 [dependencies]
-alloy.workspace = true
+alloy = { workspace = true, features = ["full", "json-rpc"] }
```
Ekleog-Polygon: Nit: features in the workspace cargo.toml mean fewer rebuilds whenever we use `cargo build -p something`

Ekleog-Polygon (dismissed):
LGTM! I've added some very optional nits, will open a backport of this against 0.4 :smile:

crates/agglayer-node/src/l1_tracing.rs
iljakuklic: > Nit: I think we have associated type impl trait now? Though I've been yearning for them for so long that I never remember whether they're actually here or I'm just dreaming of them 🤣 https://doc.rust-lang.org/reference/types/impl-trait.html#r-type.impl-trait.return-in-trait

Looks like impl trait is supported in trait method return types but not in associated types like here. Getting this error:
```
error[E0658]: `impl Trait` in associated types is unstable
```

crates/agglayer-node/src/l1_tracing.rs
iljakuklic: To be honest I am now not sure if either one is correct. I wanted the immediate request printout and the reply printout to be in the span but now I don't know if the span is passed through into the future. I am inclined to say it is not.

iljakuklic: Since I am not actually sure how to make the span work correctly, I added an atomic counter to allocate a unique ID (within the node run) to each request and tie the requests and responses together that way.

Ekleog-Polygon: Good point! For that you'd need to call [`instrument`](https://docs.rs/tracing/0.1.41/tracing/trait.Instrument.html#method.instrument) on the future.

Anyway, I don't think it's actually important to have such a span; it feels like your current solution works just as well :smile:

PR #1165, opened 2025-11-07
fix: better error types and matching, cleaner option tuples, misc cleanups

Ekleog-Polygon:


crates/agglayer-contracts/src/settler.rs
```diff
@@ -119,45 +119,35 @@ where
...
-                                new_max_priority_fee_per_gas
+                                previous
+                                    .saturating_mul(
+                                        DEFAULT_GAS_PRICE_REPEAT_TX_INCREASE_FACTOR_PERCENTS,
+                                    )
+                                    .div_ceil(100)
+                                // Note: we use div_ceil here, so we will always
+                                // increase the fee until ceiling
                             })
                             .unwrap_or(estimate.max_priority_fee_per_gas)
```
adria0pol, line 144: Nit: wondering if this can be simplified with something like

```rust
                   let increase_fee = |fee: u128| {
                        // Note: we use div_ceil here, so we will always
                        // increase the fee until ceiling
                        fee.saturating_mul(DEFAULT_GAS_PRICE_REPEAT_TX_INCREASE_FACTOR_PERCENTS)
                            .div_ceil(100)
                            .max(self.gas_price_params.floor)
                            .min(self.gas_price_params.ceiling)
                    };

                    // This is repeated transaction, increase the previous max_fee_per_gas and
                    // max_priority_fee_per_gas by a factor
                    // If previous_max_priority_fee_per_gas is None, set it to estimated.
                    let adjust: Eip1559Estimation = Eip1559Estimation {
                        max_fee_per_gas: increase_fee(previous_max_fee_per_gas),
                        max_priority_fee_per_gas: previous_max_priority_fee_per_gas
                            .map(increase_fee)
                            .unwrap_or(estimate.max_priority_fee_per_gas),
                    };
```

adria0pol (approved):
Not able to find anything wrong.
LGTM

crates/agglayer-contracts/src/settler.rs
Ekleog-Polygon: It'd be a subtly different result: with your adjusted code, `estimate.max_priority_fee_per_gas` (which AFAIR is provided by RPC) would not get bound by floor/ceiling :smiling_face_with_tear: 

We can probably improve on this computation further still in other ways, but this is already the outcome of review from two months ago, so I'd vote for landing this (and follow-ups that I haven't extracted from my old review yet) and finally switching to the new work :smile:

PR #1167, opened 2025-11-07
fix: automatically allow replacement of settled certificates in error

### LINKED ISSUE #1157: Fix automatically error "Unable to replace a certificate in error that has already been settled"

crStiv:
Fixes #1157

Previously, attempting to replace a certificate in error that had already been successfully settled would return an error. This change allows automatic replacement in such cases.

Changes:
- Removed error return when settlement transaction succeeded
- Added info logging for automatic replacement
- Replacement now works for both failed and successful settlement transactions

### DIFF
--- a/tests/integrations/tests/certificate_settlement/retries.rs
+++ b/tests/integrations/tests/certificate_settlement/retries.rs
@@ -180,7 +180,8 @@ async fn regression_pushing_certificate_after_settling(#[case] state: Forest) {
 }
 
 /// Test that a certificate in error that has already been settled (transaction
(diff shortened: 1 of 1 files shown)


atanmarko:
@crStiv Thank you for your effort on this PR! We’ve recently enabled external contributors with CLA signatures, but it seems we’re still experiencing some issues with the CI for external contributors - we’ll need to look into that first.

If you have the time and interest, adding an integration test to `tests/integrations`  that demonstrates this scenario and proves it works would be a great addition to this PR.

Issue #1168, opened 2025-11-07
Fix CI execution for PRs from external contributors


atanmarko:
For example PR https://github.com/agglayer/agglayer/pull/1167 has couple of issues in CI run:

https://github.com/agglayer/agglayer/actions/runs/19166215269/job/54789191322?pr=1167
https://github.com/agglayer/agglayer/actions/runs/19166215132/job/54789442494?pr=1167

Freyskeyd:
Fixed

PR #1169, opened 2025-11-07
fix(node): off-by-one in epoch synchronizer on EpochEnded

GarmashAlex:
The epoch synchronizer incorrectly set current_epoch_number to the ended epoch on Event::EpochEnded(n), which prevented the just-ended epoch from being packed until a later event. The fix sets current_epoch_number to n.next() to maintain the variable’s semantics as the current open epoch and ensure immediate packing of the ended epoch. A regression test was added to verify that an EpochEnded(k) arriving during synchronization results in epoch k being packed and epoch k+1 being opened and left unpacked.


atanmarko:
Related issue https://github.com/agglayer/agglayer/issues/1168

PR #1183, opened 2025-11-18
feat: display in the logs aggkit version

### LINKED ISSUE #1096: Display in the logs aggkit version
Display the aggkit version from the `x-client-version` and `x-client-type` headers in the agglayer/aggkit-prover logs.

adria0pol:
This PR introduces support for parsing gRPC headers to extract client information for traces.
The following headers are parsed:

- [`x-client-type`](https://github.com/agglayer/aggkit/blob/299d6e330b9ef9b33f5df012de2b3a32a1e8ff23/grpc/client.go#L38)
- [`x-client-version`](https://github.com/agglayer/aggkit/blob/299d6e330b9ef9b33f5df012de2b3a32a1e8ff23/grpc/client.go#L35)
- `x-proto-version`
- `x-provider`

`x-client-type` and `x-client-version` are expected fields, so a warning is generated if they are not found

Example of log

```
[agglayer]     in agglayer_grpc_api::configuration_service::get_epoch_configuration with client: "x-client-type='aggkit',x-client-version='v0.7.1'"
```

Closes #1096

### DIFF
--- a/crates/agglayer-grpc-api/src/configuration_service.rs
+++ b/crates/agglayer-grpc-api/src/configuration_service.rs
@@ -26,14 +26,13 @@ where
     StateStore: Send + Sync + 'static,
     EpochsStore: Send + Sync + 'static,
 {
-    #[instrument(skip(self, request), level = "debug", fields(client = tracing::field::Empty))]
+    #[instrument(skip(self, _request), level = "debug", fields(
+        client = crate::client_info_from_metadata(_request.metadata())
+    ))]
     async fn get_epoch_configuration(
         &self,
-        request: tonic::Request<GetEpochConfigurationRequest>,
+        _request: tonic::Request<GetEpochConfigurationRequest>,
     ) -> Result<tonic::Response<GetEpochConfigurationResponse>, tonic::Status> {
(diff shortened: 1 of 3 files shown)


crates/agglayer-grpc-api/src/configuration_service.rs
```diff
@@ -25,10 +26,14 @@ where
     StateStore: Send + Sync + 'static,
     EpochsStore: Send + Sync + 'static,
 {
+    #[instrument(skip(self, request), level = "debug", fields(client = tracing::field::Empty))]
```
iljakuklic: Seems putting the function invocation directly in the instrument is accepted by the compiler:

```suggestion
    #[instrument(skip(self, request), level = "debug", fields(client = crate::client_info_from_metadata(request.metadata)))]
```

Is there a problem doing that? It's more concise and avoids the gymnastics with overwriting the empty field.

adria0pol: Of course. Done in b585ccff02276b23cf567e8eb571e214e12f08be

Issue #1136, opened 2025-10-23
Implement integration test with `AggchainData::MultisigAndAggchainProof`


atanmarko:
Also include some actual bridge exits and claims.

PR #1184, opened 2025-11-18
feat: new test certificate generation

atanmarko:
Related https://github.com/agglayer/agglayer/issues/1136

### DIFF
--- a/crates/agglayer-types/src/lib.rs
+++ b/crates/agglayer-types/src/lib.rs
@@ -13,7 +13,9 @@ pub mod network_info;
 mod proof_modes;
 
 #[cfg(feature = "testutils")]
-pub use certificate::{compute_signature_info, EMPTY_ELF};
+pub mod testutils {
+    pub use crate::certificate::{compute_signature_info, EMPTY_ELF};
+}
 pub use certificate::{
     Certificate, CertificateHeader, CertificateId, CertificateIndex, CertificateStatus, Height,
     Metadata, SettlementTxHash,
--- a/crates/pessimistic-proof-test-suite/src/forest.rs
+++ b/crates/pessimistic-proof-test-suite/src/forest.rs
@@ -1,8 +1,8 @@
 use agglayer_tries::smt::Smt;
 use agglayer_types::{
     aggchain_proof::AggchainData,
-    compute_signature_info,
     primitives::{keccak::keccak256, Hashable},
+    testutils::compute_signature_info,
     Address, Certificate, Digest, Height, LocalNetworkStateData, Signature, U256,
 };
 use alloy::signers::{local::PrivateKeySigner, SignerSync};
--- a/tests/integrations/tests/certificate_settlement/happy_path.rs
+++ b/tests/integrations/tests/certificate_settlement/happy_path.rs
@@ -48,7 +48,7 @@ async fn successfully_push_certificate(#[case] state: Forest) {
 #[case::type_0_ecdsa(crate::common::type_0_ecdsa_forest())]
 async fn send_multiple_certificates(#[case] mut state: Forest) {
     use agglayer_contracts::contracts::PolygonRollupManager::VerifyPessimisticStateTransition;
-    use agglayer_types::{aggchain_proof::AggchainData, compute_signature_info};
+    use agglayer_types::{aggchain_proof::AggchainData, testutils::compute_signature_info};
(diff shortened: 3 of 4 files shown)


crates/agglayer-types/src/certificate/testutils.rs
```diff
@@ -111,11 +112,185 @@ impl Certificate {
...
 
+    /// Generate a certificate with random bridge exits and custom parameters.
+    ///
+    /// # Arguments
+    /// * `network_id` - The network ID
+    /// * `height` - The certificate height
+    /// * `prev_local_exit_root` - The previous local exit root
+    /// * `num_bridge_exits` - Number of random bridge exits to generate
+    /// * `aggchain_data_type` - Type of AggchainData variant to use
+    /// * `version` - Signature commitment version
```
Freyskeyd: nit: Not sure that those info are necessary as the types are self explanatory

crates/agglayer-types/src/certificate/testutils.rs
```diff
@@ -111,11 +112,185 @@ impl Certificate {
...
+    pub fn new_for_test_custom(
```
iljakuklic, line 109: Can `new_for_test_with_version` be now implemented in terms of `new_for_test_custom`?

crates/agglayer-types/src/certificate/testutils.rs
```diff
@@ -111,11 +112,185 @@ impl Certificate {
...
+}
+
+/// Create a dummy STARK proof for testing purposes.
+/// This creates a minimal SP1 proof that can be used in tests.
+fn create_dummy_stark_proof() -> agglayer_interop_types::aggchain_proof::Proof {
+    use sp1_sdk::Prover;
 
+    // Use empty ELF file for minimal setup
+    const EMPTY_ELF: &[u8] =
+        include_bytes!("../../../agglayer-storage/src/types/certificate/tests/empty.elf");
```
iljakuklic: Maybe the empty elf file could be moved here in the `testutils` subfolder. and exported as `EMPTY_ELF` constant from the crate if `testutils` feature is enabled.

crates/agglayer-types/src/certificate/testutils.rs
```diff
@@ -111,11 +112,185 @@ impl Certificate {
         }
     }
 
+    /// Generate a certificate with random bridge exits and custom parameters.
```
hadjiszs, line 108: I wonder whether what @Ekleog-Polygon put in place some time ago with all the derive `Arbitrary` wasn't for this purpose? (initially for fuzzing iiuc)

in particular I can see that `Proof` has a custom impl already in `interop` for Arbitrary

```rust
#[cfg(feature = "testutils")]
impl<'a> arbitrary::Arbitrary<'a> for Proof {
```

I'm not as fluent as @Ekleog-Polygon in all that but I wonder if we can leverage these existing implementation in any way in your tests here?

atanmarko: @hadjiszs I have tried it, but have this `Failed to generate arbitrary Proof: IncorrectFormat`. Seems `Arbitrary` implementation is trying to deserialize from bincode from random values. Creating a valid proof that can be deserialized is quite complex I think, this functions needs to be fixed @Ekleog-Polygon

crates/agglayer-types/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub mod network_info;
 mod proof_modes;
 
 #[cfg(feature = "testutils")]
-pub use certificate::compute_signature_info;
+pub use certificate::{compute_signature_info, EMPTY_ELF};
```
iljakuklic: Nit: May be cleaner to keep these under testutils namespace. Tests will need to be adjusted for this to work though.
```suggestion
pub mod testutils {
    pub use certificate::{compute_signature_info, EMPTY_ELF};
}
```

iljakuklic (dismissed):

crates/agglayer-types/src/lib.rs
atanmarko: I agree

crates/agglayer-types/src/certificate/testutils.rs
hadjiszs: yeah it seems that it is primarily for fuzzing, which feels a bit what you're doing manually here in the end

PR #1194, opened 2025-11-19
fix: get_tx_status to return Pending when receipt.block_number is None

VolodymyrBg:
Ethereum JSON-RPC represents in-flight transactions with receipts that have blockNumber set to null, which denotes a pending (not yet mined) state. Our get_tx_status previously mapped a present receipt with a None block_number to NotFound, conflating known-but-unmined with unknown hashes. This change returns Pending for None, preserves Done once block_number < current L1 block and Pending for inclusion in the current block (0 confirmations), and keeps the early TxNotFound error when the receipt itself is absent. This aligns behavior with common client semantics and our existing tests that expect “pending” or “done” for real transactions and an error for fake hashes.


atanmarko (approved):
Make sense

crates/agglayer-jsonrpc-api/src/service/mod.rs
```diff
@@ -154,7 +154,7 @@ where
         let status = match receipt.block_number {
             Some(block_number) if block_number < current_block => TxStatus::Done,
             Some(_) => TxStatus::Pending,
-            None => TxStatus::NotFound,
+            None => TxStatus::Pending,
```
iljakuklic, line 157: Seems the `NotFound` variant can now be removed from `TxStatus`.

iljakuklic (approved):

crates/agglayer-jsonrpc-api/src/service/mod.rs
VolodymyrBg: > Seems the `NotFound` variant can now be removed from `TxStatus`.

Ah, I see, corrected

PR #1195, opened 2025-11-21
refactor: remove agglayer-prover

Freyskeyd:
CONFIG-CHANGE: `prover_entrypoint` is removed.

CONFIG-CHANGE: `prover` is now using a `ProverType` from `provers` repo

CONFIG-CHANGE: `grpc` config is now mandatory.

### DIFF
--- a/crates/agglayer-config/src/lib.rs
+++ b/crates/agglayer-config/src/lib.rs
@@ -125,14 +125,6 @@ pub struct Config {
 
     #[serde(default)]
     pub grpc: grpc::GrpcConfig,
-
-    /// Extra Certificate signer per network.
-    /// Signatures is expected to be performed on the same commitment as
-    /// the certificate signature, which is the V2 commitment for now.
-    #[serde(default)]
-    #[serde_as(as = "HashMap<DisplayFromStr, _>")]
-    #[serde(skip_serializing_if = "HashMap::is_empty")]
-    pub extra_certificate_signer: HashMap<u32, Address>,
 }
 
 impl Config {
@@ -188,7 +180,6 @@ impl Config {
             debug_mode: false,
             mock_verifier: false,
             grpc: Default::default(),
-            extra_certificate_signer: Default::default(),
         }
     }
 
--- a/crates/agglayer-node/src/node.rs
+++ b/crates/agglayer-node/src/node.rs
@@ -10,10 +10,7 @@ use agglayer_jsonrpc_api::{
 };
 use agglayer_signer::ConfiguredSigner;
 use agglayer_storage::{
-    storage::{
-        backup::{BackupClient, BackupEngine},
-        DB,
-    },
+    storage::backup::{BackupClient, BackupEngine},
     stores::{
         debug::DebugStore, epochs::EpochsStore, pending::PendingStore, state::StateStore,
         PerEpochReader as _,
@@ -89,14 +86,8 @@ impl Node {
         }
 
         // Initializing storage
-        let pending_db = Arc::new(DB::open_cf(
-            &config.storage.pending_db_path,
-            agglayer_storage::storage::pending_db_cf_definitions(),
-        )?);
-        let state_db = Arc::new(DB::open_cf(
-            &config.storage.state_db_path,
-            agglayer_storage::storage::state_db_cf_definitions(),
-        )?);
+        let pending_db = Arc::new(PendingStore::init_db(&config.storage.pending_db_path)?);
+        let state_db = Arc::new(StateStore::init_db(&config.storage.state_db_path)?);
 
         // Initialize backup engine
         let backup_client = if let BackupConfig::Enabled {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -274,7 +274,6 @@ dependencies = [
  "eyre",
  "http 1.3.1",
  "prost 0.13.5",
- "rstest",
  "serde",
  "serde_json",
  "tokio",
--- a/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
@@ -2,7 +2,10 @@ use std::{collections::HashSet, sync::Arc};
 
 use agglayer_storage::{
     columns::latest_settled_certificate_per_network::SettledCertificate,
-    stores::{PendingCertificateReader, PendingCertificateWriter, StateReader, StateWriter},
+    stores::{
+        PendingCertificateReader, PendingCertificateWriter, StateReader, StateWriter,
+        UpdateEvenIfAlreadyPresent, UpdateStatusToCandidate,
+    },
 };
 use agglayer_types::{
     Certificate, CertificateHeader, CertificateStatus, CertificateStatusError, Digest,
@@ -271,7 +274,8 @@ where
                                 if let Err(error) = self.state_store.update_settlement_tx_hash(
                                     &certificate_id,
                                     contract_settlement_tx_hash,
-                                    true,
+                                    UpdateEvenIfAlreadyPresent::Yes,
+                                    UpdateStatusToCandidate::Yes,
                                 ) {
                                     error!(
                                         ?error,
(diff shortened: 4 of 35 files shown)


crates/agglayer-node/src/node.rs
```diff
@@ -219,11 +220,18 @@ impl Node {
         );
         tracing::debug!("RollupManager created");
 
+        let (_vkey, prover_executor) =
+            prover_executor::Executor::create_prover(config.prover.clone(), pessimistic_proof::ELF)
+                .await
+                .unwrap();
+
+        let prover_buffer = Buffer::new(prover_executor, 1024);
```
adria0pol: nit: Add a comment about why `bound=1024` is selected in ` let prover_buffer = Buffer::new(prover_executor, 1024);`

Cargo.toml
```diff
@@ -63,6 +60,9 @@ prover-logger = { git = "https://github.com/agglayer/provers.git", tag = "v1.8.0
 prover-executor = { git = "https://github.com/agglayer/provers.git", tag = "v1.8.0" }
 prover-engine = { git = "https://github.com/agglayer/provers.git", tag = "v1.8.0" }
 prover-utils = { git = "https://github.com/agglayer/provers.git", tag = "v1.8.0" }
+# agglayer-prover = { path = "crates/agglayer-prover" }
```
atanmarko: Cleanup

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -37,27 +33,30 @@ mod l1_context;
...
 pub struct CertifierClient<PendingStore, L1Rpc> {
     /// The pending store to fetch and store certificates and proofs.
     pending_store: Arc<PendingStore>,
-    /// The prover service client.
-    prover: PessimisticProofServiceClient<Channel>,
     /// The local CPU verifier to verify the generated proofs.
     verifier: Arc<CpuProver>,
     /// The verifying key of the SP1 proof system.
     verifying_key: SP1VerifyingKey,
+    prover: ProverService,
```
atanmarko, line 49: Could you add the doc comment here to be consistent?

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -237,95 +230,45 @@ where
...
-                            }
-                        }
-                        ErrorKind::Unspecified => {
-                            CertificationError::InternalError(source_error.message().to_string())
-                        }
-                    }
+                if let Some(error) = source_error.downcast_ref::<prover_executor::Error>() {
+                    error!("Failed to generate proof: {}", error);
+
+                    CertificationError::ProverFailed(error.to_string())
```
atanmarko, line 264: Should we keep the source `prover_executor` error here as part of `ProverFailed`, now that we have the type and details?

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -237,95 +230,45 @@ where
...
                 } else {
-                    warn!(
-                        "Failed to deserialize the error details coming from the prover: \
-                         {source_error:?}"
-                    );
+                    error!("Failed to generate proof: {:?}", source_error);
 
-                    CertificationError::InternalError(source_error.message().to_string())
+                    CertificationError::ProverReturnedUnspecifiedError
```
atanmarko: Could you keep the `BoxError` string here as detail of `ProverReturnedUnspecifiedError`?

crates/agglayer-aggregator-notifier/src/certifier/tests.rs
```diff
@@ -26,31 +25,19 @@ use crate::{CertifierClient, ELF};
...
 
-    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    let (_vkey, prover) = prover_executor::Executor::create_prover(
+        ProverType::MockProver(MockProverConfig::default()),
+        ELF,
+    )
+    .await
+    .unwrap();
+
+    let buffer = Buffer::new(prover, 1024);
```
atanmarko: Some const for buf size would be great.

crates/agglayer-node/src/node.rs
```diff
@@ -219,11 +220,18 @@ impl Node {
         );
         tracing::debug!("RollupManager created");
 
+        let (_vkey, prover_executor) =
+            prover_executor::Executor::create_prover(config.prover.clone(), pessimistic_proof::ELF)
+                .await
+                .unwrap();
```
atanmarko: Should we return here with the error details? Or at least inspect them?

crates/agglayer-config/src/grpc.rs
```diff
@@ -0,0 +1,40 @@
...
+    #[serde(
+        skip_serializing_if = "same_as_default_max_decoding_message_size",
+        default = "default_max_decoding_message_size"
+    )]
+    pub max_decoding_message_size: usize,
+    #[serde(
+        skip_serializing_if = "same_as_default_max_encoding_message_size",
+        default = "default_max_encoding_message_size"
+    )]
+    pub max_encoding_message_size: usize,
```
iljakuklic, line 17: Would be nice to use some sort of [human-readable format](https://crates.io/crates/byte-unit) for the size specification. The config values here will be huge numbers which are rather difficult to read.

crates/agglayer-config/src/lib.rs
```diff
@@ -128,7 +121,7 @@ pub struct Config {
     pub mock_verifier: bool,
 
     #[serde(default, skip_serializing_if = "crate::is_default")]
-    pub grpc: GrpcConfig,
+    pub grpc: grpc::GrpcConfig,
```
iljakuklic, line 127: Not strictly related to this PR.

I would suggest that `#[serde(skip_serializing_if = "crate::is_default")]` should only be used if the default value is an obvious trivial value.

That way, non-obvious values will be visible when invoking `agglayer config`, potentially avoiding surprises. It also makes the configuration values more visible, increasing the user awareness about the settings the software supports.

On the other hand, generating config file will "fix" those values so changing the defaults in the code will not be reflected when using a config file generated from an older version. There might be other corner cases I had not considered.

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
Freyskeyd: Unfortunately, it'll be a breaking for the storage

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
Freyskeyd: I'll fallback on `ProverFailed`, `ProverReturnedUnspecifiedError` doesn't support any params

crates/agglayer-config/src/grpc.rs
Freyskeyd: Let's create a ticket for that: #1204

crates/agglayer-config/src/lib.rs
Freyskeyd: If we remove this, it makes the config parsing failing if the value is missing, meaning that we need to update all env configs, right?

iljakuklic: > If we remove this, it makes the config parsing failing if the value is missing, meaning that we need to update all env configs, right?

I meant just removing the `skip_seralizing_if` part, `#[serde(default)]` would stay there. So `agglayer config` would print out these config values but they would still be optional.

atanmarko (approved):

atanmarko (approved):

crates/agglayer-config/src/lib.rs
```diff
@@ -111,13 +108,12 @@ pub struct Config {
     #[serde(default)]
     pub storage: storage::StorageConfig,
 
-    /// AggLayer prover entrypoint.
-    #[serde(default = "default_prover_entrypoint")]
-    #[serde(skip_serializing_if = "String::is_empty")]
-    pub prover_entrypoint: String,
```
iljakuklic, line 117: Do we issue an error or warning if this key is still present, indicating that the config was not fully updated?

iljakuklic (approved):

crates/agglayer-config/src/lib.rs
Freyskeyd: the config at start will not be parsable and return: `unknown field `prover-entrypoint`, expected one of `

Issue #1090, opened 2025-10-06
Cleanup the `outbound.rpc.settle` agglayer configuration


atanmarko:
Remove the `max-retries` and `retry-interval` fields. Rename `settlement-timeout` to `settlement-tx-timeout`. 
Update the defaults for confirmations.

PR #1203, opened 2025-11-25
feat: add second account for validium transactions

adria0pol:
closes https://github.com/agglayer/agglayer/issues/1158
closes https://github.com/agglayer/agglayer/issues/1090

> [!NOTE]
> https://github.com/0xPolygon/kurtosis-cdk/blob/main/templates/bridge-infra/agglayer-config.toml needs to be updated

### CONFIG-CHANGE: Add new GCP KMS env vars and config vars

1) Adds new environment variables for selecting the signing key for certificate settlement and tx settlement that **MUST live in the same KMS keyring**.
	
	- Certificate settlement
	  - `GOOGLE_KEY_NAME_PP_SETTLEMENT`
	  - `GOOGLE_KEY_VERSION_PP_SETTLEMENT`
	- Tx settlement
	  - `GOOGLE_KEY_NAME_TX_SETTLEMENT`
	  - `GOOGLE_KEY_VERSION_TX_SETTLEMENT`
	
2) If these env variables are not set, we fall back to the previous existing keys to ensure smooth migration, is expected that these env vars will be deprecated
	
	- `GOOGLE_KEY_NAME`
	- `GOOGLE_KEY_VERSION`
	
3) If they are not set, it will fallback to configuration variables in `[auth.gcp-kms]`
	
	- `pp-settlement-key-name`
	- `pp-settlement-key-version`
	- `tx-settlement-key-name`
	- `tx-settlement-key-version`
	
4) If they are not set, it will fallback to configuration variables in `[auth.gcp-kms]` that is expected to be deprecated
	
	- `key-name`
	- `key-version`

### CONFIG-CHANGE:  Configuration file MUST define explicit outbounds for certificate settlement and tx settlement

Previously there was the only entry
	
- `[outbound.rpc.settle]`
	
now this entry can be replaced with 
	
- `[outbound.rpc.settle-cert]` configuration will be applied to `GOOGLE_KEY_NAME_PP_SETTLEMENT` or the specified one in fallback.
	
- `[outbound.rpc.settle-tx]` configuration will be applied to `GOOGLE_KEY_NAME_TX_SETTLEMENT` or the specified one in fallback.

### Signing Keystore now accepts one or two signing keys

Since we have to provide ways to emulate the possibility of using one or two signing keys:

- if only one key is found in keystore it will work for both certificate and tx settlement signing keys.
- if more that one key is found in keystore the first will be used for certificate settlement and the second for tx settlement.

### DIFF
--- a/crates/agglayer-config/tests/snapshots/auth_config__auth_transition.snap
+++ b/crates/agglayer-config/tests/snapshots/auth_config__auth_transition.snap
@@ -2,7 +2,6 @@
 source: crates/agglayer-config/tests/auth_config.rs
 expression: config
 ---
-prover-entrypoint = "http://127.0.0.1:8080"
 prover-buffer-size = 100
 
 [full-node-rpcs]
--- a/crates/agglayer-config/tests/snapshots/auth_config__auth_update.snap
+++ b/crates/agglayer-config/tests/snapshots/auth_config__auth_update.snap
@@ -2,7 +2,7 @@
 source: crates/agglayer-config/tests/auth_config.rs
 expression: config
 ---
-prover-entrypoint = "http://127.0.0.1:8080"
+prover-buffer-size = 100
 
 [full-node-rpcs]
 


crates/agglayer-config/src/auth.rs
```diff
@@ -70,6 +70,25 @@ pub struct GcpKmsConfig {
     #[serde_as(as = "NoneAsEmptyString")]
     #[serde(default)]
     pub keyring: Option<String>,
+
+    // Added to support distinct keys for cert and tx signing, falling back to
+    // the older single key if not specified
+    #[serde(alias = "CertSettlementKeyName")]
+    #[serde_as(as = "NoneAsEmptyString")]
+    #[serde(default)]
+    pub cert_settlement_key_name: Option<String>,
```
iljakuklic: I don't think you need to introduce the serde alias. Some keys have it for beckwards compatibility with some pre-historic version of agglayer config. New keys do not need it.

Correct me if I'm wrong @Freyskeyd

crates/agglayer-signer/src/lib.rs
```diff
@@ -75,6 +72,39 @@ impl ConfiguredSigner {
...
+    // The signer for certificate settlement.
+    pub cert_settlement: ConfiguredSigner,
+    // The signer for transaction settlement, if not defined is expected
+    // that `cert_settlement` signer will be used.
+    pub tx_settlement: Option<ConfiguredSigner>,
+}
+
+impl ConfiguredSigners {
+    /// Get either a local wallet or GCP KMS signer based on the configuration.
+    pub async fn new(config: Arc<Config>) -> Result<Self, Error> {
```
iljakuklic: Nit: I don't see any `Arc`-specific operations in the method. This is a bit more general:
```suggestion
    pub async fn new(config: &Config) -> Result<Self, Error> {
```

crates/agglayer-signer/src/lib.rs
```diff
@@ -39,24 +39,21 @@ impl ConfiguredSigner {
...
-            }
-        }
+        let signer2 = if let Some(pk2) = local.private_keys.get(1) {
+            Some(
+                PrivateKeySigner::decrypt_keystore(&pk2.path, &pk2.password)?
+                    .with_chain_id(Some(chain_id)),
+            )
+        } else {
+            None
+        };
```
iljakuklic: Nit: Would this work too?
```suggestion
        let signer2 = local.private_keys.get(1).map(|pk2| {
            PrivateKeySigner::decrypt_keystore(&pk2.path, &pk2.password)?
                .with_chain_id(Some(chain_id)),
        });
```

adria0pol: The '?'  inside the closure when it does not return an Error ( it's an Option ) makes stuff little complicated.
I also do not like the `else { None }` pattern but the alternative I find by using a map is

```
let signer2 = local.private_keys.get(1)
    .map(|pk2| {
        PrivateKeySigner::decrypt_keystore(&pk2.path, &pk2.password)
            .map(|s| s.with_chain_id(Some(chain_id)))
    })
    .transpose()?;
```

Anyway, I'm going to change it to 

```
        let mut signer2 = None;
        if let Some(pk2) = local.private_keys.get(1) {
            signer2 = Some(
                PrivateKeySigner::decrypt_keystore(&pk2.path, &pk2.password)?
                    .with_chain_id(Some(chain_id)),
            )
        };
```

that I think that makes everything cleaner.

iljakuklic: > The '?' inside the closure when it does not return an Error ( it's an Option ) makes stuff little complicated. I also do not like the `else { None }` pattern but the alternative I find by using a map is

Ah did not notice the `?`, sorry about that.

crates/agglayer-signer/src/lib.rs
adria0pol: 100%, done in b1133be

crates/agglayer-signer/src/lib.rs
adria0pol: changed in b1133be

crates/agglayer-node/src/node.rs
```diff
@@ -183,31 +182,59 @@ impl Node {
...
+                )
+            };
+
+            let signers = ConfiguredSigners::new(&config).await?;
+            let provider_cert = fn_build_provider(signers.cert_settlement);
+
+            let provider_tx = if let Some(tx_settlement) = signers.tx_settlement {
+                fn_build_provider(tx_settlement)
+            } else {
+                provider_cert.clone()
```
atanmarko, line 203: Nit: warning that we are using cert signer for tx provider.

atanmarko (approved):
LGTM

crates/agglayer-config/src/auth.rs
Freyskeyd: Yes we do not need aliases

Freyskeyd: Maybe this could be name `pp` instead of `cert` ?

crates/agglayer-config/src/auth.rs
```diff
@@ -70,6 +70,25 @@ pub struct GcpKmsConfig {
...
+    #[serde(alias = "CertSettlementKeyVersion")]
+    #[serde(default)]
+    pub cert_settlement_key_version: Option<u64>,
+    #[serde(alias = "TxSettlementKeyName")]
+    #[serde_as(as = "NoneAsEmptyString")]
+    #[serde(default)]
+    pub tx_settlement_key_name: Option<String>,
+    #[serde(alias = "TxSettlementKeyVersion")]
+    #[serde(default)]
+    pub tx_settlement_key_version: Option<u64>,
```
Freyskeyd: Missing doc for those fields

crates/agglayer-gcp-kms/src/lib.rs
```diff
@@ -25,6 +37,52 @@ pub struct KMS {
...
+// 1. The primary environment variable.
+// 2. The fallback environment variable (if provided).
+// 3. The primary configuration value.
+// 4. The fallback configuration value (if provided).
+// If none are found, it returns an error.
+fn from_env_or_conf<T>(
+    env_key: &str,
+    env_key_fallback: Option<&str>,
+    config_value: &Option<T>,
+    config_value_fallback: &Option<T>,
```
iljakuklic: General nit: Passing `Option<&T>` is usually preferable to `&Option<T>`. It is a bit more flexible. The caller can easily use `.as_ref()` to get `Option<&T>` but going the other way around is not always easy.

crates/agglayer-gcp-kms/src/lib.rs
```diff
@@ -25,6 +37,52 @@ pub struct KMS {
     config: GcpKmsConfig,
 }
 
+// Helper function to get a value from environment variables or configuration
+// with fallbacks.
+// The function returns the value in the following order of precedence:
+// 1. The primary environment variable.
+// 2. The fallback environment variable (if provided).
+// 3. The primary configuration value.
+// 4. The fallback configuration value (if provided).
```
iljakuklic: Hmm, I don't have a super rigorous justification for that but the following order seems a bit more natural for me:

1. The primary environment variable.
2. The primary configuration value.
3. The fallback environment variable (if provided).
4. The fallback configuration value (if provided).

I.e. if something is in the configuration, it is taken as is unless overridden by an env variable for that specific config.

iljakuklic (commented):
Some more random observations. Will have a proper look next week.

crates/agglayer-gcp-kms/src/lib.rs
adria0pol: The reasoning about this order was deduced from the original comment:

```
    /// This method will first attempt to use the environment variables, and if
    /// they are not set, it will fall back to the values specified in the
    /// configuration.
```

So fallback was the configuration.

Also it matches with what I suspect that secrets are managed in GCP KMS - in the environment instead configuration file as a primary soure.

Anyway, if it still sounds more natural to you this order, I'll change it!

crates/agglayer-gcp-kms/src/lib.rs
adria0pol: This is a fantastic nit, thanks Lukáš !

crates/agglayer-node/src/node.rs
adria0pol: Yes, I have to use the same provider to avoid problems about having two objects managing the same nonce.

atanmarko: I mean add the warn log :)

adria0pol: > I mean add the warn log :)

Ah, ok! 😆

crates/agglayer-gcp-kms/src/lib.rs
adria0pol: Since it's an internal helper function and turning `&Option<T>` to  `Option<&T>` makes code a little less ergonomic, I keep as it.

Anyway, something new to take into consideration :)

crates/agglayer-config/src/auth.rs
adria0pol: Done in db909e434778d80f35abe09aebb3dae470c6b40a

adria0pol: Done in db909e434778d80f35abe09aebb3dae470c6b40a

crates/agglayer-config/src/auth.rs
adria0pol: Done in db909e434778d80f35abe09aebb3dae470c6b40a.

crates/agglayer-node/src/node.rs
adria0pol: Done in db909e434778d80f35abe09aebb3dae470c6b40a

crates/agglayer-gcp-kms/src/lib.rs
iljakuklic: Ok, makes sense

crates/agglayer-config/src/outbound.rs
```diff
@@ -18,7 +18,8 @@ pub struct OutboundConfig {
 #[serde(rename = "rpc", rename_all = "kebab-case")]
 pub struct OutboundRpcConfig {
     /// Outbound configuration of the RPC settle function call.
-    pub settle: OutboundRpcSettleConfig,
+    pub settle_tx: OutboundRpcSettleConfig,
+    pub settle_cert: OutboundRpcSettleConfig,
```
iljakuklic, line 23: What happens if the config still contains the `settle` key? It would be nice to add a test checking an error is raised in that case.

crates/agglayer-config/tests/snapshots/auth_config__auth_transition.snap
```diff
@@ -28,13 +28,20 @@ send-tx = "unlimited"
...
-[outbound.rpc.settle]
+[outbound.rpc.settle-tx]
 max-retries = 30
 retry-interval = "10s"
 confirmations = 1
-settlement-timeout = "20m"
 
-[outbound.rpc.settle.gas-price]
+[outbound.rpc.settle-tx.gas-price]
+ceiling = "100gwei"
```
iljakuklic, line 37: IIRC, the gas limits currently only apply to certificates. We should at some point introduce a module that handles submitting transactions and gas escalation generically for both. However, in the meantime, we should probably warn if these settings are set to non-default values for `settle-tx`.

crates/agglayer-config/src/outbound.rs
adria0pol: ~Absolutely. Done in [1587075](https://github.com/agglayer/agglayer/pull/1203/commits/1587075c117f5b986c4a9dc14b25ae594fb31550)~

adria0pol:
> @adria0pol I was refering to backward compatibility test for `outbound.rpc.settle` - to initialize properly both account providers if only deprecated config is provided, e.g.:
> 
> ```
> [outbound.rpc.settle]
>     max-retries = 60
>     retry-interval = "60s"
>     confirmations = 5
>     settlement-timeout = "20m"
>     gas-multiplier-factor = 130
> 
> 
> [outbound.rpc.settle.gas-price]
> 	multiplier = 1.5
> 	floor = "1gwei"
> 	ceiling = "100gwei"
> ```
> 
> So that we don't break current setups.

Ok, I followed @atanmarko  https://github.com/agglayer/agglayer/pull/1203#discussion_r2580200288 comment, but AFAIU this should work smoothly with any change in the configuration. Let's me change this. I will move it from `BREAKING-CHANGE` to `CONFIG-CHANGE`

crates/agglayer-config/tests/snapshots/auth_config__auth_transition.snap
adria0pol: Done in c90d26a7ba2c3148fa2219a5715a991d85e1c0fa

adria0pol:
@iljakuklic @Freyskeyd @atanmarko update

https://github.com/agglayer/agglayer/pull/1203/commits/f417b4000213ff8b3a4ab5375372e6c6b22413b7 changed the way to handle deprecated fields: now is done via custom deserializers, it keeps the `Config` struct clean and legacy parameters removal is just about removing the custom deserialization and derive `Deserialize` again.

crates/agglayer-node/src/node.rs
```diff
@@ -174,31 +175,68 @@ impl Node {
...
+        // Create RPC clients, note that they can be the same signer
+        // or not depending on the configuration. If is the same signer
+        // we share the nonce management too.
+
+        if config.outbound.rpc.settle_tx != OutboundRpcSettleConfig::default() {
+            warn!(
+                "Using non-default values for outbound settle_tx: {:?}",
+                config.outbound.rpc.settle_tx
+            );
+        }
```
atanmarko: Nit: not sure that we need this warning?

atanmarko (dismissed):
LGTM

crates/agglayer-node/src/node.rs
adria0pol: Suggested in https://github.com/agglayer/agglayer/pull/1203#discussion_r2580215438

> [iljakuklic](https://github.com/iljakuklic)
IIRC, the gas limits currently only apply to certificates. We should at some point introduce a module that handles submitting transactions and gas escalation generically for both. However, in the meantime, we should probably warn if these settings are set to non-default values for settle-tx.

atanmarko: @adria0pol Ok, but this is about gas limit - this would also issue warning if gas limit is the same and confirmations, timeout, fees are different (and both are configurable).

adria0pol: > @adria0pol Ok, but this is about gas limit - this would also issue warning if gas limit is the same and confirmations, timeout, fees are different (and both are configurable).

Mmmm.. could you clarify, plz?

- if **these settings are set to non-default values** for settle-tx or
- if **gas limit is set to non-default values** for settle-tx or
- another thing

atanmarko: Ok, I was also confusing. We have tx creation on two places - one for `send_tx` in `fn settle()`, other for certificates settlement in `verify_pessimistic_trusted_aggregator`. Those two tx creation code is  different. For example, confirmations/timeout are used on both places, so if they are different in config, we should not have a warning. Also `gas_price_params` - `multiplier`, `floor`, `ceiling` are explicitly set in both tx preparation code, so it is fine if they are different - they do not need to be same as default and we do not need warning there.  Other config properties are used only for certificate tx creation, for example gas limit (`gas_multiplier_factor`) I think (that is what @iljakuklic was reffering to - we don't set gas limit for `send_tx`) and even `OutboundRpcSettleConfig::default()` gas limit would not be used there, as it is not explicitly set (default `alloy` library gas limit would be used in `async fn settle`).

atanmarko: Having all that in mind, I think having this warning is more confusing than not having it. Fox `send_tx`, custom config values or `OutboundRpcSettleConfig::default()` will be set for most of the paramters, but for transaction gas limit none of those  would be used for `gas_multiplier_factor` but some internal alloy `Provider` value.

adria0pol: @iljakuklic  ?

atanmarko: @iljakuklic I guess @adria0pol could just add [this block](https://github.com/agglayer/agglayer/blob/main/crates/agglayer-contracts/src/settler.rs#L97) to [settle()](https://github.com/agglayer/agglayer/blob/main/crates/agglayer-jsonrpc-api/src/kernel/mod.rs#L340) and all would be good, no warnings and confusions anymore.

iljakuklic: > @iljakuklic I guess @adria0pol could just add [this block](https://github.com/agglayer/agglayer/blob/main/crates/agglayer-contracts/src/settler.rs#L97) to [settle()](https://github.com/agglayer/agglayer/blob/main/crates/agglayer-jsonrpc-api/src/kernel/mod.rs#L340) and all would be good, no warnings and confusions anymore.

Yea that would work too, also adding this afterwards as the block you posted only applies the multiplier, not the limits:

```rust
    .max(self.gas_price_params.floor)
    .min(self.gas_price_params.ceiling)
```

atanmarko: @iljakuklic `gas_price_params.floor` and `gas_price_params.ceiling` are limits for gas price, not gas amount.

iljakuklic: > @iljakuklic `gas_price_params.floor` and `gas_price_params.ceiling` are limits for gas price, not gas amount.

Yea but I see the gas price floor and ceiling in the config under the `send-tx` section so it should be applied too. Probably not in the way I described, apologies for that.

Another possibility is to just ignore the issue for the time of being and implement it in a principled way once the dedicated transaction submission component is implemented, and use it for both validium txs and for certificates.

atanmarko (approved):

crates/agglayer-node/src/node.rs
adria0pol: Done in https://github.com/agglayer/agglayer/pull/1203/commits/47a94f5500e7ceb5cf93a333218620d142d0aa3b

crates/agglayer-config/src/outbound.rs
```diff
@@ -14,11 +15,51 @@ pub struct OutboundConfig {
...
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(rename = "rpc", rename_all = "kebab-case")]
+        pub struct Intermediate {
+            pub settle_tx: Option<OutboundRpcSettleConfig>,
+            pub settle_cert: Option<OutboundRpcSettleConfig>,
+            pub settle: Option<OutboundRpcSettleConfig>,
+        }
```
iljakuklic, line 37: Note: You could probably implement the match logic below using something like:
```suggestion
        #[derive(Deserialize)]
        #[serde(untagged, rename_all = "kebab-case")]
        pub enum IntermediateInner {
            PerType {
                pub settle_tx: OutboundRpcSettleConfig,
                pub settle_cert: OutboundRpcSettleConfig,
            },
            Unified {
                pub settle: OutboundRpcSettleConfig,
            },
        }
        
        #[derive(Deserialize)]
        pub struct Intermediate {
            rpc: IntermediateEnum,
        }
```

BUT this would probably result in worse error messages and not sure the code is cleaner either. Still useful to know the trick.

adria0pol: For sure it's clean. Anyway let's me finish this PR, I'll keep in mind for next things :)

PR #1205, opened 2025-11-27
refactor: database initialization

iljakuklic:
A small database initialization refactoring. The idea is to abstract the broader code from the low-level column definitions to accommodate the database migration code a bit better.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,6 +201,7 @@ name = "agglayer-contracts"
 version = "0.1.0"
 dependencies = [
  "agglayer-primitives",
+ "agglayer-types",
  "alloy",
  "async-trait",
  "eyre",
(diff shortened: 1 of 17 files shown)


crates/agglayer-storage/src/stores/pending/mod.rs
```diff
@@ -26,16 +26,16 @@ pub struct PendingStore {
 }
 
 impl PendingStore {
+    pub fn init_db(path: &Path) -> Result<DB, crate::storage::DBError> {
```
Freyskeyd, line 29: Should it be `init_db` or `init_store` or even `new` ?

atanmarko: Both `init_db` and `init_store` are fine to me. `new` may be a bit misleading in this context, as it associate more on software runtime component than on storage on disk.

atanmarko (approved):

crates/agglayer-storage/src/stores/pending/mod.rs
iljakuklic: > Should it be `init_db` or `init_store` or even `new` ?

I would expect `new` to return a `PendingStore`. This is just initializing the internal database. Ideally this would be abstracted from the clients even more but for this PR, the focus was just that the clients don't manipulate the column family descriptors directly so that the db migration logic may be fully encapsulated within the `agglayer-storage` crate.

Freyskeyd (approved):

PR #1210, opened 2025-12-03
fix(rpc): propagate errors from get_network_info settled enrichment

Galoretka:
Replaced the Option::map usage in get_network_info with an explicit if-let to ensure errors in the settled certificate enrichment path are propagated as GetNetworkInfoError::InternalError. Previously, the closure returned Result but the Option was ignored, swallowing errors. Added tests to assert errors from read_local_network_state and get_latest_settled_claim now bubble up.


iljakuklic (commented):
Hi @Galoretka. Thank you for your contribution. This looks good to me at first glance.

atanmarko (approved):

atanmarko:
Tests need fix

iljakuklic (approved):

Ekleog-Polygon (approved):
LGTM! Though CI does need fixes.

@atanmarko I don't recall, was it on purpose that you were ignoring errors there? I remember some sort of discussion like this. Right now this change looks correct to me, but I'm wondering if I haven't lost some context.

PR #1211, opened 2025-12-03
feat: database migration mechanism core function

### LINKED ISSUE #1212: Implement storage migration core functionality
* PR #1211

iljakuklic:
The first version of the database migration framework. It follows the design described in agglayer/protocol-research#206. The user-facing core functionality is ready. It allows one to use a database builder to define a number of migration steps, starting at the "original" version of the database.

See the tests for example usage for now. Proper docs will come later.

The migration integrity feature is largely missing. The idea is to take a hash of the schema at each migration step and persist it in the migration record column, together with some other metadata. Then verify it during db initialization to check the migration steps have not been tampered with. This is planned as a follow-up.

* design agglayer/protocol-research#206
* fixes #1212 
* related #1185

### DIFF
--- a/crates/agglayer/src/main.rs
+++ b/crates/agglayer/src/main.rs
@@ -38,9 +38,7 @@ fn main() -> eyre::Result<()> {
                 .enable_all()
                 .build()?
                 .block_on(async move {
-                    let vkey_hex = compute_program_vkey(ELF)
-                        .await
-                        .context("Failed to compute program vkey");
+                    let vkey_hex = compute_program_vkey(ELF).await;
                     match vkey_hex {
                         Ok(vkey_hex) => println!("{vkey_hex}"),
                         Err(error) => eprintln!("{error:?}"),


crates/agglayer-storage/src/storage/migration/error.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[error(
+        "Fewer migration steps declared in the code than recorded in database \
+         (declared: {declared}, recorded: {recorded}). This indicates existing migration steps \
+         were removed from the code, which is not allowed."
+    )]
+    FewerStepsDeclared { declared: u32, recorded: u32 },
+}
+
+#[derive(Debug, Error)]
+#[error("Migration failed at step {step}")]
```
atanmarko, line 32: Maybe you could also add display source error printout here?

crates/agglayer-storage/src/storage/migration/error.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[source]
+    pub details: DBMigrationErrorDetails,
+}
+
+#[derive(Debug, Error)]
+pub enum DBMigrationErrorDetails {
+    #[error(transparent)]
+    Database(#[from] DBError),
+
+    #[error("Custom migration error")]
```
atanmarko, line 44: Print error display output.

crates/agglayer-storage/src/storage/migration/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+mod error;
+mod migration_cf;
+mod record;
+
+pub use error::{DBMigrationError, DBMigrationErrorDetails, DBOpenError};
+use migration_cf::MigrationRecordColumn;
+use record::MigrationRecord;
+
+/// Database builder taking care of database migrations.
+pub struct Builder {
```
atanmarko, line 19: Nit: is `Builder` appropriate name? Maybe `DbMigrator` or similar?

crates/agglayer-storage/src/storage/migration/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+
+        if step >= self.start_step {
+            info!("Running migration step {step}");
+            step_fn(&mut self)?;
+            self.db.put::<MigrationRecordColumn>(&self.step, &())?;
+        } else {
+            debug!("Step already recorded, skipping");
+        }
+
+        self.step += 1;
```
atanmarko, line 247: Maybe you could move `self.step` management to `perform_step` function. Or move whole `perform_step_impl` implementation to `perform_step`?

crates/agglayer-storage/src/storage/migration/test/corrupt.rs
```diff
@@ -0,0 +1,103 @@
...
+            .finalize(CFS_V1)?;
+
+        // Verify migration record contains 4 steps at the end of phase 1
+        let migration_record_count = db.keys::<MigrationRecordColumn>()?.count();
+        assert_eq!(migration_record_count, 4);
+    }
+
+    // Phase 2: Corrupt the migration record by introducing a gap
+    {
+        let db = Builder::open_sample(db_path)?
```
atanmarko: Do you need directory cleanup here before the test? Or just reuse db from step 1.

crates/agglayer-storage/src/storage/migration/mod.rs
iljakuklic: > Maybe you could move `self.step` management to `perform_step` function. Or move whole `perform_step_impl` implementation to `perform_step`?

The idea here is for `_impl` to do the work and return a `ErrorDetails` on failure. `perform_step` just adds the step number to the error. This allows to use `?` in the `_impl` method. Otherwise the error handling is a bit too bureaucratic.

crates/agglayer-storage/src/storage/migration/mod.rs
iljakuklic: > Nit: is `Builder` appropriate name? Maybe `DbMigrator` or similar?

Maybe, I don't have a strong opinion. Will let others express theirs.

crates/agglayer-storage/src/storage/migration/test/corrupt.rs
iljakuklic: > Do you need directory cleanup here before the test? Or just reuse db from step 1.

It is supposed to open the same database and artificially screw up the migration record. Although phase 1 and 2 could be done in one step, will simplify the test.

crates/agglayer-storage/src/storage/migration/error.rs
iljakuklic: > Print error display output.

I don't thinks it's a good idea for `#[source]` errors. The mechanisms for printing errors often follow the error sources to display the whole trace.

crates/agglayer-storage/src/storage/migration/mod.rs
atanmarko: Lets stick with builder then.

crates/agglayer-storage/src/storage/migration/error.rs
atanmarko: I'll not block PR about this comment

crates/agglayer-storage/src/error.rs
```diff
@@ -1,12 +1,15 @@
...
-use crate::storage::DBError;
+use crate::storage::{DBError, DBOpenError};
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("DB error: {0}")]
     DBError(#[from] DBError),
 
+    #[error("DB open error: {0}")]
+    DBOpenError(DBOpenError),
```
Ekleog-Polygon, line 11: Nit: standard rust casing would be `DbOpenError`. We already have `DBError` so I'm not sure which one of the options would be worse, and renaming it here probably would not make sense. Feel free to ignore this comment :smile:

Ekleog-Polygon:
(Just in case, I'm in the progress of reviewing this; as I see recent merges of main probably for landing I'll hopefully have the review done by EOD depending on meetings :smile: )

crates/agglayer-storage/src/storage/migration/error.rs
```diff
@@ -0,0 +1,45 @@
...
+    Migration(#[from] DBMigrationError),
+
+    #[error(transparent)]
+    Database(#[from] DBError),
+
+    #[error("Unexpected database schema")]
+    UnexpectedSchema,
+
+    #[error("Default column family is not empty")]
+    DefaultCFNotEmpty,
```
Ekleog-Polygon: Nit:
```suggestion
    DefaultCfNotEmpty,
```

crates/agglayer-storage/src/storage/migration/error.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    #[error("Migration record gap detected at step {0}")]
+    MigrationRecordGap(u32),
+
+    #[error(
+        "Fewer migration steps declared in the code than recorded in database \
+         (declared: {declared}, recorded: {recorded}). This indicates existing migration steps \
+         were removed from the code, which is not allowed."
```
Ekleog-Polygon:
```suggestion
         were removed from the code, or an older version of agglayer-node is being used, which is not allowed."
```

crates/agglayer-storage/src/storage/migration/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    ) -> Result<DB, DBOpenError> {
+        if self.step < self.start_step {
+            return Err(DBOpenError::FewerStepsDeclared {
+                declared: self.step,
+                recorded: self.start_step,
+            });
+        }
+
+        let expected_schema: BTreeSet<_> = expected_schema.into_iter().collect();
+        warn!(
```
Ekleog-Polygon: Can we have this as a TODO or an issue instead of a warning? It feels like it's not worth ringing infra alerts for this :sweat_smile:

crates/agglayer-storage/src/storage/migration/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    #[instrument(skip(self, step_fn), fields(step = self.step))]
+    fn perform_step_impl(
+        mut self,
+        step_fn: impl FnOnce(&mut Self) -> Result<(), DBMigrationErrorDetails>,
+    ) -> Result<Self, DBMigrationErrorDetails> {
+        let step = self.step;
+
+        if step >= self.start_step {
+            info!("Running migration step {step}");
+            step_fn(&mut self)?;
```
Ekleog-Polygon, line 240: I think this is actually not correct even if each step is idempotent. Imagine the following scenario:
1. We upgrade agglayer-node to a version with a new migration
2. The migration fails due to unexpected data
3. We revert agglayer-node to the previous version while fixing the migration code
4. In the meantime, the source data changes
5. We re-upgrade agglayer-node to the version with the new migration
6. Then we may see leftover data from the first migration polluting our second migration

I'm thinking it might make sense to force each migration to indicate which new CF it fills in; so that the builder can actually make sure the CF is empty before running the migration. Either this, or we have create_cf actually make sure the CF is fully empty after creation — I think that would work too and probably be simpler, but I haven't thought of all the possible corner cases.

Disclaimer: I'm not 100% sure of the issue I'm claiming, because rocksdb may error out if there's one CF in excess at step 3. This being said, in this second scenario, we also have a problem: if the migration ever fails we need some way to run old versions of agglayer-node until we fix it.

Does what I'm saying make sense? I must say I'm writing between meetings, so I hope I didn't miss anything in this analysis.

crates/agglayer-storage/src/storage/migration/record.rs
```diff
@@ -0,0 +1,5 @@
+/// Metadata recorded about each migration step performed.
+///
+/// Reserved for future extensions.
+pub type MigrationRecord = ();
```
Ekleog-Polygon: Do we want to have this be an empty protobuf struct, for future convenience? I'm not sure whether an empty protobuf struct serializes to an empty byte string or not.

Ekleog-Polygon (commented):
Overall LGTM! I think my main concern is regarding migration idempotency: it's described more in depth below, but I think simply idempotency is actually not enough to have secure migrations. We should in addition either ensure that the migration is reversed on failure (including power loss), or have each migration step be run in a rocksdb transaction so that power loss does not leave half-written data there.

Does my reasoning make sense to you?

crates/agglayer-storage/src/storage/migration/mod.rs
iljakuklic: I think I have some sense of what you are talking about but not exactly sure. Just a couple of points:

> I think this is actually not correct even if each step is idempotent. Imagine the following scenario:
> 
> 1. We upgrade agglayer-node to a version with a new migration
> 2. The migration fails due to unexpected data
> 3. We revert agglayer-node to the previous version while fixing the migration code
> 4. In the meantime, the source data changes
> 5. We re-upgrade agglayer-node to the version with the new migration
> 6. Then we may see leftover data from the first migration polluting our second migration

The migration system is not really prepared for the scenario of migration steps changing, I that case, I think it would be easier to produce a backup of the database before migrating and use that while the code is being fixed. Not the partially migrated storage. We might also add a small command line tool for migrating offline on a copy of the database, so we can try if it works ahead of time. Not as nice as the ability to recover from various scenarios automatically but seems simpler to me.

> I'm thinking it might make sense to force each migration to indicate which new CF it fills in; so that the builder can actually make sure the CF is empty before running the migration. Either this, or we have create_cf actually make sure the CF is fully empty after creation — I think that would work too and probably be simpler, but I haven't thought of all the possible corner cases.

The idea is that the migration step should only write to the new CF, not read it. So leftover data from the previous migration run are irrelevant. The only case I can think of is a key being deleted in the original CF between the two migration runs. Is that the case you are concerned about?

crates/agglayer-storage/src/storage/migration/record.rs
iljakuklic: > Do we want to have this be an empty protobuf struct, for future convenience? I'm not sure whether an empty protobuf struct serializes to an empty byte string or not.

I thought I would hack it around with a quick `if bytes.is_empty() { return MigrationData::default(); }` in the decoder :laughing:

crates/agglayer-storage/src/storage/migration/mod.rs
iljakuklic: Maybe there could be more guardrails in the interface to make sure the state starts in the clean state.

One idea would be to combine `.create_cfs()` and `.migrate()` into something like:

```rust
DB::builder(path, init_schema)
    .new_cf("new_cf_name", |db_access| {
        // perform migration here
        Ok(())
    })
```
where:
* the `new_cf ` function
  * creates the new family if it does not already exist
  * wipes all the data if it does
  * then calls the closure to populate the new CF with data
* The `db_access` object is a wrapper around the db, giving:
  * read-write access to the new cf
  * read-only access to the rest of the db

Would this help with the issue you raised?

Ekleog-Polygon: > The only case I can think of is a key being deleted in the original CF between the two migration runs. Is that the case you are concerned about?

Yes exactly!

And I do think `new_cf` would solve my concern, though we may want it to be `new_cfs` so that we can migrate one CF to multiple CFs "atomically"

Ekleog-Polygon: > The only case I can think of is a key being deleted in the original CF between the two migration runs. Is that the case you are concerned about?

Yes exactly!

And I do think `new_cf` would solve my concern, though we may want it to be `new_cfs` so that we can migrate one CF to multiple CFs "atomically"

iljakuklic: > And I do think `new_cf` would solve my concern, though we may want it to be `new_cfs` so that we can migrate one CF to multiple CFs "atomically"

There is a tension between having the plural `new_cfs` and the idea of `db_access` only providing write access to the new column(s).

The way I envisioned it, the `db_access.put(&key, &val)` method would implicitly write into the new CF. Enabling general `db_access.put::<ColFamily>(&key, &val)` would make it more difficult to check it is the new CF being written to. It would require either rather complicated trait machinery or turning it into a runtime check. That's why I have fallen back on adding one CF at a time with the new proposal.

Ekleog-Polygon: Hmm maybe a runtime check on the CF name would be reasonable enough?

If we wanted to over-engineer we could take a variadic generics list of CF definitions as a type-level parameter and have `put` identify which of these it should use; but it feels to me like that'd probably be too much even for me — after all this limited accessor will only ever be used from migration code; and the `new_cfs` logic already deals with anything that could go wrong while migrating.

Ekleog-Polygon: (And I do think we need the use case, eg. the cert header migration we've already started discussing would require splitting the cert header CF into two CFs; and if the migration is not atomic it could lead to headaches should any new cert be added in-between)

iljakuklic: > (And I do think we need the use case, eg. the cert header migration we've already started discussing would require splitting the cert header CF into two CFs;

That could be handled by multiple steps, e.g.:
```rust
    .add_cf("first_cf", |db_access| { /* populate the first cf */ })
    .add_cf("second_cf", |db_access| { /* populate the second cf */ })
    .drop_cfs(&["old_cf"])
```

> and if the migration is not atomic it could lead to headaches should any new cert be added in-between)

This is currently detected by checking the record. If there are more steps recorded than there are in the code, it will error out.

I.e. if the guidelines about migrations described in the docs are followed, there are two cases:

* Migration fails in the first step. In this case, the step will be re-run in full after upgrade which will be correct.
* Migration fails in one of the subsequent steps. In this case, the fact that the step has been performed will be recorded in the db and older version of the software will refuse to open it, forcing a manual cleanup or taking a database form backup.

crates/agglayer-storage/src/error.rs
iljakuklic: > Nit: standard rust casing would be `DbOpenError`. We already have `DBError` so I'm not sure which one of the options would be worse, and renaming it here probably would not make sense. Feel free to ignore this comment 😄

Since we have multiple inconsistencies in this regards, I filed it under #1248.

crates/agglayer-storage/src/storage/migration/mod.rs
iljakuklic: > Can we have this as a TODO or an issue instead of a warning? It feels like it's not worth ringing infra alerts for this 😅

Removed the log, although the hope is to implement this before infra get their hands on it. The issue has already been filed as #1214.

crates/agglayer-storage/src/storage/migration/mod.rs
Ekleog-Polygon: Good point! But I still think it'd still be somewhat a dangerous situation. Considering "older version of the software will refuse to open it", I think we need to basically ensure that we release the software after adding each migration step.

If we don't, imagine there's a migration that does basically `cp from to_1 && cp from to_2`, thus simply duplicating `from` into two CFs. With only one `new_cf` function, we could hit the following situation:
1. The `cp from to_1` migration executes properly
2. The `cp from to_2` migration fails to execute for any reason
3. Then we cannot boot the new release, but we also cannot revert to the old release. We need manual cleanup.

To avoid this, I think the only solution is to have a proper mechanism for backups before running any migration steps, easily recovering the backup if we can't open the database, and purging the backup after all migration steps successfully completed. I originally thought `new_cfs`-level atomicity would be enough, but it's actually not: we also could migrate state db and then fail during the epoch db migration, which would similarly leave us blocked.

I guess we can postpone the backup-for-migration mechanism to some follow-up PR?

This being said, if only for convenience in writing the migrations, I do think it'd be helpful to have something akin to `new_cfs` — it'd be easier to write one migration step that writes two CFs, than to have to iterate multiple times on the source CF to write each of the destination CFs. Though I don't have strong opinions here, hopefully we'll pretty rarely be writing migration steps, so a bit more work in doing it should be fine — though it's code that's hard to test, so the more straightforward it can be the better.

crates/agglayer-storage/src/storage/migration/record.rs
iljakuklic: Changed this to protobuf

crates/agglayer-storage/src/storage/migration/mod.rs
iljakuklic: > To avoid this, I think the only solution is to have a proper mechanism for backups before running any migration steps, easily recovering the backup if we can't open the database, and purging the backup after all migration steps successfully completed.

Had a quick look at the existing backup mechanism and it does not fit into this use case very nicely. Will have to think about it more.

> I originally thought `new_cfs`-level atomicity would be enough, but it's actually not: we also could migrate state db and then fail during the epoch db migration, which would similarly leave us blocked.

Do you think there is still merit to the `new_cfs` design if full pre-migration backup is implemented? It puts some more guardrails in but at a cost of increased complexity. It would also be more robust if migration is interrupted and run again (without restoring the backup).

> I guess we can postpone the backup-for-migration mechanism to some follow-up PR?

I think the design needs a bit more thought. Filed an issue for it:

* #1250

Ekleog-Polygon: Considering the latest exchanges, I think these two paths are reasonably doable:
1. Stay with your initial `add_cf` / `migrate` mechanism, but add backups. If anything goes wrong, we can just restore from backup
2. Switch to `new_cfs` and add backups. We need backups anyway, `new_cfs` just becomes a way to make it less likely that we'll have to restore backups

I don't have strong opinions between the two; but I'm thinking both would be better than a `new_cf` that only supports a single CF creation, as a `new_cf` call would make writing migrations less convenient.

Does my reasoning make sense to you?

iljakuklic: > Considering the latest exchanges, I think these two paths are reasonably doable:
> 
> 1. Stay with your initial `add_cf` / `migrate` mechanism, but add backups. If anything goes wrong, we can just restore from backup
> 2. Switch to `new_cfs` and add backups. We need backups anyway, `new_cfs` just becomes a way to make it less likely that we'll have to restore backups

Just a meta-point about this: once we have migration records in production databases, this will become harder to change.

Ekleog-Polygon: Good point! Though we'll always be able to add new ways and deprecate the old ones I think, and maybe even remove the code to do the migration if we wait for long enough after the last system was migrated — we can always crop it from the top.

iljakuklic: Changed the migration interface to `add_cfs` / `drop_cfs`. Did not yet restrict the closure in `add_cfs` to only be able to write into the new CFs (just by a comment :laughing:). That could be done as a follow up without much disruption.

Ekleog-Polygon: Sounds perfect to me! I re-requested myself a review to try and look at it when I get a chance; but if I'm late feel free to land without waiting for it 😄 And sorry for my recent delays in review, I haven't had as much time as I'd like to do reviews properly 🥲

atanmarko (approved):

PR #1224, opened 2025-12-09
fix: avoid double witness_generation and heavy state clone in recompute_state

phrwlk:
Avoid running witness_generation twice in recompute_state by reusing the recomputed state when the contract pp root matches. Replace the inline &mut state.clone() temporary with an explicit recovered_state to make ownership clear and prevent extra allocations. When reconciliation succeeds, update the header status to Candidate, send CertificateExecuted with the recovered state, and return early. In process_impl, route to process_from_candidate after recompute to avoid resubmitting a settlement tx. This reduces redundant work and aligns in-memory status with DB updates.

### DIFF
--- a/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
+++ b/crates/agglayer-certificate-orchestrator/src/certificate_task.rs
@@ -161,7 +161,11 @@ where
             CertificateStatus::Pending => self.process_from_pending().await,
             CertificateStatus::Proven => {
                 self.recompute_state().await?;
-                self.process_from_proven().await
+                if self.header.status == CertificateStatus::Candidate {
+                    self.process_from_candidate().await
+                } else {
+                    self.process_from_proven().await
+                }
             }
             CertificateStatus::Candidate => {
                 self.recompute_state().await?;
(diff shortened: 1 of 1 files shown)


crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -161,7 +161,11 @@ where
             CertificateStatus::Pending => self.process_from_pending().await,
             CertificateStatus::Proven => {
                 self.recompute_state().await?;
-                self.process_from_proven().await
+                if self.header.status == CertificateStatus::Candidate {
```
Ekleog-Polygon, line 164: I'm not sure I understand how this makes sense?

phrwlk: > I'm not sure I understand how this makes sense?

We move the status update into recompute_state precisely in the case where L1 already “promoted” the certificate to the equivalent of Candidate, and we route to process_from_candidate to reuse that information instead of resubmitting settlement or recomputing the state a second time.

Freyskeyd:
Hello,

Thanks for the contribution, after looking at the code suggestion, I'll close this PR as a lot of changes happened on this module. Feel free to reopen a PR if you think it's still valuable to address this.

Thanks !

PR #1226, opened 2025-12-09
fix(cli): unify config parameter name across commands

forkfury:
## Fix

Unifies the configuration file parameter name across CLI commands. Previously, `Run` command used `cfg` while `Backup::List` and `Backup::Restore` used `config_path`, creating inconsistent CLI API.

## Changes

- Renamed `config_path` to `cfg` in `Backup::List` and `Backup::Restore` commands
- Updated pattern matching in `main.rs` to use the unified parameter name
- All commands now consistently use `--cfg` flag for configuration file path


crates/agglayer/src/cli.rs
```diff
@@ -55,13 +55,13 @@ pub(crate) enum Backup {
     /// List all backups.
     List {
         #[arg(long, short, value_hint = ValueHint::FilePath, default_value = "agglayer.toml", env = "CONFIG_PATH")]
-        config_path: PathBuf,
+        cfg: PathBuf,
```
iljakuklic, line 58: This changes the command line switch from `--config-path` to `--cfg`. Seems to me the former is a bit nicer... Also, while unlikely, existing scripts may already use it.

atanmarko (commented):
I agree with @iljakuklic; I would also prefer a unified `--config-path` argument over `--cfg`.

Ekleog-Polygon (commented):
I am of the same opinion as @iljakuklic and @atanmarko. Ideally we could keep an alias for both `--cfg` and `--config-path` for the variants that would need a change.

iljakuklic (approved):
Looks good now, thanks

Freyskeyd (approved):

Issue #1214, opened 2025-12-04
Add storage migration integrity checks


iljakuklic:
The migration steps are only supposed to be added at the end of the list in the code. We should make best effort to detect and report if some of the steps already performed have changed.

At least, the migration code should verify if the schema matches the expected one, at each recorded step. 

Unfortunately, it is not easy to check the user-supplied closure passed to `.migrate()` has changed. Maybe keeping them in separate files, and using `include!("step.rs")` together with `hash(include_bytes!("step.rs")` but that seems a bit too restrictive as it would prevent even renames and changes to whitespace and such.

As a nice to have, each migration record may contain a timestamp with the time when the step was performed.

PR #1246, opened 2025-12-15
test: check storage initialization against production dbs

iljakuklic:
Just a number of tests to check that the storage schema assumed in the code corresponds to the storage schema we have running in production.

Also includes a tool to extract schema and sample data from production databases.

* builds on #1211

### DIFF
--- a/crates/agglayer-clock/src/block.rs
+++ b/crates/agglayer-clock/src/block.rs
@@ -47,8 +47,6 @@ pub struct BlockClock<P> {
     block_height: Arc<AtomicU64>,
     /// The Epoch duration in Blocks.
     epoch_duration: Arc<NonZeroU64>,
-    /// The current local Epoch number.
-    current_epoch: Arc<AtomicU64>,
     /// The last seen block number.
     latest_seen_block: u64,
 }
@@ -97,7 +95,6 @@ impl<P> BlockClock<P> {
             genesis_block,
             block_height: Arc::new(AtomicU64::new(0)),
             epoch_duration: Arc::new(epoch_duration),
-            current_epoch: Arc::new(AtomicU64::new(0)),
             latest_seen_block: 0,
         }
     }
@@ -142,8 +139,6 @@ pub enum BlockClockError {
     BlockHeightAlreadySet(u64),
     #[error("Failed to set the current block height, already set to: {0}")]
     SetBlockHeight(u64),
-    #[error("Failed to set the current Epoch number: previous={0}, expected={1}")]
-    SetEpochNumber(EpochNumber, EpochNumber),
     #[error("Failed to notify the start of the Clock task")]
     UnableToNotifyStart,
     #[error("Transport initialization: {0}")]
@@ -224,13 +219,14 @@ where
             Ordering::Relaxed,
         ) {
             Ok(0) => {
-                let current_epoch = self.current_epoch.load(Ordering::Acquire);
+                // Calculate epoch on demand
+                let current_epoch =
+                    <Self as Clock>::calculate_epoch_number(current_block, *self.epoch_duration);
                 info!(
                     initial_block_height = current_block,
                     initial_epoch = current_epoch,
                     "Initialized block clock state"
                 );
-                self.reinitialize_epoch_number(current_block);
 
                 // Update initial metrics
                 agglayer_telemetry::clock::record_current_block_height(current_block);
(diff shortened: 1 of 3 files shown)


hadjiszs (approved):

crates/agglayer-storage/examples/sample-storage-for-tests.rs
```diff
@@ -0,0 +1,140 @@
...
+        output_opts.create_if_missing(true);
+        output_opts.create_missing_column_families(true);
+
+        DB::open_cf_descriptors(&output_opts, output, cf_descriptors)?
+    };
+
+    // Process each column family
+    for cf_name in &cf_names {
+        info!("Processing column family: {cf_name:?}");
+        sample_column_family(&input_db, &output_db, cf_name, sample_size, &mut rng)
```
atanmarko, line 91: Would it be convenient to have an option to copy whole database?

crates/agglayer-storage/examples/sample-storage-for-tests.rs
```diff
@@ -0,0 +1,140 @@
...
+        .ok_or_else(|| eyre::eyre!("Column family not found: {cf_name:?}"))?;
+
+    let iter = input_db.iterator_cf(cf_handle, IteratorMode::Start);
+
+    // Collect all keys from the column family
+    let mut all_keys: Vec<Vec<u8>> = Vec::new();
+    for item in iter {
+        let (key, _value) = item?;
+        all_keys.push(key.to_vec());
+    }
```
atanmarko, line 121: Nit - is there any alternative to read all the key/value pairs in the loop? Could you randomly skip number of iterator entries to get samples?

crates/agglayer-storage/src/tests.rs
```diff
@@ -37,14 +36,33 @@ impl TempDBDir {
...
 
 impl Drop for TempDBDir {
     fn drop(&mut self) {
-        _ = std::fs::remove_dir_all(&self.path);
+        _ = fs::remove_dir_all(&self.path);
     }
 }
+
+/// Helper to extract tarball and return path to extracted directory
+pub fn extract_tarball(tarball_path: &Path, extract_to: &Path) -> Result<(), eyre::Error> {
```
atanmarko: Nit - you can probably move this one to the `agglayer-utils`.

crates/agglayer-storage/src/tests/migration.rs
```diff
@@ -0,0 +1,51 @@
...
+        .join(tarball_name);
+    extract_tarball(tarball_path.as_path(), &dir.path).unwrap();
+    dir
+}
+
+#[rstest::rstest]
+fn open_state(#[values("bali", "cardona", "mainnet")] env: &str) {
+    let tarball_name = format!("schema-{env}-state.tar.gz");
+    let db_dir = extract_tmp_db(tarball_name);
+    StateStore::init_db(&db_dir.path).unwrap();
```
atanmarko, line 29: Here we have set `create_if_missing(true)` in the `init_db` (`default_db_cf_definitions`). Is that expected for this test?

crates/agglayer-storage/examples/sample-storage-for-tests.rs
iljakuklic: > Nit - is there any alternative to read all the key/value pairs in the loop? Could you randomly skip number of iterator entries to get samples?

[DBIterator::advance_by](https://docs.rs/rocksdb/0.24.0/rocksdb/type.DBIterator.html#method.advance_by) is unstable. Now it is possible to avoid loading all keys in the vector by using a more advanced algortihm that just keeps the running set of sampled keys. However, I did not bother with that for testing code.

crates/agglayer-storage/examples/sample-storage-for-tests.rs
iljakuklic: > Would it be convenient to have an option to copy whole database?

Possibly. But at that point, would it not be simpler to just `cp -r`?

crates/agglayer-storage/src/tests/migration.rs
iljakuklic: > Here we have set `create_if_missing(true)` in the `init_db` (`default_db_cf_definitions`). Is that expected for this test?

Good point. However, I plan to add stricter database integrity checks. It will be more appropriate to update this as part of these changes.

* #1214

atanmarko (approved):

crates/agglayer-storage/examples/sample-storage-for-tests.rs
```diff
@@ -0,0 +1,140 @@
+//! Sample storage for tests
```
adria0pol, line 1: Why is under `./examples` ? does not makes more sense to put into `./bin` ?

crates/agglayer-storage/examples/sample-storage-for-tests.rs
```diff
@@ -0,0 +1,140 @@
...
+    } = Args::try_parse()?;
+
+    run(input.as_path(), output.as_path(), sample_size, seed)
+}
+
+fn run(input: &Path, output: &Path, sample_size: usize, seed: u64) -> eyre::Result<()> {
+    info!(?input, ?output, %sample_size, %seed, "Running...");
+
+    // Initialize RNG with seed for deterministic sampling
+    let mut rng = StdRng::seed_from_u64(seed);
```
adria0pol: In the [StdRng documentation](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html) says:

> Non-[portable](https://rust-random.github.io/book/crate-reprod.html): any future library version may replace the algorithm and results may be platform-dependent. (For a portable version, use the [rand_chacha](https://crates.io/crates/rand_chacha) crate directly.)`

Maybe better to use `rand_chacha` instead

crates/agglayer-storage/src/tests/migration.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[rstest::rstest]
+fn open_pending(#[values("bali", "cardona", "mainnet")] env: &str) {
+    let tarball_name = format!("schema-{env}-pending.tar.gz");
```
adria0pol, line 34: I do not get how we are going to operate with this.

Is expected that when we update the database in these environments we grab a copy, extract the schema and update the db/*.tar.gz files?

iljakuklic: > I do not get how we are going to operate with this.
> 
> Is expected that when we update the database in these environments we grab a copy, extract the schema and update the db/*.tar.gz files?

I think it's sufficient to just take the initial schema with dbs in the state they were in when the migration framework was introduced. Then let the migrator perform all the migration steps to move it to the most up to date version. The storage will go through all the intermediate states in the process. Tests for particular migration steps would have to be developed on an ad-hoc basis.

crates/agglayer-storage/examples/sample-storage-for-tests.rs
iljakuklic: > Why is under `./examples` ? does not makes more sense to put into `./bin` ?

I would say for something only used for testing, `examples` is a bit of a better match. Things in `bin` are installable. I think it would make sense to introduce a command line utility to manipulate/inspect agglayer storage, and such utility would go into `bin` (pun intended). This is more of an ad-hoc thing.

crates/agglayer-storage/examples/sample-storage-for-tests.rs
iljakuklic: > In the [StdRng documentation](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html) says:
> 
> > Non-[portable](https://rust-random.github.io/book/crate-reprod.html): any future library version may replace the algorithm and results may be platform-dependent. (For a portable version, use the [rand_chacha](https://crates.io/crates/rand_chacha) crate directly.)`
> 
> Maybe better to use `rand_chacha` instead

Great point. Switched to `rand_chacha`.

crates/agglayer-storage/src/tests/db/schema-bali-epoch.tar.gz
Freyskeyd, line 1: what are all those files? Do we need them synced?

iljakuklic: > what are all those files? Do we need them synced?

These are database containing only the schema, just to check the migrator expectations match the production DB. I think for this purpose it's better to leave them as they are without ever updating, in the state when the migration stuff was introduced. The test will then exercise all the migration steps starting with the initial schema.

crates/agglayer-storage/src/tests/migration.rs
adria0pol: nit: Add this information somewhere.

PR #1247, opened 2025-12-15
fix: refactor rate limiter core asserts

sashass1315:
Removed duplicated log_assert checks in RateLimiterCore::reserve and RateLimiterCore::record. The occupancy invariant is already enforced in RateLimiterCore::query, and the “record into a full limiter” invariant is enforced inside State::record. Keeping only the inner checks avoids double evaluation and duplicate logging while preserving the same safety guarantees and behavior verified by existing tests.
Allowed dead_code on State::raw, which is only used indirectly from test-only helpers.

### DIFF
--- a/crates/agglayer-rate-limiting/src/local/limiter/core.rs
+++ b/crates/agglayer-rate-limiting/src/local/limiter/core.rs
@@ -37,6 +37,7 @@ impl<S: RawState> RateLimiterCore<S> {
     /// Reserve a rate limiting slot.
     pub fn reserve(&mut self, time: S::Instant) -> Result<SlotTracker, S::LimitedInfo> {
         let occupancy = self.query(time);
+        debug_assert!(occupancy <= self.state.max_events());
 
         if occupancy < self.state.max_events() {
             self.reserved += 1;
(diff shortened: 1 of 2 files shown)


crates/agglayer-rate-limiting/src/local/state/wrapper.rs
```diff
@@ -48,6 +48,7 @@ impl<S: RawState> State<S> {
     }
 
     /// Get non-modifying access to the underlying raw state.
+    #[allow(dead_code)]
```
hadjiszs: i think you can use `#[cfg(feature = "testutils")]` instead if only used for tests

crates/agglayer-rate-limiting/src/local/limiter/core.rs
```diff
@@ -57,7 +57,6 @@ impl<S: RawState> RateLimiterCore<S> {
 
     /// Record a rate limiting event.
     pub fn record(&mut self, time: S::Instant, slot: SlotTracker) {
-        log_assert!(self.state.raw().query() < self.state.max_events());
```
hadjiszs, line 60: i believe @iljakuklic suggestion of keeping it as debug assert applies also to this one

hadjiszs (commented):
thanks @sashass1315 👍

atanmarko (approved):

Freyskeyd:
Hello,

Thanks for your contribution ! As this PR is fairly outdated, I'll close it. If you think the changes are still valuable, feel free to open a new PR against up to date `main` branch.

PR #1208, opened 2025-12-03
feat: add udeps to ci

adria0pol:
Add udeps checking in CI.
Closes #862

CONFIG-CHANGE: 
- Updates`PP_VKEY`

PR #1249, opened 2025-12-16
feat: expose the token balances on the admin API

### LINKED ISSUE #1103: Expose the LBT balance values for the L2 smart contracts
# Context

- The token balances tracked by the PP are also tracked by the L2 contracts since this issue :
   - https://github.com/agglayer/protocol-research/discussions/133#discussioncomment-12615009

- Some L2 smart contract migration (erigon to sovereign) has the need to re-initialize the LBT balances on the L2
   - https://github.com/agglayer/protocol-team-kanban/issues/725

# To do

The solution proposed by protocol team is to expose the LBT values from the agglayer-node API through these 2 endpoints:

### Endpoint 1 (balance for all tokens on a given network)

> For the initialization, we’ll need data in a structure similar to: originNetwork[], originTokenAddress[], amount[]
>
> Therefore, it would be useful to have an endpoint that, given a rollupID, returns all the corresponding LBT information:
> 
> 
> Input: rollupID
> Ouptut: Complete LBT data

### Endpoint 2 (balance for a given token on a given network)

> To get the information of a specific token, we could have another endpoint that, given a rollupID, originNetwork, and originAddress, returns only the amount associated with that token:
> 
> Input: rollupID + originNetwork + originAddress
> Output: The amount of a specific token

# Questions

1. how do we want to represent the pointer to a specific L2 state in the response?
   - in other words: how to explicit that the returned balance values are the ones at this very specific L2 block?
   - considering that:
     - the agglayer has no notion of L2 block number
     - if we provide the certificate height, how the lookup certificate height => L2 block number would be done? maybe the aggsender expose that?
2. which specific state do we want to expose?
   - should we return the values of the latest settled state in the L1? Or should we return the latest state known by the agglayer (which is still under process and not settled yet)?

hadjiszs:
Expose the token balances of the balance tree on the admin API

```
admin_getTokenBalance <network_id> [token_info]
```

if the optional `token_info` isn't provided, then it returns all the non-zero token balances for the given network

closes #1103 
needs https://github.com/agglayer/interop/pull/155


Cargo.toml
```diff
@@ -45,13 +45,13 @@ pessimistic-proof-core = { path = "crates/pessimistic-proof-core" }
...
-agglayer-primitives = "0.12.0"
-agglayer-tries = "0.12.0"
-unified-bridge = "0.12.0"
+agglayer-bincode = { git = "https://github.com/agglayer/interop.git", branch = "feat/expose-lbt" }
+agglayer-interop = { git = "https://github.com/agglayer/interop.git", branch = "feat/expose-lbt" }
+agglayer-interop-types = { git = "https://github.com/agglayer/interop.git", branch = "feat/expose-lbt" }
+agglayer-elf-build = { git = "https://github.com/agglayer/interop.git", branch = "feat/expose-lbt" }
+agglayer-primitives = { git = "https://github.com/agglayer/interop.git", branch = "feat/expose-lbt" }
+agglayer-tries = { git = "https://github.com/agglayer/interop.git", branch = "feat/expose-lbt" }
+unified-bridge = { git = "https://github.com/agglayer/interop.git", branch = "feat/expose-lbt" }
```
hadjiszs: todo: put tag once released

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -29,6 +30,13 @@ pub enum ProcessNow {
 
 #[rpc(server, namespace = "admin")]
 pub(crate) trait AdminAgglayer {
+    #[method(name = "getTokenBalance")]
+    async fn get_token_balance(
+        &self,
+        network_id: NetworkId,
+        token_info: Option<TokenInfo>,
+    ) -> RpcResult<Vec<unified_bridge::TokenBalanceEntry>>;
```
adria0pol: AFAIS in [interop](https://github.com/agglayer/interop/blob/e4b50e5a1699a84931abea63b2113fc538e086a4/crates/unified-bridge/src/balance_tree.rs#L41) there's the


```rust
/// Token balance entry structure in order to display the balance tree values.
#[serde_as]
#[derive(Serialize)]
#[serde(rename_all = "camelCase")]
pub struct TokenBalanceEntry {
    pub origin_token_address: Address,
    #[serde_as(as = "DisplayFromStr")]
    pub amount: U256,
}
```

Should not be this struct ( as a Data Transfer Object ) into this crate instead forcing everyone that uses the interop crate to use these struct fields json-friendly-camelCase?

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -191,6 +199,37 @@ where
     StateStore: StateReader + StateWriter + 'static,
     DebugStore: DebugReader + DebugWriter + 'static,
 {
+    #[instrument(skip(self), level = "debug")]
+    async fn get_token_balance(
```
adria0pol, line 239: I'm thinking that if in the future we find that this function is going to be called _somehow_ continuously, it will be nice to return something else that is not JSON, the result could be really huge in some cases. Anyway, not wondering about a timeout because I see that is set to [180](https://github.com/agglayer/agglayer/blob/780415d9be63c0b22444c99cbda81c40b332417c/crates/agglayer-config/src/rpc.rs#L123) secs.

crates/pessimistic-proof-test-suite/tests/snapshots/vkey_selector__vkey_snapshot.snap
```diff
@@ -4,5 +4,5 @@ expression: message
 ---
 # If this test fails, it means the PP vkey has changed.
 # When that happens, consider updating the selector by bumping the PP version.
-| PP_VKEY          | 0x00139b6c32b3748fae95a38c289d46c30730c3f900e3fff3ad6db2aa95aaa142 |
+| PP_VKEY          | 0x0010b2df15a644a2fa6de74da874992e378c25693495bee30e6c384715354af7 |
```
adria0pol: IIUC  you have to bump the `PP_VKEY_SELECTOR` here

crates/agglayer-jsonrpc-api/src/admin.rs
hadjiszs: yeah true 👍 I moved it

crates/agglayer-jsonrpc-api/src/admin.rs
hadjiszs: right now it's really just for rare and manual use through this admin API

crates/pessimistic-proof-test-suite/tests/snapshots/vkey_selector__vkey_snapshot.snap
hadjiszs: just saw that https://github.com/agglayer/agglayer/pull/1208#discussion_r2610260180
I havent followed which selector was for the latest version but then it means that the selector shouldnt change

crates/pessimistic-proof/src/local_balance_tree.rs
```diff
@@ -38,3 +43,22 @@ impl From<LocalBalanceTree> for pessimistic_proof_core::local_balance_tree::Loca
...
+
+pub struct BalanceTree(pub Smt<LOCAL_BALANCE_TREE_DEPTH>);
+
+impl BalanceTree {
+    /// Returns all the non-zero token balance contained in this balance tree.
+    pub fn get_all_balances(
+        &self,
+    ) -> Result<Vec<(SmtPath<LOCAL_BALANCE_TREE_DEPTH>, Digest)>, SmtError> {
+        Ok(self.0.entries()?.into_iter().collect())
+    }
```
iljakuklic, line 55: Could this return an iterator instead of creating an intermediate Vec? The caller may still decide to `collect`.

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -27,8 +30,43 @@ pub enum ProcessNow {
...
+}
+
 #[rpc(server, namespace = "admin")]
 pub(crate) trait AdminAgglayer {
+    #[method(name = "getTokenBalance")]
+    async fn get_token_balance(
+        &self,
+        network_id: NetworkId,
+        token_info: Option<TokenInfo>,
+    ) -> RpcResult<Vec<TokenBalanceEntry>>;
```
iljakuklic: In the interest of future extensibility, it may make sense to return a struct with a named field `balances` containing the Vec. This way, you will be able to add more fields to the response later relatively easily.

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -27,8 +30,43 @@ pub enum ProcessNow {
...
+/// Token balance entry structure in order to display the balance tree values.
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TokenBalanceEntry {
+    pub origin_network: NetworkId,
+    pub origin_token_address: Address,
+    #[serde_as(as = "DisplayFromStr")]
+    pub amount: U256,
+}
```
iljakuklic, line 44: Wondering if it would make sense to turn this into a doubly nested json dictionary, i.e. `BTreeMap<NetworkId, BTreeMap<Address, U256>>`. Unless there are some compatibility concerns.

iljakuklic (approved):
Looks good. Some suggestions regarding the json encoding but can be ignored.

crates/agglayer-jsonrpc-api/src/admin.rs
hadjiszs: hm I mostly tried to match the format requested by protocol team in their [issue](https://github.com/agglayer/agglayer/issues/1103) (hence the pascalCase for example)

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -191,6 +235,46 @@ where
...
+            }]
+        } else {
+            balance_tree
+                .get_all_balances()
+                .map_err(|error| {
+                    error!(?error, "Failed to get all balances");
+                    Error::internal("Unable to get all balances")
+                })?
+                .map(TokenBalanceEntry::from)
+                .collect()
```
atanmarko, line 272: How big could this vector get, should we have some limit? Does it make sense to have an api that would return all the token balances from all the networks?

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -191,6 +235,46 @@ where
     StateStore: StateReader + StateWriter + 'static,
     DebugStore: DebugReader + DebugWriter + 'static,
 {
+    #[instrument(skip(self), level = "debug")]
+    async fn get_token_balance(
+        &self,
+        network_id: NetworkId,
+        token_info: Option<TokenInfo>,
```
atanmarko, line 242: Maybe token info should not be optional? Could we make this a vector of `TokenInfo`?

hadjiszs: that's for manual process requested by protocol team there: https://github.com/agglayer/protocol-team-kanban/issues/725

this interface seems to be the closest to what they want (either all balances or just the balance for a given TokenInfo which can then be called multiple times for all the tokens they want)

crates/agglayer-jsonrpc-api/src/admin.rs
hadjiszs: >Does it make sense to have an api that would return all the token balances from all the networks?

not requested by the protocol team, I think that's for very specific network in the need of a specific migration (like just for a specific network recently as far as I understood)

>How big could this vector get, should we have some limit? 

I dont expect this to be an issue tbh, also that's not the public API but only the admin one, so I'd go as-is

atanmarko:
@hadjiszs Dependencies are updated here - you could rebase on top of this PR https://github.com/agglayer/agglayer/pull/1287

iljakuklic (approved):

PR #1251, opened 2025-12-16
feat: protobuf definition for the new settlement logic's column families

### LINKED ISSUE #1227: Reach a conclusion on the exact format of settlement jobs as serialized to RocksDB
See the three options at https://github.com/agglayer/agglayer/issues/1216. Currently, my best suggestion would be two protobuf column families, looking like:
```
SETTLEMENT_JOBS_CF:
Key: (wallet pubkey, nonce)
Value: {
    transaction-specific configuration (gas params, retry params, etc.), everything that differs between sign_tx and certificate;
    settlement job id as ULID
    oneof {
        bincode encoding of alloy v1 CallBuilder bytes
    }
}

SETTLEMENT_JOB_IDS_CF:
Key: settlement job id as ULID
Value: (wallet pubkey, nonce)
```

### LINKED ISSUE #1229: [Design] Scaffold settlement service and settlement job

### LINKED ISSUE #1245: Design the protobuf format for settlement storage

Ekleog-Polygon:
Fixes #1227 
Fixes #1245 
Fixes #1229 : the design will be the same as the protobuf CF for settlement_jobs; though implementation is not part of this yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,6 +173,7 @@ dependencies = [
 name = "agglayer-config"
 version = "0.1.0"
 dependencies = [
+ "agglayer-config",
  "agglayer-primitives",
  "agglayer-types",
  "alloy-primitives",
@@ -553,6 +554,7 @@ dependencies = [
  "chrono",
  "criterion",
  "eyre",
+ "fail",
  "hex",
  "insta",
  "mockall",
@@ -570,6 +572,7 @@ dependencies = [
  "sp1-sdk",
  "strum 0.27.2",
  "strum_macros 0.27.2",
+ "test-log",
  "thiserror 2.0.16",
  "tokio",
  "tokio-util",
--- a/crates/agglayer-config/Cargo.toml
+++ b/crates/agglayer-config/Cargo.toml
@@ -8,6 +8,7 @@ dirs.workspace = true
 alloy-primitives.workspace = true
 educe.workspace = true
 humantime-serde = "1.1.1"
+insta = { workspace = true, optional = true }
 jsonrpsee.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_with.workspace = true
@@ -24,11 +25,12 @@ agglayer-types.workspace = true
 prover-config.workspace = true
 
 [dev-dependencies]
-insta.workspace = true
 pretty_assertions = "1.4.0"
 rstest.workspace = true
 serde_json = { workspace = true }
 
+agglayer-config = { workspace = true, features = ["testutils"] }
+
 [features]
 default = []
-testutils = []
+testutils = ["dep:insta"]
--- a/crates/agglayer-config/src/auth.rs
+++ b/crates/agglayer-config/src/auth.rs
@@ -43,7 +43,7 @@ pub struct LocalConfig {
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
-#[cfg_attr(any(test, feature = "testutils"), derive(Default))]
+#[cfg_attr(feature = "testutils", derive(Default))]
 pub struct PrivateKey {
     #[serde(alias = "Path")]
     pub path: PathBuf,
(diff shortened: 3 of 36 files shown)


proto/agglayer/storage/v0/settlement_service.proto
```diff
@@ -0,0 +1,153 @@
...
+  string error_message = 2;
+}
+
+// Type of client error.
+enum ClientErrorType {
+  // Unspecified error type.
+  CLIENT_ERROR_TYPE_UNSPECIFIED = 0;
+
+  // Network error.
+  CLIENT_ERROR_TYPE_NETWORK_ERROR = 1;
```
Ekleog-Polygon: Note for the reviewer: I don't expect this to be complete, I've just added one variant and we can add the other ones as needed when we write the actual implementation.

Though TBH I'm not entirely sure we should even have "network error". Maybe we should just leave everything as "unspecified" until we get a use case for matching on this enum?

proto/agglayer/storage/v0/settlement_service.proto
```diff
@@ -0,0 +1,153 @@
...
+
+  // Ceiling for max priority fee per gas.
+  Uint128 max_priority_fee_per_gas_ceiling = 9;
+
+  // Floor for max priority fee per gas.
+  Uint128 max_priority_fee_per_gas_floor = 10;
+
+  // Percent increase for max priority fee per gas: each retry will multiply max priority fee per gas by N / 100.
+  uint32 max_priority_fee_per_gas_increase_percents = 11;
+}
```
atanmarko, line 57: You could add gas_limit here also (should be set from the configuration).

proto/agglayer/storage/v0/settlement_service.proto
atanmarko: You could add `provider error` here instead of network error?

proto/agglayer/storage/v0/settlement_service.proto
```diff
@@ -0,0 +1,153 @@
...
+
+  // Gas price parameters used for this attempt.
+  Uint128 max_priority_fee_per_gas = 4;
+
+  // Hash of the submitted transaction.
+  TxHash tx_hash = 5;
+
+  // Result of the attempt, if available.
+  TxResult result = 6;
+}
```
atanmarko, line 156: Gas specified for the transaction call should be added here.

proto/agglayer/storage/v0/settlement_service.proto
Ekleog-Polygon: Considering your :+1: on my message above, I'll just remove it and we can iterate on that with time as we find ways in which we'd want to match on the enums — feels better than having variants that we'll need to keep forever for backwards compat that we may never use :smile: 

I did add `CLIENT_ERROR_TYPE_UNSPECIFIED_TRANSIENT` and `CLIENT_ERROR_TYPE_UNSPECIFIED_PERMANENT` that we'll certainly need; and we can add other variants as we need to when we need to — I'm thinking that for the first version we can just do with these two variants and the error message and that should probably be enough!

atanmarko (dismissed):

proto/agglayer/storage/v0/settlement_service.proto
```diff
@@ -0,0 +1,162 @@
...
+
+// Result for a successfully-executed contract call.
+message ContractCallResult {
+  // The outcome of the contract call.
+  ContractCallOutcome outcome = 1;
+
+  // Additional metadata about the call outcome.
+  bytes metadata = 2;
+}
```
iljakuklic, line 105: Would it make sense to also note the block hash where the transaction was included? Could be used to reliably detect reorgs if that is needed.

iljakuklic (dismissed):

proto/agglayer/storage/v0/settlement_service.proto
Ekleog-Polygon: Makes sense! I just added it there :smile:

Ekleog-Polygon:
I handled the clippy lints in a pretty yolo way; but I think we can land this and iterate on it later — we should be able to delete the `pub` I added as soon as we finish implementing the related `agglayer-storage` changes, and especially the codecs.

PR #1255, opened 2025-12-17
feat: restrict write access to new CFs during migration

iljakuklic:
Just add a layer enforcing only newly created CFs are written to during migration. This ensures `drop_cfs` is the only destructive operation.

* builds on #1211


crates/agglayer-storage/src/storage/migration/test/corrupt.rs
```diff
@@ -94,3 +97,42 @@ fn unexpected_schema() -> Result<(), eyre::Error> {
...
+    }
+
+    // Phase 2: Try to migrate but write to the old (read-only) CF
+    {
+        let result = Builder::open_sample(db_path)?
+            .add_cfs([NetworkInfoV1Column::COLUMN_FAMILY_NAME], |db| {
+                // This should FAIL - trying to write to old CF during migration
+                let v0_value = &DATA_V0[1].1;
+                db.put::<NetworkInfoV0Column>(&NetworkId::new(42), v0_value)?;
+                Ok(())
```
atanmarko: Nit: just return `db.put()` result

iljakuklic: > Nit: just return `db.put()` result

The error type is slightly different.

Edit: Actually, let me double check 😄

iljakuklic: Yea, I thought this referred to a different bit of code, changed now. Thanks.

atanmarko (approved):

crates/agglayer-storage/src/storage/migration/test/corrupt.rs
```diff
@@ -94,3 +96,43 @@ fn unexpected_schema() -> Result<(), eyre::Error> {
...
+                // Check that the error is WritingReadOnlyCf
+                match migration_err.details {
+                    crate::storage::DBMigrationErrorDetails::WritingReadOnlyCf(cf) => {
+                        assert_eq!(NetworkInfoV0Column::COLUMN_FAMILY_NAME, cf);
+                    }
+                    err => panic!("Unexpected error {err:?}"),
+                }
+            }
+            Err(err) => panic!("Unexpected error {err:?}"),
+            Ok(_) => panic!("Expected error"),
```
adria0pol, line 133: nit

```suggestion
            Ok(_) => panic!("Test must fail"),
```

adria0pol (approved):
This is a fantastic idea, thanks for making migrations less painful 🙌

crates/agglayer-storage/src/storage/migration/test/corrupt.rs
iljakuklic: > nit

Hmm not sure, I like the current wording better. The test should pass by checking the migration framework issues an error.

PR #1264, opened 2025-12-18
perf: optimize add_leaf with direct layer access

CreeptoGengar:
Cache empty hashes array and use direct layer access in add_leaf() instead of calling get() method. 
This eliminates repeated calls to empty_hash_array_at_height in the hot path and reduces function call overhead when accessing sibling nodes.


Freyskeyd:
Hello,

Thanks for your contribution ! As this PR is fairly outdated, I'll close it. If you think the changes are still valuable, feel free to open a new PR against up to date `main` branch.

Double check if you want @hadjiszs and @Ekleog-Polygon

PR #1276, opened 2025-12-22
test: add edge case coverage for EthAmount deserialization

mattvaux:
## Summary

Adds test cases for zero values, empty strings, very large numbers, invalid units, and whitespace handling in EthAmount deserialization. These edge cases are important for gas price configuration validation and were previously untested despite being used in production configs.

## Changes

- Added test cases for zero values (`0gwei`, `0wei`, `0eth`) which are used in production configuration files
- Added test case for empty string input to verify `MissingUnit` error handling
- Added test case for `u128::MAX` value to ensure overflow handling works correctly
- Added test cases for invalid unit names (`xyz`, `invalid`) to verify error reporting
- Added test cases for leading whitespace to ensure proper rejection
- Added test cases for multiple spaces between number and unit (verified that alloy's `parse_units` handles this correctly)

## Testing

All 23 tests pass, including 11 new edge case tests. The code passes `cargo fmt`, `cargo clippy`, and builds successfully.

## CONFIG-CHANGE

None. This change only adds test coverage and does not modify configuration behavior or defaults.

## BREAKING-CHANGE

None. This is a test-only change that does not affect the public API or behavior of the `EthAmount` deserializer.


iljakuklic (approved):
Cool, thanks

PR #1280, opened 2025-12-23
fix(signer): remove obsolete KmsSigner::set_chain_id

Olexandr88:
Removes unused KmsSigner::set_chain_id after signer implementation refactor.


crates/agglayer-gcp-kms/src/signer.rs
```diff
@@ -24,6 +24,10 @@ impl KmsSigner {
         Self { signer }
     }
 
+    pub fn set_chain_id(&mut self, chain_id: Option<u64>) {
+        self.signer.set_chain_id(chain_id);
+    }
+
```
atanmarko: Please elaborate why do you need this function implementation in `impl KmsSigner` section when it is implemented as part of `Signer for KmsSigner` implementation?

Olexandr88: I agree with you. This method no longer makes sense as a public API

PR #1282, opened 2025-12-24
feat: add settlement service stub

atanmarko:
Crate and stub for the the agglayer settlement service.

Related ticket https://github.com/agglayer/agglayer/issues/1258

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1694,6 +1694,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-ff"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a177aba0ed1e0fbb62aa9f6d0502e9b46dad8c2eab04c14258a1212d2557ea70"
+dependencies = [
+ "ark-ff-asm 0.5.0",
+ "ark-ff-macros 0.5.0",
+ "ark-serialize 0.5.0",
+ "ark-std 0.5.0",
+ "arrayvec",
+ "digest 0.10.7",
+ "educe",
+ "itertools 0.13.0",
+ "num-bigint 0.4.6",
+ "num-traits",
+ "paste",
+ "zeroize",
+]
+
+[[package]]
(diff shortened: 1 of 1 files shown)


crates/agglayer-config/src/settlement_service.rs
```diff
@@ -0,0 +1,6 @@
+use serde::{Deserialize, Serialize};
+
+/// The settlement service configuration.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct SettlementServiceConfig {}
```
iljakuklic, line 6: Is the settlement service supposed to handle both certificates and validiums? If so, is the config envisioned to be shared between the two or separate?

hadjiszs (approved):

crates/agglayer-config/src/settlement_service.rs
atanmarko: @iljakuklic It will be shared if they use same wallet. As discussed previously we will make per tx config to set those things for validiums/certificates. This configuration will keep the default settings. When needed, every tx submission will have an optional tx configuration to override default values.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -0,0 +1,12 @@
+/// Internal configuration for settlement tasks.
+/// This is not exposed to users and is configured programmatically by the
+/// service.
+pub struct SettlementTaskConfig {}
```
Freyskeyd, line 4: Does it needs to be `pub` if it's not exposed ?

Freyskeyd (approved):

crates/agglayer-settlement-service/src/settlement_task.rs
atanmarko: Maybe not. We will iterate on this one. We may need it public for the integration tests.

EDIT: we need it public for the integration tests #1291

PR #1284, opened 2025-12-25
fix: cargo audit

atanmarko:
Bump `ruint` to 1.17.2 to fix audit

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2546,7 +2546,7 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.12.1",
  "log",
  "prettyplease",
  "proc-macro2",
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -10574,7 +10640,7 @@ version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0978bf7171b3d90bac376700cb56d606feb40f251a475a5d6634613564460b22"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.48.0",
```
iljakuklic, line 10643: Hmm `windows-sys` got downgraded?

iljakuklic (approved):

Cargo.lock
atanmarko: I was surprised too, that is sub-dependency of ruint

PR #1288, opened 2025-12-26
fix: check proof existence before insertion to prevent overwrite

PivasDesant:
Check if proof already exists before inserting to avoid overwriting.

Fixes the TODO. If proof exists, we log a warning and skip insertion
instead of silently overwriting it.


Freyskeyd:
Thanks for your contribution, but after some tests, we do not want to prevent proof overwrite.

Issue #838, opened 2025-06-06
Start using eyre::Report some more


Ekleog-Polygon:
See design doc at https://docs.google.com/presentation/d/1v9j5NVjjp9t6mtuYp-mUtmrz6dmhFnxWZXzQsrbSC7I/edit?usp=sharing ; the decision is in the last slide.

We'll want to mostly use `eyre::Report` everywhere, use `wrap_err` where appropriate to add lots of context, have a wrapper for `&eyre::Report` that implements `Valuable` so JSON logging also has span trace and stack traces.

We still want to be sure we stay able to match on recoverable errors properly, be it with an eyre downcast or using an enum. This means that for now interop will likely stay using error enums as a rule of thumb.

The plan of action for this PR is to start with using it in agglayer-prover, and seeing from there how well it's working in practice and see to expand it to the rest of our codebase.

Ekleog-Polygon:
Note from chatting with @adria0pol : we should probably also add mnemonics / error codes to each of our errors.

PR #1289, opened 2025-12-26
fix: use correct error type for pending certificate validation

letmehateu:
Fixes incorrect error type usage in pending certificate insertion. The code was
  using `CertificateCandidateError::UnexpectedHeight` for pending certificate validation,
  but this error type is semantically intended for candidate certificates in the epoch
  context.


atanmarko (dismissed):

Freyskeyd:
@Ekleog-Polygon If you have time to look for this PR

Ekleog-Polygon (approved):
AFAICT we never actually match on the error, so this should be good to land.

FWIW this is part of the reasons why I'm advocating for https://github.com/agglayer/agglayer/issues/838 : we don't actually care about the type 99% of the time, so just acknowledging that, using `eyre::Error` almost everywhere and downcasting only when appropriate would sound like the best option to me 😄

PR #1291, opened 2025-12-29
feat: settlement service config

### LINKED ISSUE #1278: Add settlement service configuration definition

atanmarko:
Resolves #1278

### DIFF
--- a/crates/agglayer-config/src/settlement_service.rs
+++ b/crates/agglayer-config/src/settlement_service.rs
@@ -16,24 +16,23 @@ use crate::Multiplier;
 /// ```toml
 /// [settlement.pessimistic-proof-tx-config]
 /// confirmations = 16
-/// finality = "justified"
+/// finality = "SafeBlock"
 ///
 /// [settlement.validium-tx-config]
 /// confirmations = 6
-/// finality = "immediate"  # Faster settlement for validium data
+/// finality = "LatestBlock"  # Faster settlement for validium data
 /// ```
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
-#[serde(rename_all = "lowercase")]
 pub enum Finality {
     /// Transaction is considered settled immediately after the specified
     /// number of confirmation blocks.
     ///
     /// **Security**: Vulnerable to chain reorganizations beyond the
     /// confirmation count.
-    Immediate,
+    LatestBlock,
 
     /// Transaction is considered settled when the containing block has been
-    /// justified, ie. considered "safe."
+    /// considered "safe."
     ///
     /// **Security**: Very strong. Reversing a safe block would require
     /// a significant portion of validators to be slashed.
@@ -41,35 +40,43 @@ pub enum Finality {
     /// **Time**: Typically up to ~7 minutes on mainnet. Worst case scenario
     /// 12-13 minutes.
     #[default]
-    Justified,
+    SafeBlock,
 
     /// Transaction is considered settled only when the containing block has
     /// been fully finalized.
     ///
     /// **Time**: Typically between 7-13 minutes on mainnet. Worst case scenario
     /// ~19 minutes.
-    Finalized,
+    FinalizedBlock,
 }
 
 /// Transaction retry policy for failed or pending settlement transactions.
 ///
 /// Defines the strategy used when retrying failed or pending transactions.
-/// The retry policy works in conjunction with `tx_retry_interval` to determine
-/// the timing between retry attempts.
 ///
 /// Future versions may include additional policies such as:
 /// - **Exponential**: Exponential backoff with increasing intervals
 /// - **Jittered**: Random jitter to avoid thundering herd issues
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
 #[serde(rename_all = "kebab-case")]
-pub enum TxRetryPolicy {
-    /// Linear retry policy with fixed intervals between attempts.
-    ///
-    /// This is the default and recommended policy for most use cases.
-    /// It provides predictable retry behavior and works well with the
-    /// settlement service's transaction monitoring.
-    #[default]
-    Linear,
+pub struct TxRetryPolicy {
+    /// Initial retry interval.
+    #[serde(with = "crate::with::HumanDuration")]
+    pub initial_interval: Duration,
+
+    /// Interval multiplier for each subsequent retry.
+    #[serde(default, skip_serializing_if = "crate::is_default")]
+    pub interval_multiplier_factor: Multiplier,
+
+    /// Maximum interval between retries.
+    #[serde(default = "default_max_interval")]
+    #[serde(with = "crate::with::HumanDuration")]
+    pub max_interval: Duration,
+
+    /// Jitter factor to add randomness to retry intervals.
+    #[serde(default, skip_serializing_if = "crate::is_default")]
+    #[serde(with = "crate::with::HumanDuration")]
+    pub jitter: Duration,
 }
 
 /// The settlement transaction configuration.
@@ -82,15 +89,13 @@ pub struct SettlementTransactionConfig {
     #[serde(default = "default_rpc_max_retries")]
     pub max_retries: usize,
 
-    /// Retry policy/approach for the transaction.
-    #[serde(default = "default_rpc_retry_policy")]
-    pub tx_retry_policy: TxRetryPolicy,
+    /// Retry policy for the transaction when there is a transient failure.
+    #[serde(default = "default_transient_rpc_retry_policy")]
+    pub retry_on_transient_failure: TxRetryPolicy,
 
-    /// Retry interval.
-    /// Used together with `tx_retry_policy`.
-    #[serde(default = "default_tx_retry_interval")]
-    #[serde(with = "crate::with::HumanDuration")]
-    pub tx_retry_interval: Duration,
+    /// Retry policy for the transaction when it is not included on L1.
+    #[serde(default = "default_non_inclusion_rpc_retry_policy")]
+    pub retry_on_not_included_on_l1: TxRetryPolicy,
 
     /// Number of block confirmations required for
     /// the transaction to resolve a receipt.
@@ -107,9 +112,9 @@ pub struct SettlementTransactionConfig {
     #[serde(default, skip_serializing_if = "crate::is_default")]
     pub gas_limit_multiplier_factor: Multiplier,
 
-    /// Gas limit for the transaction.
-    #[serde(default = "default_gas_limit")]
-    pub gas_limit: U256,
+    /// Ceiling for the gas limit for the transaction.
+    #[serde(default = "default_gas_limit_ceiling")]
+    pub gas_limit_ceiling: U256,
 
     /// Gas price multiplier for the transaction.
     /// The gas price is calculated as follows:
@@ -137,12 +142,12 @@ impl Default for SettlementTransactionConfig {
     fn default() -> Self {
         Self {
             max_retries: default_rpc_max_retries(),
-            tx_retry_policy: default_rpc_retry_policy(),
-            tx_retry_interval: default_tx_retry_interval(),
+            retry_on_transient_failure: default_transient_rpc_retry_policy(),
+            retry_on_not_included_on_l1: default_non_inclusion_rpc_retry_policy(),
             confirmations: default_confirmations(),
             finality: Finality::default(),
             gas_limit_multiplier_factor: Multiplier::default(),
-            gas_limit: default_gas_limit(),
+            gas_limit_ceiling: default_gas_limit_ceiling(),
             gas_price_multiplier_factor: Multiplier::default(),
             gas_price_floor: 0,
             gas_price_ceiling: default_gas_price_ceiling(),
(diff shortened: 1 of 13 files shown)


crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,312 @@
...
+
+/// Finality level for settlement transactions on Ethereum.
+///
+/// Defines when a transaction should be considered settled on the Ethereum
+/// network, determining the security guarantees for the settlement operation.
+///
+/// # Example
+///
+/// ```toml
+/// [settlement.certificate-tx-config]
```
Freyskeyd: Nit: settlement isn't about certificate but pessimistic-proof

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,312 @@
...
+    pub tx_retry_interval: Duration,
+
+    /// Number of block confirmations required for
+    /// the transaction to resolve a receipt.
+    #[serde(default = "default_confirmations")]
+    pub confirmations: usize,
+
+    /// Finality level required for the transaction to be considered settled.
+    #[serde(default)]
+    pub finality: Finality,
```
Freyskeyd, line 107: Could you explain the difference in terms of behavior for those two?

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,312 @@
...
+/// and focuses on overall service behavior and integration points.
+///
+/// # Future Configuration Options
+///
+/// This structure is designed to hold settlement service-specific values that
+/// are not related to individual transactions. The service relies on other
+/// configuration sources for:
+///
+/// - **L1 Provider**: RPC connection to the L1 blockchain
+/// - **Agglayer Contracts**: Smart contract addresses and ABIs
```
Freyskeyd: I though that the contract address / ABIs were agnostic and we were only dealing with address and encoded tx?

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,312 @@
...
+/// - **Transaction Signer**: Private keys or KMS configuration for signing
+///
+/// Future additions to this configuration may include:
+/// - Service health check intervals
+/// - Monitoring and metrics endpoints
+/// - Settlement batching strategies
```
Freyskeyd: I don't think we should define this here, and I'm also not sure that this is going to be the responisbility of this service to handle that

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,312 @@
...
+///   validium data (off-chain data availability).
+/// - **Settlement Service Config**: General settlement service configuration.
+///
+/// # Example TOML Configuration
+///
+/// ```toml
+/// [settlement]
+/// [settlement.certificate-tx-config]
+/// max-retries = 1024
+/// tx-retry-interval = "10s"
```
Freyskeyd, line 208: Those retries are covering which part of the transaction lifecycle?

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: It is; I think @atanmarko meant exactly that we're not configuring this in the settlement service. I've just deleted these paragraphs, that don't bring much value

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: I'm not sure I understand this comment?

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: See the thread at https://0xpolygon.slack.com/archives/C089DQN6WPL/p1767628513259279

Ekleog-Polygon (dismissed):
Overall LGTM; I'll push my commits directly to the branch :smile:

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+/// [settlement.pessimistic-proof-tx-config]
+/// confirmations = 16
+/// finality = "SafeBlock"
+///
+/// [settlement.validium-tx-config]
+/// confirmations = 6
+/// finality = "LatestBlock"  # Faster settlement for validium data
+/// ```
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
+pub enum Finality {
```
iljakuklic, line 26: Small style / formatting comment:

Should the `serde(rename_all = "kebab-case")` attribute be applied here? Seems it would be more consistent with the rest of the config.

I.e. the format would become

```toml
[settlement.validium-tx-config]
finality = "latest-block"
```

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+    #[serde(default = "default_rpc_max_retries")]
+    pub max_retries: usize,
+
+    /// Retry policy for the transaction when there is a transient failure.
+    #[serde(default = "default_transient_rpc_retry_policy")]
+    pub retry_on_transient_failure: TxRetryPolicy,
+
+    /// Retry policy for the transaction when it is not included on L1.
+    #[serde(default = "default_non_inclusion_rpc_retry_policy")]
+    pub retry_on_not_included_on_l1: TxRetryPolicy,
```
iljakuklic, line 98: Not sure if it's worth the effort but there could also be a shared general retry policy and `retry-on-*` would just contain overrides.

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+    #[serde(default, skip_serializing_if = "crate::is_default")]
+    #[serde_as(as = "crate::with::EthAmount")]
+    pub gas_price_floor: u128,
+
+    /// Maximum gas price ceiling (in wei) for the transaction.
+    /// Use for both EIP1559 `max_fee_per_gas` and `max_priority_fee_per_gas`.
+    /// Can be specified with units: "100gwei", "0.01eth", "10000000000wei"
+    #[serde(default = "default_gas_price_ceiling")]
+    #[serde_as(as = "crate::with::EthAmount")]
+    pub gas_price_ceiling: u128,
```
iljakuklic, line 138: Would it be nicer to have a separate section for gas-related settings?

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+    pub settlement_service_config: SettlementServiceConfig,
+}
+
+const fn default_rpc_max_retries() -> usize {
+    16 * 1024
+}
+
+const fn default_max_interval() -> Duration {
+    Duration::from_secs(60 * 60 * 24 * 36525) // effectively infinite: ~100
+                                              // years
```
iljakuklic, line 286: Conceptually it would make sense to pick `Duration::MAX` as "effectively infinite" duration. Are the reasons to not do it? Overflow concerns? Or just `Option<Duration>` for even more accurate modelling.

(Edit: not relevant if the default is removed as discussed in the other comment.)

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+    pub initial_interval: Duration,
+
+    /// Interval multiplier for each subsequent retry.
+    #[serde(default, skip_serializing_if = "crate::is_default")]
+    pub interval_multiplier_factor: Multiplier,
+
+    /// Maximum interval between retries.
+    #[serde(default = "default_max_interval")]
+    #[serde(with = "crate::with::HumanDuration")]
+    pub max_interval: Duration,
```
iljakuklic, line 74: I'm a bit concerned with the defaulting behaviour here. If the retry policy section is missing, either `default_transient_rpc_retry_policy` or `default_non_inclusion_rpc_retry_policy` gets applied, depending on context. These are 120s and 600s respectively.

However, introducing the retry policy section, *even without specifying `max-interval`* would change it to `default_max_interval` (100yrs). That is somewhat unexpected and counter-intuitive behaviour.

Ideally, introducing the section without overriding the value would not change the value. Alternatively, just remove the default here so the non-default retry policy always has to be specified in full (simpler to implement).

As far as I can tell, `interval-multiplier-factor` has the same issue.

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: I think not, for consistency with aggkit — see https://0xpolygon.slack.com/archives/C07HZQHN6RY/p1767630190850039

No strong opinions here that said.

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: I'm with you, this is suboptimal. This being said, for the sake of velocity, maybe we can land as-is for now and investigate how to improve on the defaulting behavior later?

At least with this we can start implementing the settlement logic :smile:

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: I'm with you; but as you say it might not be worth the effort

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: I have no strong opinion here :smile: though TBH I'd be all in favor of iterating on this later; because IMO this is not the most critical part of the settlement service.

Completely open to adjusting this — for full disclosure, I'm pretty sure we'll need to change some of that when we implement the settlement logic, in the same way as we'll need to change some of the protobuf design; because we'll need to make sure the two of them do match properly.

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: `Option<Duration>` was my initial choice too, but it turns out we don't have simple parsers for it. I hadn't thought of `Duration::MAX`, but my pick of 100 years was so that the display of the default duration is easy to read (see the toml snapshots). I fear that Duration::MAX would make that harder on humans reading potential logs?

iljakuklic: > `Option<Duration>` was my initial choice too, but it turns out we don't have simple parsers for it.

Have you tried `#[serde_as(as = "Option<HumanDuration>")]`? (also needs a `#[serde_as]` at the top of the struct)

iljakuklic: > but my pick of 100 years was so that the display of the default duration is easy to read (see the toml snapshots). I fear that Duration::MAX would make that harder on humans reading potential logs?

Good point. Should be something more descriptive, e.g. `null` in case of `Option` or even an enum with an `"unlimited"` arm like we have in rate limiter config.

Ekleog-Polygon:
Note to self: let's rename max_retries into max_expected_retries

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+///
+/// Defines when a transaction should be considered settled on the Ethereum
+/// network, determining the security guarantees for the settlement operation.
+///
+/// # Example
+///
+/// ```toml
+/// [settlement.pessimistic-proof-tx-config]
+/// confirmations = 16
+/// finality = "SafeBlock"
```
hadjiszs, line 19: i'm not fan of the usage of `finality`, `Finality level` and also the name of the enum `Finality` there. I think "finality" refers to a boolean state, either a tx hasnt reached finality yet, or it has and is now finalized (the `FinalizedBlock` value, every other values are pre-finality and not finality I think)

in particular reading a toml with `finality = "LatestBlock"` feels odd to me

I'd rename to `settlement-level` or `settlement-policy`, the block level at which we consider the tx settled (but not necessarily final unless `settlement-level = "FinalizedBlock"`)

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+/// [settlement.validium-tx-config]
+/// max-retries = 512
+/// tx-retry-interval = "5s"
+/// confirmations = 16
+/// finality = "justified"
+/// gas-limit = 30000000
+/// gas-multiplier-factor = 1.05
+/// gas-price-floor = "1gwei"
+/// gas-price-ceiling = "50gwei"
+/// ```
```
hadjiszs, line 225: nit: if we iterate on the config, the toml in the `snapshots` folder will stay up-to-date but not this doc no? do we want to spell out the config example there?

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,321 @@
...
+/// gas-limit = 60000000
+/// gas-multiplier-factor = 1.1
+/// gas-price-multiplier-factor = 1.2
+/// gas-price-ceiling = "100gwei"
+///
+/// [settlement.validium-tx-config]
+/// max-retries = 512
+/// tx-retry-interval = "5s"
+/// confirmations = 16
+/// finality = "justified"
```
hadjiszs, line 220: from some unit test names, I see `justified` corresponds to `SafeBlock` but in some toml I see directly `finality = "SafeBlock"`?

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: Great point! I didn't notice we also had `serde_as`, switched to that :smile:

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: Ended up actually fixing this, it's a bit of complexity but should handle config defaults properly.

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: Done!

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: I had the same thought, had let it here to smoothen the process, but if you're of the same opinion I'm removing it :smile:

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: This is a perfect example of the above — I adjusted the naming to match aggkit naming and forgot to update this instance :sweat_smile:

Ekleog-Polygon:
Closing as replaced by #1310

PR #1292, opened 2025-12-31
feat: introduce column descriptor separate from rocksdb

iljakuklic:
This is to:
* further isolate agglayer code from from the hairy rocksdb details
* make the interface less error prone
* make the interface more consistent

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/l1_context.rs
@@ -46,7 +46,7 @@ where
 
         let l1_info_root = self.fetch_l1_info_root(certificate).await?;
 
-        let aggchain_data_paylaod = CertificateAggchainData::try_from(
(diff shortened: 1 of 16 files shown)


crates/agglayer-storage/src/schema/options.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// Lz4 compression.
+    #[default]
+    Lz4,
+}
+
+impl ColumnCompressionType {
+    pub const DEFAULT: Self = Self::Lz4;
+
+    pub fn to_rocksdb_compression_type(self) -> rocksdb::DBCompressionType {
```
Freyskeyd, line 15: We can't use a proper `From` implementation ?

iljakuklic: Added some From impls. However, ultimately, these types should abstract away from rocksdb details.

Ekleog-Polygon (approved):

PR #1293, opened 2025-12-31
feat(config): add human-readable size format for config values

### LINKED ISSUE #1204: Switch to human-readable form for the size specification. The config values here will be huge numbers which are rather difficult to read.
Would be nice to use some sort of [human-readable format](https://crates.io/crates/byte-unit) for the size specification. The config values here will be huge numbers which are rather difficult to read.

_Originally posted by @iljakuklic in https://github.com/agglayer/agglayer/pull/1195#discussion_r2561223270_

Himess:
## Summary
- Added `HumanSize` serde helper for human-readable byte size configuration values
- Updated `GrpcConfig` to use human-readable format for `max_decoding_message_size` and `max_encoding_message_size`
- Users can now specify sizes like `"64 MiB"`, `"100MB"`, or `"1GiB"` instead of raw byte counts

### grpc.max-decoding-message-size
- **Type**: Size value (bytes or human-readable string)
- **Default**: `64 MiB` (67108864 bytes)
- **Description**: Maximum size for incoming gRPC messages
- **Example**: `max-decoding-message-size = "100 MiB"` or `max-decoding-message-size = 104857600`

### grpc.max-encoding-message-size
- **Type**: Size value (bytes or human-readable string)
- **Default**: `64 MiB` (67108864 bytes)
- **Description**: Maximum size for outgoing gRPC messages
- **Example**: `max-encoding-message-size = "100 MiB"` or `max-encoding-message-size = 104857600`

## Test plan
- [x] Added unit tests for HumanSize serialization/deserialization
- [x] Updated snapshot tests for new format
- [ ] CI verification (pending Cargo.lock update - unable to generate on Windows due to file naming restrictions)

Closes #1204

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,6 +176,7 @@ dependencies = [
  "agglayer-primitives",
  "agglayer-types",
  "alloy-primitives",
+ "byte-unit",
  "educe",
  "humantime-serde",
  "insta",
@@ -687,6 +688,17 @@ name = "agglayer-utils"
 version = "0.1.0"
 dependencies = [
  "tracing",
+]
+
+[[package]]
+name = "ahash"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
+dependencies = [
+ "getrandom 0.2.16",
+ "once_cell",
+ "version_check",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


Himess:
Hi @iljakuklic, thanks for the review!

I've updated the snapshot for the human-readable size format. However, I'm unable to update the `Cargo.lock` file from my Windows machine due to the repository containing files with `::` in their names (which Windows doesn't support).

Could you please help regenerate the `Cargo.lock` 

I'll update the PR description with the CONFIG-CHANGE section shortly.

PR #1294, opened 2025-12-31
refactor(storage): rename DB types to Db for Rust naming consistency

### LINKED ISSUE #1248: Database objects naming consistency
The `rocksdb` crate defines names such as `DB` and `DBError`. A more Rusty convention would be `Db` and `DbError`.

Currently we follow the rocksdb convention. However, rocksdb is mostly an implementation detail so it would be nicer to follow the Rust convention for consistency with the rest of the code base. This would touch many names in the `agglayer-storage` crate and some outside of it too, so large diff and conflicts are expected.

* raised in https://github.com/agglayer/agglayer/pull/1211#discussion_r2613937520

Himess:
Closes #1248


Himess:
@iljakuklic Could you look ?

atanmarko (approved):

PR #1298, opened 2026-01-03
docs: update description to reference SettlementClient

letmehateu:
Updates description in README.md to reference the current `SettlementClient` trait instead of the deleted `EpochPacker`.


atanmarko:
Settlement client does not handle packing of the epoch.

PR #1301, opened 2026-01-03
refactor: extract log clone in try_new for clarity

PivasDesant:
Extract log clone to a variable before decoding. Makes the code clearer and reuses the clone for logging. Clone is still needed since into() takes ownership.


atanmarko:
I don’t think a separate line for the clone operation is worth it. I would say it makes the code less readable.

PR #1303, opened 2026-01-06
feat: implement settlement service logic and settlement task api

Ekleog-Polygon:
Closes https://github.com/agglayer/protocol-research/issues/221
Closes https://github.com/agglayer/agglayer/issues/1257
Closes https://github.com/agglayer/protocol-research/issues/222
Closes https://github.com/agglayer/agglayer/issues/1258

### DIFF
--- a/crates/agglayer-settlement-service/src/settlement_service.rs
+++ b/crates/agglayer-settlement-service/src/settlement_service.rs
@@ -10,6 +10,8 @@ use crate::settlement_task::{
     SettlementJob, SettlementJobResult, SettlementTask, TaskAdminCommand,
 };
 
+const ADMIN_CHANNEL_BUFFER_SIZE: usize = 10;
+
 pub enum ServiceAdminCommand {
     TaskCommand {
         job_id: Ulid,
@@ -19,10 +21,23 @@ pub enum ServiceAdminCommand {
 
 pub struct SettlementService {
     admin_command_senders: Arc<Mutex<HashMap<Ulid, mpsc::Sender<TaskAdminCommand>>>>,
-    result_watchers: Arc<Mutex<HashMap<Ulid, SettlementJobWatcher>>>,
+    result_watchers: Arc<Mutex<HashMap<Ulid, watch::Receiver<Option<SettlementJobResult>>>>>,
 }
 
-pub type SettlementJobWatcher = watch::Receiver<Option<SettlementJobResult>>;
+pub struct SettlementJobWatcher {
+    watcher: watch::Receiver<Option<SettlementJobResult>>,
+    job_id: Ulid,
+}
+
+impl SettlementJobWatcher {
+    pub fn watcher(&mut self) -> &mut watch::Receiver<Option<SettlementJobResult>> {
+        &mut self.watcher
+    }
+
+    pub fn job_id(&self) -> Ulid {
+        self.job_id
+    }
+}
 
 pub enum RetrievedSettlementResult {
     Pending(SettlementJobWatcher),
@@ -102,8 +117,8 @@ impl SettlementService {
     pub async fn request_new_settlement(
         &self,
         job: SettlementJob,
-    ) -> eyre::Result<(Ulid, SettlementJobWatcher)> {
-        let (admin_sender, admin_receiver) = mpsc::channel(10);
+    ) -> eyre::Result<SettlementJobWatcher> {
+        let (admin_sender, admin_receiver) = mpsc::channel(ADMIN_CHANNEL_BUFFER_SIZE);
         let (result_sender, result_receiver) = watch::channel(None);
         let (job_id, mut task) = SettlementTask::create(job, admin_receiver).await?;
         self.admin_command_senders
@@ -124,7 +139,10 @@ impl SettlementService {
                 );
             }
         });
-        Ok((job_id, result_receiver))
+        Ok(SettlementJobWatcher {
+            watcher: result_receiver,
+            job_id,
+        })
     }
 
     #[tracing::instrument(skip(self))]
@@ -134,7 +152,10 @@ impl SettlementService {
     ) -> eyre::Result<RetrievedSettlementResult> {
         if let Some(watcher) = self.result_watchers.lock().await.get(&job_id) {
             return match watcher.borrow().as_ref() {
-                None => Ok(RetrievedSettlementResult::Pending(watcher.clone())),
+                None => Ok(RetrievedSettlementResult::Pending(SettlementJobWatcher {
+                    job_id,
+                    watcher: watcher.clone(),
+                })),
                 Some(result) => Ok(RetrievedSettlementResult::Completed(result.clone())),
             };
         }
--- a/crates/agglayer-settlement-service/src/settlement_task.rs
+++ b/crates/agglayer-settlement-service/src/settlement_task.rs
@@ -1,11 +1,12 @@
 use std::{sync::OnceLock, time::SystemTime};
 
 use agglayer_config::Multiplier;
+use agglayer_types::SettlementTxHash;
 use alloy::primitives::{Address, BlockHash, Bytes, U128, U256};
 use tokio::sync::mpsc;
 use ulid::Ulid;
 
-#[derive(Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct SettlementJob {
     contract_address: Address,
     calldata: Bytes,
@@ -21,39 +22,41 @@ pub struct SettlementJob {
     max_priority_fee_per_gas_multiplier: Multiplier,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum SettlementJobResult {
     ClientError(ClientError),
-    ContractCallResult(ContractCallResult),
-    ReorganizedResult(ReorganizedResult),
+    ContractCall(ContractCallResult),
+    Reorganized(ReorganizedResult),
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ClientError {
     pub kind: ClientErrorType,
     pub message: String,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum ClientErrorType {
     Transient,
     Permanent,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ContractCallResult {
     pub outcome: ContractCallOutcome,
     pub metadata: Bytes,
     pub block_hash: BlockHash,
+    pub block_number: u64,
+    pub tx_hash: SettlementTxHash,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum ContractCallOutcome {
     Success,
     Revert,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ReorganizedResult {
     pub reorg_detection_time: SystemTime,
     pub previous_result: Box<SettlementJobResult>,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -530,6 +530,7 @@ name = "agglayer-settlement-service"
 version = "0.1.0"
 dependencies = [
  "agglayer-config",
+ "agglayer-types",
  "alloy",
  "eyre",
  "tokio",
--- a/crates/agglayer-settlement-service/Cargo.toml
+++ b/crates/agglayer-settlement-service/Cargo.toml
@@ -6,6 +6,7 @@ license.workspace = true
 
 [dependencies]
 agglayer-config.workspace = true
+agglayer-types.workspace = true
 
 alloy.workspace = true
 eyre.workspace = true
--- a/crates/agglayer-storage/src/types/generated/agglayer.storage.v0.rs
+++ b/crates/agglayer-storage/src/types/generated/agglayer.storage.v0.rs
@@ -260,6 +260,13 @@ pub struct BlockHash {
     #[prost(bytes="bytes", tag="1")]
     pub hash: ::prost::bytes::Bytes,
 }
+/// Block number.
+#[derive(Clone, Copy, PartialEq, ::prost::Message)]
+pub struct BlockNumber {
+    /// Block number.
+    #[prost(uint64, tag="1")]
+    pub number: u64,
+}
 /// Settlement job ID.
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct SettlementJobId {
@@ -353,11 +360,20 @@ pub struct ContractCallResult {
     #[prost(enumeration="ContractCallOutcome", tag="1")]
     pub outcome: i32,
     /// Additional metadata about the call outcome.
+    ///
+    /// This will be either the success outcome or the raw revert reason.
+    /// At this stage we cannot parse this further, due to not knowing which ABI was used.
     #[prost(bytes="bytes", tag="2")]
     pub metadata: ::prost::bytes::Bytes,
-    /// Block number where the transaction was included.
+    /// Block hash where the transaction was included.
     #[prost(message, optional, tag="3")]
     pub block_hash: ::core::option::Option<BlockHash>,
+    /// Block number where the transaction was included.
+    #[prost(message, optional, tag="4")]
+    pub block_number: ::core::option::Option<BlockNumber>,
+    /// Transaction hash of the settlement transaction.
+    #[prost(message, optional, tag="5")]
+    pub tx_hash: ::core::option::Option<TxHash>,
 }
 /// Result indicating that the transaction's previous result was reorganized out of the chain.
 #[derive(Clone, PartialEq, ::prost::Message)]
--- a/proto/agglayer/storage/v0/settlement_service.proto
+++ b/proto/agglayer/storage/v0/settlement_service.proto
@@ -98,10 +98,19 @@ message ContractCallResult {
   ContractCallOutcome outcome = 1;
 
   // Additional metadata about the call outcome.
+  //
+  // This will be either the success outcome or the raw revert reason.
+  // At this stage we cannot parse this further, due to not knowing which ABI was used.
   bytes metadata = 2;
 
-  // Block number where the transaction was included.
+  // Block hash where the transaction was included.
   BlockHash block_hash = 3;
+
+  // Block number where the transaction was included.
+  BlockNumber block_number = 4;
+
+  // Transaction hash of the settlement transaction.
+  TxHash tx_hash = 5;
 }
 
 // On-chain outcome of the contract call.
--- a/proto/agglayer/storage/v0/types.proto
+++ b/proto/agglayer/storage/v0/types.proto
@@ -49,3 +49,9 @@ message BlockHash {
   // Hash bytes.
   bytes hash = 1;
 }
+
+// Block number.
+message BlockNumber {
+  // Block number.
+  uint64 number = 1;
+}


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+}
+
+#[derive(Clone)]
+pub enum ContractCallOutcome {
+    Success,
+    Revert,
+}
+
+#[derive(Clone)]
+pub struct ReorganizedResult {
```
iljakuklic: Some of the types here would benefit from `#[derive(Debug)]` for logging and such and `#[derive(Eq)]` for checking test values. However, that could be done on demand as we go.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+}
+
+#[derive(Clone)]
+pub struct ClientError {
+    pub kind: ClientErrorType,
+    pub message: String,
+}
+
+#[derive(Clone)]
+pub enum ClientErrorType {
```
adria0pol, line 39: I like this!

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,144 @@
...
+        }
+    }
+
+    #[tracing::instrument(skip_all)]
+    async fn cancellation_token_proxy(
+        cancellation_token: CancellationToken,
+        senders: Arc<Mutex<HashMap<Ulid, mpsc::Sender<TaskAdminCommand>>>>,
+    ) {
+        cancellation_token.cancelled().await;
+        let senders = senders.lock().await;
```
iljakuklic, line 104: Out of curiosity, why did you decide to do this rather than just cloning the cancellation token into the job tasks?

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,144 @@
...
+                );
+            }
+        }
+    }
+
+    #[tracing::instrument(skip(self))]
+    pub async fn request_new_settlement(
+        &self,
+        job: SettlementJob,
+    ) -> eyre::Result<(Ulid, SettlementJobWatcher)> {
```
iljakuklic: Nit: might be nicer to return a Handle struct that bundles the Ulid and the watcher together. I suspect these will sometimes be used together.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,144 @@
...
+            result_watchers: Arc::new(Mutex::new(HashMap::new())),
+        };
+        tokio::task::spawn(Self::admin_command_proxy(
+            admin_commands,
+            this.admin_command_senders.clone(),
+        ));
+        tokio::task::spawn(Self::cancellation_token_proxy(
+            cancellation_token,
+            this.admin_command_senders.clone(),
+        ));
```
iljakuklic, line 64: We do not take handles of these tasks which means we have no way (as far as I can tell) of checking/handling if any of them panicked. Are there plans to address this issue?

iljakuklic (commented):
Good start. Some random thoughts below:

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+use std::{sync::OnceLock, time::SystemTime};
 
-pub struct SettlementTask {}
+use agglayer_config::Multiplier;
+use alloy::primitives::{Address, BlockHash, Bytes, U128, U256};
+use tokio::sync::mpsc;
+use ulid::Ulid;
+
+#[derive(Debug)]
+pub struct SettlementJob {
```
adria0pol, line 10: Usually found that `Job` is a set of `Tasks`, so it's a little confusing for me.
Can we rename it like `SettlementParams` or something like this?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+}
+
+#[derive(Clone)]
+pub enum ClientErrorType {
+    Transient,
+    Permanent,
+}
+
+#[derive(Clone)]
+pub struct ContractCallResult {
```
adria0pol, line 45: What about adding `tx_hash` and `block_number` ?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+    pub outcome: ContractCallOutcome,
+    pub metadata: Bytes,
+    pub block_hash: BlockHash,
+}
+
+#[derive(Clone)]
+pub enum ContractCallOutcome {
+    Success,
+    Revert,
```
adria0pol, line 56:
```suggestion
    RevertWithReason(String),
```

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+}
+
+#[derive(Clone)]
+pub struct ReorganizedResult {
+    pub reorg_detection_time: SystemTime,
```
adria0pol, line 61: Mmm.... are we going to handle reorganizations?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+pub enum TaskAdminCommand {
+    Abort,
+    ReloadAndRestart,
+}
+
+pub struct SettlementTask {
+    id: Ulid,
+    job: SettlementJob,
+    admin_commands: mpsc::Receiver<TaskAdminCommand>,
+}
```
adria0pol, line 79:
```suggestion
   gas : TaskParam<U128>,
   gas_price : TaskParam<U128>,
   max_fee_per_gas : TaskParam<U128>,
   max_priority_fee_per_gas : TaskParam<U128>,
}
```

I think that it will be nice to keep the values that we have to modify with admin premissions, allowing them to set a fixed value without applying `max_` paramerrs.

```
enum TaskParam<T> {
   Dynamic(T),
   Fixed(T)
}
```

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+
+pub enum StoredSettlementJob {
+    Pending(SettlementTask),
+    Completed(SettlementJob, SettlementJobResult),
+}
+
+pub enum TaskAdminCommand {
+    Abort,
+    ReloadAndRestart,
+}
```
adria0pol, line 73:
```suggestion
    BumpGasPriceTo(U128),
    // I really do not know if we are going to use them
    BumpGasTo(U128)
    BumpMaxFeePerGas(U128),
    BumpMaxPriorityFeePerGas(U128)
}
```

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: We will certainly want the admin API to be able to stop a specific settlement task anyway, for any human-determined reason. By merging the two inputs together we can reduce the API surface for the settlement task, and make its future `select!`s a bit simpler :smile:

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: Hmm… so my expectation was that the `Ulid` would be basically immediately saved to disk and then forgotten about; but the watcher would stay alive forever. This being said, I don't have any argument against a specific struct! I'll probably just make the fields `pub` to avoid too much boilerplate.

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: I think we will know of any panic: the watcher's sender end will automatically close, which will notify any client that's waiting on it.

This being said, there is indeed no way of handling it. Do you think we'd need any specific handling of panics? TBH the default rust logging of panics plus notifying any client that'd be waiting for the result sounds good enough for me, at least for the first iteration :smile:

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: Hmm so this is a continuation of #1251 and #1227 in particular.

I think it also matches your expectation: a `Job` is one settlement we want to perform, and we'll spin up a `Task` on each agglayer-node reboot to actually execute it.

Considering this defines an "unit of work to do", I don't think `Params` would make much sense.

WDYT?

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: No strong opinions, I went with whatever had been defined in #1251 here. I'll probably add these at both places in a follow-up PR to unblock concurrent work :smile:

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: AFAICT we can't actually have this, because decoding the reason requires knowing which contract function was called… I think? That's the reason why we have the `metadata` field in the outer `ContractCallResult`, that contains the unparsed details of the outcome

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: Considering that we're assigning nonces ourselves we don't have much choice, as after a reorg we'll need to resend all the txes that had been included before, hence why it's being tracked at [[Design] Design the reorg detection logic for the settlement service (optional)](https://github.com/agglayer/protocol-research/issues/220#top)#220

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: I'm not sure I understand? These parameters are already configured in `SettlementJob`. As for admin permissions to modify them, my thinking was that the admin would call some RPC to edit them in database, and then send the admin command "reload and restart" to force the settlement task to read the new settlement job from database.

(Also, if we ever do need fixed parameters, we would probably do that by just setting `*_ceiling = *_floor` — I'm not sure adding code to handle the rare case of fixed fees would be worth it)

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: See https://github.com/agglayer/agglayer/pull/1303#discussion_r2669647538 for my reasoning — each admin command adds complexity, so for the first iteration I'm thinking we can go with the simplest option that more or less works and covers all use cases; and we can iterate on it if we find out that we frequently change some specific parameters.

crates/agglayer-settlement-service/src/settlement_service.rs
iljakuklic: > Do you think we'd need any specific handling of panics?

Not really, just paranoia. These tasks are fairly simple loops so the risk is low. I am thinking if for example we want to restart them in case of a panic or something like that. We currently only do it for network task so maybe it's not something to think about now.

crates/agglayer-settlement-service/src/settlement_service.rs
iljakuklic: > Hmm… so my expectation was that the `Ulid` would be basically immediately saved to disk and then forgotten about

Fair enough, if it's not used in conjunction with the watcher, we don't need to bundle them together. The particular case I was thinking about is if the watcher errors out for some reason, we may want to log ulid on the wathcher side for debugging purposes but maybe it could be done differently or is not necessary.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,144 @@
...
+    }
+
+    #[tracing::instrument(skip_all)]
+    async fn cancellation_token_proxy(
+        cancellation_token: CancellationToken,
+        senders: Arc<Mutex<HashMap<Ulid, mpsc::Sender<TaskAdminCommand>>>>,
+    ) {
+        cancellation_token.cancelled().await;
+        let senders = senders.lock().await;
+        for (job_id, sender) in senders.iter() {
```
polykeith, line 105: IIRC this works:

```suggestion
        for (job_id, sender) in &senders {
```

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,144 @@
...
+            }
+        }
+    }
+
+    #[tracing::instrument(skip(self))]
+    pub async fn request_new_settlement(
+        &self,
+        job: SettlementJob,
+    ) -> eyre::Result<(Ulid, SettlementJobWatcher)> {
+        let (admin_sender, admin_receiver) = mpsc::channel(10);
```
polykeith: Might want to store this 10 as a constant, rather than hardcoding it here, or even better -- have it be configurable.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+    max_priority_fee_per_gas_ceiling: U128,
+    max_priority_fee_per_gas_floor: U128,
+    max_priority_fee_per_gas_multiplier: Multiplier,
+}
+
+#[derive(Clone)]
+pub enum SettlementJobResult {
+    ClientError(ClientError),
+    ContractCallResult(ContractCallResult),
+    ReorganizedResult(ReorganizedResult),
```
polykeith: nit: The enum itself already has `Result`, so having discriminants that still say `Result` is redundant.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1,12 +1,117 @@
...
+    Pending(SettlementTask),
+    Completed(SettlementJob, SettlementJobResult),
+}
+
+pub enum TaskAdminCommand {
+    Abort,
+    ReloadAndRestart,
+}
+
+pub struct SettlementTask {
```
polykeith, line 75: Honestly, it now sounds like to me that this isn't just a "task" in the sense that it's a unit of work, but a thread altogether.

I propose renaming as such:
 - `SettlementTask` -> `SettlementThread`
 - `SettlementJob` -> `SettlementTask`

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: Good point, I hadn't thought about that; just did it :smile: And I didn't make the fields `pub` after all, because this way we're sure users won't eg. mistakenly edit the job id.

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: Hmm so I gave it a bit of thought, and I'm actually thinking it might be better to just let the task die than to auto-restart it. Otherwise if there's an unexpected panic, we could enter a busy-loop of restarting the task and it crashing again forever. Or we'd need an actually intelligent restart logic with timeouts… but unfortunately rust doesn't yet have erlang-level supervisor quality :smiling_face_with_tear:

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: Actually did it now :smile:

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: Unfortunately here `senders` is a `MutexGuard`, so we'd need `&*senders` to go through the `impl Deref for MutexGuard`. At this stageTB, I'm thinking a `.iter()` would look less confusing, though I don't have strong opinions. WDYT?

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: Made it a constant, TBH I don't think it's worth being configurable — I don't expect us to ever submit 10 different admin commands before the settlement task manages to read them, which should usually happen within a few seconds, maybe minutes at worst :sweat_smile:

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: Done!

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: I don't think thread would be the correct word here — Rust has stabilized on [the task naming](https://docs.rs/futures/latest/futures/task/) (like [in tokio](https://docs.rs/tokio/latest/tokio/task/index.html)) for one green thread running on a futures executor.

That's actually also the reason why we have `CertificateTask` and `NetworkTask` in `agglayer-certificate-orchestrator` :smile: 

So while I'm completely open to renaming `SettlementJob`, though I don't have any better idea; I don't think it'd be a good idea to rename `SettlementTask`. Does that make sense to you?

crates/agglayer-settlement-service/src/settlement_task.rs
adria0pol: `would call some RPC to edit them in database`
`*_ceiling = *_floor` 

This combinarion will also work  +1!

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,165 @@
...
+pub struct SettlementJobWatcher {
+    watcher: watch::Receiver<Option<SettlementJobResult>>,
+    job_id: Ulid,
+}
 
-pub struct SettlementService {}
+impl SettlementJobWatcher {
+    pub fn watcher(&mut self) -> &mut watch::Receiver<Option<SettlementJobResult>> {
+        &mut self.watcher
+    }
```
iljakuklic, line 35: Note: having this watcher in a struct will allow us to introduce domain-specific methods, like `.wait_for_settlement()` or `.is_settled()` and such.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,165 @@
...
+use tokio::sync::{mpsc, watch, Mutex};
+use tokio_util::sync::CancellationToken;
+use tracing::error;
+use ulid::Ulid;
+
+use crate::settlement_task::{
+    SettlementJob, SettlementJobResult, SettlementTask, TaskAdminCommand,
+};
+
+const ADMIN_CHANNEL_BUFFER_SIZE: usize = 10;
```
Freyskeyd, line 13: Shouldn't it be a configuration params with a default value?

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,165 @@
...
+
+pub enum ServiceAdminCommand {
```
Freyskeyd, line 15: Missing comment

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,165 @@
...
+    TaskCommand {
+        job_id: Ulid,
+        command: TaskAdminCommand,
+    },
+}
+
+pub struct SettlementService {
```
Freyskeyd, line 22: Missing comment

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -1,9 +1,165 @@
...
+    admin_command_senders: Arc<Mutex<HashMap<Ulid, mpsc::Sender<TaskAdminCommand>>>>,
```
Freyskeyd, line 23: this isn't supposed to be a tower service?

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: I don't think it'd be worth it; I can't imagine any scenario where we would actually want to configure it :sweat_smile: 

See https://github.com/agglayer/agglayer/pull/1303#discussion_r2672147690 for more details

PR #1308, opened 2026-01-08
fix: update the aggchain hash mismatch check to be for all path

### LINKED ISSUE #1045: Update the aggchain hash mismatch check to be for all path
# Context

We already have a check on mismatch of the aggchain hash here:

https://github.com/agglayer/agglayer/blob/6c507f1f3125b019d0f36eade8c1b9e621bdc2a6/crates/agglayer-aggregator-notifier/src/certifier/mod.rs#L425-L430

However, this check is verified only for `AggchainData::Generic` (the old path for katana). We ideally want this check to be performed for all the paths.

> [!NOTE] 
> This check isn't critical to have but nice to have:
> - without the check: the mismatch occurs upon settlement attempt on the L1: `InvalidProof`
> - with the check: the mismatch occurs as an error on the agglayer side.

# To do 

We need to update it to have this mismatch being checked for all the paths, because now all the chains are `consensus_type == 1`, so they normally all expose one `aggchain_hash` (unlike before).

adria0pol:
Closes #1045

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/tests.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/tests.rs
@@ -51,15 +51,6 @@ async fn happy_path() {
     let signer = state.get_signer();
     let certificate_id = certificate.hash();
 
-    let agglayer_types::aggchain_proof::AggchainData::ECDSA { signature } =
-        certificate.aggchain_data
-    else {
-        panic!("this test expects one ecdsa as aggchain data")
-    };
-
-    let aggchain_hash =
-        pessimistic_proof::core::AggchainData::LegacyEcdsa { signer, signature }.aggchain_hash();
-
     pending_store
         .expect_get_certificate()
         .once()
@@ -79,13 +70,8 @@ async fn happy_path() {
 
     l1_rpc
         .expect_get_rollup_contract_address()
-        .times(2)
-        .returning(|_| Ok(Address::ZERO));
-
-    l1_rpc
-        .expect_get_aggchain_hash()
         .once()
-        .returning(move |_, _| Ok(*aggchain_hash));
+        .returning(|_| Ok(Address::ZERO));
 
     l1_rpc
         .expect_default_l1_info_tree_entry()
@@ -165,13 +151,8 @@ async fn prover_timeout() {
 
     l1_rpc
         .expect_get_rollup_contract_address()
-        .times(2)
-        .returning(|_| Ok(Address::ZERO));
-
-    l1_rpc
-        .expect_get_aggchain_hash()
         .once()
-        .returning(|_, _| Ok([1u8; 32]));
+        .returning(|_| Ok(Address::ZERO));
 
     l1_rpc
         .expect_default_l1_info_tree_entry()


crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -353,45 +353,56 @@ where
...
-            ..
-        } = &certificate.aggchain_data
+        // Check if the certificate multisig hash matches the one stored in L1
+
+        if matches!(certificate.aggchain_data, AggchainData::MultisigOnly { .. })
+            || matches!(
+                certificate.aggchain_data,
+                AggchainData::MultisigAndAggchainProof { .. }
+            )
+            || matches!(certificate.aggchain_data, AggchainData::Generic { .. })
```
Freyskeyd:
```suggestion
        if matches!(certificate.aggchain_data,
          AggchainData::MultisigOnly { .. } |
          AggchainData::MultisigAndAggchainProof { .. } |
          AggchainData::Generic { .. }
        )
```

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -353,45 +353,56 @@ where
...
-            // Fetching rollup contract address
             let rollup_address = self
                 .l1_rpc
                 .get_rollup_contract_address(certificate.network_id.to_u32())
                 .await
                 .map_err(CertificationError::RollupContractAddressNotFound)?;
 
-            // Verify matching on the aggchain hash between the L1 and the agglayer
-            let l1_aggchain_hash: Digest = self
+            let l1_hash = self
```
atanmarko: Nit: you could probably leave the variable name as `l1_aggchain_hash` for more clarity.

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -353,45 +353,56 @@ where
...
         .map_err(|source| CertificationError::NativeExecutionFailed { source })?;
 
-        // Verify consistency on the aggchain proof public values if provided in the
-        // optional context
-        if let AggchainData::Generic {
-            public_values: Some(pv_from_proof),
-            aggchain_params,
-            ..
-        } = &certificate.aggchain_data
+        // Check if the certificate multisig hash matches the one stored in L1
```
atanmarko: Aggchain hash contains other things beside multisig hash (aggchain params and aggchain vkey ) so we are not checking only multisig hash.
```suggestion
        // Check if the certificate aggchain hash matches the one stored in L1
```

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
adria0pol: [5b49c1d](https://github.com/agglayer/agglayer/pull/1308/commits/5b49c1ddac9562023259cbb279cde8c05b3293ad)

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -353,45 +353,56 @@ where
...
-            ..
-        } = &certificate.aggchain_data
+        // Check if the certificate aggchain hash matches the one stored in L1
+
+        if matches!(certificate.aggchain_data, AggchainData::MultisigOnly { .. })
+            || matches!(
+                certificate.aggchain_data,
+                AggchainData::MultisigAndAggchainProof { .. }
+            )
+            || matches!(certificate.aggchain_data, AggchainData::Generic { .. })
```
hadjiszs: normally it can be also for `LegacyEcdsa` now, those networks got upgraded to use as well the `aggchain_hash` on the smart contracts

we can see there how this `aggchain_hash` is computed for the `LegacyEcdsa` : 
https://github.com/agglayer/agglayer/blob/be8b3f912981e4a8f311e1fc4625333b85eb081d/crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs#L76-L85

cc @invocamanman to confirm just in case

--

so maybe we can have a `match` here instead where all existing branches does the check with the value in the L1, so that if we have a new enum variant, we get a compiling error here to adapt the check for the newly added variant?

Freyskeyd (approved):

PR #1310, opened 2026-01-08
feat: settlement service config

Ekleog-Polygon:
Resolves https://github.com/agglayer/agglayer/issues/1278

Second attempt at #1291, but based on top of the settlement service logic in order to avoid future merge conflicts — settlement service logic will unblock more parallel follow-up tasks.

### DIFF
--- a/crates/agglayer-config/src/settlement_service.rs
+++ b/crates/agglayer-config/src/settlement_service.rs
@@ -41,17 +41,11 @@ pub enum SettlementPolicy {
     FinalizedBlock,
 }
 
-/// Transaction retry policy for failed or pending settlement transactions.
-///
-/// Defines the strategy used when retrying failed or pending transactions.
-///
-/// Future versions may include additional policies such as:
-/// - **Exponential**: Exponential backoff with increasing intervals
-/// - **Jittered**: Random jitter to avoid thundering herd issues
+/// Transaction retry policy.
 #[serde_as]
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq, Default)]
 #[serde(rename_all = "kebab-case")]
-pub struct ConfigTxRetryPolicy {
+struct ConfigTxRetryPolicy {
     /// Initial retry interval.
     #[serde(default, skip_serializing_if = "crate::is_default")]
     #[serde_as(as = "Option<HumanDuration>")]


crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,357 @@
+use std::time::Duration;
+
+use agglayer_primitives::U256;
 use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, serde_conv};
+
+use crate::{with::HumanDuration, Multiplier};
+
+/// Finality level for settlement transactions on Ethereum.
```
hadjiszs:
```suggestion
/// Policy for considering the transactions as settled on Ethereum.
```

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,357 @@
...
+
+    /// Retry policy for the transaction when it is not included on L1.
+    #[serde(default = "TxRetryPolicy::default_non_inclusion_on_l1")]
+    #[serde_as(as = "NonInclusionRetryPolicy")]
+    pub retry_on_not_included_on_l1: TxRetryPolicy,
+
+    /// Number of block confirmations required for
+    /// the transaction to resolve a receipt.
+    #[serde(default = "default_confirmations")]
+    pub confirmations: usize,
```
hadjiszs, line 187: I am bit confused around this field given the `settlement_policy` field.

Does that mean that we can have

```toml
settlement-policy = "LatestBlock"
confirmations = 32 # (default value)
```

then, `agglayer-node` would resolve the tx receipt only after 32 confirmations despite the settlement policy set to `LatestBlock`? (which I read as `confirmations = 1` normally no?)

Just wondering whether the `settlement-policy` wouldn't be sufficient to entirely define when a receipt has to be resolved (in other words, resolving the tx receipt simply when it's considered settled at the latest?)

crates/agglayer-config/tests/fixtures/settlement/finality_justified.toml
```diff
@@ -0,0 +1,3 @@
+# Configuration with justified finality (default)
+confirmations = 32
+settlement-policy = "SafeBlock"
```
hadjiszs: nit: filename uses old naming right?

crates/agglayer-config/tests/snapshots/settlement_service__deserialize_custom_certificate_config.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: crates/agglayer-config/tests/settlement_service.rs
+expression: config
+---
+max-retries = 2048
+tx-retry-policy = "linear"
+tx-retry-interval = "15s"
+confirmations = 64
+finality = "finalized"
```
hadjiszs: not reflecting the latest config definition I think, but the CI is green apparently, does it mean that some snapshots arent properly test covered?

proto/agglayer/storage/v0/settlement_service.proto
iljakuklic: Would it make sense to wrap this in a separate message too, like the other fields? I, for one, am not a big fan of excessive wrapping, effectively introducing an equivalent of `Option<Option<_>>`. However, for plain data types, the case is a bit stronger. Also to be consistent with the rest.

Edit: referring to the `metadata` field

crates/agglayer-config/tests/fixtures/settlement/custom_tx_config_1.toml
```diff
@@ -0,0 +1,11 @@
+# Custom settlement configuration with finalized finality
+max-expected-retries = 2048
+retry-on-transient-failure.initial-interval = "15s"
+confirmations = 64
+settlement-policy = "FinalizedBlock"
+gas-limit-ceiling = "0x5f5e100"
```
iljakuklic, line 6: Is it common for gas limit to be specified in hex? Should we accept a decimal number too? I guess this is just a side effect of using the `U256` type. Maybe `u128` would be sufficient in config, converting to `U256` on demand or using a utility method.

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,357 @@
...
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
 #[serde(rename_all = "kebab-case")]
-pub struct SettlementServiceConfig {}
+pub struct SettlementConfig {
+    /// Configuration for certificate settlement transactions.
+    ///
+    /// This controls how certificates (proofs of state transitions) are
+    /// submitted to the L1 settlement layer.
+    #[serde(default)]
+    pub pessimistic_proof_tx_config: SettlementTransactionConfig,
```
iljakuklic, line 318: This is a bit more consistent with the terminology in the comments:

```suggestion
    pub certificate_tx_config: SettlementTransactionConfig,
```

crates/agglayer-settlement-service/src/settlement_task.rs
iljakuklic: Is the link publicly accessible? If not, perhaps it should not be in code comments.

Edit: This refers to a TODO with an excalidraw link.

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,357 @@
...
+use agglayer_primitives::U256;
 use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, serde_conv};
+
+use crate::{with::HumanDuration, Multiplier};
+
+/// Finality level for settlement transactions on Ethereum.
+///
+/// Defines when a transaction should be considered settled on the Ethereum
+/// network, determining the security guarantees for the settlement operation.
```
adria0pol, line 12:
```suggestion
/// network, determining the security guarantees for the settlement operation.
/// This finality statuses comes from the `latest`, `safe` and `finalized` JSON-RPC block tags.
```

crates/agglayer-settlement-service/src/settlement_service.rs
iljakuklic, line 13: Just to note @Freyskeyd asked for this to be configurable on the previous PR, and this is a configuration PR. (I don't have a strong opinion about this matter.)

Edit: referring to the mpsc channel size constant.

iljakuklic, line 24: Descriptive type aliases for the value types of these hashmaps would be nice. (Likely not relevant for this PR.)

proto/agglayer/storage/v0/settlement_service.proto
iljakuklic: Hmm the code context disappeared. Hopefully it's clear enough from the file and comment :confused:

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: This code was originally from @atanmarko, but to the best of my understanding this is basically like aggkit's `LatestBlock+n` settings.

So `LatestBlock` would be equivalent to `LatestBlock+0` (ie. what we currently would count as `confirmations = 0` I think), we could set `LatestBlock+4` to get the equivalent of `confirmations = 4`, or we could use `SafeBlock` to get any completed epoch without a specific number of confirmations.

Does that make sense to you?

crates/agglayer-config/tests/snapshots/settlement_service__deserialize_custom_certificate_config.snap
Ekleog-Polygon: Good catch thanks! I've removed all the snapshots and regenerated them :+1:

crates/agglayer-config/tests/fixtures/settlement/custom_tx_config_1.toml
Ekleog-Polygon: I have no idea, it was suggested by @atanmarko who's most up-to-date there :sweat_smile: 

I'd argue for iterating on details later; as soon as we land this the structs that are being defined here can be used to parallelize all the follow-up work, like interaction with rocksdb :smile:

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: This is actually the outcome of a review comment from @Freyskeyd on https://github.com/agglayer/agglayer/pull/1291 :sweat_smile: 

I have fixed the comments!

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: It's not, but it'll be gone by the time the settlement service is done — I'm thinking that having it here will be helpful for whoever takes up the task of implementing this function

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: For more context: https://github.com/agglayer/agglayer/pull/1303#discussion_r2672403830

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,360 @@
...
+        ConfigTxRetryPolicy {
+            initial_interval: Some(value.initial_interval),
+            interval_multiplier_factor: Some(value.interval_multiplier_factor),
+            max_interval: Some(value.max_interval),
+            jitter: Some(value.jitter),
+        }
+    }
+}
+
+serde_conv!(pub TransientRetryPolicy, TxRetryPolicy, TransientRetryPolicyImpl::new, TransientRetryPolicyImpl::get);
```
atanmarko, line 80: Any particular reason to use `serde_conv` here? Seems to be less readable (and requiring developers to go and look for `serde_conv` definition) - I don't see some particular advantage of it. Why not having just `TransientRetryPolicy` with implementation instead of `TransientRetryPolicyImpl`? Would we ever convert from `TxRetryPolicy` to `TransientRetryPolicy` ?

Alternatively, we could have a `TxRetryPolicy` builder with various recipies (e.g. trait `TxRetryPolicyBuilder` implementing `TxRetryPolicy` builders)

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,360 @@
...
+                .0
+                .interval_multiplier_factor
+                .unwrap_or(Multiplier::from_u64_per_1000(1500)),
+            max_interval: self.0.max_interval.unwrap_or(Duration::from_secs(120)),
+            jitter: self.0.jitter.unwrap_or(Duration::from_secs(1)),
+        })
+    }
+}
+
+serde_conv!(pub NonInclusionRetryPolicy, TxRetryPolicy, NonInclusionRetryPolicyImpl::new, NonInclusionRetryPolicyImpl::get);
```
atanmarko, line 103: Same question as above

crates/agglayer-config/src/settlement_service.rs
```diff
@@ -1,6 +1,360 @@
...
+    }
+}
+
+/// Transaction retry policy for failed or pending settlement transactions.
+///
+/// Defines the strategy used when retrying failed or pending transactions.
+///
+/// Future versions may include additional policies such as:
+/// - **Exponential**: Exponential backoff with increasing intervals
+/// - **Jittered**: Random jitter to avoid thundering herd issues
```
atanmarko, line 132: I think this comment is deprecated. I guess we would define policy with the members that we have added to the `TxRetryPolicy`.

crates/agglayer-config/src/settlement_service.rs
atanmarko: Yes, this is `LatestBlock`+confirmations. It would make sense to move confirmations to `SettlementPolicy`, and then only using confirmations where it is needed (with `LatestBlock` not  `SafeBlock` nor `FinalizedBlock`) - we just have to make configuration specification clear about that.

crates/agglayer-config/tests/fixtures/settlement/custom_tx_config_1.toml
atanmarko: @iljakuklic  @Ekleog-Polygon Yes, hex is just consequence of using `U256`. I agree we could iterate on this later to make it more user friendly.

crates/agglayer-config/src/settlement_service.rs
hadjiszs: ok so do I understand correctly that having those two config variables as-is are redundant, or worse: let user provide non-sensical config values? (e.g., `confirmations = 2; settlement-policy = Finalized`)?

If yeah then I'd be in favor of rearranging the fields a bit as suggested by @atanmarko right above if I understood correctly, maybe with another enum variant named maybe like `Custom(usize)` so that we avoid the need for infra and external people to hack the semantic around the `LatestBlock` value?

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: The reason is @iljakuklic's comment here: https://github.com/agglayer/agglayer/pull/1291#discussion_r2664665921

I didn't find any way other than `serde_conv` to deal with this without having to duplicate the `TxRetryPolicy` or the `SettlementServiceConfig` struct. That said I'm not super happy with the result either; but I'm hoping that it's the kind of things we need to write once and then we'll be able to reuse them forever :sweat_smile:

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: :+1: For the sake of increasing parallelism, I also suggest we iterate on this later — moving this config to `SettlementPolicy` would probably require implementing a nontrivial deserializer (or at least figuring out what kind of serde (un)tagged enum attributes to use), which is probably not the most pressing thing to do

crates/agglayer-config/src/settlement_service.rs
Ekleog-Polygon: (To be precise, my other idea was making `TxRetryPolicy` generic on some type parameter that defines the actual defaults… but serde's defaulting attribute seems to not work very well with that unfortunately)

atanmarko: Ok, lets have it that way then.

crates/agglayer-config/src/settlement_service.rs
atanmarko: @Ekleog-Polygon @hadjiszs Added ticket https://github.com/agglayer/pm/issues/257

PR #1326, opened 2026-01-11
fix: unify transaction cloning in sign_transaction_typed

PivasDesant:
Local signer now clones transactions in a helper method, matching KmsSigner behavior.

### DIFF
--- a/crates/agglayer-signer/src/lib.rs
+++ b/crates/agglayer-signer/src/lib.rs
@@ -186,10 +186,6 @@ impl TxSigner<Signature> for ConfiguredSigner {
 
 impl ConfiguredSigner {
     /// Signs a transaction using a local signer.
-    ///
-    /// This helper method clones the transaction internally to match the
-    /// behavior of `KmsSigner::sign_transaction`, ensuring consistent API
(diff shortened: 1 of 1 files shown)


crates/agglayer-signer/src/lib.rs
```diff
@@ -185,20 +185,31 @@ impl TxSigner<Signature> for ConfiguredSigner {
 }
 
 impl ConfiguredSigner {
+    /// Signs a transaction using a local signer.
+    ///
+    /// This helper method clones the transaction internally to match the
+    /// behavior of `KmsSigner::sign_transaction`, ensuring consistent API
+    /// across both signer types.
```
iljakuklic: I'm not quite convinced by utility of this part of the comment. It basically describes internal implementation details, no reason for it to show up in API docs.

PivasDesant: > I'm not quite convinced by utility of this part of the comment. It basically describes internal implementation details, no reason for it to show up in API docs.

I'm convinced now :) Fixed

atanmarko (approved):

PR #1328, opened 2026-01-11
feat: use settlement job ID instead of tx hash until certificate is s…

### LINKED ISSUE #1237: Update certificate and network task logic and storage to deal with settlement job ID instead of settlement tx hash until marked as Settled

mattvaux:
Track settlement using job ID until certificate is marked as Settled, then store transaction hash.

- Add `SettlementJobId` type and field to `CertificateHeader`
- Update `SettlementClient` to return/accept job IDs
- Update certificate and network tasks to use job IDs until settled
- Clear job ID and set tx hash when status becomes Settled

Closes #1237

### DIFF
--- a/crates/agglayer-types/src/certificate/header/mod.rs
+++ b/crates/agglayer-types/src/certificate/header/mod.rs
@@ -2,9 +2,11 @@ use crate::{
     CertificateId, CertificateIndex, EpochNumber, Height, LocalExitRoot, Metadata, NetworkId,
 };
 
+mod settlement_job_id;
 mod settlement_tx_hash;
 mod status;
 
+pub use settlement_job_id::SettlementJobId;
 pub use settlement_tx_hash::SettlementTxHash;
 pub use status::CertificateStatus;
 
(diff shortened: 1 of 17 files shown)


crates/agglayer-types/src/certificate/header/mod.rs
```diff
@@ -19,5 +21,9 @@ pub struct CertificateHeader {
     pub new_local_exit_root: LocalExitRoot,
     pub metadata: Metadata,
     pub status: CertificateStatus,
+    /// Settlement job ID used to track settlement until marked as Settled.
+    /// Once settled, this is cleared and settlement_tx_hash is set.
+    pub settlement_job_id: Option<SettlementJobId>,
```
Ekleog-Polygon, line 26: Unfortunately this is not possible without significant other changes: this change would change the on-disk serialization of certificate headers, which would break backward-compatibility and make everything explode on upgrade :smiling_face_with_tear: 

We'll need to handle the storage side of changes with a bit more care in order to keep everything stable.

(Note to self. I just did a very quick skim over this and basically haven't reviewed anything significant here yet)

Ekleog-Polygon (changes requested):

atanmarko:
@mattvaux Thank you for your PR. Unfortunately, this task needs to be handled by internal team members, as it is a critical one. We need to organize better on our side - we will clearly mark tasks that are open to the public separately from critical ones that must be handled internally.

mattvaux:
NO problem, thanks for the review your honesty

PR #1336, opened 2026-01-12
fix: confirmations in settlement policy

atanmarko:
Resolves https://github.com/agglayer/agglayer/issues/1335

### DIFF
--- a/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
+++ b/crates/agglayer-aggregator-notifier/src/certifier/mod.rs
@@ -150,7 +150,6 @@ where
         info!(
             %prev_pp_root,
             %new_pp_root,
-            %certificate_id,
             "Successfully generated the witness for the PP for certificate",
         );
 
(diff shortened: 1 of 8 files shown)


Ekleog-Polygon:
The reason I was suggesting `LatestBlock+4` is I think it's what aggkit uses — might be best to check with them, but for the sake of consistency we should probably take the same naming on both sides, regardless of which one we choose to adjust 😄

Ekleog-Polygon:
I'd argue for not making it snake-case, just because this way infra can use infrastructure-as-code to automatically set the same value in both aggkit and agglayer-node configs. Not a very strong opinion that said, but it'd feel like a nice-to-have :smile:

atanmarko:
> I'd argue for not making it snake-case, just because this way infra can use infrastructure-as-code to automatically set the same value in both aggkit and agglayer-node configs. Not a very strong opinion that said, but it'd feel like a nice-to-have 😄

It would be counter intuitive to use snake-case for all other fields and not here. Same would apply to all other settings in aggkit that are CamelCase on their side and snake-case on our side.

atanmarko:
@iljakuklic Ok I'll do it, but you have to approve it immediately :D

PR #1344, opened 2026-01-14
feat: wait for tx to be mined

### LINKED ISSUE #1315: Helper function to wait for a given tx hash to be included on L1 with a timeout

The return will need to classify three possible answers:
- Transient error, that needs a retry with the same nonce (eg. RPC network error)
- Nonce already used error, that needs a retry with another nonce
- Tx included, with contract call success / revert data

atanmarko:
Resolves #1315

### DIFF
--- a/crates/agglayer-config/src/settlement_service.rs
+++ b/crates/agglayer-config/src/settlement_service.rs
@@ -91,8 +91,8 @@ impl<'de> Deserialize<'de> for SettlementPolicy {
                 E: de::Error,
             {
                 match value {
-                    "safe-block" => Ok(SettlementPolicy::SafeBlock),
-                    "finalized-block" => Ok(SettlementPolicy::FinalizedBlock),
+                    "safe-block" | "SafeBlock" => Ok(SettlementPolicy::SafeBlock),
+                    "finalized-block" | "FinalizedBlock" => Ok(SettlementPolicy::FinalizedBlock),
                     s if s.starts_with("latest-block/") => {
                         let confirmations_str = s.strip_prefix("latest-block/").unwrap();
                         let confirmations = confirmations_str.parse::<usize>().map_err(|_| {
@@ -103,15 +103,26 @@ impl<'de> Deserialize<'de> for SettlementPolicy {
                         })?;
                         Ok(SettlementPolicy::LatestBlock { confirmations })
                     }
-                    "latest-block" => {
-                        // Support "latest-block" without confirmations, use default
+                    s if s.starts_with("LatestBlock/") => {
+                        let confirmations_str = s.strip_prefix("LatestBlock/").unwrap();
+                        let confirmations = confirmations_str.parse::<usize>().map_err(|_| {
+                            E::custom(format!(
+                                "invalid confirmations value '{}' in 'LatestBlock/N'",
+                                confirmations_str
+                            ))
+                        })?;
+                        Ok(SettlementPolicy::LatestBlock { confirmations })
+                    }
+                    "latest-block" | "LatestBlock" => {
+                        // Support both cases without confirmations, use default
                         Ok(SettlementPolicy::LatestBlock {
                             confirmations: default_latest_block_confirmations(),
                         })
                     }
                     _ => Err(E::custom(format!(
-                        "unknown settlement policy '{}', expected 'safe-block', \
-                         'finalized-block', 'latest-block', or 'latest-block/N'",
+                        "unknown settlement policy '{}', expected 'safe-block', 'SafeBlock', \
+                         'finalized-block', 'FinalizedBlock', 'latest-block', 'LatestBlock', \
+                         'latest-block/N', or 'LatestBlock/N'",
                         value
                     ))),
                 }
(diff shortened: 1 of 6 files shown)


crates/agglayer-settlement-service/src/utils/transactions.rs
```diff
@@ -0,0 +1,198 @@
...
+    // Timeout for the transaction receipt.
+    timeout: Duration,
+    // Required number of blocks for confirmation
+    confirmations: usize,
+) -> Result<TransactionReceipt, TransactionReceiptError>
+where
+    P: Provider + 'static,
+{
+    let mut attempts: usize = 0;
+    let start_time = Instant::now();
```
adria0pol: Never used it, but what about using [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html)  instead, AFAIU it will be usefull for stuck RPC calls and sleeps also

crates/agglayer-settlement-service/src/utils/transactions.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let tx_receipt = wait_for_transaction_receipt(rpc_provider.clone(), tx_hash, timeout).await?;
+
+    // Wait for the required number of confirmations
+    if confirmations > 0 {
+        let receipt_block = tx_receipt.block_number.ok_or_else(|| {
+            // `wait_for_transaction_receipt` should return receipt with existing block
+            // number.
+            error!("Tx {tx_hash} receipt has no block number");
+            TransactionReceiptError::InvalidReceipt { tx_hash }
```
adria0pol, line 163: block_number.is_some() is already enforced by `wait_for_transaction_receipt`, defensive programming here?

wait_for_transaction_receipt could return the block number in addition, if we do not want to unwrap.

crates/agglayer-settlement-service/src/utils/transactions.rs
```diff
@@ -0,0 +1,198 @@
...
+                    let current_confirmations = current_block
+                        .saturating_sub(receipt_block)
+                        .saturating_add(1);
+                    if current_confirmations >= confirmations as u64 {
+                        info!(
+                            current_confirmations,
+                            confirmations,
+                            current_block,
+                            attempts,
+                            "Tx {tx_hash} confirmed with required confirmations"
```
adria0pol:
```suggestion
                            "Tx {tx_hash} confirmed with {current_confirmations} confirmations"
```

crates/agglayer-settlement-service/src/utils/transactions.rs
```diff
@@ -0,0 +1,198 @@
...
+    time::{Duration, Instant},
+};
+
+use agglayer_types::primitives::alloy_primitives::TxHash;
+use alloy::{providers::Provider, rpc::types::TransactionReceipt};
+use tracing::{debug, error, info, warn};
+
+use crate::utils::error::TransactionReceiptError;
+
+const RECEIPT_RETRY_INTERVAL: Duration = Duration::from_secs(10);
```
adria0pol:
```suggestion
const RETRY_INTERVAL: Duration = Duration::from_secs(10);
```

Since it's used for Receipts as also Blocks, maybe better just to put a generic name her

crates/agglayer-settlement-service/src/utils/transactions.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        // Wait until we have the required number of confirmations.
+        // Block where the transaction is included is the first confirmation.
+        loop {
+            attempts += 1;
+            match rpc_provider.get_block_number().await {
+                Ok(current_block) => {
+                    let current_confirmations = current_block
+                        .saturating_sub(receipt_block)
+                        .saturating_add(1);
```
adria0pol: nice detail! 🙌

crates/agglayer-settlement-service/src/utils/transactions.rs
```diff
@@ -0,0 +1,198 @@
...
+                    if current_confirmations >= confirmations as u64 {
```
adria0pol:
```suggestion
                    if current_confirmations >= confirmations as u64 {
                    let fresh_tx_receipt = fetch_transaction_receipt(tx_hash)?;
                    if tx_receipt != fresh_tx_receipt {
                        return TransactionReceiptError::SmallReorgDetected(...)
                    }
```
                
IMO it will be nice to catch 99.9999% of the reorgs here.
Large reorgs ( that are going to be black swans) eventually will need some manual intervention AFAIS.

crates/agglayer-settlement-service/src/utils/transactions.rs
atanmarko: Yeah, that may be more idiomatic and a bit cleaner. I'll switch to tokio::time::timeout.

crates/agglayer-settlement-service/src/utils/transactions.rs
atanmarko: Yeah, just defensive programming. Felt weird to return extra block number if we already have a receipt with block number.

crates/agglayer-settlement-service/src/utils/transactions.rs
atanmarko: Done

PR #1349, opened 2026-01-15
feat: add retry_callback_until_success helper to TxRetryPolicy

### LINKED ISSUE #1343: Helper function: TxRetryPolicy::retry_callback_until_success
Will wrap most of the other network-level helper functions. API looks something like
```rust
async fn retry_callback_until_success<T>(&self, impl AsyncFn() -> Result<T, TransientError>) -> T
```

l3ft0ver:
Implements `retry_callback_until_success` helper function for `TxRetryPolicy` with exponential backoff and jitter.

- Add `TransientError` trait
- Add retry helper function
- Add dependencies: `rand`, `tokio`

Closes #1343

### DIFF
--- a/crates/agglayer-config/Cargo.toml
+++ b/crates/agglayer-config/Cargo.toml
@@ -9,9 +9,11 @@ educe.workspace = true
 humantime-serde = "1.1.1"
 insta = { workspace = true, optional = true }
 jsonrpsee.workspace = true
+rand = "0.8"
 serde = { workspace = true, features = ["derive"] }
 serde_with.workspace = true
 thiserror.workspace = true
+tokio = { workspace = true, features = ["time"] }
 toml.workspace = true
 tracing.workspace = true
 tracing-appender.workspace = true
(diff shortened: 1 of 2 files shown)


Ekleog-Polygon (commented):
Thank you for the PR! From a quick glance it looks mostly fine, but I do have a few comments:
1. Can you take current_interval from final_interval, instead of capped_interval_secs, and simplify the code accordingly? This would help desynchronizing retries more and more as the multiplier goes
2. Instead of having a TransientError trait, can you introduce something that works well with `eyre::Report`? I'm thinking of most likely a struct that we can `wrap_err` and then use `agglayer_errors::match_err` to identify — see https://github.com/agglayer/interop/pull/106 / https://github.com/agglayer/agglayer/issues/838 for the context
3. Can you cut down on the comments? They look very AI-generated and not like they bring much value, for most of them 🥲

Ekleog-Polygon:
Closing as replaced by https://github.com/agglayer/agglayer/pull/1476

PR #1364, opened 2026-01-26
feat: implement basic settlement task logic

### LINKED ISSUE #1233: Implement the transaction submission logic for settlement task
See https://app.excalidraw.com/s/65UEf35l1DW/8Fy17zCAQNl

Doing this will certainly also close #1234 and #1272

### LINKED ISSUE #1234: Implement the retry logic for settlement task (including gas adjustments)
https://app.excalidraw.com/s/65UEf35l1DW/8Fy17zCAQNl

### LINKED ISSUE #1236: [Design] Determine whether external nonce use will automatically heal or needs a specific Admin RPC interface

### LINKED ISSUE #1272: Implement basic flow for the settlement task
Implement the basic state machine/flow for the settlement task, as designed in #1267

Ekleog-Polygon:
Closes #1233
Closes #1234
Closes #1272
Closes #1236 by actually implementing retry-with-new-nonce today

### DIFF
--- a/crates/agglayer-settlement-service/src/settlement_task.rs
+++ b/crates/agglayer-settlement-service/src/settlement_task.rs
@@ -176,7 +176,7 @@ impl SettlementTask {
                 .min_by_key(|(_, result)| result.block_number);
             if let Some((attempt, result)) = earliest_included_tx {
                 let included_tx_hash = self.attempt(attempt).hash;
-                if !self.wait_for_finalization(included_tx_hash).await {
+                if !self.wait_for_settlement(included_tx_hash).await {
                     // TODO: admin commands handling
                     // Finalization failed, go back to the topmost loop
                     continue;
@@ -362,9 +362,9 @@ impl SettlementTask {
         todo!()
     }
 
-    async fn wait_for_finalization(&self, _tx: SettlementTxHash) -> bool {
-        // TODO: wait for finalization of the given transaction on L1
-        // XREF: https://github.com/agglayer/agglayer/issues/1316
+    async fn wait_for_settlement(&self, _tx: SettlementTxHash) -> bool {
+        // TODO: wait for settlement of the given transaction on L1 according to the
+        // settlement policy XREF: https://github.com/agglayer/agglayer/issues/1316
         todo!()
     }
 


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -100,21 +142,238 @@ impl SettlementTask {
...
+            self.write_attempt_to_db(&next_attempt).await;
+            self.submit_attempt_to_l1(&next_attempt).await;
+            // Loop back to the topmost loop, checking for inclusion on L1
+        }
+    }
+
+    fn attempt(&self, attempt_number: SettlementAttemptNumber) -> &SettlementAttempt {
+        self.attempts
+            .get(&attempt_number)
+            .expect("Attempt number not found internally to the SettlementTask itself")
```
hadjiszs, line 264: do we want to panic here or bubble up as error?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -72,10 +104,20 @@ pub enum TaskAdminCommand {
     ReloadAndRestart,
 }
 
+pub struct SettlementAttempt {
+    pub sender_wallet: Address,
+    pub nonce: Nonce,
+    pub gas_limit: u128,
```
hadjiszs, line 110: isn't it always supposed to be the `gas_limit` value defined in the upper level `SettlementJob`?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -100,21 +142,239 @@ impl SettlementTask {
...
+    pub async fn run(&mut self) -> SettlementJobResult {
+        loop {
+            // Is any of the previous settlement attempts included on L1?
+            let included_attempts = self.attempts_included_on_l1().await;
+            let earliest_included_tx = included_attempts
+                .into_iter()
+                .min_by_key(|(_, result)| result.block_number);
+            if let Some((attempt, result)) = earliest_included_tx {
+                let included_tx_hash = self.attempt(attempt).hash;
+                if !self.wait_for_finalization(included_tx_hash).await {
```
hadjiszs: nit: I'd rename to `wait_for_confirmations` or `wait_for_settlement`, for consistency given that `SettlementPolicy` can be set to `Safe` which doesnt correspond to `Finalized`, in which case `wait_for_finalization` would be misleading for someone lacking history/context

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: I think panicking here is the better choice! This would be a developer mistake and not an external condition; plus if any settlement task were unable to settle it'd result in absolutely no network being able to settle ever again.

So all things considered, I do think that panicking is the better option, here and probably in most of the settlement tasks for anything that can't be retried.

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: I'm hesitating a bit here. Overall, currently we have three different places where all these structs are being defined: here, in the protobuf schema, and in the config structs. In particular, the config structs currently do expose a `gas_limit_ceiling` and multiplier factor, which I'm not totally sure what it's about.

I plan on going through all of them between the work on settlement task and storage, to make the whole system more consistent. So hopefully we can iterate on this in a follow-up. WDYT? :smile:

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: Makes total sense to me! I've updated the PR

PR #1366, opened 2026-01-28
fix: make sure that RPC being unaware of tx does not block settlement task

### LINKED ISSUE #1358: Adjust the settlement task control flow
Adjust the control flow for the design changes from https://github.com/agglayer/agglayer/issues/1357

Ekleog-Polygon:
Closes #1358 

The core of the idea here is, rather than checking whether RPC thinks our txes landed, instead we ask RPC which exact tx landed, and we identify ourselves whether it's one of ours.

### DIFF
--- a/crates/agglayer-settlement-service/src/settlement_task.rs
+++ b/crates/agglayer-settlement-service/src/settlement_task.rs
@@ -181,7 +181,7 @@ impl SettlementTask {
                 // We know that only the first tx has a chance to settle, as any other tx
                 // included later would necessarily conflict with the first included tx.
                 let included_tx_hash = self.attempt(attempt).hash;
-                if !self.wait_for_finalization(included_tx_hash).await {
+                if !self.wait_for_settlement(included_tx_hash).await {
                     // TODO: admin commands handling
                     // Finalization failed, go back to the topmost loop
                     continue;
@@ -396,7 +396,7 @@ impl SettlementTask {
         todo!()
     }
 
-    async fn wait_for_finalization(&self, _tx: SettlementTxHash) -> bool {
+    async fn wait_for_settlement(&self, _tx: SettlementTxHash) -> bool {
         // TODO: wait for finalization of the given transaction on L1
         // XREF: https://github.com/agglayer/agglayer/issues/1316
         todo!()


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -174,9 +176,12 @@ impl SettlementTask {
             let earliest_included_tx = included_attempts
                 .into_iter()
                 .min_by_key(|(_, result)| result.block_number);
+            // TODO: Also filter by index-in-block
             if let Some((attempt, result)) = earliest_included_tx {
+                // We know that only the first tx has a chance to settle, as any other tx
+                // included later would necessarily conflict with the first included tx.
                 let included_tx_hash = self.attempt(attempt).hash;
-                if !self.wait_for_settlement(included_tx_hash).await {
+                if !self.wait_for_finalization(included_tx_hash).await {
```
hadjiszs: rebase issue no? it feels it reverted the review changes of the previous PR which renamed this `wait_for_finalization` into `wait_for_settlement`

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -174,9 +176,12 @@ impl SettlementTask {
             let earliest_included_tx = included_attempts
                 .into_iter()
                 .min_by_key(|(_, result)| result.block_number);
+            // TODO: Also filter by index-in-block
             if let Some((attempt, result)) = earliest_included_tx {
+                // We know that only the first tx has a chance to settle, as any other tx
```
hadjiszs, line 181: I see in the description of `attempts_included_on_l1()` that we expect to return also the attempts which ended up as "in error"

Then not sure to understand why we wait for full settlement of this `earliest_included_tx` if it is a failed attempt? Don't we want to only take the earliest included tx such that `result.outcome == ContractCallOutcome::Success`?

Another related question regarding `// TODO: Also filter by index-in-block`, considering the above, we won't need to consider the index-in-block given that we cannot have 2 successful attempts included in the chain (regardless of being in the same block or not) right?

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: Nice catch, thanks! Fixed :smile:

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: This is very related to the recent thinking on contract call reverts I've been speaking about on slack. I've also written some thoughts in https://github.com/agglayer/agglayer/pull/1373, which are not complete either — please bear with me until I can finish the exact design and make the (hopefully last) PR to adjust the settlement task logic :smile: 

TL;DR: this was originally to handle contract call reverts properly. However, in order to handle them more properly (especially in case of a reorg), I'll iterate again on the task logic, and update this again.

At that time, the lines you're pointing to will have been rewritten, and I'll hopefully be able to be more clear on the answer!

hadjiszs (approved):

PR #1374, opened 2026-02-04
fix: make log_assert_eq panic only in debug

MozirDmitriy:
Reason for changes: `log_assert_eq!` used `assert_eq!`, which caused a panic in both debug and release when the invariant was violated, and the block with error! was never actually executed. This was inconsistent with the doc comment and tests, which expected a panic only when debug assertions were enabled and logging in release.
What was done: replaced `assert_eq!` with `debug_assert_eq!` in both branches of `log_assert_eq!` so that the macro only panics in debug builds and logs via error! when the invariant is violated in release, bringing its behavior in line with log_assert! and the documentation.

### DIFF
--- a/crates/agglayer-utils/src/assertions.rs
+++ b/crates/agglayer-utils/src/assertions.rs
@@ -37,7 +37,7 @@ macro_rules! log_assert_eq {
     ($lhs:expr, $rhs:expr $(,)?) => {
         match (&$lhs, &$rhs) {
             (lhs, rhs) => {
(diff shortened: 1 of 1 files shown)


Ekleog-Polygon (approved):
This definitely feels like a typo. I looked through the git history, and it was this way since the initial commit introducing the macros.

That said, I'll wait for a second opinion to be sure I'm not missing anything!

cc @Freyskeyd @hadjiszs

PR #1386, opened 2026-02-11
feat: recover properly from a reorg transforming a revert into a success

### LINKED ISSUE #1367: Settlement Task - Control flow & main logic

### LINKED ISSUE #1377: Update the settlement task to reflect the new logic
See the logic defined in https://github.com/agglayer/agglayer/issues/1376

Ekleog-Polygon:
Fixes #1377
Fixes #1367

This is basically a complete rewrite of the logic, following the new diagram https://app.excalidraw.com/s/65UEf35l1DW/8oLi71w4c9f

### DIFF
--- a/crates/agglayer-settlement-service/src/settlement_task.rs
+++ b/crates/agglayer-settlement-service/src/settlement_task.rs
@@ -255,6 +255,12 @@ impl SettlementTask {
                 // All nonces were seen on L1, but we didn't get a successful settlement result
                 // for any of them. Also, there was at least one revert.
                 // We can wait for finalization without submiting a new attempt.
+                let earliest_revert_result = reverts
+                    .values()
+                    .map(|(_, _, result)| result)
+                    .min_by_key(|result| result.block_number)
+                    .unwrap() // No panic: we checked `!reverts.is_empty()` just before.
+                    .clone();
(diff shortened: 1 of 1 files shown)


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -159,265 +165,382 @@ impl SettlementTask {
...
+                        );
                     }
                 }
-            };
+                let earliest_revert_result = reverts
+                    .values()
+                    .map(|(_, _, result)| result)
+                    .min_by_key(|result| result.block_number)
+                    .unwrap(); // No panic: we checked `!reverts.is_empty()` above.
+                self.write_job_revert_to_db(earliest_revert_result).await;
```
hadjiszs: regarding the `No panic:`, in the for loop above, there is one `else if` clause where we mutate `reverts` through `remove()` so maybe we want to compute this `earliest_revert_result` before the for loop?

Ekleog-Polygon: You're entirely right, thank you! :pray: 
I guess at some point in writing this PR, a refactor changed the order of code blocks, causing this issue :/

PR #1388, opened 2026-02-13
feat: add test for distinct PP and TX settlement keys

Freyskeyd:
Add a regression test to ensure GCP KMS auth parsing preserves distinct PP and TX settlement key fields when both are explicitly provided.

## Changes
- Added `/Users/spaitrault/work/polygon/agglayer/crates/agglayer-config/tests/fixtures/auth/distinct_settlement_keys.toml`
- Added `/Users/spaitrault/work/polygon/agglayer/crates/agglayer-config/tests/auth_config.rs:88` with `auth_distinct_pp_and_tx_settlement_keys_are_preserved`
- Asserted:
  - `pp_settlement_key_name = "pp-distinct-key-name"`
  - `pp_settlement_key_version = 11`
  - `tx_settlement_key_name = "tx-distinct-key-name"`
  - `tx_settlement_key_version = 22`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6071,9 +6071,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Ekleog-Polygon (approved):
Oh so we had actually already implemented distinct settlement keys? That's awesome 🎉

Issue #1312, opened 2026-01-09
Storage function to list recorded settlement attempts for a given settlement job


Ekleog-Polygon:
If there is already a function, list it here.
If not, write one and list it here

PR #1391, opened 2026-02-18
feat: add settlement storage schema with column families and interfaces

Freyskeyd:
Introduces database schema for settlement operations with four column
families: jobs, attempts, attempt results, and attempts per wallet.

Includes read/write traits for settlement operations with insert-only
semantics and comprehensive test coverage for data persistence and
retrieval patterns.

### DIFF
--- a/crates/agglayer-storage/src/stores/state/settlement/mod.rs
+++ b/crates/agglayer-storage/src/stores/state/settlement/mod.rs
@@ -21,7 +21,7 @@ impl StateStore {
     fn with_settlement_write_lock<T>(
         &self,
         settlement_job_id: &Ulid,
-        f: impl FnOnce() -> Result<T, Error>,
+        callback: impl FnOnce() -> Result<T, Error>,
     ) -> Result<T, Error> {
         let key_lock = {
             let mut settlement_write_locks = self.settlement_write_locks.lock().map_err(|_| {
@@ -40,7 +40,7 @@ impl StateStore {
             ))
         })?;
 
-        f()
+        callback()
     }
 }
 
@@ -120,11 +120,12 @@ impl SettlementWriter for StateStore {
         attempt_sequence_number: u64,
         settlement_attempt: &SettlementAttempt,
     ) -> Result<(), Error> {
-        self.with_settlement_write_lock(settlement_job_id, || {
-            let job_exists = match self.db.get::<SettlementJobsColumn>(settlement_job_id) {
-                Ok(value) => value.is_some(),
-                Err(err) => return Err(err.into()),
-            };
+        self.with_settlement_write_lock(settlement_job_id, || -> Result<(), Error> {
+            let job_exists = self
+                .db
+                .get::<SettlementJobsColumn>(settlement_job_id)?
+                .is_some();
+
             if !job_exists {
                 return Err(Error::UnprocessedAction(format!(
                     "Settlement job does not exist for id {settlement_job_id}"
@@ -136,10 +137,7 @@ impl SettlementWriter for StateStore {
                 attempt_sequence_number,
             };
 
-            let attempt_exists = match self.db.get::<SettlementAttemptsColumn>(&key) {
-                Ok(value) => value.is_some(),
-                Err(err) => return Err(err.into()),
-            };
+            let attempt_exists = self.db.get::<SettlementAttemptsColumn>(&key)?.is_some();
             if attempt_exists {
                 return Err(Error::UnprocessedAction(format!(
                     "Settlement attempt already exists for job {settlement_job_id} and attempt \
@@ -181,24 +179,17 @@ impl SettlementWriter for StateStore {
             let attempt_per_wallet_value = attempt_per_wallet::Value;
 
             let mut batch = WriteBatch::default();
-            if let Err(err) = self.db.multi_insert_batch::<SettlementAttemptsColumn>(
+            self.db.multi_insert_batch::<SettlementAttemptsColumn>(
                 [(&key, settlement_attempt)],
                 &mut batch,
-            ) {
-                return Err(err.into());
-            }
-            if let Err(err) = self
-                .db
+            )?;
+            self.db
                 .multi_insert_batch::<SettlementAttemptPerWalletColumn>(
                     [(&attempt_per_wallet_key, &attempt_per_wallet_value)],
                     &mut batch,
-                )
-            {
-                return Err(err.into());
-            }
-            if let Err(err) = self.db.write_batch(batch) {
-                return Err(err.into());
-            }
+                )?;
+
+            self.db.write_batch(batch)?;
 
             Ok(())
         })


Ekleog-Polygon (commented):
Removing my request for review, please re-request when ready

crates/agglayer-storage/benches/extract_settlement_latest_attempt_results.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python3
```
Ekleog-Polygon: TBH I'm a bit surprised that there's a need of so much custom python code to extract and plot criterion benchmarks together, when criterion already includes HTML report generation. Was what you wanted to do really impossible with upstream tools?

Regardless, this is not a blocker, benches are internal-only tools anyway. For that same reason, I reviewed them very quickly and may have missed important issues, but 🤷 

If you want a proper review there too, it might be best to split it into another PR to avoid blocking the main changes!

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -27,6 +27,26 @@ pub const PER_EPOCH_PROOFS_CF: &str = "per_epoch_proofs_cf";
 pub const PER_EPOCH_END_CHECKPOINT_CF: &str = "per_epoch_end_checkpoint_cf";
 pub const PER_EPOCH_START_CHECKPOINT_CF: &str = "per_epoch_start_checkpoint_cf";
 
+// Settlement related CFs
+pub const SETTLEMENT_ATTEMPTS_CF: &str = "settlement_attempts_cf";
+pub const SETTLEMENT_ATTEMPT_PER_WALLET_CF: &str = "settlement_attempt_per_wallet_cf";
+pub const SETTLEMENT_ATTEMPT_RESULTS_CF: &str = "settlement_attempt_results_cf";
+pub const SETTLEMENT_JOBS_CF: &str = "settlement_jobs_cf";
```
Ekleog-Polygon, line 34: I think this is missing `settlement_job_results_cf`. I've thought a bit more about that question, and I think there's one reason why we definitively want it: future-compatibility. If we ever change the logic to how we consider a settlement job's result to be, we don't want it to implicitly change the considered results of previous jobs.

For that reason (plus the fact that the logic is nontrivial and already implemented in #1386), I'm pretty sure that we'll want to store the settlement job result separately from the attempt results. Or at least, that we'll want to store for each job, which attempt does have the definitive result — but the added code complexity for that indirection might not be worth it, one result should be less than 1KiB, we shouldn't be seeing so many settlements that it'd ever reach a non-negligible size.

TL;DR: I'm pretty sure we should add either `settlement_job_results_cf`, or at least `settlement_job_conclusive_attempt_cf` — but probably the former for the first step, for code simplicity.

crates/agglayer-storage/src/columns/mod.rs
```diff
@@ -27,6 +27,26 @@ pub const PER_EPOCH_PROOFS_CF: &str = "per_epoch_proofs_cf";
...
+    compression: crate::schema::options::ColumnCompressionType::Lz4,
+    prefix_extractor: crate::schema::options::PrefixExtractor::Fixed {
+        size: 16, // settlement_job_id (Ulid)
+    },
+};
+
+pub const SETTLEMENT_ATTEMPT_PER_WALLET_COLUMN_OPTIONS: ColumnOptions = ColumnOptions {
+    compression: crate::schema::options::ColumnCompressionType::Lz4,
+    prefix_extractor: crate::schema::options::PrefixExtractor::Fixed {
+        size: 28, // address (20 bytes) + nonce (u64)
```
Ekleog-Polygon, line 47: I think this should be 20 for just the address? Otherwise you're considering the whole key as a prefix, which would sound wrong to me.

Edit: actually the design changed from the original design; with the new design this prefix makes sense to me.

crates/agglayer-storage/src/stores/interfaces/reader/settlement_reader.rs
```diff
@@ -0,0 +1,36 @@
...
+///
+/// This trait is intentionally limited to point lookups and metadata-style
+/// reads. Missing records are returned as `Ok(None)`.
+pub trait SettlementReader: Send + Sync {
+    /// Returns the settlement job for `settlement_job_id`, if present.
+    fn get_settlement_job(&self, settlement_job_id: &Ulid) -> Result<Option<SettlementJob>, Error>;
+
+    /// Returns the settlement attempt identified by
+    /// `(settlement_job_id, attempt_sequence_number)`, if present.
+    fn get_settlement_attempt(
```
Ekleog-Polygon: FWIW I think this function wouldn't actually be useful and could be deleted — #1312 should be the only way we actually load attempts in practice, I think. I thus did not review implementation.

crates/agglayer-storage/src/stores/interfaces/reader/settlement_reader.rs
```diff
@@ -0,0 +1,36 @@
...
+    ) -> Result<Option<SettlementAttempt>, Error>;
+
+    /// Returns the stored result for an attempt identified by
+    /// `(settlement_job_id, attempt_sequence_number)`, if present.
+    ///
+    /// Note: If a result doesn't exist for the attempt, it can mean two things:
+    /// 1. The attempt doesn't exist
+    /// 2. The attempt exists but hasn't been finalized yet (i.e., its result
+    ///    hasn't been stored yet).
+    fn get_settlement_attempt_result(
```
Ekleog-Polygon: Same here (#1312 also includes loading the attempt results)

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        generated::agglayer::storage::v0::{SettlementAttempt, SettlementJob, TxResult},
+        settlement::{attempt::Key as SettlementAttemptKey, attempt_per_wallet},
+    },
+};
+
+mod telemetry;
+
+impl StateStore {
+    fn lock_settlement_writes(&self) -> Result<std::sync::MutexGuard<'_, ()>, Error> {
+        self.settlement_write_lock
```
Ekleog-Polygon: Hmm I'm not sure I'm convinced by this lock? AFAICT we don't have any for other stores, and this one doesn't seem more dangerous than the others. Maybe just having the `settlement_job_id` / `attempt_number` generation be performed directly inside the storage layer would handle the potential concerns?

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    fn insert_settlement_attempt(
+        &self,
+        settlement_job_id: &Ulid,
+        attempt_sequence_number: u64,
+        settlement_attempt: &SettlementAttempt,
+    ) -> Result<(), Error> {
+        let _settlement_write_lock = self.lock_settlement_writes()?;
+        let mut metrics = AttemptWriteMetrics::start();
```
Ekleog-Polygon: Trying to have metrics already seems a bit eager to me; I'll probably skim over the metrics part of this PR for now, for the sake of review efficiency but TBH it'd have been great if it had been split into a separate PR 😅 (FWIW if I don't lose it in the refactorings, I had actually written an AI agent rule to ensure that PRs were of review-able size and having it reduce the scope as needed to reach that goal)

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        let job_exists = match self.db.get::<SettlementJobsColumn>(settlement_job_id) {
```
Ekleog-Polygon: I'm not against this sanity-check, but TBH I'm not sure it brings much value — `SettlementJobId` type creation should basically always coincide with database read or writes, so I'd think it's not required. Though I'm not sure we have `SettlementJobId` yet? That might be the best

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+            Err(err) => return Err(err.into()),
+        };
+        if attempt_exists {
+            return Err(Error::UnprocessedAction(format!(
+                "Settlement attempt already exists for job {settlement_job_id} and attempt \
+                 sequence number {attempt_sequence_number}"
+            )));
+        }
+
+        let sender_wallet = match settlement_attempt.sender_wallet.as_ref() {
```
Ekleog-Polygon: I'd think this should take the settlement attempt as a proper Rust type (where it's not yet `Option`), this would make this whole code much simpler? And we can convert from the Rust type to a Protobuf type inside of this function, should be a one-liner, and we would be sure that there always are all the required parameters, instead of checking them for 25 lines.

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+                    "Settlement attempt sender_wallet must be 20 bytes".to_string(),
+                ));
+            }
+        };
+
+        let attempt_per_wallet_key = attempt_per_wallet::Key {
+            address,
+            nonce: nonce.nonce,
+            settlement_job_id: *settlement_job_id,
+            attempt_sequence_number,
```
Ekleog-Polygon: Hmm I'm hesitating a bit here: this new schema makes it possible to have multiple settlement jobs for a single nonce, while [the initial design](https://github.com/agglayer/agglayer/issues/1227#issuecomment-3655369905) did not. No strong opinions that said, because in exchange the CF becomes write-only. Was this a conscious choice?

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+            .get::<SettlementAttemptResultsColumn>(&key)?
+            .is_some()
+        {
+            return Err(Error::UnprocessedAction(format!(
+                "Settlement attempt result already exists for job {settlement_job_id} and attempt \
+                 sequence number {attempt_sequence_number}"
+            )));
+        }
+
+        if self.db.get::<SettlementAttemptsColumn>(&key)?.is_none() {
```
Ekleog-Polygon: Same here, this sanity check probably makes sense but doesn't seem strictly speaking required to me

crates/agglayer-storage/src/stores/state/tests/settlement.rs
```diff
@@ -0,0 +1,441 @@
+use std::sync::Arc;
```
Ekleog-Polygon, line 1: I skimmed over tests for the sake of review efficiency, nothing looked obviously wrong.

crates/agglayer-telemetry/src/settlement_storage.rs
```diff
@@ -0,0 +1,38 @@
...
+
+lazy_static! {
+    pub static ref ATTEMPT_WRITES_TOTAL: Counter<u64> =
+        global::meter(AGGLAYER_STORAGE_OTEL_SCOPE_NAME)
+            .u64_counter("settlement_attempt_writes_total")
+            .with_description("Total number of settlement-attempt write operations")
+            .build();
+    pub static ref ATTEMPT_WRITE_DURATION_MS: Histogram<f64> =
+        global::meter(AGGLAYER_STORAGE_OTEL_SCOPE_NAME)
+            .f64_histogram("settlement_attempt_write_duration_ms")
```
Ekleog-Polygon: I'm not sure the [default histogram boundaries](https://docs.rs/opentelemetry/latest/opentelemetry/metrics/struct.HistogramBuilder.html#method.with_boundaries) make much sense for this, we'd almost certainly always be between 0 and 10ms so everything would fall in only two buckets. Probably deserves configuration!

crates/agglayer-storage/benches/extract_settlement_latest_attempt_results.py
Freyskeyd: The automatic export was not readable at all.
I can remove this if needed

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Freyskeyd: @Ekleog-Polygon it depends where we want the ownership of creating those values.

If it's outside the store: we need a lock}
if it's in the store: we don't

edit: And in fact, we need concurrency safety (be sure that there we're not overwritting a pre-existing value), so for me, we need this logical lock as rocksdb doesn't provide any way to do so

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Freyskeyd: Going to try to extract that in another PR

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+    ) -> Result<T, Error> {
+        let key_lock = {
+            let mut settlement_write_locks = self.settlement_write_locks.lock().map_err(|_| {
+                Error::Unexpected("Settlement write lock map is poisoned".to_string())
+            })?;
+
+            settlement_write_locks
+                .entry(*settlement_job_id)
+                .or_insert_with(|| std::sync::Arc::new(std::sync::Mutex::new(())))
+                .clone()
```
hadjiszs, line 34: nit: it seems that this hashmap is ever growing, is it ok or should we have entries dropped somewhere?

Freyskeyd: This will be refactor later to a sharding lock hash that prevent that

crates/agglayer-storage/src/columns/settlement_job_results/tests.rs
```diff
@@ -0,0 +1,46 @@
...
+    let encoded_key = key.encode().expect("Unable to encode key");
+    let decoded_key = Key::decode(&encoded_key).expect("Unable to decode key");
+    assert_eq!(decoded_key, key);
+
+    let encoded_value = value.encode().expect("Unable to encode value");
+    let decoded_value = Value::decode(&encoded_value).expect("Unable to decode value");
+    assert_eq!(decoded_value, value);
+}
+
+fn mk_tx_result_success() -> TxResult {
```
Ekleog-Polygon, line 28: Nit: if this were a method of `TxResult`, it'd avoid duplication with the same in `settlement_attempt_results/tests.rs`

crates/agglayer-storage/src/columns/settlement_jobs/tests.rs
```diff
@@ -0,0 +1,58 @@
...
+fn settlement_job_roundtrip_codec() {
+    let key = Ulid::from(42u128);
+    let value = mk_settlement_job();
+
+    let encoded_key = key.encode().expect("Unable to encode key");
+    let decoded_key = Key::decode(&encoded_key).expect("Unable to decode key");
+    assert_eq!(decoded_key, key);
+
+    let encoded_value = value.encode().expect("Unable to encode value");
+    let decoded_value = Value::decode(&encoded_value).expect("Unable to decode value");
```
Ekleog-Polygon, line 23: Nit: This should be fuzzers (and same for all others): we could have the fuzzers auto-generate random structs, and validate that encode/decode is a no-op.

Level 2 would be generating random bytes, and if they decode properly, then decode/encode/decode results in the same result as decode.

Can this be done in a follow-up maybe?

You can find examples of exactly such fuzzers in `crates/agglayer-grpc-types/src/compat/v1/tests.rs`. Maybe it'd make sense to add "try to write bolero fuzzers before you define hardcoded data" in some "how to write tests" AI skill 😄

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        };
+
+        let _settlement_write_lock = key_lock.lock().map_err(|_| {
+            Error::Unexpected(format!(
+                "Settlement write lock is poisoned for job {settlement_job_id}",
+            ))
+        })?;
+
+        f()
```
Ekleog-Polygon: Nit: I'd rename that `callback` or similar, on first quick read I had missed this three-letter call and was wondering why you took a lock to release it immediately 😅 a longer name would make the call more obvious!

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+}
+
+impl SettlementReader for StateStore {
+    fn get_settlement_job(&self, settlement_job_id: &Ulid) -> Result<Option<SettlementJob>, Error> {
+        Ok(self.db.get::<SettlementJobsColumn>(settlement_job_id)?)
```
Ekleog-Polygon, line 49:
```suggestion
        self.db.get::<SettlementJobsColumn>(settlement_job_id)
```

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    fn get_settlement_job_result(
+        &self,
+        settlement_job_id: &Ulid,
+    ) -> Result<Option<TxResult>, Error> {
+        Ok(self
+            .db
+            .get::<SettlementJobResultsColumn>(settlement_job_id)?)
```
Ekleog-Polygon, line 58:
```suggestion
        self
            .db
            .get::<SettlementJobResultsColumn>(settlement_job_id)
```

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn insert_settlement_attempt(
+        &self,
+        settlement_job_id: &Ulid,
+        attempt_sequence_number: u64,
+        settlement_attempt: &SettlementAttempt,
+    ) -> Result<(), Error> {
+        self.with_settlement_write_lock(settlement_job_id, || {
+            let job_exists = match self.db.get::<SettlementJobsColumn>(settlement_job_id) {
+                Ok(value) => value.is_some(),
+                Err(err) => return Err(err.into()),
```
Ekleog-Polygon: Nit: this is exactly `?`. You may just need to replace `|| {` with `|| -> Result<(), Error> {` if type inference is not doing its job properly

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+
+            let key = SettlementAttemptKey {
+                settlement_job_id: *settlement_job_id,
+                attempt_sequence_number,
+            };
+
+            let attempt_exists = match self.db.get::<SettlementAttemptsColumn>(&key) {
+                Ok(value) => value.is_some(),
+                Err(err) => return Err(err.into()),
```
Ekleog-Polygon: Nit: Same here (except I'm surprised type inference wouldn't work here, there's a plain `return Err` above so I'd understand the first instance to fail but the second seems surprising to me)

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            };
+            if attempt_exists {
+                return Err(Error::UnprocessedAction(format!(
+                    "Settlement attempt already exists for job {settlement_job_id} and attempt \
+                     sequence number {attempt_sequence_number}"
+                )));
+            }
+
+            let sender_wallet = match settlement_attempt.sender_wallet.as_ref() {
```
Ekleog-Polygon, line 148: Was not handling https://github.com/agglayer/agglayer/pull/1391#discussion_r2891187683 on purpose? It can come as a follow-up, but I think it'd both make the code simpler and smaller to have it, so it'd probably be best to do it sooner than later 😄

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+                .multi_insert_batch::<SettlementAttemptPerWalletColumn>(
+                    [(&attempt_per_wallet_key, &attempt_per_wallet_value)],
+                    &mut batch,
+                )
+            {
+                return Err(err.into());
+            }
+            if let Err(err) = self.db.write_batch(batch) {
+                return Err(err.into());
+            }
```
Ekleog-Polygon: Nit: These two `if let` could also be `?`

crates/agglayer-storage/src/types/settlement/attempt.rs
```diff
@@ -0,0 +1,33 @@
...
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Key {
+    pub(crate) settlement_job_id: ulid::Ulid,
+    pub(crate) attempt_sequence_number: u64,
+}
+
+pub type Value = crate::types::generated::agglayer::storage::v0::SettlementAttempt;
+
+impl Codec for Value {
+    fn encode_into<W: io::Write>(&self, mut writer: W) -> Result<(), crate::schema::CodecError> {
```
Ekleog-Polygon, line 17: Nit: we're starting to have quite a few CFs that encode protobuf, it'd be really nice to have this be replaced by eg. `define_protobuf_codec!(Value);`, that'd be defined in `agglayer-storage/src/types/mod.rs` probably. It'd avoid all that code duplication and copy-paste risks

crates/agglayer-storage/src/columns/settlement_jobs/tests.rs
Freyskeyd: Created a ticket for that

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Freyskeyd: will address that in the conversion layer PR

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Freyskeyd: mmmh `RocksDBError` are not `agglayer-storage::Error`

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Freyskeyd: same here, we need to convert DBError to Error

crates/agglayer-storage/src/columns/settlement_job_results/tests.rs
Freyskeyd: created a followup ticket for that

Issue #1254, opened 2025-12-17
[Design] Design an admin api for the settlement service


Ekleog-Polygon:
- Allow pausing settlement attempts, reporting when all ongoing settlement tasks have paused
- Allow adding settlement attempts / manually editing settlement attempts
- Allow manually marking a settlement attempt as "definitely failed / retry with new nonce" — required if we rotate the agglayer wallet and we don't have access to the private key any longer to fill pending nonces that will never be retried again. Also required if we don't implement "wait for finalization of a given nonce" for lack of available RPC.

Ekleog-Polygon:
Closing in favor of #1675, that lists the design made as part of #1663

PR #1393, opened 2026-02-19
feat: bind settlement service with certificate/network tasks

### LINKED ISSUE #1237: Update certificate and network task logic and storage to deal with settlement job ID instead of settlement tx hash until marked as Settled

### LINKED ISSUE #1595: Helper function to build the actual calldata for the settlement job
## Goal

Add a helper that builds the real `verifyPessimisticTrustedAggregator` calldata and populates a `SettlementJob` from a certificate's proof outputs, so a settlement job can actually be submitted to L1 instead of carrying empty/zero call data.

## Target type

The job the helper must populate — today `calldata` is just a `Bytes` field with no producer wiring it from a proof:

https://github.com/agglayer/agglayer/blob/8c5103839d1e1f92d4b5e820060d4277c0d40d23/crates/agglayer-types/src/settlement.rs#L82-L88

## Work

- [ ] Populate `SettlementJob`: set `calldata` to the encoded call and `contract_address` to the RollupManager (instead of empty/zero values)
- [ ] Check how to test

hadjiszs:
Fixes https://github.com/agglayer/agglayer/issues/1237


crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -167,7 +175,7 @@ where
             }
             CertificateStatus::Candidate => {
                 self.recompute_state().await?;
-                self.process_from_candidate().await
+                self.process_from_candidate(None).await
```
Ekleog-Polygon: Following up from our chat, this would be problematic due to losing the link between settlement job and certificate in case of reboot.

Probably the best would be to save the settlement job id at the end of `process_from_proven` and load it again at the beginning of `process_from_candidate`, the write is required anyway and the read will certainly not have any meaningful performance impact.

I'll review the rest of the PR later, thank you for the PR!

Ekleog-Polygon (commented):
We discussed this and improvements on it in the last meeting, please re-request my review when ready

crates/agglayer-certificate-orchestrator/src/network_task/tests/status.rs
```diff
@@ -1,5 +1,9 @@
 use std::time::Duration;
```
Ekleog-Polygon, line 1: P1 nit: The refactoring of the tests looks great 🎉 
That said, it makes it pretty hard to see which changes were actually required by the updates to the certificate / network task. Do you think it'd be possible to split the tests refactoring into another PR (that we could land immediately), so that we could see more easily which changes are actually required for this PR and confirm they're correct?

crates/agglayer-certificate-orchestrator/src/network_task/tests.rs
```diff
@@ -1,28 +1,47 @@
 use std::{collections::VecDeque, sync::Mutex, time::Duration};
```
Ekleog-Polygon, line 1: (Exactly like this file, where it's obvious that the only changes are replacing the included settlement logic with the new settlement service 😄)

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -29,7 +26,10 @@ const MAX_TX_RETRY: usize = 5;
 /// related to the certificate until it gets finalized, including exchanging the
 /// required messages with the network task to both get required information
 /// from it and notify it of certificate progress.
-pub struct CertificateTask<StateStore, PendingStore, CertifierClient> {
+pub struct CertificateTask<StateStore, PendingStore, CertifierClient, SettlementSvc>
+where
+    SettlementSvc: SettlementServiceTrait,
```
Ekleog-Polygon: P2 nit: usually we don't have `where` clauses in struct definitions, unless strictly required.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -38,26 +38,29 @@ pub struct CertificateTask<StateStore, PendingStore, CertifierClient> {
...
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         certificate: Certificate,
         network_task: mpsc::Sender<NetworkTaskMessage>,
         state_store: Arc<StateStore>,
         pending_store: Arc<PendingStore>,
         certifier_client: Arc<CertifierClient>,
         cancellation_token: CancellationToken,
+        settlement_service: Arc<SettlementSvc>,
+        settlement_config: Arc<SettlementTransactionConfig>,
```
Ekleog-Polygon: P2 nit: keep cancellation token as the last argument, grouping all the trait things together.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -424,84 +290,81 @@ where
...
-        self.process_from_candidate().await
+        // Wait for result — the settlement service owns its own timeouts
+        let result = watcher
+            .wait_for_result()
+            .await
+            .map_err(|e| CertificateStatusError::SettlementError(e.to_string()))?;
+
+        match result {
+            SettlementJobResult::ContractCall(contract_result)
+                if contract_result.outcome == ContractCallOutcome::Revert =>
```
Ekleog-Polygon: P1 nit: I'd rather have the "default" case be Revert, and have the Success case be explicitly checked for success, to avoid any issues should we ever add an outcome kind. WDYT?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -424,84 +290,81 @@ where
...
+                let tx_hash = contract_result.tx_hash;
+                info!("Settlement successful: {}", tx_hash);
+
+                self.header.settlement_tx_hash = Some(tx_hash);
+
+                // The settlement service may have already marked the
+                // certificate as Settled (e.g. RpcSettlementClient's
+                // wait_for_settlement assigns it to the epoch). Tolerate the
+                // "already settled" error and skip straight to finalization in
+                // that case.
```
Ekleog-Polygon: P1: The settlement service will not actually mark the certificate as Settled; it won't know anything about the link between the settlement tx data and the certificate id — settlement service is agnostic from what kind of settlement tx data it handles, and doesn't know anything about certificates.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -424,84 +290,81 @@ where
...
+                match self.state_store.update_settlement_tx_hash(
+                    &certificate_id,
+                    tx_hash,
+                    UpdateEvenIfAlreadyPresent::Yes,
+                    UpdateStatusToCandidate::Yes,
+                ) {
+                    Ok(()) => {
+                        self.header.status = CertificateStatus::Candidate;
```
Ekleog-Polygon: P0: Uhh no the settlement service watcher will complete only after settlement has completed.
Instead, we should:

Set status to Candidate immediately after having requested the settlement
At that time, we should also record the settlement job id to database, AFAICT currently we only have a println
Here we should set to Settled, and not UpdateStatusToCandidate

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -512,100 +375,7 @@ where
...
-        .await?;
-
-        if self.header.status != CertificateStatus::Settled {
-            return Err(CertificateStatusError::InternalError(
-                "CertificateTask completed with a non-settled certificate".into(),
-            ));
-        }
-
-        Ok(())
+        self.finalize_settlement().await
```
Ekleog-Polygon, line 470: P0: This will be called after a reboot if there's any certificate that has a settlement job but hasn't been settled yet. We should request the result with the saved job id here, and wait for the watcher if the result isn't ready yet

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -72,17 +73,17 @@ pub type SettlementTasks = FuturesUnordered<
...
-    Sc,
     CertifierClient,
     PendingStore,
     EpochsStore,
     PerEpochStore,
     StateStore,
-> {
+    SettlementSvc,
+> where
+    SettlementSvc: SettlementServiceTrait,
```
Ekleog-Polygon: P2 nit: we don't usually have where clauses on structs, unless required

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -98,33 +72,19 @@ pub enum NetworkTaskMessage {
...
-pub(crate) struct NetworkTask<CertifierClient, SettlementClient, PendingStore, StateStore> {
+pub(crate) struct NetworkTask<
+    CertifierClient,
+    PendingStore,
+    StateStore,
+    SettlementSvc,
+    PerEpochStoreT,
+> where
+    SettlementSvc: SettlementServiceTrait,
+    PerEpochStoreT: PerEpochWriter + PerEpochReader + 'static,
```
Ekleog-Polygon: P2 nit: same here, no `where` clauses on structs unless required to build

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -319,13 +288,15 @@ where
             .map(|exit| exit.hash())
             .collect::<Vec<Digest>>();
         let task = tokio::spawn(
-            CertificateTask::new(
+            CertificateTask::<_, _, _, SettlementSvc>::new(
```
Ekleog-Polygon: P3 nit: this should be auto-inferred, I'm surprised that it's being required?

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -363,166 +334,18 @@ where
...
+                                self.local_state.get_roots().display_to_hex()
+                            )));
                         };
-                        debug!(
-                            old_state = self.local_state.get_roots().display_to_hex(),
-                            new_state = new.get_roots().display_to_hex(),
-                            old_pp_root_v3 = self.pending_pessimistic_root(Height::new(height.as_u64().saturating_sub(1)), PessimisticRootCommitmentVersion::V3, &self.local_state.get_roots()).to_string(),
-                            new_pp_root_v3 = self.pending_pessimistic_root(height, PessimisticRootCommitmentVersion::V3, &new.get_roots()).to_string(),
-                            "Updated the state following certificate settlement",
-                        );
```
Ekleog-Polygon, line 563: P0 nit: Can we keep at least old/new state here, even if we no longer have the pp roots? It'd probably still be useful for debugging

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -533,6 +356,36 @@ where
                                 certificate_id,
                                 error: e.to_string(),
                             })?;
+
+                        const MAX_EPOCH_ASSIGNMENT_RETRIES: usize = 5;
+                        let mut max_retries = MAX_EPOCH_ASSIGNMENT_RETRIES;
+                        let (epoch_number, certificate_index) = loop {
+                            max_retries -= 1;
```
Ekleog-Polygon: P2 nit: Uhh I guess this is copy-pasted from the previous code, but this is fragile: it's subject to either panic or overflow depending on the configuration, if the bug `match` below were ever to change — that's a far-away interaction that's likely to cause bugs in the future.

Can we, instead of having the `max_retries == 0` test in the match block below, instead just have it above the subtraction here, to make the code invariants more local?

crates/agglayer-config/tests/snapshots/auth_config__auth_distinct_pp_and_tx_settlement_keys_are_preserved.snap
```diff
@@ -86,3 +86,43 @@ proving-timeout = "5m"
...
+max-interval = "2m"
+jitter = "1s"
+
+[settlement.pessimistic-proof-tx-config.retry-on-not-included-on-l1]
+initial-interval = "1m"
+interval-multiplier-factor = 2.0
+max-interval = "10m"
+jitter = "10s"
+
+[settlement.validium-tx-config]
```
Ekleog-Polygon: P3 followup: considering the latest exchanges we'll actually not need to have a validium tx config I think

crates/agglayer-node/src/node.rs
```diff
@@ -259,32 +260,32 @@ impl Node {
         let core = Kernel::new(rpc_tx_settlement.clone(), config.clone()).unwrap();
 
         let current_epoch_store = Arc::new(arc_swap::ArcSwap::new(Arc::new(current_epoch_store)));
-        let epoch_packing_aggregator_task = RpcSettlementClient::new(
-            Arc::new(config.outbound.rpc.settle_cert.clone()),
-            state_store.clone(),
-            pending_store.clone(),
-            Arc::clone(&rollup_manager),
-            current_epoch_store.clone(),
-        );
```
Ekleog-Polygon, line 268: P1 followup: AFAICT we'll no longer need the `RpcSettlementClient` after this PR. Probably we can remove it altogether?

crates/agglayer-rate-limiting/src/local/limiter/slot_tracker.rs
```diff
@@ -36,7 +36,7 @@ mod tests {
     use super::SlotTracker;
 
     #[test]
-    #[should_panic = "slots not released"]
+    #[cfg_attr(debug_assertions, should_panic = "slots not released")]
```
Ekleog-Polygon: P1: Why was this change required? This disables a test that AFAICT should have nothing to do with this PR

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -34,21 +35,90 @@ impl SettlementJobWatcher {
...
+    /// Wait for the settlement job to complete and return the result.
+    pub async fn wait_for_result(&mut self) -> eyre::Result<SettlementJobResult> {
+        self.watcher
+            .changed()
+            .await
+            .map_err(|_| eyre::eyre!("Settlement job watcher closed"))?;
+        self.watcher
+            .borrow()
+            .clone()
+            .ok_or_else(|| eyre::eyre!("Settlement job completed with no result"))
```
Ekleog-Polygon: P0 issue: I think this is wrong: there could be a result before `.changed()` is called and it'd be ignored. Instead, use `wait_for(Option::is_some)` or something like that, I think that'd be the proper way to handle it!

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -34,21 +35,90 @@ impl SettlementJobWatcher {
...
+        Self { watcher, job_id }
+    }
+
+    #[cfg(any(test, feature = "testutils"))]
+    pub fn new_for_test(
+        watcher: watch::Receiver<Option<SettlementJobResult>>,
+        job_id: Ulid,
+    ) -> Self {
+        Self { watcher, job_id }
+    }
```
Ekleog-Polygon: P3 nit: Uhh I think this is pointless considering there's already `from_channel` above? Though `from_channel` could maybe be `pub(crate)` instead of `pub`

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -34,21 +35,90 @@ impl SettlementJobWatcher {
...
 }
 
 pub enum RetrievedSettlementResult {
     Pending(SettlementJobWatcher),
     Completed(SettlementJobResult),
 }
 
+#[cfg_attr(feature = "testutils", mockall::automock)]
+#[async_trait::async_trait]
+pub trait SettlementServiceTrait: Send + Sync {
```
Ekleog-Polygon: P2 nit: can we have a dedicated file for the trait, considering it's the most important exposed API? It'd make changes to the API more obvious

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -224,3 +294,47 @@ impl tower::Service<AdminCommand> for SettlementService {
         })
     }
 }
+
+#[cfg(any(test, feature = "testutils"))]
+pub mod testutils {
```
Ekleog-Polygon: P2 nit: this whole module feels like it could be static methods on `SettlementJobWatcher` directly

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -30,20 +30,45 @@ impl Nonce {
...
+    pub(crate) max_fee_per_gas_multiplier: Multiplier,
+    pub(crate) max_priority_fee_per_gas_ceiling: U128,
+    pub(crate) max_priority_fee_per_gas_floor: U128,
+    pub(crate) max_priority_fee_per_gas_multiplier: Multiplier,
+
+    pub(crate) settlement_config: Arc<SettlementTransactionConfig>,
+}
+
+impl SettlementJob {
+    pub fn new(settlement_config: Arc<SettlementTransactionConfig>) -> Self {
```
Ekleog-Polygon: P1 question: Can we have `todo!()` rather than a placeholder here? It'd make it more obvious that this is known wrong and just used as a placeholder.

crates/agglayer-utils/src/assertions.rs
```diff
@@ -37,7 +37,7 @@ macro_rules! log_assert_eq {
     ($lhs:expr, $rhs:expr $(,)?) => {
         match (&$lhs, &$rhs) {
             (lhs, rhs) => {
-                ::std::assert_eq!(lhs, rhs);
+                ::std::debug_assert_eq!(lhs, rhs);
```
Ekleog-Polygon, line 40: P2 nit: Can you cancel these changes? They have nothing to do with this PR and are already handled by https://github.com/agglayer/agglayer/pull/1374, which is just waiting for a second pair of eyes to confirm 😝

tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs
```diff
@@ -162,16 +162,13 @@ async fn transaction_with_receipt_timeout_many_times(#[case] state: Forest) {
...
     match result.status {
         CertificateStatus::InError { error } => {
             let error_message = error.to_string();
             assert!(
-                error_message.contains(
-                    "Too many different settlement transactions submitted for the same certificate"
-                ),
-                "Expected error message about too many settlement transactions, but got: \
+                error_message.contains("Timeout waiting for"),
+                "Expected error message about timeout waiting for settlement, but got: \
```
Ekleog-Polygon: P1 question: Hmm I'm not sure I understand what the changes to this file are about exactly? There should no longer be any timeout after the switch to the settlement service, but OTOH I also don't expect the test to make sense at all any longer anyway 🤔

Ekleog-Polygon (commented):
Thank you for the updates to the PR! I tried to classify priority for each comment below 😄

crates/agglayer-certificate-orchestrator/src/network_task/tests/status.rs
hadjiszs: reverted the refactoring, I'll propagate those in a dedicated PR later on

tests/integrations/tests/certificate_settlement/retries.rs
```diff
@@ -20,6 +20,7 @@ const SEND_CERTIFICATE_ERROR: i32 = -10006;
 #[tokio::test]
 #[timeout(Duration::from_secs(180))]
 #[case::type_0_ecdsa(crate::common::type_0_ecdsa_forest())]
+#[ignore = "failpoints live only in the bypassed RpcSettlementClient; pending migration"]
```
Freyskeyd: why this is ignored?

crates/agglayer-certificate-orchestrator/src/lib.rs
```diff
@@ -72,17 +73,15 @@ pub type SettlementTasks = FuturesUnordered<
 /// The Certificate Orchestrator collects the generated proofs and settles
 /// them on the L1 on the go.
 pub struct CertificateOrchestrator<
-    Sc,
     CertifierClient,
     PendingStore,
     EpochsStore,
     PerEpochStore,
     StateStore,
+    SettlementSvc,
```
Freyskeyd:
```suggestion
    SettlementService,
```

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -98,33 +73,16 @@ pub enum NetworkTaskMessage {
...
 }
 
 /// Network task that is responsible to certify the certificates for a network.
-pub(crate) struct NetworkTask<CertifierClient, SettlementClient, PendingStore, StateStore> {
+pub(crate) struct NetworkTask<
+    CertifierClient,
+    PendingStore,
+    StateStore,
+    SettlementSvc,
+    PerEpochStoreT,
```
Freyskeyd:
```suggestion
    SettlementService,
    PerEpochStore,
```

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -519,134 +336,144 @@ where
...
+            *output.new_pessimistic_root,
+            proof_with_selector.into(),
+            self.certificate.custom_chain_data.clone().into(),
+        );
+
+        Ok(SettlementJob {
+            contract_address: self.certifier_client.rollup_manager_address(),
+            calldata,
+            eth_value: U256::ZERO,
+            gas_limit: self
```
Ekleog-Polygon: P0 issue: resolve the settlement gas limit before persisting the job

This stores only the configured gas-limit ceiling in `SettlementJob.gas_limit`, and the actual `eth_estimateGas * multiplier` resolution is deferred until `SettlementTask::build_next_attempt_*`, after the settlement job and certificate -> job-id link have already been persisted. That makes pre-submit simulation part of pending-job recovery: if `eth_estimateGas` deterministically fails for this calldata, for example because the call would revert against the current L1 state, the build path returns a non-transient error, the settlement task panics, and the persisted job can be reloaded into the same failure loop without ever recording a terminal result.

Can we resolve the effective gas limit inside the settlement-service job-creation path before the durable job/link is created? `SettlementTask` already owns the L1 provider and settlement wallet needed for `eth_estimateGas`; for example, `SettlementTask::create` or a writer helper could build the settlement call with the service wallet, run `estimate_gas`, apply `gas_limit_multiplier_factor` and `gas_limit_ceiling`, and store that resolved value into `SettlementJob.gas_limit` before `save_settlement_job_to_db` persists the job. Recovered jobs would then build attempts from the persisted gas limit without re-running `eth_estimateGas`, and deterministic simulation failures would fail before the durable certificate/job link is created.

Origin: introduced by this PR. The PR records the new settlement job from the certificate path and defers gas estimation to the persisted settlement-task attempt builder.

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -574,6 +402,52 @@ where
                                 certificate_id,
                                 error: e.to_string(),
                             })?;
+
+                        // Assign the settled certificate to the current epoch,
+                        // retrying a bounded number of times to ride out a
+                        // transient epoch rollover (packed epoch or add failure).
+                        const MAX_EPOCH_ASSIGNMENT_RETRIES: usize = 5;
+                        let (epoch_number, certificate_index) = 'assign: {
```
Ekleog-Polygon, line 392: P1 issue: assign the certificate to the epoch before persisting the advanced state

This writes the local network state before `current_epoch.add_certificate(...)`. If epoch assignment exhausts its retries after this write, storage has already applied the certificate while the epoch/header/latest-settled metadata still describes the previous certificate. On restart, the network task reads that advanced state but re-enters at the same height, so Candidate recovery is broken because it no longer has the state before this certificate.

Can we restore the previous ordering and call `add_certificate` before `write_local_network_state`? That keeps this failure on the older, known recovery side instead of checkpointing state ahead of settlement metadata. Also, once `add_certificate` succeeds, the explicit `update_certificate_header_status(&certificate_id, &CertificateStatus::Settled)` below should be removed or the comment fixed: `add_certificate` already calls `assign_certificate_to_epoch`, which persists `Settled` with the epoch/index.

Origin: introduced by this PR.

crates/agglayer-node/src/node.rs
```diff
@@ -259,15 +259,18 @@ impl Node {
...
 
-        info!("Epoch packing aggregator task created.");
+        let settlement_config = Arc::new(config.settlement.pessimistic_proof_tx_config.clone());
```
Ekleog-Polygon, line 265: P1 follow-up: error when legacy settle-cert tuning would be ignored

This switches certificate settlement to `config.settlement.pessimistic_proof_tx_config`, while the old `[outbound.rpc.settle-cert]` block is still accepted and serialized. Existing operators who tuned certificate confirmations, retry intervals, or gas/fee settings under the old block will now run the new settlement service with defaults unless they know to add the new block. That changes L1 finality and transaction behavior silently.

Can we follow up with a logged error, or configuration-validation failure, when the legacy `outbound.rpc.settle-cert` block is set but would be ignored by the new settlement service?

Origin: introduced by this PR.

crates/agglayer-settlement-service/src/settlement_service_trait.rs
```diff
@@ -0,0 +1,71 @@
...
+
+use crate::settlement_service::{RetrievedSettlementResult, SettlementService};
+
+#[cfg_attr(feature = "testutils", mockall::automock)]
+#[async_trait::async_trait]
+pub trait SettlementServiceTrait: Send + Sync {
+    /// Submit a settlement job for `certificate_id` and return its id once the
+    /// service has accepted and persisted it.
+    ///
+    /// The service records the certificate <-> job-id link atomically when it
```
Ekleog-Polygon, line 22: P1 follow-up: recover dangling certificate settlement job links

This contract is what the certificate task now relies on for crash recovery, but the implementation still writes the certificate -> job id mapping in `SettlementTask::reserve_settlement_job_id()` before `save_settlement_job_to_db()` writes the job. A crash or write failure between those two operations leaves a certificate mapped to a non-existent settlement job. On restart, settlement-service recovery scans no job to resume, while the certificate task sees the mapping, moves to `Candidate`, and `wait_for_settlement()` fails with “No settlement job found”.

Can we follow up by teaching certificate recovery to handle this dangling-link case by recreating the missing settlement job in storage and then waiting on it, instead of failing the certificate? That would keep this specific recovery window aligned with the broader RocksDB non-atomicity model we already have elsewhere.

Origin: pre-existing broader storage/settlement-service atomicity gap, but this PR newly documents and depends on this invariant from the certificate recovery path.

tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs
```diff
@@ -13,6 +13,7 @@ use tokio_util::sync::CancellationToken;
 #[tokio::test]
 #[timeout(Duration::from_secs(180))]
 #[case::type_0_ecdsa(crate::common::type_0_ecdsa_forest())]
+#[ignore = "failpoints live only in the bypassed RpcSettlementClient; pending migration"]
```
Ekleog-Polygon: P1 test follow-up: restore certificate-settlement failure coverage on the new service path

This PR switches certificate settlement from `RpcSettlementClient` to `SettlementService`, but it disables the integration tests that covered reverted receipts, missing receipts, low-gas retry, timeout/retry, and retry-on-error because their failpoints are still on the old client. I don't see equivalent integration coverage added for the new service/orchestrator/storage boundary; the settlement-service unit tests don't exercise the certificate status transitions, job-id recovery, or local-state/epoch updates.

Can we port these failpoints to `SettlementTask`/the new service, or add equivalent certificate-settlement integration tests, as a follow-up? Otherwise the bridge-safety cases most likely to regress are no longer covered end to end.

Origin: introduced by this PR.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -519,134 +336,144 @@ where
...
-                ))?,
-            settlement_submitted_notifier,
-        })
-        .await?;
+    /// Build the real `verifyPessimisticTrustedAggregator` settlement calldata
+    /// from the certificate's proof, to hand to the settlement service.
+    async fn build_settlement_job(&self) -> Result<SettlementJob, CertificateStatusError> {
+        let certificate_id = self.header.certificate_id;
+
+        let proof = match self.pending_store.get_proof(certificate_id)? {
```
Ekleog-Polygon, line 343: P2 issue: remove the old settlement client path

`node.rs` no longer constructs `RpcSettlementClient`, so the old `SettlementClient` trait, `RpcSettlementClient`, `NonceInfo`, `TxReceiptStatus`, and the two `settlement_client` modules are now production-dead aside from their own exports/tests. Keeping them around also leaves a second implementation of the settlement input pipeline in `RpcSettlementClient` — proof loading, `PessimisticProofOutput` decoding, verifier-type handling, `PESSIMISTIC_PROOF_PROGRAM_SELECTOR` prefixing, and `verifyPessimisticTrustedAggregator` encoding — even though the live path now does that in `certificate_task::build_settlement_job`.

Can we remove the old client path and its tests in this PR? That avoids maintaining duplicate L1 settlement ABI logic for code that is no longer reachable in production.

Origin: introduced by this PR. The old path existed before, but this PR makes it production-dead and adds the live replacement.

tests/integrations/src/l1_setup.rs
```diff
@@ -12,7 +12,7 @@ const L1_FIXTURE_ENV: &str = "AGGLAYER_INTEGRATION_L1_FIXTURE";
 const ANVIL_BACKEND: &str = "anvil";
 const DOCKER_BACKEND: &str = "docker";
 const DEFAULT_ANVIL_FIXTURE: &str =
-    concat!(env!("CARGO_MANIFEST_DIR"), "/fixtures/anvil-l1/state.hex");
+    concat!(env!("CARGO_MANIFEST_DIR"), "/fixtures/anvil-l1/state.json");
```
Ekleog-Polygon, line 15: P2 doc: update Anvil fixture docs and remove the orphaned state.hex

This changes the default Anvil fixture from `state.hex` loaded through `anvil_loadState` to `state.json` loaded through the CLI `--load-state` path, which the comment below says is required to preserve historical states for `eth_getTransactionBySenderAndNonce`. However, `tests/integrations/fixtures/anvil-l1/state.hex` is still committed, and the surrounding docs still point contributors at the old artifact: `README.md` still says the regeneration script rebuilds/commits `state.hex`, and `.github/workflows/test.yml` still says CI pins Anvil so it can decode the committed `state.hex` fixture.

Can we update the README/CI comments to refer to `state.json`, and drop the orphaned `state.hex` fixture so contributors don't regenerate or review the wrong artifact?

Origin: introduced by this PR.

tests/integrations/tests/certificate_settlement/retries/recovery.rs
```diff
@@ -29,13 +29,23 @@ async fn sent_transaction_recover(#[case] failpoints: &[&str], #[case] state: Fo
...
     let scenario = FailScenario::setup();
+    let cancellation_token = CancellationToken::new();
 
+    // Cancel (graceful shutdown) then panic when the failpoint fires, so the node
+    // actually stops and `agglayer_shutdowned` resolves -- a plain panic only kills
+    // the runtime thread without shutting the node down.
     for f in failpoints {
-        fail::cfg(*f, "panic(killing node)").expect("Failed to configure failpoint");
+        let token = cancellation_token.clone();
+        fail::cfg_callback(*f, move || {
```
Ekleog-Polygon, line 39: P2 follow-up: remove the deleted settlement_submitted failpoint from this case

This case still configures `network_task::make_progress::settlement_submitted`, but this PR deleted that failpoint along with the old network-task settlement-submission path. The callback never fires now, so this scenario only exercises `certificate_task::process_impl::about_to_record_candidate` while the test name/comment still imply it kills the node after sending the settlement transaction.

Can we remove the dead failpoint from the case, or replace it with a failpoint in the new settlement-service path that actually fires after the settlement transaction is submitted?

Origin: introduced by this PR.

crates/agglayer-settlement-service/src/lib.rs
```diff
@@ -2,14 +2,21 @@
 //! AggLayer.
 //!
 //! This service consists of:
+//! - [`SettlementServiceTrait`]: the public API the certificate orchestrator
```
Ekleog-Polygon, line 5: P2 follow-up: remove the stale crate-wide dead_code allow

This PR integrates `agglayer-settlement-service` into the node and certificate orchestrator, which is exactly the TODO condition for removing this crate-wide `dead_code` allow. Keeping it now can hide genuinely dead settlement-service code while this PR is wiring the live certificate settlement path.

Can we remove the crate-level allow and add targeted allows only for any remaining intentionally unused admin/test helpers?

Origin: pre-existing allow, but this PR satisfies the TODO condition by integrating the settlement service.

crates/agglayer-certificate-orchestrator/src/network_task/tests/status.rs
```diff
@@ -291,8 +408,8 @@ async fn from_proven_to_settled() {
 
 #[rstest]
 #[test_log::test(tokio::test)]
-#[timeout(Duration::from_secs(2))]
-async fn from_candidate_to_settled() {
+#[timeout(Duration::from_secs(30))]
+async fn from_candidate_to_settle_via_pending() {
```
Ekleog-Polygon, line 409: P3 follow-up: remove the duplicate Candidate recovery test

This test is effectively the same as `from_candidate_to_settled` above: both start from a `Candidate` certificate, insert the same persisted settlement job and certificate -> job-id link, expect one `witness_generation`, return the same successful settlement result, and assert the header becomes `Settled`. The name suggests a different “via pending” path, but the setup does not drive a pending/proven transition.

Can we either collapse the duplicate test or change this setup so it exercises the distinct path the name describes?

Origin: introduced by this PR.

crates/agglayer-certificate-orchestrator/src/network_task/tests.rs
```diff
@@ -220,15 +329,22 @@ async fn repeated_unreadable_proof_errors_certificate() {
 
     pending
         .expect_set_latest_proven_certificate_per_network()
-        .times(2)
+        .once()
```
Ekleog-Polygon, line 326: P3 follow-up: rename or restore the repeated unreadable-proof case

This test name still says `repeated_unreadable_proof_errors_certificate`, but the updated expectations make it a single-shot failure: `certify()` runs once, `get_proof()` returns `UnreadableProof` once, and the certificate goes `Proven -> InError`. The previous retry/reprove behavior is no longer covered by this test, so the name now overstates what it exercises.

Can we either rename the test to match the new single-failure behavior, or restore a true repeated/retry scenario if that coverage is still intended?

Origin: introduced by this PR.

Ekleog-Polygon (changes requested):
Thanks for the update. I reviewed the PR again from scratch at the current head.

The main blocker I found is that the effective settlement gas limit is still resolved during settlement-task attempt building, after the settlement job and certificate -> job-id link have already been persisted. If gas simulation deterministically fails, the job can be left as a persisted pending job that fails again on recovery instead of failing before the durable job/link is created.

I also left a P1 comment around settlement finalization ordering, where persisting the advanced local state before epoch/header/latest-settled metadata would break Candidate recovery.

There are several follow-up comments for migration debt, cleanup, and test coverage. If you agree with those, it would be great to open tracking issues for the ones that should not block this PR.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -519,134 +336,144 @@ where
...
+        );
+
+        Ok(SettlementJob {
+            contract_address: self.certifier_client.rollup_manager_address(),
+            calldata,
+            eth_value: U256::ZERO,
+            gas_limit: self
+                .settlement_config
+                .gas_limit_ceiling
+                .saturating_to::<u128>(),
```
Freyskeyd: Could we replace that by a `self.settlement_service.create_job` that already hold the config instead of spreading it accross the whole dependency tree?

tests/integrations/tests/certificate_settlement/retries.rs
```diff
@@ -16,54 +16,6 @@ mod recovery;
 
 const SEND_CERTIFICATE_ERROR: i32 = -10006;
 
-#[rstest]
-#[tokio::test]
-#[timeout(Duration::from_secs(180))]
-#[case::type_0_ecdsa(crate::common::type_0_ecdsa_forest())]
-async fn retry_on_error(#[case] state: Forest) {
```
Freyskeyd, line 23: this test seems still valuable to me

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -574,6 +402,52 @@ where
...
+                            for attempt in 1..=MAX_EPOCH_ASSIGNMENT_RETRIES {
+                                let related_epoch = self.current_epoch.load_full();
+                                if related_epoch.is_epoch_packed() {
```
Freyskeyd: This is not needed, you can try adding the certificate and then if error is `Error::AlreadyPacked(epoch_number)` you can retry with delay

crates/agglayer-certificate-orchestrator/src/network_task.rs
```diff
@@ -574,6 +402,52 @@ where
...
+                        // the next run, never durably settled with no epoch.
+                        self.state_store
+                            .update_certificate_header_status(
+                                &certificate_id,
+                                &CertificateStatus::Settled,
+                            )
+                            .map_err(|e| Error::PersistenceError {
+                                certificate_id,
+                                error: e.to_string(),
+                            })?;
```
Freyskeyd: This is not needed, it's already done by the `add_certificate` method

Freyskeyd (changes requested):

tests/integrations/tests/certificate_settlement/retries.rs
hadjiszs: No longer ignored. The test was adapted and re-enabled (commit a95158d0); retries.rs has no #[ignore] now and both regression tests run.

crates/agglayer-certificate-orchestrator/src/network_task.rs
hadjiszs: Done

tests/integrations/tests/certificate_settlement/retries.rs
hadjiszs: taken over by `l1_settlement/
failure.rs::transaction_with_receipt_status_0_retry`

crates/agglayer-certificate-orchestrator/src/network_task.rs
hadjiszs: `add_certificate` sets the header status to `Settled` (via `assign_certificate_to_epoch`), so the status part is indeed redundant, but it does not write `CertificatePerNetworkColumn`, the cursor index the RPC reads via `get_certificate_header_by_cursor` (`agglayer-rpc/src/lib.rs:360,692`).

probable follow-up : move the cursor write into `assign_certificate_to_epoch` so this call can be dropped

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: Done

crates/agglayer-certificate-orchestrator/src/network_task.rs
hadjiszs: >Can we restore the previous ordering and call add_certificate before write_local_network_state?

Done 👍 

>update_certificate_header_status(&certificate_id, &CertificateStatus::Settled) below should be removed or the comment fixed: add_certificate already calls assign_certificate_to_epoch, which persists Settled with the epoch/index.

Not done https://github.com/agglayer/agglayer/pull/1393#discussion_r3512645397

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -148,6 +146,19 @@ where
...
+            // Resume that job rather than re-proving and re-submitting (which the
+            // at-most-once guard rejects), so it recovers instead of erroring.
+            if let Some(job_id) = self
+                .state_store
+                .get_certificate_settlement_job_id(&certificate_id)?
+            {
+                info!(%job_id, "Proven certificate already has a settlement job; resuming");
+                self.set_status(CertificateStatus::Candidate)?;
+                return self.process_from_candidate(true).await;
+            }
```
Freyskeyd, line 160: Would it makes sense to have an optional field `job_id` ?

Freyskeyd (dismissed):
just a nit

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: included as part of this follow-up: https://github.com/agglayer/agglayer/issues/1603

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: tackled in this commit: https://github.com/agglayer/agglayer/commit/3226cfbc

tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs
hadjiszs: tackled in this commit: https://github.com/agglayer/agglayer/commit/a95158d0

tests/integrations/src/l1_setup.rs
hadjiszs: tackled in this commit: https://github.com/agglayer/agglayer/commit/6760bb28

crates/agglayer-node/src/node.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1602

crates/agglayer-settlement-service/src/settlement_service_trait.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1603

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1604

tests/integrations/tests/certificate_settlement/retries/recovery.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1604

crates/agglayer-settlement-service/src/lib.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1604

crates/agglayer-certificate-orchestrator/src/network_task/tests/status.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1605

crates/agglayer-certificate-orchestrator/src/network_task/tests.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1605

crates/agglayer-node/src/node.rs
```diff
@@ -259,15 +259,18 @@ impl Node {
...
+        let settlement_service = Arc::new(
+            agglayer_settlement_service::SettlementService::start(
+                config.settlement.settlement_service_config.clone(),
+                settlement_config.clone(),
+                rpc_tx_settlement.clone(),
```
Ekleog-Polygon: P0 issue: certificate settlement now signs with the tx-settlement key

The legacy path settled certificates through `rollup_manager` (`L1RpcClient`), which is built on `rpc_pp_settlement` — the provider whose signer we log as "Cert signer address" and whose config docs call "the signer for PP settlement". This line hands the settlement service `rpc_tx_settlement` instead, so every `verifyPessimisticTrustedAggregator` tx is signed and funded by the tx-settlement wallet.

With distinct pp/tx keys configured (supported — see the `auth_config__auth_distinct_pp_and_tx_settlement_keys_are_preserved` snapshot), gas estimation runs with `from` = the tx-settlement address — which, on standard deployments where only the pp key holds the rollup manager's trusted-aggregator role, means a deterministic revert → `submit_settlement_job` fails → every certificate goes InError. Even with a single key, this leaves certificate settlement sharing the Kernel's `interop_sendTx` wallet with a second, uncoordinated nonce allocator (adjacent races are tracked in #1575/#1597; passing `rpc_pp_settlement` also re-separates the wallets). CI can't catch this: the integration setup configures one key, so both providers are the same `Arc`.

```suggestion
                rpc_pp_settlement.clone(),
```

Ekleog-Polygon: P1 followup: warn at startup when the pp- and tx-settlement wallets are the same

The provider-construction fallback already warns when no `tx_settlement` signer is configured ("Using the same provider for certificate and tx settlement"), but when both signers are configured and resolve to the same address, startup is silent — and the settlement service and the Kernel then share one wallet with two uncoordinated nonce allocators (#1575/#1597 territory) without any operator-visible signal.

Can we add a `warn!` at startup (next to the existing "Cert signer address" / "Tx signer address" logs) when `rpc_pp_settlement` and `rpc_tx_settlement` report the same signer address, so shared-wallet deployments are explicit?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -975,11 +1033,19 @@ impl<
...
             .get_transaction_receipt(tx_hash.into())
-            .await?
-        {
+            .await?;
+
+        // Test-only failpoint: hide an otherwise-available receipt so the run loop
+        // treats it as indexing lag and keeps polling. Compiled out of production.
+        #[cfg(feature = "testutils")]
+        let receipt = receipt.filter(|_| {
+            !fail::eval("settlement::receipt_transiently_unavailable", |_| true).unwrap_or(false)
```
Ekleog-Polygon, line 1045: P0 follow-up: investigate reported panic-loop on conflicting settlement-attempt results (AI-reported, unverified)

Flagged by AI-assisted review — I don't fully understand it yet, and the mechanism pre-dates this PR (it becomes production-reachable now that the settlement service is the live path), so filing as a follow-up to investigate rather than something to resolve in this PR.

The reported failure mode: attempt results are persisted before the terminal job result (the TODO at :1633 acknowledges the window), and `SettlementAttemptResult::can_be_replaced_by` never allows replacing a `ContractCall` result. A job with ≥2 nonces that durably records `ContractCall(Revert)`/`NonceAlreadyUsed` for one nonce, then escapes the all-nonces-seen branch via a shallow-reorg `continue 'start` (:637/:649/:652), can later have a successful attempt whose `write_job_result_to_db` rewrites all other attempts with `ClientError(settlement_succeeded_elsewhere)` (:1650-1661, no terminal-result filter) — hitting the stored unreplaceable result and panicking at :1707 *before* `insert_settlement_job_result`. The conflict is durable and rehydrated on load, so startup recovery reportedly respawns into the same panic on every restart: settlement succeeded on L1, job never completes, certificate InError, no repair API.

Needs its own issue: confirm the trigger walk (multi-nonce job + pre-safe reorg during the finalization wait), attempt a repro, and if it holds, treat unreplaceable-result conflicts as data (finalize the job with the on-L1 outcome or a terminal error) instead of panicking.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -519,134 +332,142 @@ where
...
+        // re-derive it from the hash the service actually settled -- not the header's
+        // recorded hash, which can be stale after a reboot. Skipped on the live path.
+        if recompute_local_state {
+            self.recompute_state(Some(Digest::from(tx_hash))).await?;
+        }
+
+        // Record the settlement tx hash before marking Settled: the storage
+        // layer rejects tx-hash updates on already-settled certificates.
+        self.header.settlement_tx_hash = Some(tx_hash);
+        self.state_store.update_settlement_tx_hash(
```
Ekleog-Polygon, line 463: P1 issue: refuse certificate replacement while a pending settlement job exists

The settlement tx hash is now written to the header only after a terminal Success, so any pre-Success failure leaves `InError` + `settlement_tx_hash: None` — and `validate_pre_existing_certificate`'s `None` branch then accepts a replacement certificate while the original broadcast tx is still live in the mempool (or gets re-broadcast by startup recovery). The replacement's job takes a higher nonce on the same wallet, so L1 must mine the orphaned tx first: L1 settles the *old* certificate's state transition while our height records the replacement, whose own tx then reverts against the advanced root — both certs end InError and latest-settled freezes. (Legacy wrote the hash at submission, so the gate caught this window via `TransactionNotYetMined`.)

Can we make the replacement gate also consult the settlement service and refuse replacement while the certificate's settlement job has not reverted (link exists and the job result is absent or Success)? That restores the in-flight protection without changing the write-on-success design (the RPC's state store is already the concrete `StateStore` implementing `SettlementReader`, so this should only need a trait bound in agglayer-rpc).

crates/agglayer-node/src/node.rs
```diff
@@ -259,15 +259,18 @@ impl Node {
...
+                state_store.clone(),
+                cancellation_token.clone(),
+            )
+            .await?,
```
Ekleog-Polygon, line 274: P1 follow-up: don't fail node boot on a single unloadable settlement job

`resume_pending_settlement_jobs` scans every settlement job id ever written and `?`-propagates per-job load errors from inside the loop (`settlement_service.rs:112-122`); this `await?` makes that a hard precondition of node boot. One undecodable/corrupt row — the CF is insert-only with no pruning, and the settlement schema is still young and evolving (e.g. #1599) — means the node can't start at all, not even read-only RPC, until manual DB surgery. The in-code TODO covers the scan cost but not the fail-closed behavior.

Follow-up, to be handled together with the settlement-service admin RPC work (#1254/#1259/#1408): skip-and-log (with a metric) on per-job load failures so an unloadable job becomes a loudly-reported stuck job that the admin API can inspect and repair, rather than a node that won't boot. (The scan-size half of this is already tracked in #1592.)

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -353,6 +398,13 @@ impl<
...
         control: TaskControl,
     ) -> eyre::Result<(SettlementJobId, Self)> {
+        let job = resolve_settlement_gas_limit(
+            provider.as_ref(),
+            tx_config.as_ref(),
+            job,
+            &control.cancellation_token,
+        )
+        .await?;
         let id = Self::reserve_settlement_job_id(store.as_ref(), certificate_id).await?;
```
Ekleog-Polygon, line 408: P1 followup: the #1599 reverse-mapping CF is never written

`CertificateIdPerSettlementJobIdColumn` — the job→certificate reverse mapping added by the breaking storage change in #1599 in preparation for this feature line — has no writer or reader at this head: `insert_certificate_settlement_job_id` writes only the forward `SettlementJobIdPerCertificateIdColumn`, and the only references to the reverse CF are its definition and a schema-presence test. Every settlement job created by this PR is therefore persisted without a reverse link, so whatever consumes the CF later (e.g. job→cert resolution for the admin API's job-id-keyed abort/reload commands, or incident tooling) will find it empty for all jobs created in the meantime and need a backfill migration.

Can we write the reverse link alongside the forward one at link-creation time (ideally in the same batch), or note the backfill debt in the tracking issue if it's deliberately deferred?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -519,134 +332,142 @@ where
...
+            })?;
+
+        let rollup_id = output.origin_network.to_u32();
+        let verifier_type = self
+            .certifier_client
+            .verifier_type(rollup_id)
+            .await
+            .map_err(|error| {
+                CertificateStatusError::SettlementError(format!(
+                    "Failed to resolve verifier type: {error}"
```
Ekleog-Polygon, line 366: P2 follow-up: verifier-type failures lost their error-level log

At the base commit, a failing verifier-type lookup surfaced as `Error::UnableToGetVerifierType` → `CertificateStatusError::InternalError`, which `process()` logs with `error!("Internal error in certificate processing")`. This rewrap into `SettlementError` routes it to the `_ => debug!` arm instead — so an L1 RPC endpoint that's down or misconfigured at settlement-build time sends every proven certificate to InError with no error-level log line for the cause, exactly when a whole network stops settling. Tooling that branches on the InError kind from `getCertificateHeader` also sees the classification change.

Can we either keep this as `InternalError` (it is an infrastructure failure, not a settlement outcome), or add an explicit `error!` at this site so ops alerting keyed on error-level logs still fires?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -519,134 +332,142 @@ where
...
-                // number of confirmations, and have not done that here yet.
-                self.header.status = CertificateStatus::Candidate;
-                return Box::pin(self.process_from_candidate()).await;
-            }
-        };
+        // The settlement job id is the only link the orchestrator keeps to the
+        // in-flight settlement, and it survives reboots.
+        let job_id = self
+            .state_store
+            .get_certificate_settlement_job_id(&certificate_id)?
```
Ekleog-Polygon, line 425: P2 followup: remove the now-inert `set-settlement-tx-hash` admin op

With settlement resolved exclusively through the cert→job-id link, the Candidate path no longer reads `header.settlement_tx_hash`, so `admin_forceEditCertificate`'s `set-settlement-tx-hash` operation (admin.rs:591) no longer drives Candidate settlement recovery: a Candidate cert with no job link goes straight back to InError regardless of the supplied hash, and one with a link ignores it in favor of the job's terminal result. (It isn't a complete no-op — the op still writes the header, and the hash still feeds the Proven arm's recompute block-pinning — but the incident runbook use is gone.) The endpoint still accepts it without warning, and the pre-existing `admin_fixup_tx_hash` test (send_certificate.rs:399, untouched by this PR) only runs with `process-now = false` and asserts the header field, so nothing surfaces the change.

Recovery tooling for stuck settlements is being handled by the settlement-service admin API currently under design (#1254/#1259/#1408), so no replacement is needed here — can we remove or deprecate `set-settlement-tx-hash` as a recovery tool (e.g. reject it for certificates whose settlement is job-managed), so operators aren't led into a no-op runbook during an incident?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -340,6 +340,51 @@ pub struct SettlementTask<L1Provider, SettlementStore> {
...
+            eyre::eyre!("settlement gas estimation cancelled before completion")
+        }
+        RetryCallbackError::Error(error) => {
+            eyre::eyre!("failed to estimate settlement gas limit: {error}")
+        }
+    })?;
+
+    job.gas_limit = tx_config
+        .gas_limit_multiplier_factor
+        .saturating_mul_u128(gas_estimate as u128)
```
Ekleog-Polygon, line 383: P2 follow-up: default gas headroom for a frozen estimate

`gas_limit_multiplier_factor` defaults to `Multiplier::ONE`, and this estimate is taken once at job creation and reused for every attempt, fee bump, and post-restart resubmission (`resolve_base_gas_params` passes it through; bumps change fees only; `load()` never re-estimates). `eth_estimateGas` leaves ~0–1.5% slack, and the call's real gas can drift upward between estimation and inclusion — other networks' settlements from the same wallet are invisible to the nonce-less estimate, and rollup-manager/GER state grows. An included OOG is receipt-status-0 → terminal Revert → InError, and with the at-most-once link that certificate id can't be re-settled.

Suggest defaulting the multiplier to 1.3.

crates/agglayer-certificate-orchestrator/src/network_task/tests.rs
```diff
@@ -1,43 +1,147 @@
...
+                .map_err(|error| eyre::eyre!("{error}"))?;
+            // The real service records the certificate <-> job-id link as part of
+            // creating the job; mirror it so process_from_candidate can resume.
+            store
+                .insert_certificate_settlement_job_id(&certificate_id, &id)
+                .map_err(|error| eyre::eyre!("{error}"))?;
+            Ok(id)
+        });
+    settlement_service
+        .expect_wait_for_settlement()
```
Ekleog-Polygon, line 111: P2 test follow-up: restore argument pinning on the settlement mocks

The pre-PR tests pinned `wait_for_settlement`'s arguments with `.withf(|t, i| ...)` (tx hash + certificate id); the rewrite stubs all 9 expectation sites (network_task/tests.rs + tests/status.rs) with `|_|`, and the 4 inline `expect_submit_settlement_job` mocks return constant `job_id(1)`. The `mock_settlement_persisting` helper does persist distinct job ids and the cert→job-id link, but no `wait_for_settlement` expectation anywhere checks the job id it receives, so a refactor that waits on the wrong or stale job id (e.g. reusing another certificate's, as in `changing_epoch_triggers_certify` where submit returns `job_id(1)` for both certificates) passes every orchestrator unit test and would only surface in e2e or production.

Can we `.withf`-pin `wait_for_settlement(job_id)` (against what the store stub / submit mock returns), and give the two-certificate test distinct job ids — e.g. by reusing `mock_settlement_persisting`, which already mints distinct ids?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1930,7 +1997,9 @@ mod tests {
     async fn create_generates_settlement_job_id() {
         let mut store = MockStateStore::new();
         let job = mk_job();
-        let expected_job = job.clone();
+        // `create` resolves the gas limit via estimateGas (mock returns 200_000).
+        let mut expected_job = job.clone();
+        expected_job.gas_limit = 200_000;
```
Ekleog-Polygon, line 2002: P2 test follow-up: no test exercises the gas-limit multiplier or ceiling anymore

Before the config-propagation rework, these create-path tests pinned the capping interaction (mock estimate 200_000 exceeded the job's 100_00x cap, and the capped value was asserted on the persisted job). After the rework they assert the raw estimate, and with default config (multiplier 1.0, ceiling 60M) both knobs are identity — a grep confirms neither knob binds behaviorally in any test anymore (the only remaining references are agglayer-config TOML parse tests that assert the parsed values). The `.min(ceiling)` is the only bound protecting settlement txs from an absurd or buggy `estimateGas` result, so a refactor that drops the min or the multiplier application ships green.

Can we set a non-default multiplier and a ceiling strictly between the raw estimate and `multiplier × estimate` in at least one of these tests? The asserted value is then the ceiling, and dropping either the multiplier or the min changes the result — both knobs bind.

tests/integrations/src/agglayer_setup.rs
```diff
@@ -121,6 +121,22 @@ pub async fn start_agglayer(
             .parse()
             .unwrap();
 
+    // Tune settlement for the test L1: it mines ~1s blocks but its "safe" head
+    // lags many blocks, so the prod SafeBlock policy would idle each certificate
+    // for tens of seconds. Settle on latest with one confirmation, and poll a few
+    // seconds apart (kept above the ~1s block time so an attempt is never
+    // abandoned before it can mine, which would resubmit and double-settle).
+    {
+        let tx_config = &mut config.settlement.pessimistic_proof_tx_config;
```
Ekleog-Polygon, line 130: P2 test follow-up: allow integration tests to opt out of the settlement-config override

This block runs unconditionally *after* `config.unwrap_or_else(...)`, so the fields it sets (settlement-policy, confirmations, retry intervals) are clobbered for every caller: the production defaults this PR turns live (SafeBlock, 12 confirmations) have zero integration coverage, and no future test can exercise those fields via `Some(config)` — the override silently rewrites them, so the test would exercise the override instead of its intended config.

Can we apply these test defaults only when the caller passed `None` (or hoist them into a helper the caller can override), so a SafeBlock/confirmations regression is at least testable end-to-end?

tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs
```diff
@@ -49,54 +44,46 @@ async fn transaction_with_receipt_status_0(#[case] state: Forest) {
...
     let tmp_dir = TempDBDir::new();
     let scenario = FailScenario::setup();
     let cancellation_token = CancellationToken::new();
 
-    fail::cfg(
-        "notifier::packer::settle_certificate::gas_estimate::low_gas",
-        "return",
-    )
-    .expect("Failed to configure failpoint");
+    fail::cfg("settlement::force_revert", "return").expect("Failed to configure failpoint");
```
Ekleog-Polygon, line 62: P2 test follow-up: the revert tests never produce a real on-chain revert

`settlement::force_revert` only flips the interpretation in `contract_call_result_from_receipt` (utils.rs:158) — the settlement tx itself executes and lands successfully on L1. So this test settles its "corrected" certificate against a rollup-manager state that already advanced from the supposedly-reverted tx, a state that's impossible after a real revert (which leaves L1 unchanged). The base suite's `gas_estimate::low_gas` failpoint — the only fixture that produced a genuine on-chain failure — lost its last users in this rewrite and sits on the legacy path whose removal is tracked in #1604, so no integration test now exercises a real revert or OOG settlement end to end (unchanged-L1 resubmission, revert-reason plumbing, the frozen-gas-limit OOG path). This is distinct from #1604's cleanup: it needs a new fixture on the new pipeline.

As part of the failure-coverage follow-up already discussed on this PR: can we add a fixture that genuinely fails on-chain (e.g. a low-gas failpoint in the settlement task's attempt builder, or intentionally corrupt calldata for one attempt), so the corrected-certificate flow is tested against a truthful L1 state?

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -750,14 +781,25 @@ mod tests {
                 }
             });
 
+        // `create` runs `estimateGas` before persisting; answer it above the
```
Ekleog-Polygon, line 784: P3 test followup: this comment now states the inverted contract

The new comment at :750 (added in "rework config propagation") correctly says `create` resolves the stored gas limit to the 200_000 estimate — but this remaining comment at :784 still claims the mock answers "above the ceiling so the stored limit is unchanged". Both halves are now wrong: 200_000 is far below the real 60M config ceiling (it was above the *job's own* 100_007 cap only under the previous semantics), and the stored limit *does* change (100_007 → 200_000). Whoever reads or copies this test next will encode the pre-rework contract.

Can we replace the stale sentence to match :750 — e.g. "`create` runs `estimateGas` before persisting; the 200_000 estimate is below the config ceiling (60M), so it becomes the stored limit."?

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -519,134 +332,142 @@ where
...
+            proof_with_selector.into(),
+            self.certificate.custom_chain_data.clone().into(),
+        );
+
+        Ok(SettlementJob {
+            contract_address: self.certifier_client.rollup_manager_address(),
+            calldata,
+            eth_value: U256::ZERO,
+            // Resolved and capped at the ceiling by the settlement service.
+            gas_limit: 0,
```
Ekleog-Polygon, line 404: P3 followup: have `SettlementJob` construction fill in its own gas limit

Can we enforce "a `SettlementJob` always carries a resolved gas limit" at construction time instead of `SettlementTask::create` fixing up a placeholder — e.g. a service-owned builder, or an `UnresolvedSettlementCall` params type the orchestrator hands over so that only the settlement service can mint a `SettlementJob` (running the estimation as part of construction)? (A literal `SettlementJob::new(provider, config)` can't live in agglayer-types, since agglayer-config depends on it.)

Ekleog-Polygon (changes requested):
Thanks for the update — I re-reviewed from scratch at the current head (47bf803c), this round with AI assistance.

One blocker: the settlement service is wired with the tx-settlement provider instead of the pp-settlement one (P0 inline) — a one-line fix, but with distinct keys configured every settlement would revert on the trusted-aggregator role check.

Everything else is follow-up material: recovery/ops gaps that the settlement-service admin API work should absorb (boot-time job-load failures, removing the now-inert `set-settlement-tx-hash` op, the unwritten #1599 reverse-mapping CF), a replacement-gate issue (an InError cert with a live in-flight settlement tx is currently replaceable), a P0-follow-up to investigate an AI-reported panic-loop failure mode in the settlement task, and test follow-ups (gas-knob coverage, mock argument pinning, the harness settlement-config override, and a fixture that produces a genuine on-chain revert).

Related tracked issues, referenced below where relevant (none fully covers these comments): #1575/#1597 (nonce allocation across settlement jobs), #1592 (bounding recovery scans), #1603 (dangling cert→job links), #1604 (dead legacy-path cleanup), #1254/#1259/#1408 (settlement admin API).

The asks from my previous round are all addressed at this head (gas resolved before persisting the job/link, epoch-assignment ordering, `wait_for` watcher semantics) — thank you!

crates/agglayer-node/src/node.rs
hadjiszs: great catch, thank you! 

tackled in 1cb257d1d4c4372c9a77b2692bdfe9d523c63706

hadjiszs: tackled in 1cb257d1d4c4372c9a77b2692bdfe9d523c63706

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1607

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1608

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1609

crates/agglayer-node/src/node.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1609

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1610

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1610

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1611

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1611

crates/agglayer-settlement-service/src/settlement_service.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1611

crates/agglayer-certificate-orchestrator/src/network_task/tests.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1612

tests/integrations/src/agglayer_setup.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1613

tests/integrations/tests/certificate_settlement/l1_settlement/failure.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1613

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
hadjiszs: included in this follow-up issue: https://github.com/agglayer/agglayer/issues/1614

Ekleog-Polygon (approved):
Thanks for the quick fix — the wallet wiring and the startup warning both look right (1cb257d1).

Approving with one ask: please log the remaining follow-ups from the previous reviews as GitHub issues before landing. With those tracked, it should be fine to generate the RC from this commit.

PR #1417, opened 2026-03-02
fix(grpc-api): correct error code for missing BlockClock config

eeemmmmmm:
get_epoch_configuration returns InvalidArgument when BlockClock is not configured, but this is misleading - the client can't fix their request, it's a server config issue. Changed to FailedPrecondition which correctly signals a server-side precondition failure.


Ekleog-Polygon (approved):
LGTM! cc @Freyskeyd as I know you investigated the tonic error codes quite a bit

Freyskeyd (approved):

PR #1420, opened 2026-03-05
feat(settlement): add proto conversion layer for settlement storage

Freyskeyd:
Introduce a settlement compat module that maps service domain structs
to storage proto records with strict required-field and enum
validation.

Centralize reusable primitive proto conversions in agglayer-storage
to keep encoding rules consistent across crates.

Keep runtime-only config fields out of persisted proto records by
requiring explicit injection when reconstructing SettlementJob.


Ekleog-Polygon (commented):
Giving a very quick look: I'm thinking this'd be better placed in `agglayer-storage`, to have all of protobuf encapsulated there and not exposed externally. WDYT?

crates/agglayer-storage/src/stores/interfaces/writer/settlement_writer.rs
```diff
@@ -52,6 +50,6 @@ pub trait SettlementWriter: Send + Sync {
         &self,
         settlement_job_id: &Ulid,
         attempt_sequence_number: u64,
-        tx_result: &TxResult,
+        tx_result: &SettlementJobResult,
```
Ekleog-Polygon: P0 followup: Hmm I thought you wanted to separate the settlement attempt result and settlement job result types?

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -42,20 +43,52 @@ impl StateStore {
 
         callback()
     }
+
+    fn encode_settlement_job(settlement_job: &SettlementJob) -> Result<v0::SettlementJob, Error> {
+        v0::SettlementJob::try_from(settlement_job).map_err(|error| {
+            Error::Unexpected(format!(
```
Ekleog-Polygon: P3 followup: this feels like it'd be a perfect place to use `eyre::Result` rather than the custom type. Probably doesn't even need to be encoded as an issue, we already have an issue to switch to eyre.

Just one thing: it'd feel cleaner if there was simply a `impl From<SettlementJobTryFromError> for Error`, this way we could completely remove these functions and just directly use `try_from(...)?`

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -42,20 +43,52 @@ impl StateStore {
...
+    fn decode_settlement_job(settlement_job: v0::SettlementJob) -> Result<SettlementJob, Error> {
+        SettlementJob::try_from(settlement_job).map_err(|error| {
+            Error::Unexpected(format!(
+                "Failed to decode settlement job from storage: {error}"
+            ))
+        })
+    }
+
+    fn encode_result(result: &SettlementJobResult) -> v0::TxResult {
+        result.clone().into()
```
Ekleog-Polygon: P3 nit: this also feels unnecessary, but 🤷
Also see the comment below about cloning at the beginning of a function.

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -120,6 +157,10 @@ impl SettlementWriter for StateStore {
         attempt_sequence_number: u64,
         settlement_attempt: &SettlementAttempt,
     ) -> Result<(), Error> {
+        let sender_wallet = settlement_attempt.sender_wallet;
+        let nonce = settlement_attempt.nonce.0;
+        let proto_settlement_attempt: v0::SettlementAttempt = settlement_attempt.clone().into();
```
Ekleog-Polygon: P3 nit: either take `settlement_attempt: SettlementAttempt` rather than cloning inside the function, or have `v0::SettlementAttempt::from(&SettlementAttempt)` rather than `from(SettlementAttempt)`. See https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

crates/agglayer-storage/src/stores/state/tests/settlement.rs
```diff
@@ -34,76 +37,51 @@ fn mk_ulid(seed: u128) -> Ulid {
...
+fn mk_settlement_job_proto(job: &SettlementJob) -> v0::SettlementJob {
+    v0::SettlementJob::try_from(job).expect("settlement job conversion should succeed")
+}
+
+fn mk_settlement_attempt_proto(attempt: &SettlementAttempt) -> v0::SettlementAttempt {
+    attempt.clone().into()
+}
+
+fn mk_result_proto(result: &SettlementJobResult) -> v0::TxResult {
+    result.clone().into()
```
Ekleog-Polygon: P2 followup: should we introduce one line in our rust code writing skill, to tell AI to avoid helper functions that are less than 3 lines, unless adding a name adds significant clarity for the reader? Here these functions feel meh

crates/agglayer-storage/src/types/settlement/compat/contract_call.rs
```diff
@@ -0,0 +1,118 @@
...
+
+fn parse_contract_call_metadata(value: v0::ContractCallMetadata) -> Result<Vec<u8>, Error> {
+    Ok(value.metadata.to_vec())
+}
+
+impl From<ContractCallOutcome> for v0::ContractCallOutcome {
+    fn from(value: ContractCallOutcome) -> Self {
+        match value {
+            ContractCallOutcome::Success => v0::ContractCallOutcome::Success,
+            ContractCallOutcome::Revert => v0::ContractCallOutcome::Reverted,
```
Ekleog-Polygon: P1 nit: can we update the proto to name it `Revert` instead? Sorry for missing that in the original proto file, your names are definitely better and we should encode them properly.

crates/agglayer-storage/src/types/settlement/compat/error.rs
```diff
@@ -0,0 +1,34 @@
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
```
Ekleog-Polygon, line 1: P2 followup: can we unify this with the same error type in agglayer-grpc-types? Or should we just replace both with `eyre::Error`? Here we're speaking of a root-cause-level error so I'd lean towards keeping it a specific type; but it'd probably make sense to move both copies to eg. agglayer-types. Among other reasons because IMHO the pre-existing type has better error formatting 😝

crates/agglayer-storage/src/types/settlement/compat/mod.rs
```diff
@@ -0,0 +1,24 @@
+macro_rules! required_field {
+    ($from:expr, $field:ident => $converter:expr) => {
```
Ekleog-Polygon: P3 nit: considering this takes `$converter:expr` and not `:ident`, the functions that are being used as `$converter` could just be lambdas.

crates/agglayer-storage/src/types/settlement/compat/primitives.rs
```diff
@@ -0,0 +1,102 @@
+use std::time::SystemTime;
+
+use agglayer_types::{Address, SettlementTxHash, B256, U256};
+
+use super::Error;
+use crate::{schema::CodecError, types::generated::agglayer::storage::v0};
+
+fn from_codec_error(error: CodecError) -> Error {
```
Ekleog-Polygon: P1 nit/followup: All these helpers feel iffy, that's too many one-liner functions IMO. Can we just implement From as needed, instead of having so many one-liners? eg. `parse_calldata` has a name that's actively misleading, hence the P0

crates/agglayer-storage/src/types/settlement/compat/primitives.rs
```diff
@@ -0,0 +1,102 @@
...
+
+pub(super) fn parse_address(value: v0::Address) -> Result<Address, Error> {
+    value.try_into().map_err(from_codec_error)
+}
+
+pub(super) fn to_proto_address(value: Address) -> v0::Address {
+    value.into()
+}
+
+pub(super) fn parse_calldata(value: v0::Calldata) -> Result<Vec<u8>, Error> {
```
Ekleog-Polygon: (eg. this one has a name that's actively misleading)

crates/agglayer-storage/src/types/settlement/compat/primitives.rs
```diff
@@ -0,0 +1,102 @@
...
+    Ok(value.into())
+}
+
+pub(super) fn to_proto_calldata(value: &[u8]) -> v0::Calldata {
+    value.to_vec().into()
+}
+
+pub(super) fn parse_uint128_to_u128(value: v0::Uint128) -> Result<u128, Error> {
+    value.try_into().map_err(from_codec_error)
```
Ekleog-Polygon: (BTW, if we have Error::from(CodecError) as suggested above, this can become just `.try_into()?`)

crates/agglayer-storage/src/types/settlement/compat/primitives.rs
```diff
@@ -0,0 +1,102 @@
...
+        .try_into()
+        .map_err(|error| Error::invalid_data(format!("invalid timestamp: {error}")))
+}
+
+pub(super) fn to_proto_timestamp(value: SystemTime) -> prost_types::Timestamp {
+    prost_types::Timestamp::from(value)
+}
+
+#[cfg(test)]
+mod tests {
```
Ekleog-Polygon: P3 nit: IMO these tests are pointless, because the functions should not exist in the first place

crates/agglayer-storage/src/types/settlement/compat/settlement_attempt.rs
```diff
@@ -0,0 +1,120 @@
...
+    use std::time::SystemTime;
+
+    use agglayer_types::{
+        Address, ClientError, ClientErrorType, Digest, SettlementJobResult, SettlementTxHash,
+    };
+
+    use super::*;
+
+    #[test]
+    fn settlement_attempt_round_trip_ignores_result() {
```
Ekleog-Polygon: P1 nit/followup: From the type perspective this feels very weird. Could we instead not have `result` at all in `SettlementAttempt`, and just expose a pair `(SettlementAttempt, SettlementAttemptResult)` whenever needed?

crates/agglayer-storage/src/types/settlement/compat/settlement_job.rs
```diff
@@ -0,0 +1,125 @@
...
+        let mut proto = v0::SettlementJob::try_from(sample_job()).unwrap();
+        proto.contract_address = None;
+
+        let result = SettlementJob::try_from(proto);
+
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn settlement_tx_hash_primitive_conversion_smoke() {
```
Ekleog-Polygon: P3 nit: this test feels useless and not actually belonging here?

crates/agglayer-types/src/settlement.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct SettlementAttemptNumber(pub u64);
+
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, derive_more::Display)]
+pub struct Nonce(pub u64);
+
+impl Nonce {
+    pub fn previous(&self) -> Option<Nonce> {
+        self.0.checked_sub(1).map(Nonce)
+    }
+}
```
Ekleog-Polygon, line 17: P0 issue: AFAICT we already have these structs as part of the agglayer-settlement-service crate, as well as the ones below. Can we deduplicate from there, probably by making agglayer-settlement-service use these types directly? Having duplicated code this way would be quite likely to cause bugs 🥲 

Considering it'll be much easier to review once we can see the structs have just been moved from one place to another, I've skipped reviewing this file this time.

crates/agglayer-types/src/settlement.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SettlementAttempt {
+    pub sender_wallet: Address,
+    pub nonce: Nonce,
+    pub max_fee_per_gas: u128,
+    pub max_priority_fee_per_gas: u128,
+    pub hash: SettlementTxHash,
+    pub submission_time: SystemTime,
+    pub result: Option<SettlementJobResult>,
```
hadjiszs: it was meant to be removed completely no?

in particular, I see in `impl From<&SettlementAttempt> for v0::SettlementAttempt {` that the `result` is indeed dropped

proto/agglayer/storage/v0/settlement.proto
```diff
@@ -40,18 +40,24 @@ message SettlementJob {
...
 
     // Result of a successfully-executed contract call.
     ContractCallResult contract_call_result = 2;
   }
 }
 
+// Terminal result of a settlement job.
+message SettlementJobResult {
+  // Result of the final on-chain contract call for the job.
+  ContractCallResult contract_call_result = 1;
```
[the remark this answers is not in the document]
Freyskeyd: This is not a problem as nothing is deployed yet

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -5,35 +5,24 @@ use std::{
...
-    }
-}
-
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub struct SettlementJob {
-    contract_address: Address,
-    calldata: Bytes,
-    eth_value: U256,
-
+pub struct ActiveSettlementJob {
```
Ekleog-Polygon: P1 issue: do we need `ActiveSettlementJob` at all anymore?

My understanding is that the settlement policy and fee parameters are persisted on `SettlementJob` on purpose, so that reloaded jobs and jobs modified through the admin RPC keep using the saved values rather than the node's current config.

If that's the intended model, then `ActiveSettlementJob` feels like an unnecessary second layer: it duplicates data that is already in `SettlementJob`, and it even reintroduces `settlement_config` as another apparent source of truth.

Would it make sense to remove `ActiveSettlementJob` entirely and keep using `SettlementJob` directly here? AFAICT once `settlement_config` goes away, I don't see any runtime-only state left in this wrapper.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -348,6 +304,7 @@ impl SettlementTask {
     }
 
     fn is_any_attempt_pending_for_nonce(&self, wallet: Address, nonce: Nonce) -> bool {
+        // TODO: Replace by Storage call
```
Ekleog-Polygon: P1 issue: this TODO points future work in the wrong direction.

The intended design here is that storage is loaded once when the settlement task starts, and from then on the task operates on its in-memory state. So this helper is doing the right kind of thing already: it is querying `self.attempts`, not re-fetching from storage.

As a result, this TODO is misleading and risks pushing a later change toward the wrong architecture. Would it make sense to remove it rather than leaving an instruction that contradicts the intended design?

crates/agglayer-storage/src/types/settlement/compat/tx_result.rs
```diff
@@ -0,0 +1,180 @@
+use agglayer_types::{SettlementAttemptResult, SettlementJobResult};
+
+use super::Error;
+use crate::types::generated::agglayer::storage::v0;
+
+enum SettlementAttemptResultKind {
```
Ekleog-Polygon: P2 nit: why do we need `SettlementAttemptResultKind` as an intermediate type here?

AFAICT this is structurally identical to `SettlementAttemptResult`, so the conversion now becomes `proto -> SettlementAttemptResultKind -> SettlementAttemptResult` without gaining much from the extra step.

Would it make sense to convert directly from `v0::SettlementAttemptResult` / `v0::settlement_attempt_result::Result` into `SettlementAttemptResult`? This feels like the kind of indirection that makes the compat layer a bit harder to read without really simplifying the logic.

crates/agglayer-storage/src/error.rs
```diff
@@ -48,6 +48,12 @@ pub enum Error {
     SmtNodeNotFound,
 }
 
+impl From<crate::types::settlement::compat::Error> for Error {
```
Ekleog-Polygon: P2 followup: can we avoid stringifying the compat error here?

One thing I liked in the rest of this refactor is that the compat layer now carries fairly precise field-level errors. But this conversion immediately flattens all of that back into `Unexpected(String)`, which means we lose the structure again as soon as we cross into `agglayer-storage`.

Given that we've agreed as a team that we'd like to move more towards `eyre::Error`, this looks like a good opportunity to at least avoid baking the information into a string too early. Would it make sense either to keep the compat error as a proper source error here, or to use an `eyre::Error`-style wrapper rather than `format!`ing it away?

crates/agglayer-storage/src/types/settlement/compat/settlement_job.rs
```diff
@@ -0,0 +1,109 @@
...
+                value,
+                max_priority_fee_per_gas_floor => try_into::<u128>
+            ),
+            max_priority_fee_per_gas_increase_percents: value
+                .max_priority_fee_per_gas_increase_percents,
+        })
+    }
+}
+
+impl TryFrom<&SettlementJob> for v0::SettlementJob {
```
Ekleog-Polygon: P3 nit: should this still be `TryFrom`?

AFAICT this conversion is now fully infallible: every field is converted with `into()`, and the body just wraps the result in `Ok(...)`.

If that's right, using `TryFrom` here makes the API look more complex than it really is, and forces callers to handle an error path that does not actually exist. Would it make sense to switch this one to `From<&SettlementJob>` (and maybe `From<SettlementJob>` too if we still want the owned variant)?

crates/agglayer-settlement-service/src/settlement_task.rs
Freyskeyd: If the goal is to not hit the storage (which for me, we could do in regard of the amount of call we'll have), this cache should be placed in the store directly, not in the service

Ekleog-Polygon: Thanks, that clarifies the alternative you have in mind.

The intended design here is that the settlement task loads persisted state once when it starts, and from that point on the source of truth for active settlement state lives in the task itself. In that model, querying `self.attempts` here is the right layering: this is task-local runtime state, not store state.

Moving that cache into the store would mix persisted data with live task state, which is what I want to avoid here.

If we want to revisit the storage/task boundary more broadly, I'd rather do that explicitly in a separate change.

crates/agglayer-storage/src/types/settlement/compat/error.rs
```diff
@@ -0,0 +1,46 @@
...
```
Ekleog-Polygon, line 1: P2 followup: can we avoid introducing a second generic compat error type here?

AFAICT this is the same kind of field-annotated conversion error we already have on the gRPC side, and `crates/agglayer-storage/src/types/settlement/compat/mod.rs` also duplicates the same `required_field!` pattern from `crates/agglayer-grpc-types/src/compat/v1/mod.rs`.

Keeping two near-identical compat error/helper stacks means they can drift on formatting, variants, and helper APIs over time, and any cleanup has to be done twice. Would it make sense to move this generic compat error machinery to a shared crate and have both storage and gRPC conversions reuse it?

Freyskeyd: I would prefer a shared crate, I'll do that in a followup PR

Issue #803, opened 2025-05-16
Sensitive WebSocket token exposed in logs: secret included in connection URL


obynonwane:
### Description

When a user runs the agglayer-node, the logs may include the full WebSocket connection URL — including an embedded secret or token as the last path segment. This becomes a security/privacy risk when users share logs to support team for troubleshooting and help or even to community channel like discord for other developers to help resolve issue, unaware that the secret is being exposed.

### Example

Logs currently show something like: {"message":"Starting BlockClock with provider: wss://websocket-url/secret-key"}

This can leak sensitive credentials unintentionally.

### Expected Behavior

Sensitive tokens in URLs should be redacted in logs. For example:  {"message":"Starting BlockClock with provider: wss://websocket-url/xxxx..."}

### Recommendation

Use a sanitizer (e.g., `sanitize_ws_url`) before printing any WebSocket URLs to the console or log. Ensure this is applied consistently across all modules that output connection details.

### Impact

- Leaks secrets in public/shared logs
- Users may unknowingly compromise their node setup
- Affects troubleshooting and security hygiene

### Additional Context

This can be mitigated by adopting a standard log-sanitizing wrapper or utility for sensitive data.

Freyskeyd:
Another solution could be to remove the URL from the logs directly.

PR #1432, opened 2026-03-10
feat: hide urls in alloy errors

Ekleog-Polygon:
Fixes https://github.com/agglayer/agglayer/issues/803
Fixes https://github.com/agglayer/security/issues/205


crates/agglayer-node/src/node.rs
Freyskeyd, line 181: This is only for RPC call linked to settlement, what's about Clock (it uses websocket IIRC) ?

Issue #1383, opened 2026-02-11
Helper function to find the tx hash on L1 currently consuming a given wallet/nonce


Ekleog-Polygon:
If there is already a function, list it here.
If not, write one and list it here.

Search in the code:
```
// XREF: https://github.com/agglayer/agglayer/issues/1383
```

PR #1439, opened 2026-03-13
feat: document kurtosis running for ai agents

Ekleog-Polygon:
Part of #1383: first step towards trying out reth for kurtosis, to get the required RPC available.


docs/knowledge-base/e2e-tests.md
```diff
@@ -0,0 +1,159 @@
...
+mkdir -p /tmp/agglayer-e2e
+
+git clone https://github.com/0xPolygon/kurtosis-cdk.git /tmp/agglayer-e2e/kurtosis-cdk
+cd /tmp/agglayer-e2e/kurtosis-cdk
+git fetch origin "$KURTOSIS_CDK_REF" && git checkout "$KURTOSIS_CDK_REF"
+
+git clone https://github.com/agglayer/e2e.git /tmp/agglayer-e2e/agglayer-e2e
+cd /tmp/agglayer-e2e/agglayer-e2e
+git fetch origin "$AGGLAYER_E2E_REF" && git checkout "$AGGLAYER_E2E_REF"
+```
```
hadjiszs, line 83: nit: I think we can use directly `kurtosis run github.com/<>/<>@tag` https://docs.kurtosis.com/run/

>If you want to run a non-main branch, tag or commit use the following syntax kurtosis run github.com/package-author/package-repo@tag-branch-commit

docs/knowledge-base/e2e-tests.md
```diff
@@ -0,0 +1,159 @@
...
+### 4. Start the kurtosis network
+
+```bash
+kurtosis engine start
+kurtosis clean --all
+kurtosis run \
+  --enclave cdk \
+  --args-file /tmp/agglayer-e2e/kurtosis-args.json \
+  /tmp/agglayer-e2e/kurtosis-cdk
+```
```
hadjiszs, line 105: just to be sure I guess that's mock mode right?

docs/knowledge-base/e2e-tests.md
Ekleog-Polygon: Hmm… I don't thkni that'd be possible with our specific kurtosis config? We need to define custom yml files in the repository before running kurtosis to configure everything properly :thinking: 

Will log looking into it more as a follow-up if you think that's important!

docs/knowledge-base/e2e-tests.md
Ekleog-Polygon: It's whatever is defined in the kurtosis-args.json! Which should default to mock mode with the CI config, that the agents should be using. In practice considering how I've been trying it locally, it does seem to be mock mode, or I'd never have seen tests succeed :smile:

Issue #1450, opened 2026-03-18
Upgrade to the latest `agglayer-contracts` ABI


hadjiszs:
# Context

The `agglayer-node` refers to the ABI of a topic branch `feature/ongoing-0.3.0` from `agglayer-contracts`.

The latest version of `agglayer-contracts` is [v12.2.3](https://github.com/agglayer/agglayer-contracts/releases/tag/v12.2.3) as of this issue date creation

In between, a renaming of the contracts happened, so doing this upgrade to `>12.2.3` will come with quite a large PR since the rust structures need to be renamed accordingly as well.

Renaming documented there: https://agglayer.github.io/protocol-team-docs/smart-contracts/v12/Rename%20%26%20reorg/renaming/

# To Do

- Upgrade on the `agglayer-node`
- Upgrade on the `aggkit-prover` (witness generation for the aggchain proof requires some fetching from the agglayer contracts)

PR #1460, opened 2026-03-25
chore: migrate agglayer to SP1 v6 and interop 0.14

Freyskeyd:
This PR upgrades agglayer to SP1 v6 and interop 0.14 as part of the migration tracked in #1451. It aligns agglayer with the SP1-v6-compatible provers revision (`c3e3e32a4a070f56b1dd6cf963d808020561c08f`) and updates proving-related dependencies to the new versions required by that stack.

The certifier and pessimistic-proof paths are migrated to SP1 v6 APIs, including `Elf::Static(...)`, by-value `SP1Stdin` usage in execute/prove flows, `EnvProver` for mock/cpu setup, and proving-key access through `ProvingKey::verifying_key()`. Error mapping was also updated for new SP1 verification variants.

To keep artifact generation and checks aligned with SP1 v6, pessimistic-proof ELF references are moved to the rv64 target paths and corresponding CI checksum inputs are updated. Test/support code that constructs mock proofs or invokes SP1 runners was updated accordingly so the full local proving/verification flow remains consistent with v6 behavior.

Related: #1451  
Follow-up: #1459

BREAKING-CHANGE: agglayer proving/verification codepaths now rely on SP1 v6 APIs and rv64 ELF artifacts. SP1 v5 assumptions (including previous ELF target/path and v5 proving call patterns) are no longer compatible with this branch.


crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -91,24 +95,79 @@ impl<PendingStore, L1Rpc> CertifierClient<PendingStore, L1Rpc> {
...
+            SP1Proof::Plonk(plonk) => &plonk.public_inputs,
+            SP1Proof::Groth16(groth16) => &groth16.public_inputs,
+            _ => return Ok(()),
+        };
+
+        let vkey_hash = public_inputs.first().ok_or_else(|| {
+            eyre::eyre!(
+                "mock proof is missing required public inputs: expected at least 2, got {}",
+                public_inputs.len()
+            )
```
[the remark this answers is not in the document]
Freyskeyd: This helper is intentionally mirroring SP1 v6 mock verification semantics rather than adding stricter checks.

In sp1-sdk v6, mock `Plonk/Groth16` verification delegates to `verify_mock_public_inputs`, which only validates:
- `public_inputs[0]` against `verifying_key.hash_bn254()`
- `public_inputs[1]` against `public_values.hash_bn254()`

 It does not validate the remaining mock-only slots (exit_code, vk_root, proof_nonce).

Also, once the proof is typed as `PlonkBn254Proof / Groth16Bn254Proof`, `public_inputs` is already a fixed `[String; 5]`, so malformed shorter layouts are not representable at this point.

Since this failpoint path is specifically meant to preserve the existing mock-verification behavior under `return()`, I kept it aligned with the upstream SP1 mock verifier instead of introducing stricter, repo-specific checks.

scripts/make/Makefile.contracts.toml
```diff
@@ -1,6 +1,7 @@
 [env]
-CONTRACT_VERSION = "feature/ongoing-v0.3.0"
+CONTRACT_VERSION = "12.2.3"
 CONTRACT_REPO = "https://api.github.com/repos/agglayer/agglayer-contracts"
+CONTRACT_IMAGE_REPO = "ghcr.io/agglayer/agglayer-contracts"
```
hadjiszs, line 4: hm but then this PR doesnt contain the update of `cargo make contracts` tied to this new version right? If so then I'd keep the version tied to the current contracts considered by the agglayer codebase, and then make the proper upgrade of contracts in a dedicated PR. In particular I think it'd require to update to the new contract names of the latest version referred by #1450

Freyskeyd: I had a hard time making the integration tests pass, so I explore the upgrade of the contract.
I also ran the `cargo make contracts` but they seems up to date (at least for the contract we're using)

Do you see any major issue here?

hadjiszs: ah but I dont see the ABI files changed no? those at `crates/agglayer-contracts/src/contracts/*.json` arent supposed to be updated if `cargo make contracts` is successful?

when I tried it didn't work because the contract names changed, hence the dedicated ticket for the renaming etc

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -91,24 +95,79 @@ impl<PendingStore, L1Rpc> CertifierClient<PendingStore, L1Rpc> {
...
+        #[cfg(any(test, feature = "testutils"))]
         fail::fail_point!(
             "notifier::certifier::certify::before_verifying_proof",
-            |_| {
-                let verifier = sp1_sdk::ProverClient::builder().mock().build();
-                let (_, verifying_key) = verifier.setup(ELF);
-
-                Ok(verifier.verify(proof, &verifying_key)?)
-            }
+            |_| testutils::verify_mock_proof_inputs(verifying_key, proof)
```
Ekleog-Polygon, line 106: P1 issue: preserve proof-verification error classification in the failpoint verifier

`verify_mock_proof_inputs` now returns plain `eyre` errors, but this caller only maps `SP1VerificationError` into `CertificationError::ProofVerificationFailed`. When `notifier::certifier::certify::before_verifying_proof` is enabled, bad proofs therefore get reported as `CertificationError::Other` instead of a proof-verification failure. That changes the downstream certificate-status mapping, and this path is exercised in integration startup via `tests/integrations/src/agglayer_setup.rs`, which enables this failpoint with `return()`. Can we have the failpoint path return the same SP1 verification error type as the normal verifier so tests and recovery paths keep the same semantics?

crates/agglayer-storage/src/columns/proof_per_certificate/tests.rs
```diff
@@ -37,44 +35,60 @@ fn can_parse_value() {
...
-        }
-    }
-}
-
+/// Guard against accidental changes in the storage encoding of [`Proof`].
+///
+/// The previous fixture was serialised with SP1 v3 types which are binary-
+/// incompatible with SP1 v6 (`PlonkBn254Proof.public_inputs` grew from
+/// `[String; 2]` to `[String; 5]`).  Instead of maintaining a cross-version
+/// fixture shim, we now generate a real Plonk proof via the mock prover and
```
Ekleog-Polygon: P0 issue: do not replace the persisted proof fixture with a newly generated v6 round-trip

`proof_per_certificate` stores persisted proofs that are read back later during settlement and epoch processing, including after restarts. Replacing the historical fixture with a newly generated v6 proof removes the only coverage we had for deserializing bytes written by older versions; this test now only proves that freshly encoded v6 data can be decoded by the same v6 codec. That is a backward-compatibility regression in exactly the area the old fixture was protecting. Can we keep the historical fixture test and add the new v6-native round-trip as a separate test instead? Breaking deserialization of already-persisted proofs would strand pending certificates across upgrades.

crates/agglayer-storage/src/types/certificate/tests/mod.rs
```diff
@@ -385,6 +389,35 @@ fn regressions(#[case] cert_filename: &str) {
     let _certificate = Certificate::decode(&bytes).expect("decoding failed");
 }
 
+/// Regenerate `regression_01.hex` with current (v6) SP1 types.
```
Ekleog-Polygon: P0 issue: do not rewrite the historical certificate regression fixture to current SP1 bytes

The whole point of `regression_01.hex` was to keep a real previously-encoded certificate decodable forever. Replacing that fixture with freshly generated v6 bytes turns `regressions()` from an upgrade-safety test into another current-version round-trip, even though `Certificate::decode` still treats these rows as the same storage version (`1`) and there is no migration path for old embedded SP1 proof bytes. If pre-upgrade certificates with aggchain proofs can no longer deserialize, nodes will fail to read already-persisted certificate rows after upgrade. Please keep the old regression fixture and add any new v6 fixtures as separate tests; if compatibility really must be broken here, it needs an explicit storage-version bump or migration rather than editing the regression fixture in place.

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -62,18 +63,21 @@ impl<PendingStore, L1Rpc> CertifierClient<PendingStore, L1Rpc> {
...
+                    EnvProver::Mock(ProverClient::builder().mock().build())
                 } else {
-                    sp1_sdk::ProverClient::builder().cpu().build()
+                    EnvProver::Cpu(ProverClient::builder().cpu().build())
                 };
-                let (_, verifying_key) = verifier.setup(ELF);
-                (verifier, verifying_key)
+                let proving_key = verifier
+                    .setup(Elf::Static(ELF))
+                    .map_err(|e| eyre!(e.to_string()))?;
```
Ekleog-Polygon: P2 issue: avoid stringifying the SP1 setup error here

This `eyre!(e.to_string())` drops the original error type and any chained context right at the verifier setup boundary. Since the surrounding code already uses `eyre`, can we keep the original error with `eyre!(e)` (or just propagate it directly) here and in similar new call sites? Keeping the structured error will make SP1 upgrade and debug failures much easier to diagnose.

crates/pessimistic-proof-test-suite/tests/cycle-tracker.rs
```diff
@@ -10,14 +10,14 @@ use pessimistic_proof::{
...
+#[timeout(Duration::from_secs(240))]
 fn sanity_check() {
     cycles_on_sample_inputs("s00_be000", Forest::new([]), std::iter::empty());
 }
 
 #[rstest::rstest]
-#[timeout(Duration::from_secs(60))]
-fn cycles_on_state01(#[values(0, 1, 2, 20, 50, 100, usize::MAX)] n_exits: usize) {
+#[timeout(Duration::from_secs(240))]
+fn cycles_on_state01(#[values(0, 1, 2, 20, 50)] n_exits: usize) {
```
Ekleog-Polygon: P2 question: why were the larger cycle-tracker snapshots dropped?

This used to cover `100` and the full sample (`usize::MAX`, which produced the `be223` snapshot), and this PR deletes both corresponding snapshots. I can understand trimming CI cost, but those larger inputs were the only guard we had against non-linear cycle growth once the bridge-exit set gets big. Was there a measured reason to remove them rather than keep them in a slower lane or ignored test? If the intent is just runtime reduction, I would prefer preserving the coverage somewhere instead of deleting the regression snapshots outright.

crates/agglayer-storage/src/types/certificate/tests/snapshots/agglayer_storage__types__certificate__tests__cert_v1_01.snap
```diff
@@ -1,6 +1,5 @@
...
...01000000000000000100000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000001000000000000000000000000000000000000aa04000078000000000058a74cf35ad4980c7affd0891e5991c06fbba5c96c0aefbf6879b10721c1f33f394933fb4e9cfc634f78a95a3167d86927c8cdde40587b4628ade6211e1ec6890a37882106fb8c0f3789faa47a1f5d6a47f82161713e03eb000000000000000000000005312e322e3358585858585858585858585858585858585858585858585858585858585858580000000000000020787878787878787878787878787878787878787878787878787878787878787800000000000000209a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a9a00b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9b9000000000000000000
```
Ekleog-Polygon: P0 issue: these storage snapshot updates are replacing backward-compat checks rather than extending them

Across the certificate storage tests, the snapshots for the proof-carrying v1 cases are being rewritten in place to the new SP1 v6 byte layout. That means the tests stop checking whether older version-1 rows remain decodable and instead only assert the encoding of freshly generated current data. Since `Certificate::decode` still dispatches all of these rows through the same version tag (`1`), changing the embedded proof encoding without keeping the old fixtures effectively drops the upgrade-safety coverage for persisted certificate rows. I think we should keep the existing snapshots and fixtures for old serialized bytes and add new v6-specific cases alongside them, so old storage tests continue to pass while new storage tests cover the new layout.

Ekleog-Polygon (changes requested):
Sorry this took me so long. I kept skirting around it because I expected it to be even harder to review than it actually was. It was still not easy, but it was not a random `+6000/-5000` code-motion PR; most of the churn here is test fixtures and generated artifacts.

Most of my substantive concerns are in the inline comments, but the main theme is backward compatibility: several storage-test updates look like they replace historical fixtures and snapshots with freshly generated SP1 v6 bytes, which weakens the upgrade-safety coverage exactly where persisted data matters most.

I did not review the JSON contract artifact changes in `crates/agglayer-contracts/src/contracts/*.json`.

Could @krlosMata confirm that the new contracts image/tag used for this upgrade is the intended one?

scripts/make/Makefile.contracts.toml
Freyskeyd: Contracts files have been updated :) 
good catch

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
Freyskeyd: Fixed.

The failpoint path now returns `SP1VerificationError` again, so the caller keeps mapping bad proofs to `CertificationError::ProofVerificationFailed` instead of `Other`.

I also added a regression test covering the failpoint verifier path:
`failpoint_mock_verifier_preserves_sp1_error_type`.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -280,71 +386,203 @@ impl<'a> From<&'a AggchainData> for AggchainDataV1<'a> {
...
+        other => CodecError::NonWritableCertificate {
+            reason: other.to_string(),
+        },
+    })
 }
 
 impl crate::schema::Codec for Certificate {
     fn encode_into<W: std::io::Write>(&self, writer: W) -> Result<(), CodecError> {
+        ensure_writable(self)?;
         Ok(bincode_codec().serialize_into(writer, &CurrentCertificate::from(self))?)
```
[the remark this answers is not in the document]
Freyskeyd: This one is intentional.

For this upgrade path, pending certificates carrying legacy SP1 v5 proofs are expected to be discarded rather than migrated forward through storage rewriting, so we are not preserving that re-encode path.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -280,71 +386,203 @@ impl<'a> From<&'a AggchainData> for AggchainDataV1<'a> {
...
+{
+    Certificate::try_from(deserialize_bincode::<T>(bytes)?)
+}
+
+fn ensure_storable(certificate: &Certificate) -> Result<(), CodecError> {
+    let result = match &certificate.aggchain_data {
+        AggchainData::ECDSA { .. } | AggchainData::MultisigOnly { .. } => Ok(()),
+        AggchainData::Generic { proof, .. } => proof.ensure_readable().map(|_| ()),
+        AggchainData::MultisigAndAggchainProof { aggchain_proof, .. } => {
+            aggchain_proof.proof.ensure_readable().map(|_| ())
```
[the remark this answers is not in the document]
Freyskeyd: This is intentional for now.

The storage layer currently enforces readability of accepted proof versions, but it does not require full executability at encode time.

The deeper SP1 validation still happens later in the certification path.

crates/agglayer-grpc-types/src/compat/v1/certificate.rs
```diff
@@ -1,30 +1,329 @@
...
+
+fn parse_sp1_proof(value: agglayer_interop::grpc::v1::Sp1StarkProof) -> Result<Proof, Error> {
+    let proof = SP1StarkWithContext {
+        version: value.version,
+        proof: value.proof.to_vec(),
+        vkey: value.vkey.to_vec(),
+    };
+    proof
+        .ensure_deserializable()
+        .map_err(|err| Error::invalid_data(err.to_string()))?;
```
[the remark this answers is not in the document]
Freyskeyd: This is currently intentional.

Ingress now validates that SP1 payload bytes are deserializable for accepted legacy/current versions, but it does not require executability at ingest time.

The current behavior is to accept readable/deserializable legacy payloads and let certification reject non-executable ones later.

crates/agglayer-grpc-types/src/compat/v1/certificate.rs
```diff
@@ -1,30 +1,318 @@
...
+
+fn parse_sp1_proof(value: agglayer_interop::grpc::v1::Sp1StarkProof) -> Result<Proof, Error> {
+    let proof = SP1StarkWithContext {
+        version: value.version,
+        proof: value.proof.to_vec(),
+        vkey: value.vkey.to_vec(),
+    };
+    proof
+        .ensure_readable()
+        .map_err(|err| Error::invalid_data(err.to_string()))?;
```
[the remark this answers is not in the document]
Freyskeyd: Fixed.

Ingress now requires the SP1 payload to be deserializable, not just readable by version string.
I also added a regression test for malformed SP1 bytes on ingestion.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -280,71 +386,203 @@ impl<'a> From<&'a AggchainData> for AggchainDataV1<'a> {
...
+    T: DeserializeOwned,
+    Certificate: TryFrom<T, Error = CodecError>,
+{
+    Certificate::try_from(deserialize_bincode::<T>(bytes)?)
+}
+
+fn ensure_storable(certificate: &Certificate) -> Result<(), CodecError> {
+    let result = match &certificate.aggchain_data {
+        AggchainData::ECDSA { .. } | AggchainData::MultisigOnly { .. } => Ok(()),
+        AggchainData::Generic { proof, .. } => proof.ensure_readable().map(|_| ()),
```
hadjiszs, line 566: I read `ensure_storable` as ensuring that something can be stored/written (so `ensure_writable`), but then in the checks I see that for the aggchain proof we ask for `ensure_readable`

also, a bit below we see `CodecError::NonWritableCertificate` with a reason message `... is not readable in certificate storage v2`

just to be sure if all that is intentional?

crates/agglayer-types/src/certificate/testutils.rs
```diff
@@ -244,20 +249,22 @@ pub enum AggchainDataType {
     MultisigAndAggchainProof { num_signers: usize },
 }
 
-/// Empty ELF file for testing purposes.
-/// This is a minimal ELF that can be used to create dummy SP1 proofs in tests.
-pub const EMPTY_ELF: &[u8] = include_bytes!("tests/empty.elf");
+/// Any valid riscv64 ELF works here; we only need it to derive a proving key
+/// for mock proof creation. The previous `empty.elf` was riscv32 and is not
+/// compatible with SP1 v6.
```
hadjiszs:
```suggestion
/// Any valid riscv64 ELF works here; we only need it to derive a proving key
/// for mock proof creation.
```

crates/agglayer-storage/src/columns/proof_per_certificate/tests.rs
```diff
@@ -38,43 +39,149 @@ fn can_parse_value() {
...
+    plonk_vkey_hash: [u8; 32],
+}
+
+#[derive(Deserialize)]
+struct LegacyGroth16Bn254Proof {
+    public_inputs: [String; 2],
+    encoded_proof: String,
+    raw_proof: String,
+    groth16_vkey_hash: [u8; 32],
+}
```
hadjiszs: we don't need that no? I see many occurrence like for handling legacy groth16 proofs which we dont have, the ones we'll get will already be v6 ones no?

crates/agglayer-storage/src/columns/proof_per_certificate/tests.rs
```diff
@@ -38,43 +39,149 @@ fn can_parse_value() {
...
     }
 }
 
-impl From<SP1ProofWithPublicValuesV3> for SP1ProofWithPublicValues {
-    fn from(v3: SP1ProofWithPublicValuesV3) -> Self {
+fn expand_legacy_public_inputs([vkey_hash, public_values_hash]: [String; 2]) -> [String; 5] {
+    [
+        vkey_hash,
+        public_values_hash,
+        "0".to_string(),
```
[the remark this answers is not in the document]
Freyskeyd: valid concern, but I’d prefer to defer the storage-proof type refactor to a follow-up PR and keep this one scoped to compatibility fixes: https://github.com/agglayer/agglayer/issues/1489

crates/pessimistic-proof/elf/riscv64im-succinct-zkvm-elf
hadjiszs, line 1: nit: we can maybe remove the `riscv32im-succinct-zkvm-elf` (32b one) that is still there?

crates/agglayer-storage/src/types/certificate/mod.rs
Freyskeyd: `encode_into` is definitely a write, but in this case it writes the storage v2 certificate envelope, not the latest active proof format.

The storage intent here is to preserve readable historical proofs when rewriting old rows, so the gate is “storage-readable” rather than “currently writable as the latest proof format”. That is why the check uses `ensure_readable()` and not `ensure_writable()`.

I agreed the old naming was confusing, so I renamed the helper to `ensure_storage_readable` and added a comment to make that distinction explicit.

There is also a regression test that demonstrates the intended behavior:

- `regression_01_legacy_can_be_reencoded_for_storage`

It now explicitly asserts that:

- the input fixture starts with storage version `1`
- after `Certificate::encode(...)`, the output starts with storage version `2`

So the tested contract is:
- readable legacy v1 certificate in
- current storage v2 certificate out
- no loss of semantic data across the rewrite

That is why `ensure_writable()` would be too strict here: it would reject the exact legacy proofs that storage v2 is supposed to preserve during migration/re-encoding.

crates/agglayer-grpc-types/src/compat/v1/certificate.rs
Ekleog-Polygon, line 1: `interop#202` intentionally preserves two different v1 encodings for `context["public_values"]`: the legacy `Option<Box<AggchainProofPublicValues>>` payload for `AggchainData::Generic`, and the bare `AggchainProofPublicValues` payload for `AggchainProof`. Here both directions go through the same `parse_public_values` / `encode_public_values` helpers, so the `Generic` path now collapses onto the bare shape as well.

Since this PR is explicitly trying to introduce `interop#202` behavior, could we avoid re-specifying that contract here and instead share/delegate to the interop compat implementation? The current duplication already drifted once, and this is the kind of protocol detail that is very hard to keep in sync across repos by hand.

crates/agglayer-storage/src/types/certificate/tests/mod.rs
Ekleog-Polygon, line 1: The updated `cert_v1_*` / `aggdata_v1_*` snapshots are now driven by synthetic test constructors instead of the checked-in historical bytes from main. That makes these snapshots much harder to trust as backward-compat evidence, because there are now two sources of truth: the real historical fixtures and the locally rebuilt `CertificateV1::test*` / `AggchainDataV1::test*` values.

Could we anchor these snapshots to the historical fixture bytes instead, or derive the expected bytes through the same production compatibility path? Otherwise snapshot churn here mostly tells us that the test builders changed, not that the live retrocompat behavior stayed correct.

Ekleog-Polygon, line 1: This regression test never goes through `Certificate::decode` / `legacy_proof_into_v2`. It deserializes `CertificateV1` directly and then mutates the result with `normalize_legacy_certificate_version`, which means the compatibility behavior is being reproduced in test-only code instead of checked through the live decoder.

Could we make this assertion use the production decode path, or at least a shared production helper, so the fixture actually protects the runtime migration logic? As written, a drift in the real compatibility code could still leave this test green.

crates/agglayer-storage/src/columns/proof_per_certificate/tests.rs
Ekleog-Polygon, line 1: This backward-compat check currently goes through a bespoke `LegacySP1ProofWithPublicValues` test struct and a test-only `Into` conversion, so it never exercises the production storage decode path. That makes the assertion much weaker: the test can stay green even if the live decoder/migration logic drifts, because it is snapshotting the shim's reconstruction instead.

Could we route this through an existing production code path instead, or extract the historical-proof loading logic into shared non-test code if production really depends on it? Right now the retrocompat coverage here is mostly proving that the test shim still knows how to read the fixture.

Ekleog-Polygon (changes requested):
Reviewed against commit `d51f2d26c8ab7bc134f5ab7963be61490cf46cf2`.

- The gRPC v1 compat logic in agglayer should not independently re-specify the `public_values` contract from `interop#202`; it already drifted on the `AggchainData::Generic` path, and this wants shared code or a single source of truth.
- I also have a few lower-priority test concerns: several of the retrocompat checks currently validate test-only reconstruction code or synthetic builders instead of the production decode/migration path, which makes the compatibility coverage less trustworthy than it looks.

crates/agglayer-grpc-types/src/compat/v1/certificate.rs
```diff
@@ -7,24 +11,108 @@ impl TryFrom<v1::Certificate> for Certificate {
...
+                    aggchain_params,
+                    signature,
+                    public_values,
+                }
+            }
+            Some(agglayer_interop::grpc::v1::aggchain_data::Data::Multisig(multisig)) => {
+                // Full multisig delegation is temporary until agglayer/interop#214
+                // hardens signer-index overflow handling in the upstream compat decoder.
+                AggchainData::MultisigOnly {
+                    multisig: multisig
```
Ekleog-Polygon, line 63: This refactor fixed the earlier Generic/public_values drift by delegating to interop, but it also delegates multisig decoding wholesale even though the upstream overflow hardening is still missing. The new ignored regression test makes that explicit: malformed signer indexes that agglayer used to reject locally are now only expected to be rejected once `interop#214` lands.

Because these protobuf payloads still arrive from untrusted gRPC clients, I think we should keep the local signer-index overflow / too-many-signers checks here until the upstream fix lands, rather than knowingly relaxing ingress validation in this branch.

crates/agglayer-grpc-types/src/compat/v1/tests.rs
```diff
@@ -74,3 +111,355 @@ make_round_trip_fuzzers!(
...
+        _ => panic!("expected generic aggchain proof"),
+    };
+
+    assert_eq!(sp1.version, sp1_sdk_v5::SP1_CIRCUIT_VERSION);
+    assert_eq!(sp1.proof, proof_bytes);
+    assert_eq!(sp1.vkey, vkey_bytes);
+}
+
+#[test]
+#[ignore = "blocked on agglayer/interop#214"]
```
Ekleog-Polygon, line 179: I think it would be better to fix `agglayer/interop#214` first and re-enable this regression test before landing this branch. Right now the ignored test documents a known ingress-validation vulnerability, and merging with it still disabled makes it too easy to forget about it by the time we cut a release. If we cannot take the upstream fix yet, I’d prefer to keep the local hardening in place until that issue is resolved.

crates/agglayer-types/src/certificate/testutils.rs
```diff
@@ -244,20 +249,21 @@ pub enum AggchainDataType {
     MultisigAndAggchainProof { num_signers: usize },
 }
 
-/// Empty ELF file for testing purposes.
-/// This is a minimal ELF that can be used to create dummy SP1 proofs in tests.
-pub const EMPTY_ELF: &[u8] = include_bytes!("tests/empty.elf");
+/// Any valid riscv64 ELF works here; we only need it to derive a proving key
+/// for mock proof creation.
+pub const EMPTY_ELF: &[u8] = pessimistic_proof::ELF;
```
Ekleog-Polygon, line 254: These storage v2 snapshots changed again in this range, and the nearby explanation is the proof-program ELF bump: the current test helpers derive mock proof keys from `pessimistic_proof::ELF`, so certificate-storage snapshots churn whenever the proof program vkey changes, even if the storage codec itself does not.

Could we switch these storage snapshot tests to a stable dedicated riscv64 test ELF or another fixed proof fixture? We will need that separation anyway, and doing it here would make backward-compat reviews much safer by keeping storage snapshot diffs focused on storage-format changes instead of unrelated proof-program updates.

crates/agglayer-grpc-types/src/compat/v1/certificate.rs
```diff
@@ -7,24 +11,108 @@ impl TryFrom<v1::Certificate> for Certificate {
...
+                .map_err(|e: Error| e.inside_field("aggchain_data"))?;
+
+                let AggchainData::Generic {
+                    proof,
+                    aggchain_params,
+                    signature,
+                    public_values,
+                } = parsed
+                else {
+                    unreachable!("interop generic conversion returned a non-generic variant")
```
Ekleog-Polygon, line 43: This `unreachable!` feels like type-plumbing noise more than a useful invariant check: we just wrapped a `Generic` variant, delegated to interop, and then immediately unwrap it again. Could we restructure this so the delegated path returns the concrete generic payload directly, instead of introducing a panic-only arm? That would make this bridge code a bit easier to read and would remove the duplicated `unreachable!` pattern in both directions.

Ekleog-Polygon (changes requested):
Requesting changes for one blocker in this range: multisig ingress validation is still knowingly relaxed until `agg.layer/interop#214`, and I do not want us to merge with that gap still open or with the regression test left ignored.

Additionally, the storage encoding snapshots are still coupled to `pessimistic_proof::ELF`, which makes backward-compat diffs much noisier than they need to be.

I also left one minor follow-up nit on the new `unreachable!` round-trip in the gRPC compat bridge.

PR #1464, opened 2026-03-27
chore: backport SP1 v6 and interop 0.14 to release/0.5

Freyskeyd:
Backport the SP1 v6 migration onto `release/0.5` so the release line matches the newer provers revision and rv64 proof artifacts already validated on the mainline SP1 upgrade branch.

This updates Agglayer proving and verification paths to SP1 v6, bumps the pessimistic-proof program selector to `10.0.0`, refreshes ELF artifacts and cycle tracker snapshots, and carries the worktree-safe `[workspace]` fix needed for standalone zkVM program builds on this maintenance branch.

Verification run on this branch:
- `cargo check --workspace --locked`
- `cargo nextest run -p agglayer-aggregator-notifier`
- `cargo nextest run -p pessimistic-proof-test-suite`

Known issue: `cargo nextest run -p agglayer-storage types::certificate::tests::regressions::case_1_regression_01` still fails with `assertion failed: n < P`, and the same failure reproduces unchanged on `origin/session/upgrade-sp1v6`.

BREAKING-CHANGE: Agglayer proving on `release/0.5` now relies on SP1 v6 APIs and rv64 ELF artifacts, and the pessimistic-proof program selector moves from `0x00000009` to `0x0000000a`.


Ekleog-Polygon (changes requested):
I think we agreed on not actually backporting, this can probably be closed?

PR #1472, opened 2026-04-01
feat(storage): list settlement attempts by job id

### LINKED ISSUE #1312: Storage function to list recorded settlement attempts for a given settlement job

Freyskeyd:
The settlement task needs to reload every recorded attempt and result
for a job, but storage only exposed point lookups. Add prefix-scoped
listing APIs, iterator seek support, and a matching results prefix
extractor so callers can restore settlement history efficiently.

Closes #1312


crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -66,6 +66,68 @@ impl SettlementReader for StateStore {
             .map(SettlementJobResult::try_from)
             .transpose()?)
     }
+
+    fn list_settlement_attempts(
+        &self,
+        settlement_job_id: &Ulid,
+    ) -> Result<Vec<(u64, SettlementAttempt)>, Error> {
+        let mut iterator = self.db.iter_with_direction::<SettlementAttemptsColumn>(
```
Ekleog-Polygon, line 74: P2 followup: Since this column family now has a fixed 16-byte prefix extractor, could we use RocksDB's `prefix_iterator_cf` here instead of adding a generic `ColumnIterator::seek()` and manually stopping once `settlement_job_id` changes? `prefix_iterator_cf` already uses `prefix_same_as_start`, so it would let RocksDB enforce the prefix boundary for us and keep the settlement-history scan logic simpler. The same applies to `list_settlement_attempt_results` below.

Ekleog-Polygon (dismissed):
Looks good to me overall.

The storage changes match the goal of restoring settlement history by job id, and the new tests cover the main success and isolation paths for both attempts and attempt results.

One follow-up suggestion inline: since these column families now have fixed prefix extractors, it may be simpler to use RocksDB's native prefix iterator directly instead of extending our wrapper with manual seek/boundary handling.

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Ekleog-Polygon: I'll handle this as a follow-up.

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -66,6 +66,68 @@ impl SettlementReader for StateStore {
...
+            ReadOptions::default(),
```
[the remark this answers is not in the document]
Ekleog-Polygon: This will change in follow-ups, already tracked in #1539

Ekleog-Polygon (approved):

hadjiszs (approved):

PR #1474, opened 2026-04-02
feat(settlement): persist settlement task results

Freyskeyd:
Wire settlement tasks to a settlement writer so terminal job and attempt outcomes are recorded as they are observed.

Keep the change scoped to result persistence and cover the write path with focused task-level tests.


Ekleog-Polygon:
Superseded by https://github.com/agglayer/agglayer/pull/1549

PR #1477, opened 2026-04-03
test: clean up phase 0 test infrastructure

Freyskeyd:
This packages the Phase 0 testing cleanup work by documenting the repository's test conventions, centralizing the notifier L1 RPC mock, and replacing sleep-based JSON-RPC shutdown waits with awaited shutdown.

It also removes dead orchestrator test scaffolding and documents ignored tests so the next coverage pass starts from a smaller, clearer base.

Verification could not be completed in this git worktree because workspace builds currently fail before reaching these crates: pessimistic-proof-program resolves workspace metadata against the parent repo instead of the worktree.


docs/TESTING.md
```diff
@@ -0,0 +1,232 @@
+# Testing guide
```
Ekleog-Polygon, line 1: P2 doc: consider moving this under `docs/knowledge-base`

Would this fit better as `docs/knowledge-base/testing.md`? Most of the human-facing how-to docs in this repo already live there, so keeping the testing guide alongside them would make the docs layout more consistent and easier to discover.

docs/TESTING.md
```diff
@@ -0,0 +1,232 @@
...
+### Unit tests
+
+Located inline (`#[cfg(test)] mod tests`) or in sibling `tests.rs` files.
+Run with:
+
+```sh
+cargo nextest run --workspace
+```
+
+Use unit tests to verify individual functions and error paths in isolation.
```
Ekleog-Polygon, line 20: P2 doc: strengthen the unit-test guidance around fuzzing

I’d be tempted to mention here that when a unit test can reasonably be expressed as a property test, it should usually be a `bolero` fuzzer instead of a growing pile of hardcoded fixtures. In practice that often just means deriving `Arbitrary`, and it tends to cover more cases with less test code.

docs/TESTING.md
```diff
@@ -0,0 +1,232 @@
...
+exercising end-to-end flows such as certificate settlement and storage backups.
+They run serially (one at a time) due to shared resource constraints.
+
+### End-to-end tests
+
+Managed externally via Kurtosis CDK.
+Triggered in CI on merge queue or manual dispatch.
+See `docs/knowledge-base/e2e-tests.md` for setup instructions.
+
+## Test runner: nextest
```
Ekleog-Polygon, line 42: P3 doc: maybe mention the `blast-radius` helper as a follow-up

As a follow-up, it might be useful to call out the repo’s `cargo make blast-radius` helper here as a lightweight first pass for scoping local verification before running the full suites. If that becomes part of the recommended workflow, it may also be worth checking whether its recommendations should account for doc-test coverage too.

docs/TESTING.md
```diff
@@ -0,0 +1,232 @@
...
+# In the providing crate's Cargo.toml:
+[features]
+testutils = []
+
+# In the consuming crate's Cargo.toml:
+[dev-dependencies]
+agglayer-storage = { path = "../agglayer-storage", features = ["testutils"] }
+```
+
+Crates that currently expose `testutils`:
```
Ekleog-Polygon, line 154: P2 doc: I’m not sure the explicit `testutils` inventory will stay accurate

The guidance around the pattern already makes sense on its own. Listing every participating crate here may create churn as crates add or remove helpers, and the list will be easy to forget to update.

Ekleog-Polygon (approved):
This cleanup direction makes sense overall: centralizing the notifier L1 RPC mock, replacing sleep-based shutdown waits with explicit shutdown, trimming dead scaffolding, and documenting ignored tests all make the test surface easier to reason about.

I left a few non-blocking documentation suggestions inline. Separately, CI is currently red, so this will still need a follow-up push before merge.

PR #1513, opened 2026-04-21
ci(contracts): update ABI release to 12.2.3

### LINKED ISSUE #1450: Upgrade to the latest `agglayer-contracts` ABI

Freyskeyd:
Refresh the bundled contract ABIs and contract download flow to use the
12.2.3 agglayer-contracts release as the single source of truth.

Also load the integration L1 image through cargo-make and add an explicit
standalone workspace to pessimistic-proof-program so nested worktrees
keep building correctly.

closes #1450


scripts/make/Makefile.contracts.toml
```diff
@@ -16,27 +17,27 @@ dependencies = [
 ]
 
 [tasks.contracts-download-AggLayerGateway]
-env = { "CONTRACT" = "AggLayerGateway.json" }
+env = { "CONTRACT_SOURCE" = "AgglayerGateway.json", "CONTRACT_OUTPUT" = "AggLayerGateway.json" }
```
Ekleog-Polygon: P3 question: `AggLayerGateway.json` is downloaded and maintained here, but no `sol!()` macro in `contracts.rs` references it -- the bindings only consume `AggchainBase.json`, `PolygonRollupManager.json`, `PolygonZkEVM.json`, and `PolygonZkEVMGlobalExitRootV2.json`. This also makes the `CONTRACT_SOURCE`/`CONTRACT_OUTPUT` rename mapping (to handle the upstream `AgglayerGateway.json` → local `AggLayerGateway.json` casing) dead code in practice.

Is this file deliberately kept for upcoming use (e.g., a future `sol!()` binding for the gateway contract), or could the download task and the file be removed to reduce maintenance surface?

Ekleog-Polygon (dismissed):
**LGTM.** The CI refactoring, cargo-make integration, and workspace fix all look correct. Removed ABI entries were verified against Rust bindings -- no breakage.

Note: I did not review the exact JSON content of the five updated ABI artifacts line-by-line; I trust they match the 12.2.3 agglayer-contracts release.

scripts/make/Makefile.contracts.toml
Freyskeyd: I'm not sure, after some research the module for `agglayer_gateway` has been removed in this [PR](https://github.com/agglayer/agglayer/pull/1434/changes#diff-a976204a359e40d64064d3a31f03dc3300250bac31faa5b8f6aabb995ffa2219L16), maybe we can remove the json file too then

hadjiszs (dismissed):

Issue #1508, opened 2026-04-21
Storage refactor: opaque envelope, migration, and v5 rejection


Freyskeyd:
Part of #1451

## Summary

Decouple agglayer persistence from SP1 types. Introduce a versioned opaque
storage envelope and a one-time migration step. Still uses the SP1 v5 SDK
internally — no SP1 version bump here.

## Scope

- New `agglayer-sp1` thin crate owning all typed SP1 SDK usage (extension
  traits, version detection, `Proof` ↔ `StorageProof` conversions with
  `ProofMode` extraction).
- **Storage proof envelope (addresses #1489)**:
  - Define `StorageProof` proto message with `proof_system`, `version`,
    `mode` (Compressed/Plonk/Groth16), opaque `proof` + `public_values`
    bytes.
  - Switch `ProofPerCertificateColumn` / `ProofPerIndexColumn` to the
    storage-local envelope.
  - Runtime callers continue to consume `pessimistic_proof::Proof` only
    via explicit conversion from the storage type — the active runtime
    proof stays typed.
  - Remove ad-hoc legacy decode fallbacks; compatibility logic is
    centralized in the storage proof type.
- Storage proto schema in `proto/agglayer/storage/` covering both the
  certificate envelope and `StorageProof` (previous bullet).
- New RocksDB column families with proto-only codecs; reader falls back to
  legacy until migration runs.
- Strip SP1 SDK deps from `agglayer-types` and `agglayer-grpc-types`.
- One-time migration step via `add_cfs()` copies v0/v1/v2 bincode rows to
  proto columns. Module-local `interop-types-v13` + `sp1-sdk-v5` for v1
  typed decode. Legacy rows untouched.
- gRPC ingress: reject v5 submissions by version string with clear
  "upgrade aggkit-prover" error.
- `catch_unwind` around `bincode::deserialize` (storage resilience).
- CI hygiene: protoc install in `quality.yml` + `cycle-tracker.yml`;
  drop unused `certificate.proto` import from `node_state.proto`.
- Unit tests against `main-preupdate-cert-v1-*.hex` fixtures.

## Acceptance criteria

- `agglayer-sp1` compiles; `agglayer-types` and `agglayer-grpc-types` have
  zero SP1 SDK deps.
- New writes go to proto columns; reads fall back to legacy when rows are
  not yet migrated.
- Migration converts v0/v1/v2 bincode rows to proto without errors; legacy
  rows untouched.
[...]

PR #1519, opened 2026-04-24
feat(storage): add proto certificate conversion boundary

Freyskeyd:
This PR adds the storage proto conversion boundary for proofs and
certificates in `agglayer-storage` without changing live column-family
wiring or migration behavior.

It introduces typed/domain <-> `agglayer.storage.v0` conversions for
proofs, certificates, aggchain data, imported bridge exits, and L1 info
tree structures. The proof conversion layer stays structural so readable
historical proofs can round-trip without mixing storage policy into
serialization.

Part of #1508.


crates/agglayer-storage/src/types/proof.rs
```diff
@@ -0,0 +1,276 @@
...
+            source,
+        })
+        .map_err(Into::into)
+}
+
+impl TryFrom<&TypedProof> for proto::Proof {
+    type Error = ProofConversionError;
+
+    fn try_from(value: &TypedProof) -> Result<Self, Self::Error> {
+        value.ensure_writable(&AcceptancePolicy::DEFAULT)?;
```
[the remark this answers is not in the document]
Freyskeyd: Fixed. This conversion boundary no longer enforces rollout policy.

The `TryFrom` impls now only reject structurally invalid proofs
(unknown version, unsupported system/mode, bad bytes), so readable historical
proofs like SP1 `v6` can roundtrip through the storage proto layer.

Added proof-level and certificate-level regression tests for the `v6` roundtrip.

hadjiszs (dismissed):

Freyskeyd:
@claude review once

proto/agglayer/storage/v0/storage_proof.proto
```diff
@@ -22,25 +22,32 @@ enum ProofMode {
...
   // Proof mode inside the selected proof system.
   ProofMode mode = 3;
 
   // Opaque serialized proof payload.
+  // Decoding requires `proof_system`, `mode`, and `version`.
   bytes proof = 4;
 
   // Opaque serialized public values.
+  // Decoding requires `proof_system`, `mode`, and `version`.
   bytes public_values = 5;
```
Ekleog-Polygon: **P1 followup: Remove unused proof public values from the storage proto**

Do we need this field in the storage proto at all? The storage path currently never generates a proof payload with `public_values` filled, and the typed storage conversion has no owner/consumer for this data. Keeping it in the persisted schema creates an attractive nuisance for future writers, because a row could contain data here even though the storage boundary does not actually use it. I think we should remove this field from the storage proto unless we have a concrete storage use case for it.

crates/agglayer-storage/src/types/proof.rs
```diff
@@ -0,0 +1,278 @@
...
+    #[error("failed to serialize SP1 verifying key bytes for version `{version}`: {source}")]
+    SerializeSp1Vkey {
+        version: String,
+        #[source]
+        source: agglayer_types::bincode::Error,
+    },
+}
+
+fn serialize_sp1_proof<T: serde::Serialize>(
+    proof: &T,
```
Ekleog-Polygon, line 33: **P1 followup: Check that the version matches the typed SP1 payload**

The `version` argument in `serialize_sp1_*` / `deserialize_sp1_*` is currently only used for diagnostics; the actual serialization format is selected by the typed `T`. That leaves room to accidentally serialize or deserialize a payload with metadata claiming a different SP1 version than the concrete type expects. Can we add a check that the parsed version matches the typed payload family, or split these helpers by version-specific type, so storage cannot emit bytes tagged with the wrong format?

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -22,15 +22,638 @@ use std::borrow::Cow;
...
+            value: value.0.to_vec().into(),
+        }
+    }
+}
+
+impl TryFrom<proto::FixedBytes32> for LocalExitRoot {
+    type Error = CertificateConversionError;
+
+    fn try_from(value: proto::FixedBytes32) -> Result<Self, Self::Error> {
+        Ok(Self::from(Digest::try_from(value)?))
```
Ekleog-Polygon: **P1 followup: Keep strong domain types wrapped in the storage proto**

Right now several strong domain types collapse directly to the same `FixedBytes32` storage shape, e.g. `Metadata`, `LocalExitRoot`, digests, and roots all share the same protobuf representation. Could we wrap these in domain-specific proto messages like `Metadata { FixedBytes32 value = 1; }` and `LocalExitRoot { FixedBytes32 value = 1; }`? It costs a few extra bytes, but gives us a cleaner persisted schema and room to evolve these concepts independently later.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -22,15 +22,638 @@ use std::borrow::Cow;
...
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
-use crate::schema::{bincode_codec, CodecError};
+use crate::{
+    schema::{bincode_codec, CodecError},
+    types::{generated::agglayer::storage::v0 as proto, proof::ProofConversionError},
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum CertificateConversionError {
```
Ekleog-Polygon, line 43: **P2 followup: Consider sharing proto building blocks with the gRPC boundary**

This conversion layer looks like it copies a fair amount of the gRPC/interoperability proto surface into storage-specific protos. Would it make sense to extract a small protobuf library of shared bridge/certificate primitives that both gRPC and storage can reuse? That would reduce drift between API and storage representations while still letting storage own its envelope/versioning.

crates/agglayer-storage/src/types/proof.rs
```diff
@@ -0,0 +1,278 @@
...
+fn panic_bincode_error() -> agglayer_types::bincode::Error {
+    Box::new(agglayer_types::bincode::ErrorKind::Custom(String::from(
+        "panic during deserialization",
+    )))
+}
+
+fn deserialize_sp1_proof<T: serde::de::DeserializeOwned>(
+    bytes: &[u8],
+    version: &str,
+) -> Result<T, ProofConversionError> {
```
Ekleog-Polygon, line 69: **P2 nit: Reuse the SP1 panic wrapper pattern consistently**

We already have the `sp1_fast` pattern for wrapping SP1 calls that may panic. Could we use that here, or move the same helper somewhere reusable, instead of open-coding `catch_unwind`? Also, this currently only wraps deserialization; proof and vkey serialization should get the same treatment, especially since we have already seen a panic around vkey serialization before.

crates/agglayer-storage/src/types/proof.rs
```diff
@@ -0,0 +1,278 @@
+use agglayer_sp1::{version_kind, ProofError, ProofExt};
+use agglayer_types::aggchain_proof::{Proof as TypedProof, SP1StarkWithContext};
+
+use crate::{schema::bincode_codec, types::generated::agglayer::storage::v0 as proto};
+
+#[derive(Debug, thiserror::Error)]
+pub enum ProofConversionError {
```
Ekleog-Polygon, line 7: **P3 followup nit: Consider using `eyre::Error` for proof conversion errors**

AFAICT production code does not match on `ProofConversionError` variants, so the structured enum mainly adds boilerplate here. Unless callers need stable variant matching, `eyre::Error` would probably be enough for this conversion boundary, with context added at the call sites that can fail.

Ekleog-Polygon (dismissed):
Thanks, this looks good to me as a conversion boundary. I left a few follow-up comments around proto shape and SP1 serialization safety. None of these need to block this PR from my side, but I'd like us to track them because this is becoming a persisted storage boundary.

crates/agglayer-storage/src/types/proof.rs
Freyskeyd: How would `eyre::Error` be used during test where we're asserting branch failure ?
Would we need to match on a particular field or string?

crates/agglayer-storage/src/types/certificate/mod.rs
Freyskeyd: This is intentional to have separation between gRPC and the storage.
The storage promise is "bytes on disk readable forever or via explicit migration."
Interop is independently versioned and externally consumed, coupling disk format to it would turn any interop refactor into a migration.
Wire/gRPC wants permissive/additive evolution, storage wants certainty and safety. The shared layer is `agglayer-types`, canonical Rust model both boundaries convert through.
The less dependencies the storage will have, the easier it'll be to maintain.

But, we can explore solutions to reduce the burden of conversion duplication, etc

crates/agglayer-storage/src/types/proof.rs
Freyskeyd: I think it will not be necessary in the future, the proof version will be tested before being inserted

crates/agglayer-storage/src/types/proof.rs
Freyskeyd: I would prefer not to bound `agglayer-storage` to `prover-executor`, let's see after we finish the migration if we need to tackle such refactorisation

proto/agglayer/storage/v0/proof.proto
```diff
@@ -0,0 +1,181 @@
...
+  // Proof mode inside the selected proof system.
+  ProofMode mode = 3;
+
+  // Opaque serialized proof payload.
+  // Decoding requires `proof_system`, `mode`, and `version`.
+  bytes proof = 4;
+
+  // Opaque serialized verifying key.
+  // Decoding requires `proof_system`, `mode`, and `version`.
+  bytes vkey = 5;
```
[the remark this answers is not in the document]
Freyskeyd: that's fine

proto/agglayer/storage/v0/certificate.proto
```diff
@@ -79,19 +57,19 @@ message BridgeExit {
 // A Merkle proof.
 message MerkleProof {
   // Root of the Merkle tree.
-  FixedBytes32 root = 1;
+  MerkleRoot root = 1;
   // Path from the leaf to the root.
-  repeated FixedBytes32 siblings = 2;
+  repeated MerkleRoot siblings = 2;
```
hadjiszs: nit: a bit weird naming tbh to see the siblings of a merkle path named as multiple `MerkleRoot`, if the concern is the usage of raw `FixedBytes32`, I'm fine having a `Digest` or `Hash` data type maybe as a wrapper? and the `root` field right above could be of this `Digest` or `Hash` type as well?

hadjiszs (dismissed):

hadjiszs (approved):

crates/agglayer-storage/src/types/proof.rs
Freyskeyd: let's do this in a followup PR if needed :)

PR #1523, opened 2026-04-28
feat(storage): migrate certificate stores to proto-backed CFs

Freyskeyd:
Switch agglayer-storage certificate-bearing stores to proto-backed column
families with reopen-time backfill from legacy bincode rows.

This keeps pending, per-epoch, and debug certificate data readable during
the migration while making runtime reads and writes use the proto-backed
families. It also co-locates legacy and proto column definitions, updates
certificate serialization snapshots, and adds coverage for migration and
readonly access behavior.


crates/agglayer-storage/src/stores/per_epoch/cf_definitions.rs
```diff
@@ -1,17 +1,32 @@
...
+    ColumnDescriptor::new::<PerEpochMetadataColumn>(),
+    ColumnDescriptor::new::<ProofPerIndexColumn>(),
+    ColumnDescriptor::new::<StartCheckpointColumn>(),
+    ColumnDescriptor::new::<EndCheckpointColumn>(),
+];
+
 /// Definitions for the column families in the epochs storage.
 pub const EPOCHS_DB: &[ColumnDescriptor] = &[
     ColumnDescriptor::new::<CertificatePerIndexColumn>(),
+    ColumnDescriptor::new::<CertificatePerIndexProtoColumn>(),
```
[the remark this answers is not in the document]
Freyskeyd: This path intentionally assumes the epoch DB has already been migrated by a
normal open.
`try_open_readonly()` is only expected to operate on the current schema, and
we do not want readonly access to carry its own legacy fallback path. The
compatibility boundary for this PR is the migration step itself: once an epoch
DB is opened through `init_db()`, legacy rows are backfilled into the proto CF,
and runtime access stays on the proto-backed layout.
So this is an operational invariant rather than a missing fallback.

crates/agglayer-storage/src/stores/per_epoch/mod.rs
```diff
@@ -175,14 +182,22 @@ impl<PendingStore, StateStore> PerEpochStore<PendingStore, StateStore> {
         } else {
             // For read-write access, calculate the next index from existing certificates
             if let Some(Ok((index, _))) = db
-                .iter_with_direction::<CertificatePerIndexColumn>(
+                .iter_with_direction::<CertificatePerIndexProtoColumn>(
                     ReadOptions::default(),
                     rocksdb::Direction::Reverse,
                 )?
```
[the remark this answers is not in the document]
Freyskeyd: This branch treats the migration step as authoritative.
Before the store starts serving writes, `init_db()` runs the backfill from the
legacy bincode CF into the proto CF. After that point, runtime logic uses the
proto-backed CF as the source of truth and does not continue consulting legacy
rows.
Under that invariant, deriving the next index from the proto CF alone is
intentional: any legacy-only rows should already have been copied during
migration, so there is no mixed-state index source to merge at runtime.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -411,16 +409,26 @@ where
...
+        Ok(())
     }
 
     fn decode(bytes: &[u8]) -> Result<Self, CodecError> {
         match bytes.first().copied() {
             None => Err(CodecError::CertificateEmpty),
             Some(0) => decode::<CertificateV0>(bytes),
             Some(1) => decode::<CertificateV1>(bytes),
-            Some(version) => Err(CodecError::BadCertificateVersion { version }),
+            Some(_) => {
```
Ekleog-Polygon, line 488: P0 issue: use explicit legacy/proto certificate codecs

`Certificate::decode()` now guesses the format from the first byte: `0`/`1` means legacy bincode, everything else means protobuf. That makes the correctness of proto decoding depend on an implicit “protobuf bytes never start with 0 or 1” assumption, and the same global `Certificate` codec is also used by both legacy and proto CF types. Please split this into explicit value codecs/types: legacy certificate CFs should decode bincode only and proto certificate CFs should encode/decode protobuf only. That makes the migration boundary unambiguous and avoids format sniffing for persisted certificate data.

crates/agglayer-storage/src/stores/debug/mod.rs
```diff
@@ -20,7 +27,12 @@ pub struct EnabledDebugStore {
 
 impl DebugStore {
     pub fn init_db(path: &Path) -> Result<DB, crate::storage::DBOpenError> {
-        DB::open_cf(path, cf_definitions::DEBUG_DB)
+        DB::builder(path, cf_definitions::DEBUG_DB_V0)?
+            .add_cfs(
+                cf_definitions::DEBUG_CERTIFICATE_PROTO_CFS,
+                backfill_debug_certificates_proto_from_legacy_bincode,
+            )?
+            .finalize(cf_definitions::DEBUG_DB)
```
Ekleog-Polygon, line 36: P1 issue: drop the legacy debug CF after backfill

Once the debug certificates are copied into the proto CF, the old bincode CF should be dropped as part of the same migration chain and the finalized schema should only contain the proto-backed runtime CF. Keeping both CFs around makes the rollback story look easier than it is, leaves two possible sources of truth, and is why `DebugReader` now has to try both sides. Since the migration framework drops half-written target CFs before retrying a step, this should be a straight `add_cfs(...).drop_cfs(...)` migration with reads/writes only using the proto CF afterward.

crates/agglayer-storage/src/stores/pending/mod.rs
```diff
@@ -29,7 +32,12 @@ pub struct PendingStore {
 
 impl PendingStore {
     pub fn init_db(path: &Path) -> Result<DB, crate::storage::DBOpenError> {
-        DB::open_cf(path, cf_definitions::PENDING_DB)
+        DB::builder(path, cf_definitions::PENDING_DB_V0)?
+            .add_cfs(
+                cf_definitions::PENDING_CERTIFICATE_PROTO_CFS,
+                backfill_pending_certificates_proto_from_legacy_bincode,
+            )?
+            .finalize(cf_definitions::PENDING_DB)
```
Ekleog-Polygon, line 41: P1 issue: drop the legacy pending queue CF after backfill

After pending certificates are copied into the proto CF, the legacy pending queue CF should be dropped and removed from the finalized runtime schema. Keeping `PendingQueueColumn` in `PENDING_DB` preserves an extra source of truth that runtime code intentionally no longer reads, and it makes future changes more likely to accidentally interact with stale data. Please chain a `drop_cfs` for the migrated pending queue CF after the backfill and finalize with only the proto-backed queue CF plus the unrelated metadata/proof CFs.

crates/agglayer-storage/src/stores/per_epoch/mod.rs
```diff
@@ -50,7 +52,12 @@ pub struct PerEpochStore<PendingStore, StateStore> {
 
 impl<PendingStore, StateStore> PerEpochStore<PendingStore, StateStore> {
     pub fn init_db(path: &std::path::Path) -> Result<DB, crate::storage::DBOpenError> {
-        DB::open_cf(path, cf_definitions::EPOCHS_DB)
+        DB::builder(path, cf_definitions::EPOCHS_DB_V0)?
+            .add_cfs(
+                cf_definitions::EPOCH_CERTIFICATE_PROTO_CFS,
+                backfill_epoch_certificates_proto_from_legacy_bincode,
+            )?
+            .finalize(cf_definitions::EPOCHS_DB)
```
Ekleog-Polygon, line 63: P1 issue: drop the legacy epoch certificate CF after backfill

After `CertificatePerIndexColumn` is backfilled into `CertificatePerIndexProtoColumn`, the legacy certificate CF should be dropped and removed from the finalized epoch schema. Keeping both CFs around adds a second certificate source for each epoch and keeps runtime code coupled to legacy state that should only exist during the migration step. Please chain a `drop_cfs` for the old certificate CF and finalize the epoch DB with the proto certificate CF plus the unchanged metadata/proof/checkpoint CFs.

crates/agglayer-storage/src/stores/per_epoch/mod.rs
```diff
@@ -239,6 +254,27 @@ impl<PendingStore, StateStore> PerEpochStore<PendingStore, StateStore> {
...
+/// bincode epoch certificate CF to the proto-backed CF.
+///
+/// This keeps the original CF untouched while populating the new one so we can
+/// finish validation before removing the legacy family in a later cleanup.
+fn backfill_epoch_certificates_proto_from_legacy_bincode(
+    db: &crate::storage::DbAccess,
+) -> Result<(), crate::storage::DBMigrationErrorDetails> {
+    let keys = db
+        .keys::<CertificatePerIndexColumn>()?
+        .collect::<Result<Vec<_>, _>>()?;
```
Ekleog-Polygon: P1 issue: stream migration rows instead of buffering every key

This copies the whole legacy keyspace into a `Vec` before writing any proto rows, so migration memory grows with the number of certificates in an epoch. The migration framework already recreates the target CF when retrying a step, so we should be able to iterate the legacy side and write each decoded certificate to the proto CF as we go, or use bounded batches if RocksDB iterator invalidation is a concern. Please keep the migration memory bounded here, and apply the same pattern to the pending and debug certificate backfills.

crates/agglayer-storage/src/stores/debug/cf_definitions.rs
```diff
@@ -1,4 +1,16 @@
...
+pub const DEBUG_DB_V0: &[ColumnDescriptor] = &[ColumnDescriptor::new::<DebugCertificatesColumn>()];
 
 /// Definitions for the column families in the debug storage.
-pub const DEBUG_DB: &[ColumnDescriptor] = &[ColumnDescriptor::new::<DebugCertificatesColumn>()];
+pub const DEBUG_DB: &[ColumnDescriptor] = &[
+    ColumnDescriptor::new::<DebugCertificatesColumn>(),
+    ColumnDescriptor::new::<DebugCertificatesProtoColumn>(),
+];
+
+pub const DEBUG_CERTIFICATE_PROTO_CFS: &[ColumnDescriptor] =
```
Ekleog-Polygon: P1 issue: make migration step CF lists immutable by construction

This name describes the current target schema shape, not a specific migration step. If a later PR migrates another debug CF to protobuf, it would be natural to append that CF here and update the backfill function, but that would silently rewrite an already-recorded migration step for existing nodes. Please either inline the one-element slice at the `add_cfs` call site or give it a step-specific name that must not be edited, for example `DEBUG_DB_MIGRATION_1_ADD_CERTIFICATE_PROTO_CFS`, with a comment saying future migrations need a new migration step and a new constant.

crates/agglayer-storage/src/stores/pending/cf_definitions.rs
```diff
@@ -2,15 +2,28 @@ use crate::{
...
+/// Definitions for the pending queue storage.
 pub const PENDING_DB: &[ColumnDescriptor] = &[
     ColumnDescriptor::new::<LatestProvenCertificatePerNetworkColumn>(),
     ColumnDescriptor::new::<LatestPendingCertificatePerNetworkColumn>(),
     ColumnDescriptor::new::<PendingQueueColumn>(),
+    ColumnDescriptor::new::<PendingQueueProtoColumn>(),
     ColumnDescriptor::new::<ProofPerCertificateColumn>(),
 ];
+
+pub const PENDING_CERTIFICATE_PROTO_CFS: &[ColumnDescriptor] =
```
Ekleog-Polygon: P1 issue: make migration step CF lists immutable by construction

This name describes the current target schema shape, not a specific migration step. If a later PR migrates another pending CF to protobuf, it would be natural to append that CF here and update the backfill function, but that would silently rewrite an already-recorded migration step for existing nodes. Please either inline the one-element slice at the `add_cfs` call site or give it a step-specific name that must not be edited, for example `PENDING_DB_MIGRATION_1_ADD_PENDING_QUEUE_PROTO_CFS`, with a comment saying future migrations need a new migration step and a new constant.

crates/agglayer-storage/src/stores/per_epoch/cf_definitions.rs
```diff
@@ -1,17 +1,32 @@
...
     ColumnDescriptor::new::<PerEpochMetadataColumn>(),
     ColumnDescriptor::new::<ProofPerIndexColumn>(),
     ColumnDescriptor::new::<StartCheckpointColumn>(),
     ColumnDescriptor::new::<EndCheckpointColumn>(),
 ];
+
+pub const EPOCH_CERTIFICATE_PROTO_CFS: &[ColumnDescriptor] =
```
Ekleog-Polygon: P1 issue: make migration step CF lists immutable by construction

This name describes the current target schema shape, not a specific migration step. If a later PR migrates another epoch CF to protobuf, it would be natural to append that CF here and update the backfill function, but that would silently rewrite an already-recorded migration step for existing nodes. Please either inline the one-element slice at the `add_cfs` call site or give it a step-specific name that must not be edited, for example `EPOCH_DB_MIGRATION_1_ADD_CERTIFICATE_PROTO_CFS`, with a comment saying future migrations need a new migration step and a new constant.

crates/agglayer-storage/src/stores/per_epoch/cf_definitions.rs
Ekleog-Polygon, line 24: P1 issue: readonly opens should not bypass required migrations

Readonly epoch access currently requires the final schema but cannot create the proto CF or backfill legacy rows. On upgrade, older historical epoch DBs may not have been opened by the normal write path before RPC reads them, so `get_certificate`/`get_proof` can fail on legacy-only epoch DBs. I think every database open path should run the declared migrations first; the cost is that a readonly open may need a read-write migration phase, then reopen read-only afterward. That seems preferable to keeping long-term readonly legacy compatibility, which would add significant tech debt. Operators should have external DB backups that `agglayer-node` cannot mutate anyway, so acknowledging that past epoch data must be migrated seems like the cleaner tradeoff.

crates/agglayer-storage/src/stores/debug/mod.rs
```diff
@@ -33,13 +45,41 @@ impl DebugStore {
...
+}
+
 impl DebugReader for DebugStore {
     fn get_certificate(
         &self,
         certificate_id: &CertificateId,
     ) -> Result<Option<Certificate>, Error> {
         match self {
             DebugStore::Enabled(store) => {
+                if let Some(certificate) = store
```
Ekleog-Polygon: P1 issue: do not keep legacy CF fallback in runtime reads

After this migration, debug certificate reads should have a single source of truth: the proto CF. Keeping a fallback to `DebugCertificatesColumn` makes it look like mixed legacy/proto runtime state is supported, but the intended model is to migrate once and then drop the old CF. Please remove this fallback and rely on the migration step to populate the proto CF before the store is used.

crates/agglayer-storage/src/stores/pending/mod.rs
```diff
@@ -29,7 +32,12 @@ pub struct PendingStore {
 
 impl PendingStore {
     pub fn init_db(path: &Path) -> Result<DB, crate::storage::DBOpenError> {
-        DB::open_cf(path, cf_definitions::PENDING_DB)
+        DB::builder(path, cf_definitions::PENDING_DB_V0)?
+            .add_cfs(
```
Ekleog-Polygon, line 37: P2 question: should migration ordering live in one documented place?

What do you think about centralizing the migration chains in a dedicated `migrations.rs` file? The per-store modules could still own the migration function bodies, but the ordered list of migration steps would live in one heavily documented place that explains the invariants: never edit a previously released migration step, always append a new step, when it is valid to drop old CFs, and how readonly opens should handle migration. That would make future schema changes harder to get wrong and keep the “migration history” separate from the current runtime store implementation.

Ekleog-Polygon (dismissed):
Thanks for pushing this migration forward. I think this needs another pass before merge, mostly around making the migration boundary explicit and operationally safe.

The blocker is that the legacy and proto certificate CFs should not share one sniffing `Certificate` codec. We should have explicit legacy-bincode and proto value codecs/types, then migrate from one to the other. That also lets us drop the old CFs after backfill and avoid keeping mixed runtime read paths around.

A few other migration details seem important before this lands: keep migration memory bounded instead of buffering all keys, make migration step constants/names clearly immutable, ensure readonly epoch access can run required migrations before reopening readonly, and document the manual stop-backup-restart upgrade flow until the draft PR/work for automated pre-migration backups exists.

I skimmed the tests, but I did not do a full test review. Iif you want a deeper pass on the tests too, please say so.

crates/agglayer-storage/src/stores/debug/mod.rs
Freyskeyd: We're not removing the old column for safety, first time we're making this migration system working...

crates/agglayer-storage/src/stores/pending/mod.rs
Freyskeyd: We're not removing the old column for safety, first time we're making this migration system working...

crates/agglayer-storage/src/stores/per_epoch/mod.rs
Freyskeyd: Same as above

crates/agglayer-storage/src/stores/pending/mod.rs
Freyskeyd: That's something we can explore in a followup ticket.

crates/agglayer-storage/src/stores/per_epoch/cf_definitions.rs
Freyskeyd: I disagree, a readonly store, shouldn't have any write mechanism. If the readonly store fails due to unapplied migration, the problem is elsewhere

crates/agglayer-storage/src/stores/per_epoch/mod.rs
```diff
@@ -173,9 +181,12 @@ impl<PendingStore, StateStore> PerEpochStore<PendingStore, StateStore> {
             // For readonly access, we don't need to track the next index
             AtomicU64::new(0)
         } else {
-            // For read-write access, calculate the next index from existing certificates
+            // For read-write access, calculate the next index from existing certificates.
+            // The proto-backed CF is the single source of truth at runtime: legacy rows
+            // are backfilled into it during `init_db`, so consulting the legacy CF here
+            // would only re-read data that has already been migrated.
             if let Some(Ok((index, _))) = db
```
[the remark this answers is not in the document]
Freyskeyd: Correcting myself: 954008ed has been reverted in c3da2961. The `max(legacy, proto)` guard was patching the symptom at the wrong layer — it would have let the binary keep running indefinitely with an incomplete proto CF, silently undermining the contract that after a successful migration the proto CF is authoritative and the legacy CF can be ignored.

The right path is upstream:
- Surface the unparsable rows in the migration report so the operator sees them all at once (improved reporting follow-up on this PR).
- Add a `storage-doctor` CLI step in #1531 that lets the operator list those rows (and purge them in a follow-up to #1531) before re-running migration.

After a clean re-run, the proto CF is complete and `next_certificate_index` consulting only it is correct.

Sorry for the back-and-forth on this thread.

crates/agglayer-storage/src/types/certificate/mod.rs
Freyskeyd: Done in 421e6005.

The split is now:
- `Certificate` — strictly proto on encode and decode (`mod.rs:429-444`). Used by the new proto CFs only.
- `LegacyCertificate` — newtype wrapper with its own codec (`mod.rs:467-499`), only attached to the legacy CFs (`CertificatePerIndexColumn`, `PendingProofColumn`, `DebugCertificatesColumn`).

`LegacyCertificate` is intentionally permissive about format because the legacy CFs are not bincode-only on disk in production: between #1519 (which switched the global `Certificate` codec's default encode to proto) and this PR (which moves runtime writes off the legacy CFs), the legacy CFs received both bincode v0/v1 rows from older binaries and proto rows from post-#1519 binaries. A bincode-only legacy decoder would refuse to read all proto rows from that window — so this isn't format sniffing on a clean dataset, it's matching the historical reality of what's on disk.

The first-byte heuristic for picking between bincode and proto is sound for this dataset:
- bincode v0/v1 always starts with the version byte (0 or 1)
- proto can never start with 0 or 1 — proto wire format encodes a tag byte as `(field_number << 3) | wire_type`, so the smallest valid tag byte is `(1 << 3) | 0 = 8`. Empty messages encode to zero bytes (caught explicitly as `CodecError::CertificateEmpty`).

So the namespaces don't overlap by construction. The doc comment on `LegacyCertificate` records this as transitional — the type and the legacy CFs come out together in the follow-up cleanup once the migration is validated in production.

crates/agglayer-storage/src/stores/per_epoch/cf_definitions.rs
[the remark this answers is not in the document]
Freyskeyd: This is by design and consistent with the resolved P1 sibling [r3165479549](https://github.com/agglayer/agglayer/pull/1523#discussion_r3165479549) (same concern, raised by @Ekleog-Polygon). The agreed model on this PR is: readonly opens require the schema to already be at head; they don't run migrations. After this PR, runtime cert/proof reads only consult the proto CF, the resolved P1 [r3165479555](https://github.com/agglayer/agglayer/pull/1523#discussion_r3165479555) explicitly removed the legacy fallback in reads, so the proto CF must exist before readonly RPC against historical epochs works.

The deployment path for that is the new `migrate-storage` CLI subcommand stacked in #1531. Operators run it once during the maintenance window before promoting the new binary; it calls `init_db` against every store and every epoch DB, populating the proto CFs in place. After that, readonly RPC works as before.

This failure mode is captured in the test `readonly_epoch_access_requires_migrated_proto_cf` (`per_epoch/tests.rs:230-249`), it documents the expectation, not a bug.

Adding a "tolerant readonly open" that silently degrades to legacy CFs would re-introduce the fallback that r3165479555 explicitly removed, and would let the binary serve historical RPC from unmigrated epoch DBs without forcing the migration to run, hiding the operational gap rather than surfacing it. The hard failure here is intentional so the migration requirement is loud.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -55,7 +55,17 @@ pub struct StateStore {
...
-        DB::open_cf(path, cf_definitions::STATE_DB)
+        // V0 is the eight-CF schema that pre-dates `disabled_networks_cf`
+        // and the settlement family. Legacy production snapshots still
+        // have only those CFs; running the current binary against them
+        // would fail the migration framework's schema gate without this
+        // path. `ensure_cfs` is idempotent: passing the full `STATE_DB`
+        // list creates whatever is missing on disk and is a no-op when
+        // every CF is already present, so legacy V0 DBs converge to the
+        // current schema and existing post-V0 DBs are unaffected.
+        DB::builder(path, cf_definitions::STATE_DB_V0)?
```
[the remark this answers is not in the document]
Freyskeyd: The specific upgrade scenario described — state DB with all 14 STATE_DB CFs and no `migration_record_cf` — only exists for binaries that adopted the post-`disabled_networks_cf` / post-settlement-CFs schema while still bypassing the migration framework. Verified against the production tag chain:

- `v0.4.4` (latest stable): `state_db_cf_definitions()` returns exactly 8 CFs — the same set as `STATE_DB_V0` here. Operators upgrading from v0.4.x land in the existing third branch of `Builder::open` (`cfs_db == cfs_v0`), bootstrap the migration record, and `ensure_cfs(STATE_DB)` adds the 6 new CFs cleanly.

- `v0.5.0-rc.1` (the only tag with the full 14 CFs and no migration framework call from `StateStore::init_db`): tagged 2026-04-29, same day as PR4's schema work landed, never promoted past internal RC.

So the strict-equality gate is fine for the only deployed upgrade path (v0.4.x → PR5+). Happy to revisit if any operator runs an interim binary that produces the 14-CF / no-record layout — please flag the specific version.

proto/agglayer/storage/v0/proof.proto
```diff
@@ -0,0 +1,178 @@
...
+  // Proof mode inside the selected proof system.
+  ProofMode mode = 3;
+
+  // Opaque serialized proof payload.
+  // Decoding requires `proof_system`, `mode`, and `version`.
+  bytes proof = 4;
+
+  // Opaque serialized verifying key.
+  // Decoding requires `proof_system`, `mode`, and `version`.
+  bytes vkey = 5;
```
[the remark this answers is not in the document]
Freyskeyd: Verified against the production tag chain:

- `v0.4.4` (latest stable): Certificate codec is bincode-only (`encode_into` uses `bincode_codec().serialize_into(...)`). Operators upgrading from v0.4.x have **no** proto Certificate rows on disk; `LegacyCertificate::decode` only ever exercises the bincode v0/v1 paths for them.

- `v0.5.0-rc.1` (the only tag containing `storage_proof.proto` with `Proof.vkey = 6` / `bytes public_values = 5`): tagged 2026-04-29, same day as PR4's schema reshape (`36ed410d` "harden proof proto conversions"). Never promoted past internal RC.

So no production operator has on-disk proto Certificate rows under the old schema.

Worth noting: the vkey field-number change is a symptom of a broader wire-shape break PR4 introduced — `Proof` went from a single 6-field envelope (with `bytes public_values = 5`) to a typed wrapper (`AggchainStoredProof { Proof proof = 1; AggchainProofPublicValues public_values = 2 }`). Old single-message rows wouldn't decode cleanly as the new top-level wrapper anyway, so a vkey-only fix wouldn't restore compat.

If a `v0.5.0-rc.1` deployment exists somewhere with on-disk proto Certificate rows that must round-trip through this PR, please flag — that would need a broader fix than this thread proposes.

crates/agglayer-storage/src/diagnostics.rs
```diff
@@ -0,0 +1,330 @@
...
+        ));
+    }
+    Ok(out)
+}
+
+fn open_readonly_v0(
+    path: &Path,
+    cfs_v0: &[crate::schema::ColumnDescriptor],
+) -> Result<DB, ScanError> {
+    DB::open_cf_readonly(path, cfs_v0).map_err(|source| ScanError::OpenError {
```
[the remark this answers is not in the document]
Freyskeyd: Fixed in `15910232`.

The diagnostics path now succeeds on migrated stores because `DB::open_cf_readonly()` opens the full on-disk CF set when the DB already exists, instead of assuming the caller-provided descriptor list is complete. I also added post-migration diagnostics tests for pending/debug/epoch stores in `crates/agglayer-storage/src/diagnostics.rs` to cover that scenario explicitly.

crates/agglayer-storage/src/stores/migration_helpers.rs
```diff
@@ -0,0 +1,103 @@
...
+                error!(
+                    cf = L::COLUMN_FAMILY_NAME,
+                    label,
+                    ?key,
+                    error = %codec_error,
+                    "skipping unparsable legacy row during proto migration; this row is also \
+                     unreadable through the current runtime codec",
+                );
+                skipped += 1;
+            }
```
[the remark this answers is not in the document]
Freyskeyd: I do not think aborting the migration on `LegacyCertificate` decode failure is the right default for this rollout.

At this boundary, the skipped rows are rows already unreadable through the current legacy runtime codec: the legacy CFs are typed as `LegacyCertificate`, and the migration helper decodes through that same path before converting to the proto-backed `Certificate`. So this is not taking previously readable rows and making them newly invisible; it is preserving availability for readable rows while surfacing pre-existing unreadable corruption.

This is also no longer silent: the migration logs each skipped row at `error!`, emits a `warn!` summary, and the new diagnostics scan lists those rows so operators can inspect and clean them up before the legacy CFs are dropped in a later step. Turning that condition into a hard startup failure would convert corrupted historical rows into a full outage for the store, which is a worse operational default for this transition.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -55,7 +55,17 @@ pub struct StateStore {
...
+            .ensure_cfs(cf_definitions::STATE_DB)?
```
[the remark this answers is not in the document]
Freyskeyd: Checked the cited upgrade path, and I do not see this specific layout coming from the immediate parent build.

The parent code initializes state storage via `DB::open_cf(path, STATE_DB)`, but `DB::open_cf()` already goes through `Builder::open(path, cfs)?.finalize(cfs)` in `crates/agglayer-storage/src/storage/mod.rs`. `Builder::open()` creates/opens the migration record CF and records step 0 before returning, so that path does not produce a full-schema/no-record state DB.

I agree the `STATE_DB_V0` schema gate would reject a full-schema/no-record layout if one existed from some out-of-band initializer, but I could not verify the specific “immediately previous build” scenario described here from the repo history.

crates/agglayer-storage/src/columns/debug_certificates/mod.rs
```diff
@@ -1,19 +1,45 @@
...
-use super::{ColumnSchema, DEBUG_CERTIFICATES_CF};
+use super::{ColumnSchema, DEBUG_CERTIFICATES_CF, DEBUG_CERTIFICATES_PROTO_CF};
+use crate::types::LegacyCertificate;
 
-/// Column family containing the certificates received.
+/// Legacy column family for debug certificates.
+///
+/// Kept readable so the proto migration can backfill existing rows. The CF
+/// historically received bincode rows (pre-#1519) and proto rows
+/// (#1519..this PR), so its `Value` codec is `LegacyCertificate`, which
```
hadjiszs, line 10: nit: do we want to keep such ref to pr numbers in rustdoc?

crates/agglayer-storage/src/stores/migration_helpers.rs
```diff
@@ -0,0 +1,103 @@
...
+                // Race-y: the key was visible to the iterator but has gone
+                // by the time we try to read it. Should not happen during
+                // normal startup migrations, but log just in case.
+                debug!(
+                    cf = L::COLUMN_FAMILY_NAME,
+                    label,
+                    ?key,
+                    "legacy CF key vanished between iterate and read"
+                );
+            }
```
hadjiszs, line 71: i'm not sure to understand if this case can really happen but we still need to have `skipped += 1` in this branch as well no? and maybe having it as `error!` log

hadjiszs (approved):

crates/agglayer-storage/src/columns/debug_certificates/mod.rs
Freyskeyd: I'll create a followup [ticket](https://github.com/agglayer/agglayer/issues/1535) to merge this PR.

crates/agglayer-storage/src/stores/migration_helpers.rs
Freyskeyd: Basically it can happen if the storage is migrated while there is activity on it.
Which is unlikely but we want to log it.

hadjiszs (approved):

PR #1525, opened 2026-04-28
feat(sp1): migrate to v6 with legacy v5 compatibility

Freyskeyd:
This completes the SP1 v6 migration on top of the storage migration
stack while keeping temporary legacy v5 read compatibility for safe
certificate and proof decoding.

It switches prover wiring and test artifacts to SP1 v6, updates gRPC
ingress checks for writable proof versions, and keeps storage V1
compatibility decode-only while proto-backed writes stay on the current
format.

CONFIG-CHANGE:
- switch guest and tooling artifacts from rv32 to rv64
- update CI and Docker wiring to the SP1 v6 toolchain


crates/agglayer-sp1/src/ext.rs
```diff
@@ -31,10 +32,106 @@ pub trait ProofExt {
...
+    sp1_hypercube::SP1PcsProofInner,
+>;
+
+#[derive(Clone)]
+pub struct CurrentSp1StarkWithContext {
+    pub proof: CurrentSp1StarkProof,
+    pub vkey: SP1VerifyingKey,
+}
+
+impl std::fmt::Debug for CurrentSp1StarkWithContext {
```
Ekleog-Polygon: Nit: the custom `Debug` here could probably be replaced by `derive_more` or `educe` so we don't have to maintain a hand-rolled impl just to mask the vkey field.

crates/agglayer-sp1/src/ext.rs
```diff
@@ -31,10 +32,106 @@ pub trait ProofExt {
...
+        message.clone()
+    } else {
+        "panic while deserializing SP1 bytes".to_owned()
+    }
+}
+
+fn deserialize_current_sp1_proof(
+    sp1: &SP1StarkWithContext,
+) -> Result<CurrentSp1StarkProof, ProofError> {
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
Ekleog-Polygon, line 92: At the `catch_unwind` call sites added in this PR (`deserialize_current_sp1_proof` here, `deserialize_legacy_v5_vkey` below, and `verifying_key` further down), can we use the existing `sp1_fast` / `sp1_blocking` helpers instead? They already wrap unwind-handling consistently, and it would avoid open-coding the panic-to-error conversion in three places.

Ekleog-Polygon (commented):
Thanks for pushing this, it's a chunky migration. Most of my comments are inline. A few cross-cutting items up here:

- I skimmed the tests rather than reviewing them properly. Ping me if you want a dedicated pass.
- Main migration question: do we know whether SP1 v6 can verify v5-era aggchain proofs inside the v6 PP proof? I saw you and Monir discussed this, but I didn't find the conclusion. If it's supported, upgraded nodes should keep certifying pre-upgrade pending certificates; if not, we should make the migration behavior on those certificates explicit.
- This PR appears to depend on interop `aa61c6c54b86feae6ef47ae1bb975c17f8cb2eee` and provers `3bec61e95d5fdfe5b71c5c7a50d0b892c25d9994`, but I don't see corresponding P0 PRs pending review on those repos. Is that expected?
- Could we document the process to rebuild the `empty.elf` / `empty_v5.elf` fixtures (e.g. alongside `docs/knowledge-base/pessimistic-proof-elf.md`)? It'll save time next time we do a large SP1 upgrade.

crates/agglayer-sp1/src/ext.rs
```diff
@@ -31,10 +32,106 @@ pub trait ProofExt {
...
+    fn vkey_hash_u32(&self) -> Result<[u32; 8], ProofError>;
+
+    /// Deserialize the current SP1 verifying key carried by this proof.
+    fn verifying_key(&self) -> Result<SP1VerifyingKey, ProofError>;
+
+    /// Deserialize the current executable SP1 proof and verifying key.
+    fn executable_sp1(&self) -> Result<CurrentSp1StarkWithContext, ProofError>;
+}
+
+pub type CurrentSp1StarkProof = sp1_core_executor::SP1RecursionProof<
```
Ekleog-Polygon: Naming question: rather than calling the new format `current_*`, could we call it `v6` directly? `current` will age the moment we start the v7 migration. Same logic on the other side: the "old" format has effectively been the format since v3, so naming it `v3` (and reserving `v5` for the SP1 SDK version it's tied to) might be more durable. WDYT?

crates/agglayer-sp1/src/ext.rs
```diff
@@ -31,10 +32,106 @@ pub trait ProofExt {
...
+        }),
+        Err(payload) => Err(ProofError::DeserializeSp1Vkey {
+            version: sp1.version.clone(),
+            source: agglayer_interop_types::bincode::ErrorKind::Custom(panic_message(payload))
+                .into(),
+        }),
+    }
+}
+
+pub fn current_sp1_stark_with_context<P: Serialize>(
```
Ekleog-Polygon: None of `deserialize_current_sp1_proof`, `deserialize_legacy_v5_vkey`, or `current_sp1_stark_with_context` actually checks that the SP1 version on the wrapper matches the format they (de)serialize. It's currently fine because callers gate on `version_kind` first, but it would be safer to assert/match the version inside these helpers so a future caller can't pass a v5 envelope through the v6 path silently.

crates/agglayer-sp1/src/ext.rs
```diff
@@ -78,11 +175,55 @@ impl ProofExt for Proof {
         policy.ensure_writable(v, &sp1.version)
     }
 
-    fn vkey_hash_bytes(&self) -> [u8; 32] {
-        self.sp1().vkey.hash_bytes()
+    fn vkey_hash_bytes(&self) -> Result<[u8; 32], ProofError> {
```
Ekleog-Polygon, line 234: This looks a bit scary: `verifying_key()` deserializes with the current (v6) format unconditionally, and the `Sp1ProofVersion::V6` arm here goes through it. If anyone ever ends up calling these helpers on a v5 envelope through the v6 arm, we'll silently mis-decode a vkey. Could we either (a) make `verifying_key()` itself version-dispatched, or (b) add a `debug_assert!` that the version really is V6 before falling into the v6 branch?

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -170,24 +222,28 @@ where
             AggchainData::ECDSA { .. } => {}
             AggchainData::MultisigOnly { .. } => {}
             AggchainData::Generic { ref proof, .. } => {
-                let agglayer_types::aggchain_proof::Proof::SP1Stark(stark_proof) = proof;
+                let stark_proof = proof
+                    .executable_sp1()
```
Ekleog-Polygon: Stepping back, I don't think we should use `executable_sp1()` here at all — and more broadly, I'm not sure `ensure_executable` / `ensure_writable` (and the `AcceptancePolicy` layer behind them) actually buy us much over a direct version check. At the call sites we always know what we want (v6 today), so this can just be:

```rust
match version_kind(&proof.sp1().version)? {
    Sp1ProofVersion::V6 => { /* deserialize and use */ }
    _ => Err(...),
}
```

That removes the policy indirection, makes the v6 assumption explicit at the place that actually depends on it, and avoids the situation where `executable_sp1` quietly rejects pending v5 certificates via `AcceptancePolicy::DEFAULT`. Same argument on the writable side. Curious what you think — happy to be wrong if there's a reason to keep the policy abstraction.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -19,9 +19,11 @@
 
 use std::borrow::Cow;
```
Ekleog-Polygon, line 39: Could we keep the version number in the local alias too (e.g. `legacy_interop_types_v13`)? Otherwise, the next time we deprecate something and `legacy_interop_types` becomes the not-quite-most-legacy one, we end up with `legacy_legacy_interop_types`. Same theme as the `current_*` naming concern in `agglayer-sp1`.

crates/agglayer-storage/src/types/certificate/mod.rs
```diff
@@ -403,9 +345,14 @@ fn deserialize_bincode<T: DeserializeOwned>(bytes: &[u8]) -> Result<T, CodecErro
...
 where
-    T: DeserializeOwned + Into<Certificate>,
+    T: DeserializeOwned,
+    Certificate: TryFrom<T, Error = CodecError>,
 {
-    Ok(deserialize_bincode::<T>(bytes)?.into())
+    Certificate::try_from(deserialize_bincode::<T>(bytes)?)
+}
+
+fn decode_v0(bytes: &[u8]) -> Result<Certificate, CodecError> {
```
Ekleog-Polygon, line 372: Why does `decode_v0` get a separate path while the v1+ branches go through proto/bincode? I'd have expected v1 to be in the same shape as v0 here, or v0 to have moved to the same dispatch — having one bincode-only branch and one proto branch makes the version dispatch a bit harder to reason about.

crates/agglayer-storage/src/types/proof.rs
```diff
@@ -144,32 +144,15 @@ impl TryFrom<proto::Proof> for TypedProof {
...
 mod tests {
     use agglayer_sp1::ProofError;
-    use agglayer_types::aggchain_proof::{Proof as TypedProof, SP1StarkWithContext};
-    use sp1_sdk::Prover;
+    use agglayer_types::{
+        aggchain_proof::Proof as TypedProof, testutils::dummy_sp1_stark_proof_with_version,
+    };
 
     use super::*;
     use crate::types::generated::agglayer::storage::v0::{ProofMode, ProofSystem};
```
Ekleog-Polygon, line 69: This `mock_proof` helper just forwards to `dummy_sp1_stark_proof_with_version`; we can drop it and call the underlying helper directly in each test.

crates/agglayer-types/src/aggchain_data/aggchain_proof.rs
```diff
@@ -46,3 +49,24 @@ impl From<PayloadWithCtx<Payload, Context>> for core::AggchainProof {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::Payload;
+    use crate::{certificate::dummy_sp1_stark_proof_with_version, Digest};
+
+    #[test]
+    fn legacy_v5_vkey_hash_helpers_work() {
```
Ekleog-Polygon: This test mostly asserts that `Vec<u8>::len() == 32` and `[u32; 8]::len() == 8`, which is structurally guaranteed. To actually validate the legacy v5 path I think we'd want a binary proof fixture and a snapshot of the expected vkey hash; otherwise this test doesn't catch a regression in the v5 deserialization path.

crates/pessimistic-proof/src/error.rs
```diff
@@ -42,7 +42,25 @@ impl From<SP1VerificationError> for ProofVerificationError {
...
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{ProofVerificationError, SP1VerificationError};
+
+    #[test]
+    fn maps_unexpected_exit_code_to_other() {
```
Ekleog-Polygon: This test essentially checks the `format!()` string of one match arm. I'd be in favor of dropping it.

crates/pessimistic-proof-program/elf/riscv64im-succinct-zkvm-elf
hadjiszs, line 1: there is the elf committed in `crates/pessimistic-proof/elf/riscv64im-succinct-zkvm-elf` as well, and they aren't the same version so not sure which one you want to keep

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
Freyskeyd: It's the role of `agglayer-sp1` to know which version can be executed or not, later it will be the role of the prover-engine so that it's transparent for `notifier`

crates/agglayer-aggregator-notifier/src/certifier/mod.rs
```diff
@@ -170,24 +222,28 @@ where
...
+                    .map_err(|source| CertificationError::Other(eyre!(source)))?;
```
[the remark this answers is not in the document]
Freyskeyd: That's fine

crates/agglayer-storage/src/types/certificate/mod.rs
Freyskeyd: Renamed the local alias to `legacy_interop_types_v13` in the certificate codec and its tests so the historical interop type version is explicit.

crates/agglayer-sp1/src/ext.rs
Freyskeyd: I kept `agglayer-sp1` independent from `prover-executor`, so I did not switch this to `sp1_fast` / `sp1_blocking`. Pulling executor helpers into this crate would add layering in the wrong direction. The unwind-to-error conversion is now centralized in `deserialize_sp1_bytes`, so the three deserialization paths share the same handling without adding that dependency.

crates/agglayer-sp1/src/ext.rs
Freyskeyd: This is now guarded by `deserialize_v6_sp1_vkey()`, which calls `ensure_v6_readable()` before deserializing. A v5 envelope is rejected before the v6 vkey decode path.

crates/agglayer-storage/src/types/certificate/mod.rs
Freyskeyd: I left this split in place. `CertificateV0 -> Certificate` is still infallible, while `CertificateV1 -> Certificate` is fallible because it may convert legacy SP1 proof data. The decode dispatch comment now documents why v0 keeps the dedicated path.

hadjiszs (dismissed):

hadjiszs (approved):

PR #1528, opened 2026-04-30
feat: start moving error types to eyre

Ekleog-Polygon:
Part of #838


crates/agglayer-signer/src/lib.rs
```diff
@@ -37,15 +34,20 @@ impl ConfiguredSigner {
     pub(crate) fn local_wallet(
         chain_id: u64,
         local: &LocalConfig,
-    ) -> Result<(PrivateKeySigner, Option<PrivateKeySigner>), Error> {
-        let pk1 = local.private_keys.first().ok_or(Error::NoPk)?;
-        let signer1 = PrivateKeySigner::decrypt_keystore(&pk1.path, &pk1.password)?
+    ) -> eyre::Result<(PrivateKeySigner, Option<PrivateKeySigner>)> {
```
hadjiszs, line 36: the `#[allow(clippy::result_large_err)]` above is still useful?

hadjiszs (dismissed):

crates/agglayer-signer/src/lib.rs
Ekleog-Polygon: Removed in 753c1b2c. Since `local_wallet` now returns `eyre::Result`, the large concrete error type is gone and the `clippy::result_large_err` allow is no longer needed. Local verification passed with the blast-radius recommended checks.

hadjiszs (approved):

PR #1548, opened 2026-06-01
fix: fallback to legacy cf on old epoch

hadjiszs:
get network info requires to retrieve the latest settled claim, this can lead to the reading of a certificate from old epoch which doesn't have the new columns, here is a workaround by falling back to the read of the legacy CF


Ekleog-Polygon (approved):
LGTM to unblock things now, but @Freyskeyd should probably have a look at this because I don't know how it interacts with the planned tool to migrate old epoch dbs — I'd personally have gone with just migrating the old epoch dbs because we have backups anyway, but I know he's opposed to it and I don't have the full context 😄

Freyskeyd (approved):

PR #1549, opened 2026-06-02
feat(settlement): persist terminal result metadata

### LINKED ISSUE #1317: Storage function to write a settlement job's result to rocksdb

Ekleog-Polygon:
Persist terminal settlement job results with the finalized wallet, nonce, and attempt number.

Record attempt-level terminal outcomes and allow client errors to be replaced by stronger final evidence.

Fixes https://github.com/agglayer/agglayer/issues/1317


proto/agglayer/storage/v0/settlement.proto
```diff
@@ -55,8 +55,17 @@ message SettlementAttemptResult {
...
+
+  // Nonce of the finalized settlement transaction.
+  Nonce nonce = 2;
+
+  // Sequence number of the finalized settlement attempt.
+  AttemptSequenceNumber attempt_number = 3;
+
   // Result of the final on-chain contract call for the job.
-  ContractCallResult contract_call_result = 1;
+  ContractCallResult contract_call_result = 4;
```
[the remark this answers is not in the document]
Ekleog-Polygon: This was never deployed yet, it's fine.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -331,13 +331,18 @@ mod tests {
...
-            outcome,
-            metadata: vec![seed, seed.wrapping_add(1)].into(),
-            block_hash: B256::from([seed; 32]),
-            block_number: seed as u64,
-            tx_hash: SettlementTxHash::new(Digest::from([seed.wrapping_add(2); 32])),
-        })
+        SettlementJobResult {
+            wallet: agglayer_types::Address::from([seed.wrapping_add(3); 20]),
+            nonce: Nonce(seed as u64 + 200),
+            attempt_number: SettlementAttemptNumber(seed as u64 + 300),
```
Freyskeyd, line 337: what's this magic numbers? `300` and `200`

proto/agglayer/storage/v0/settlement.proto
```diff
@@ -37,8 +37,17 @@ message SettlementAttemptResult {
 
 // Terminal result of a settlement job.
 message SettlementJobResult {
+  // Sender wallet of the finalized settlement transaction.
+  Address wallet = 1;
+
+  // Nonce of the finalized settlement transaction.
+  Nonce nonce = 2;
```
Freyskeyd, line 44: Why do we need those informations here?
Can't this be replaced by a txHash?

proto/agglayer/storage/v0/settlement.proto
```diff
@@ -57,6 +66,9 @@ enum ClientErrorType {
 
   // Nonce already used and finalized in other tx.
   CLIENT_ERROR_TYPE_NONCE_ALREADY_USED = 1;
+
+  // Settlement succeeded in another tx.
+  CLIENT_ERROR_TYPE_SETTLEMENT_SUCCEEDED_ELSEWHERE = 2;
```
Freyskeyd, line 71: `CLIENT_ERROR_TYPE_SETTLEMENT_SUCCEEDED_ELSEWHERE` this feels odd, what does `elsewhere` means here?

Freyskeyd (approved):

proto/agglayer/storage/v0/settlement.proto
Ekleog-Polygon: txHash (and this information) are both actually reachable through the attempt number. Attempt number is the change that we definitely need; these two are here for convenience if we ever need to debug the storage rather than as elements actually required.

I'm open to removing them in a follow-up if you think the storage would be simpler without it, or to add a txHash if you think the storage would be cleaner with it!

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: Just random values so that serialization doesn't end with all-0's and there is some actual data in there that validates eg. the fields don't get swapped after a dependency upgrade :smile:

proto/agglayer/storage/v0/settlement.proto
Ekleog-Polygon: It means another attempt of the same job got the settlement to succeed! Do you have any suggestion for a better name? If yes I'll fix in a follow-up; this way we avoid having to fix merge conflicts in the next few PRs that are already ready :smile:

PR #1552, opened 2026-06-04
fix(settlement): load persisted settlement attempts

Ekleog-Polygon:
Handle the settlement-attempt reload path that was forgotten in #1472
so #1312 is properly completed.

This hydrates stored attempts and matching results when rebuilding a
settlement task from storage. It also rejects duplicate attempt entries
or orphaned result entries so recovery does not continue with an
inconsistent settlement history.

Tests cover successful hydration and orphaned result rejection.


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -987,6 +1029,10 @@ mod tests {
         )
     }
 
+    fn mk_job_id(seed: u128) -> SettlementJobId {
+        SettlementJobId::from(ulid::Ulid::from(seed))
+    }
+
```
Freyskeyd, line 1035: this function isn't necessary, I think `SettlementJobId` implement `From<u128>`

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -765,9 +765,51 @@ impl<L1Provider: Provider + 'static, SettlementStore: SettlementReader + Settlem
     }
 
     async fn load_settlement_attempts_from_db(&mut self) -> eyre::Result<()> {
```
Freyskeyd, line 767: there is no await or async call in this function, we could make it sync to avoid confusion no?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -765,9 +765,51 @@ impl<L1Provider: Provider + 'static, SettlementStore: SettlementReader + Settlem
...
+        if let Some((attempt_number, _)) = results_by_attempt_number.first_key_value() {
+            eyre::bail!(
+                "Settlement attempt result {attempt_number} exists for job {} without a recorded \
+                 settlement attempt",
+                self.id,
+            );
+        }
+
+        self.attempts = loaded_attempts;
+        Ok(())
```
Freyskeyd, line 812: Could we move all the hydratation of `attempts` into a pure function (outside the `SettlementTask` so that it can be tested without the need of a big setup?
Basically the`load_settlement_attempts_from_db` would only deal with calling the DB and calling the hydrate method. It would make the hydration tests more easy and readable, something like:

```rust
impl SettlementTask {
	// ...

	fn load_settlement_attempts_from_db(&mut self) -> eyre::Result<()> {
	    let results = self.store.list_settlement_attempt_results(&self.id)?;
	    let attempts = self.store.list_settlement_attempts(&self.id)?;

	    self.attempts = hydrate_settlement_attempts(attempts, results, self.id)?;
	    Ok(())
	}
}

fn hydrate_settlement_attempts(attempts, results) -> eyre::Result<BTreeMap<(Address, Nonce), BTreeMap<SettlementAttemptNumber, ActiveSettlementAttempt>>> {
	// ...
}

#[cfg(test)]
mod tests {
	#[test]
	fn hydrate_rejects_result_without_attempt() {
	    let error = hydrate_settlement_attempts(
	        Vec::new(),
	        vec![(7, mk_client_error(5))],
	        mk_job_id(2),
	    )
	    .expect_err("orphaned attempt result should fail hydration");

	    assert!(error
	        .to_string()
	        .contains("without a recorded settlement attempt"));
	}
}

Freyskeyd (approved):

Ekleog-Polygon:
Thank you for your three comments! I have a commit coming in that should handle all of them; considering it's mostly test cleanup I'll land this now and submit the commit as a follow-up PR once it passes all tests :smile:

PR #1560, opened 2026-06-05
refactor(settlement): isolate attempt hydration

Ekleog-Polygon:
## Summary
- move persisted settlement-attempt hydration into a pure helper
- make settlement-attempt DB loading synchronous
- remove the test-only job-id helper and add focused hydration tests

Follow-up PR from #1552

## Verification
- cargo make blast-radius
- cargo make generate-proto
- cargo check --workspace --tests --all-features
- mdbook build docs/knowledge-base/
- cargo make ci-all
- cargo make pp-check-vkey-change
- cargo test -p agglayer-clock skipped_blocks_are_handled -- --test-threads=1

Note: cargo nextest run --workspace failed in agglayer-clock skipped_blocks_are_handled cases during the broad parallel run; the same three cases passed in the targeted serial rerun above.


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -139,15 +139,58 @@ struct ActiveSettlementAttempt {
     result: Option<SettlementAttemptResult>,
 }
 
+type ActiveSettlementAttempts =
+    BTreeMap<(Address, Nonce), BTreeMap<SettlementAttemptNumber, ActiveSettlementAttempt>>;
+
+fn hydrate_settlement_attempts(
+    attempts: Vec<(u64, SettlementAttempt)>,
+    results: Vec<(u64, SettlementAttemptResult)>,
```
Freyskeyd: This can be an iterator maybe?

```rust
attempts: impl IntoIterator<Item = (impl Into<SettlementAttemptNumber>, SettlementAttempt)>,
```

Freyskeyd (dismissed):

Ekleog-Polygon:
On the question of whether the storage functions could also return `impl Iterator`: I've left that as **future work** rather than doing it in this PR.

The `hydrate_settlement_attempts` helper now accepts `impl IntoIterator`, so the allocation concern raised in review is addressed at the helper boundary. Pushing true streaming into the storage layer is a meaningfully larger change, for these reasons:

- The storage methods are trait methods on `SettlementReader` (e.g. `list_settlement_attempts(...) -> Result<Vec<(u64, SettlementAttempt)>, Error>`). The concrete `StateStore` already uses `prefix_iterator` internally, but collects immediately so decode/storage errors surface before the method returns.
- Returning `Result<impl Iterator<...>, Error>` is only easy if we still collect first and hand back `vec.into_iter()` — that keeps the allocation, so there's little practical gain.
- A genuinely streaming `impl Iterator<Item = Result<..., Error>>` moves error handling from method-return time into iteration time, which would ripple through callers, tests, mocks, and the `hydrate_settlement_attempts` usage.
- Because `SettlementReader` is a trait backed by mocks, a streaming return likely needs a boxed iterator or a GAT-style associated type, touching `SettlementReader`, `StateStore`, `MockStateStore`, the storage tests, and the call sites.

So for now I didn't change the storage trait for this. A real storage streaming API should be a separate, deliberate refactor if/when we have a concrete need to avoid materializing large result sets.

Freyskeyd (approved):

PR #1564, opened 2026-06-08
feat: verify whether address is owned

### LINKED ISSUE #1385: Tie configuration with settlement task so that it can know for which wallets we know the privkey

Search in the code:
```
// XREF: https://github.com/agglayer/agglayer/issues/1385
```

hadjiszs:
Fixes https://github.com/agglayer/agglayer/issues/1385


Freyskeyd (approved):
Nothing special on this PR, maybe a nit would be to have an alias for the trait bound `Provider + WalletProvider + 'static` but it's not mandatory

PR #1571, opened 2026-06-09
feat: wait for any new expected `(wallet, nonce)` to be on L1

### LINKED ISSUE #1314: Helper function to wait for any tx from a given address with a given nonce to be included on L1

Context: to be used in https://app.excalidraw.com/s/65UEf35l1DW/8Fy17zCAQNl

Search in the code, this place will call this helper function in a loop
```
// XREF: https://github.com/agglayer/agglayer/issues/1314
```

hadjiszs:
Fixes https://github.com/agglayer/agglayer/issues/1314


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -41,26 +41,30 @@ fn required_settlement_head_number(receipt_block_number: u64, confirmations: usi
     receipt_block_number.saturating_add(confirmation_offset)
 }
 
+/// Error returned by the L1 polling callbacks; the "not yet" variants are
+/// transient and tell the retry loop to keep polling.
 #[derive(Debug)]
-enum WaitForSettlementError {
+enum WaitForL1Error {
```
Freyskeyd: I guess this `enum` is name `WaitForL1Error` due to the fact that it was planned to be used in `wait_for_*` methods right?

The issue now is that this error enum is not just used in `wait_for` method, should we rename it?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1134,6 +1168,28 @@ mod tests {
         }
     }
 
+    fn build_anvil_provider(anvil: &AnvilInstance) -> impl Provider + WalletProvider + 'static {
```
Freyskeyd: I think there is already a method somewhere to build this something like `build_provider`

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: it was named `WaitForSettlementError` but I extended it also for non-settlement error with the variant `NotIncludedYet`, that's why got renamed to `WaitForL1Error`.

I can rename it to something else if you have better idea, or I can also rename it back to the original name `WaitForSettlementError` and then have a proper `WaitForInclusionError` enum just for the wait for inclusion in block brought by this PR

Freyskeyd: I don't know what was the approach that @Ekleog-Polygon envision on those errors

Freyskeyd (dismissed):

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: I think it is simply for the errors that occur while waiting for a status to be reached.

initially it was `WaitForSettlement`, so only for "settlement", like this:

```rust
enum WaitForSettlementError {
    NotSettledYet,
    Transport(TransportError),
}
```

and in this PR, I need to wait for simple inclusion in a block for the implementation of `wait_for_any_nonce_on_l1`, which is not for settlement per se then, but simple inclusion in a block. 

Consequently, I added the variant `NotIncludedYet` in this enum, and renamed it to something broader `WaitForL1Error`

Freyskeyd: I don't have a strong opinion about that, WDYT @Ekleog-Polygon ?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -635,17 +663,19 @@ impl<
...
-            RetryCallbackError::Error(WaitForSettlementError::Transport(error)) => {
+            RetryCallbackError::Error(WaitForL1Error::Transport(error)) => {
                 RetryCallbackError::Error(error)
             }
-            RetryCallbackError::Error(WaitForSettlementError::NotSettledYet) => {
-                unreachable!("not-settled-yet errors are always transient")
+            RetryCallbackError::Error(
+                WaitForL1Error::NotSettledYet | WaitForL1Error::NotIncludedYet,
+            ) => {
+                unreachable!("not-ready-yet errors are always transient")
```
Freyskeyd: No risks to have that panic in production?

hadjiszs: normally it is unreachable because transient error are retried by the `retry` callback

but I switched to actual error + error log just in case.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: I switched back to `WaitForSettlementError` to unblock the merge of this PR, we can rename to something more appropriate if we want to in the next PR

Freyskeyd (dismissed):

Freyskeyd (approved):

PR #1573, opened 2026-06-15
build(deps): bump openssl, rustls-webpki, rand for security fixes

Freyskeyd:
Resolve 11 open Dependabot alerts with in-range Cargo.lock bumps:

- openssl 0.10.76 -> 0.10.80 (8 alerts, 5 High)
- rustls-webpki 0.103.12 -> 0.103.13 (High, GHSA-82j2-j2ch-gfr8)
- rand 0.8.5 -> 0.8.6 and 0.9.2 -> 0.9.3 (GHSA-cq8v-f236-94qc)

Lockfile-only change. Verified with cargo check --workspace --tests
--all-features and cargo nextest run --workspace (745 passed).

Remaining alerts (lru, git2, rustls-webpki 0.101.7) are pinned by the
SP1 toolchain and require an upstream bump; p3-symmetric has no
upstream patch yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -437,7 +437,7 @@ dependencies = [
  "mockall",
  "pessimistic-proof",
  "pin-project",
- "rand 0.9.2",
+ "rand 0.9.3",
  "reqwest 0.12.28",
  "rstest",
  "serde",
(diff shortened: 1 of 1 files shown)


hadjiszs (approved):
probably that `.cargo/audit.toml` can be trimmed a little as well then

PR #1578, opened 2026-06-15
feat(storage): store certificate settlement job ids

### LINKED ISSUE #1430: Column for certificate id -> settlement job id storage, with store/read helpers

Ekleog-Polygon:
Adds certificate ID to settlement job ID storage so certificate tasks can resolve settlement jobs directly.

The state store now has a certificate_settlement_job_cf column family with required StateReader and StateWriter helpers, insert-only semantics, and migration coverage for existing DBs. Test doubles implement the required methods for unit tests.

Fixes #1430

### DIFF
--- a/crates/agglayer-storage/src/columns/certificate_settlement_job/mod.rs
+++ b/crates/agglayer-storage/src/columns/certificate_settlement_job/mod.rs
@@ -1,0 +1,19 @@
+use agglayer_types::{CertificateId, SettlementJobId};
+
+use crate::{columns::CERTIFICATE_SETTLEMENT_JOB_CF, schema::ColumnSchema};
+
+/// Column family mapping certificates to their settlement jobs.
+///
+/// ## Column definition
+///
+/// | key             | value             |
+/// | --              | --                |
+/// | `CertificateId` | `SettlementJobId` |
+pub(crate) struct CertificateSettlementJobColumn;
+
+impl ColumnSchema for CertificateSettlementJobColumn {
+    type Key = CertificateId;
+    type Value = SettlementJobId;
+
+    const COLUMN_FAMILY_NAME: &'static str = CERTIFICATE_SETTLEMENT_JOB_CF;
+}
--- a/crates/agglayer-storage/src/stores/interfaces/reader.rs
+++ b/crates/agglayer-storage/src/stores/interfaces/reader.rs
@@ -2,7 +2,7 @@ use std::collections::BTreeMap;
 
 use agglayer_types::{
     Certificate, CertificateHeader, CertificateId, CertificateIndex, EpochNumber, Height,
-    LocalNetworkStateData, NetworkId, Proof,
+    LocalNetworkStateData, NetworkId, Proof, SettlementJobId,
 };
 
 use crate::{
@@ -88,6 +88,11 @@ pub trait StateReader: Send + Sync {
         &self,
         certificate_id: &CertificateId,
     ) -> Result<Option<CertificateHeader>, Error>;
+
+    fn get_certificate_settlement_job_id(
+        &self,
+        certificate_id: &CertificateId,
+    ) -> Result<Option<SettlementJobId>, Error>;
 
     fn get_certificate_header_by_cursor(
         &self,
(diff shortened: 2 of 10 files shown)


crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -208,6 +210,36 @@ impl StateWriter for StateStore {
...
 
+    fn insert_certificate_settlement_job_id(
+        &self,
+        certificate_id: &CertificateId,
+        settlement_job_id: &SettlementJobId,
+    ) -> Result<(), Error> {
+        if self
+            .db
+            .get::<CertificateSettlementJobColumn>(certificate_id)?
+            .is_some()
```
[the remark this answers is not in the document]
Ekleog-Polygon: There is a single certificate task that can register to the certificate/settlement job CF, so this should be fine for now. We can reconsider later as further hardening, but for now it would add unneeded complexity for something already critical.

crates/agglayer-storage/src/columns/certificate_settlement_job/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+use crate::{columns::CERTIFICATE_SETTLEMENT_JOB_CF, schema::ColumnSchema};
+
+/// Column family mapping certificates to their settlement jobs.
+///
+/// ## Column definition
+///
+/// | key             | value             |
+/// | --              | --                |
+/// | `CertificateId` | `SettlementJobId` |
+pub(crate) struct CertificateSettlementJobColumn;
```
Freyskeyd, line 12: I've a broader question related to this, this CF definition states that we only and ever have one settlement job per `certificate_id`. But what if we want to stop, cancel or a `SettlementJob` becomes unsettlable?

It could happen for various productions issues and the design of `SettlementJob never fails` may not hold

crates/agglayer-storage/src/stores/interfaces/reader.rs
```diff
@@ -89,6 +89,11 @@ pub trait StateReader: Send + Sync {
         certificate_id: &CertificateId,
     ) -> Result<Option<CertificateHeader>, Error>;
 
+    fn get_certificate_settlement_job_id(
+        &self,
+        certificate_id: &CertificateId,
+    ) -> Result<Option<SettlementJobId>, Error>;
+
```
Freyskeyd, line 96: is it really valuable ? will we ever want just the `ID` and not the full payload of the `SettlementJob` ?

crates/agglayer-storage/src/stores/interfaces/writer.rs
```diff
@@ -69,6 +70,17 @@ pub trait StateWriter: Send + Sync {
...
+    /// Inserts the settlement job id associated with `certificate_id`.
+    ///
+    /// This is an insert-only operation and must fail if `certificate_id`
+    /// already has a stored settlement job id. The settlement job must already
+    /// exist.
+    fn insert_certificate_settlement_job_id(
+        &self,
+        certificate_id: &CertificateId,
+        settlement_job_id: &SettlementJobId,
+    ) -> Result<(), Error>;
```
Freyskeyd, line 82: Same for this, the `SettlementJobId` should be inserted when we persist the `SettlementJob`

crates/agglayer-storage/src/columns/certificate_settlement_job/mod.rs
Ekleog-Polygon: In that case the certificate will end up in SettlementError.

Considering what we currently know, I'm thinking that this is enough for the medium term at least, especially as we already are moving from a single SettlementTxHash to a SettlementJobId that can match with many settlement transactions.

Does that make sense to you too?

crates/agglayer-storage/src/stores/interfaces/reader.rs
Ekleog-Polygon: We actually will most often want only the settlement job is: this column will be read by the CertificateTask, which only cares about the job id. Then it will send that job id to settlement service, which will actually read the rest.

This allows us to keep a better separation of concerns in exchange for not much loss 😄

crates/agglayer-storage/src/stores/interfaces/writer.rs
Ekleog-Polygon: This is actually the theme of #1453, which is the next PR I'm currently working on. I didn't include it here to keep the diff easier to review 😄

hadjiszs (approved):

PR #1582, opened 2026-06-18
feat(settlement): record certificate job ids before jobs

Ekleog-Polygon:
Record certificate-to-settlement-job mappings during settlement task creation so the certificate link is persisted before the settlement job row and before the task starts executing.

This keeps the settlement service on one creation path, with the tower request carrying an optional certificate id and the inherent service API taking certificate id and job as separate parameters.

The storage insert remains insert-only for duplicate certificate mappings, but no longer requires the settlement job row to exist first.

Verified with:
- cargo make ci-format
- cargo make ci-clippy
- cargo make ci-typos
- cargo test -p agglayer-settlement-service
- cargo test -p agglayer-storage stores::state::tests::settlement

Fixes #1453

### DIFF
--- a/crates/agglayer-settlement-service/src/settlement_service.rs
+++ b/crates/agglayer-settlement-service/src/settlement_service.rs
@@ -1,8 +1,8 @@
 use std::{collections::HashMap, future::Future, pin::Pin, sync::Arc};
 
 use agglayer_config::settlement_service::{SettlementServiceConfig, SettlementTransactionConfig};
-use agglayer_storage::stores::{SettlementReader, SettlementWriter};
-use agglayer_types::{SettlementJob, SettlementJobId, SettlementJobResult};
+use agglayer_storage::stores::{SettlementReader, SettlementWriter, StateReader, StateWriter};
+use agglayer_types::{CertificateId, SettlementJob, SettlementJobId, SettlementJobResult};
 use alloy::providers::{Provider, WalletProvider};
 use educe::Educe;
 use eyre::Context as _;
@@ -52,7 +52,7 @@ pub enum RetrievedSettlementResult {
 
 impl<
         L1Provider: Provider + WalletProvider + 'static,
-        SettlementStore: SettlementReader + SettlementWriter + Send + Sync + 'static,
+        SettlementStore: SettlementReader + SettlementWriter + StateReader + StateWriter + Send + Sync + 'static,
     > SettlementService<L1Provider, SettlementStore>
 {
     pub async fn start(
@@ -116,6 +116,7 @@ impl<
     #[tracing::instrument(skip(self))]
     pub async fn request_new_settlement(
         &self,
+        certificate_id: Option<CertificateId>,
         job: SettlementJob,
     ) -> eyre::Result<SettlementJobWatcher> {
         let (admin_sender, admin_receiver) = mpsc::channel(ADMIN_CHANNEL_BUFFER_SIZE);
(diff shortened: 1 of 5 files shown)


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -344,6 +335,41 @@ impl<
...
+        store: &SettlementStore,
+        certificate_id: Option<CertificateId>,
+    ) -> eyre::Result<SettlementJobId> {
+        let Some(certificate_id) = certificate_id else {
+            return Ok(Self::generate_settlement_job_id().await);
+        };
+
+        let settlement_job_id = Self::generate_settlement_job_id().await;
+        store
+            .insert_certificate_settlement_job_id(&certificate_id, &settlement_job_id)
```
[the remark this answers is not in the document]
Ekleog-Polygon: For now, crashes between two storage writes are mostly out of scope. AFAIU the worst that could happen would be a crash here, so it should be fine to keep it this way for now

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -344,6 +335,41 @@ impl<
         Ok((id, this))
     }
 
+    async fn settlement_job_id_for(
```
Freyskeyd: This method hide a persistent side effect behind a getter-like name, maybe it's better to rename it to something like: `assign_settlement_job_id` or `reserve_settlement_job_id` ?

And a broader suggestion would be to not bound `CertificateId` to the `agglayer-settlement-service` at all. It doesn't need to have that link and the requester of a settlement should own the link between `certificate_id -> settlement_job_id`.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -209,11 +211,15 @@ impl<
     }
 }
 
-pub struct RequestNewSettlement(pub SettlementJob);
+#[derive(Debug)]
+pub struct RequestNewSettlement {
+    pub certificate_id: Option<CertificateId>,
```
Freyskeyd, line 216: This should be a non-optional `SettlementJobId` instead of bounding to `CertificateId`

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: I agree with the rename, and will do it! For the broader suggestion, it is actually the original design, but with the rest of the current system (where settlement job id is generated by storage to avoid any risk of duplicates), it would be basically impossible to do it without significant API changes. So in order to avoid too much friction, I'd favor doing that as a follow-up refactoring rather than as part of the critical path towards phase 1.

Does that make sense to you too?

crates/agglayer-settlement-service/src/settlement_service.rs
Ekleog-Polygon: (see the other comment's discussion)

crates/agglayer-settlement-service/src/settlement_task.rs
Freyskeyd: I was thinking of calling the storage from `CertificateTask` to "reserve" the `SettlementJobId` before drafting the query to the `SettlementService`. Something like taking a "ticket".
I'll draft a little PR to test this

Freyskeyd (approved):

hadjiszs (approved):

PR #1584, opened 2026-06-19
fix(settlement-service): cleanup watchers after terminal settlement jobs

NikhilSharmaWe:
Fixes #1480

## Summary

- Remove `result_watchers` entries when a spawned settlement task exits (Completed, Cancelled, reload-error, reload-to-completed, panic), fixing unbounded in-memory growth on the normal completion path.
- Use a unified **`TaskRegistrationGuard`** for synchronous RAII teardown of both `result_watchers` and `task_controls` when the spawn loop exits — no async cleanup from `Drop`.
- Drop order is **watcher first, then control**, so `admin_reloadSettlementTask` cannot respawn while a control entry still exists; this eliminates the stale-teardown race where delayed cleanup could delete a respawned task after panic (per review feedback).
- Callers are notified via their cloned `watch::Receiver` before cleanup; `retrieve_settlement_result` falls back to RocksDB once the in-memory entry is gone.

## Tests

- `cargo nextest run -p agglayer-settlement-service settlement_service::tests`
- `cargo clippy -p agglayer-settlement-service --tests -- -D warnings`
- `cargo +nightly fmt --all --check`

Added coverage:
- `task_registration_guard_clears_watcher_then_control_on_drop` — sync RAII teardown
- `retrieve_uses_storage_after_active_tracking_cleanup` — storage fallback after map cleanup
- `caller_watcher_retains_result_after_active_tracking_cleanup` — cloned receiver contract
- `spawn_clears_result_watchers_after_task_completes` — spawn → Completed → both maps empty
- `reload_and_restart_clears_registrations_when_reload_finds_completed_job` — reload-to-completed also clears maps
- `panicked_task_reload_respawns_task_without_stale_teardown` — panic → admin reload → respawned task stays live

### DIFF
--- a/crates/agglayer-settlement-service/src/settlement_service.rs
+++ b/crates/agglayer-settlement-service/src/settlement_service.rs
@@ -25,65 +25,6 @@ use crate::{
     wallet_nonce_locks::WalletNonceLocks,
 };
 
-type ResultWatchersMap =
-    Arc<Mutex<HashMap<SettlementJobId, watch::Receiver<Option<SettlementJobResult>>>>>;
-type TaskControlsMap = Arc<std::sync::Mutex<HashMap<SettlementJobId, TaskControlHandle>>>;
-
-async fn remove_job_from_active_tracking(
-    job_id: SettlementJobId,
-    result_watchers: &ResultWatchersMap,
-    task_controls: &TaskControlsMap,
-) {
-    result_watchers.lock().await.remove(&job_id);
-    task_controls
-        .lock()
-        .expect("settlement task_controls lock poisoned")
-        .remove(&job_id);
-}
-
-/// Ensures in-memory job tracking is removed when a spawned settlement task
-/// ends abnormally (e.g. panics) before explicit cleanup can run.
-struct JobTrackingGuard {
-    job_id: SettlementJobId,
-    result_watchers: ResultWatchersMap,
-    task_controls: TaskControlsMap,
-    armed: bool,
-}
-
-impl JobTrackingGuard {
-    fn new(
-        job_id: SettlementJobId,
-        result_watchers: ResultWatchersMap,
-        task_controls: TaskControlsMap,
-    ) -> Self {
-        Self {
-            job_id,
-            result_watchers,
-            task_controls,
-            armed: true,
-        }
-    }
-
-    fn disarm(&mut self) {
-        self.armed = false;
-    }
-}
-
-impl Drop for JobTrackingGuard {
-    fn drop(&mut self) {
-        if !self.armed {
-            return;
-        }
-
-        let job_id = self.job_id;
-        let result_watchers = self.result_watchers.clone();
-        let task_controls = self.task_controls.clone();
-        tokio::spawn(async move {
-            remove_job_from_active_tracking(job_id, &result_watchers, &task_controls).await;
-        });
-    }
-}
-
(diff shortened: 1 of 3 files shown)


crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -16,6 +16,62 @@ use crate::settlement_task::{
...
+    fn drop(&mut self) {
+        if !self.armed {
+            return;
+        }
+
+        let job_id = self.job_id;
+        let result_watchers = self.result_watchers.clone();
+        let task_controls = self.task_controls.clone();
+        tokio::spawn(async move {
+            remove_job_from_active_tracking(job_id, &result_watchers, &task_controls).await;
```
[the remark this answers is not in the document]
NikhilSharmaWe: Valid point on the panic path, cleanup can remove the in-memory watcher before a terminal result exists, so retrieve_settlement_result may hit the existing todo!() in #1230.

This PR is mainly for #1480 (cleanup after normal completion, when the result is already persisted). @Freyskeyd Should we handle panic cleanup here, or keep that out of scope for this PR?

NikhilSharmaWe:
@Freyskeyd, please take a look. Thanks!

NikhilSharmaWe:
cc @Ekleog-Polygon @Freyskeyd @hadjiszs Resolved merge conflicts, ready for review

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -25,6 +25,65 @@ use crate::{
...
+impl Drop for JobTrackingGuard {
+    fn drop(&mut self) {
+        if !self.armed {
+            return;
+        }
+
+        let job_id = self.job_id;
+        let result_watchers = self.result_watchers.clone();
+        let task_controls = self.task_controls.clone();
+        tokio::spawn(async move {
```
hadjiszs: `Drop` spawns a detached future keyed only by `job_id`, with no check the entries are still this
task's. After a panic the runbook's recovery is `admin_reloadSettlementTask`, which creates a fresh
task and result watcher; if that future lands afterwards it deletes them. `hasLiveTask` then reports
`false` for a task that is still running, and the runbook makes `hasLiveTask: false` the
precondition for a safe respawn, so the documented procedure spawns a duplicate task on the same
wallet nonces.

Suggest making `result_watchers` a `std::sync::Mutex` and removing it in
`TaskControlRegistrationGuard`, watcher before control entry. Nothing holds that lock across an
`.await`. That drops `JobTrackingGuard` and every `disarm()`, and closes the reload -> `Completed`
leak too.

NikhilSharmaWe: Updated

- Replaced `JobTrackingGuard` with a sync `TaskRegistrationGuard` that clears both `result_watchers` and `task_controls` when the spawned task exits — Completed, Cancelled, reload error, reload→Completed, or panic.

- Moved `result_watchers` to `std::sync::Mutex` so cleanup runs synchronously in `Drop`, not via a spawned task. The guard sits outside the run loop so reload→Pending keeps both registrations; on terminal exit it removes the watcher first, then the control, since reload won't respawn while a control entry is still present.

- Removed the inline watcher removes and all `disarm()` calls. Added `panicked_task_reload_respawns_task_without_stale_teardown` for the panic→reload path. Rebased onto main.

PR #1619, opened 2026-07-03
feat: include software version in /health endpoint

krlosMata:
## Summary

The `/health` endpoint of the agglayer node now returns the running software version in addition to the existing health boolean.

Response body:

```json
{"health":true,"version":"agglayer (v0.x.x-...) [git commit timestamp: ...]"}
```

- `health` — unchanged boolean (`true`), same status code (`200 OK`)
- `version` — the running build version (already available in `agglayer_node::main`, previously only logged)

This makes it easy to check which build is deployed without shell access to the container.

## Test plan

- [x] `cargo build -p agglayer-node`
- [x] `cargo nextest run -p agglayer-node healthcheck_method_can_be_called`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/agglayer-node/src/lib.rs
+++ b/crates/agglayer-node/src/lib.rs
@@ -112,6 +112,7 @@ pub fn main(
         Node::builder()
             .config(config.clone())
             .cancellation_token(global_cancellation_token.clone())
+            .version(version.to_string())
             .start(),
     )?;
     let terminate_signal = async {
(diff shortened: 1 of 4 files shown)


Ekleog-Polygon (dismissed):
LGTM! @tnobayashi can you confirm this won't break health infra?

tnobayashi:
> LGTM! @tnobayashi can you confirm this won't break health infra?

@Ekleog-Polygon I confirmed it should be fine 👍

Ekleog-Polygon:
@krlosMata I see the formatting lint fails, if you run `cargo +nightly fmt` it should fix it and then we can land :smile:

Ekleog-Polygon (approved):

PR #1630, opened 2026-07-06
fix(storage): apply declared column options when reopening the database

Ekleog-Polygon:
## Problem

On reboot, startup recovery fails with:

```
Error: Failed to load settlement job 01KWW4KKFRFHEW4D5V3MZSA4B4 during startup recovery
Caused by: Duplicate settlement attempt 0 for job 01KWW4KKFRFHEW4D5V3MZSA4B4
```

## Root cause

The settlement attempt columns are keyed by `16-byte job id ‖ 8-byte BE attempt number` and declare a fixed 16-byte prefix extractor, but the extractor (and the rest of the declared column options) was only applied when creating a **fresh** database or a new column family. Reopening an existing database went through `rocksdb::DB::open_cf`, which opens every column family by name with **default options**.

Without a prefix extractor, RocksDB silently ignores `prefix_same_as_start` (see `DBIter`: it is disabled when the CF has no extractor), so the per-job "prefix" listing actually scanned from the requested job id to the end of the column family, sweeping in later jobs' attempts. Every job numbers its attempts from 0, so loading any pending job followed by another job with recorded attempts fails with "Duplicate settlement attempt 0". Fresh databases (first boot, e2e) were unaffected, which is why this only surfaced on reboot — the declared options are identical across v0.5.1 / v0.6.0-rc.1 / v0.6.0-rc.3, so this affects every deployed version equally.

The stored data itself is fine: keys are unique per (job id, attempt number) by construction, so **no data repair is needed** — recovery works again as soon as this fix is deployed.

## Fix

`DB::builder`/`Builder::open` now take the authoritative schema catalog (each store passes its full target schema, e.g. `STATE_DB`); existing and read-only databases open each on-disk column family with its declared options, falling back to defaults (with a debug log) for column families the binary does not know about. This also restores the declared **Lz4 compression** on reopen (new SSTs were silently written with RocksDB's default compression before).

`DB::prefix_iterator` keeps using RocksDB's native prefix seek; it now documents the invariant it relies on (the CF must be opened with an extractor covering exactly the encoded prefix). Mixed SST files from past extractor-less opens are safe: RocksDB skips bloom filters whose recorded extractor mismatches, and the boundary check happens at the iterator layer using the current extractor.

## Tests

Both written first and verified to fail before the fix:

- `list_settlement_attempts_stays_within_job_after_reopen` — end-to-end reproduction of the production failure: two jobs with attempt 0, close, reopen, list.
- `reopened_database_applies_declared_prefix_extractor` — probes the raw RocksDB prefix iterator on a reopened database, directly guarding the invariant `prefix_iterator` relies on.

## Observability

New debug logs for future incidents: the known-CF catalog (with options) at every open, per-CF declared-vs-default options when opening existing databases, options included in the create-CF migration logs, and per-attempt/per-result lines when hydrating a settlement job (emitted before the duplicate checks, so the offending entries are visible if this class of bug ever recurs).

## Verification

`cargo make generate-proto` (no diffs), `cargo check --workspace --tests --all-features`, `mdbook build docs/knowledge-base/`, `cargo make ci-all`, `cargo nextest run --workspace`, and `cargo make pp-check-vkey-change` all pass; targeted storage+settlement tests re-run green after rebasing onto main.

### DIFF
--- a/crates/agglayer-storage/src/storage/migration/mod.rs
+++ b/crates/agglayer-storage/src/storage/migration/mod.rs
@@ -44,8 +44,19 @@ impl Builder {
     /// This method initializes or opens an existing database with the provided
     /// initial schema (v0). It automatically sets up migration tracking and
     /// validates the database schema.
-    pub fn open(path: &Path, cfs_v0: &[ColumnDescriptor]) -> Result<Self, DBOpenError> {
+    ///
+    /// `known_cfs` catalogs every column family this binary declares across
+    /// all schema versions. Column family options only apply when passed at
+    /// open time, so an existing database is opened with the declared options
+    /// looked up there (RocksDB defaults for on-disk column families it does
+    /// not list).
+    pub fn open(
+        path: &Path,
+        cfs_v0: &[ColumnDescriptor],
+        known_cfs: &[ColumnDescriptor],
+    ) -> Result<Self, DBOpenError> {
         debug!("Preparing database for initialization and migration");
+        debug!(?known_cfs, "Column families known to this binary");
         let desc_v0: Vec<_> = cfs_v0.iter().map(DB::descriptor).collect();
         let cfs_v0: BTreeSet<_> = desc_v0.iter().map(|d| d.name()).collect();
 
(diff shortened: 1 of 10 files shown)


hadjiszs (approved):

crates/agglayer-storage/src/storage/migration/mod.rs
```diff
@@ -44,8 +44,19 @@ impl Builder {
...
+    ///
+    /// `known_cfs` catalogs every column family this binary declares across
+    /// all schema versions. Column family options only apply when passed at
+    /// open time, so an existing database is opened with the declared options
+    /// looked up there (RocksDB defaults for on-disk column families it does
+    /// not list).
+    pub fn open(
+        path: &Path,
+        cfs_v0: &[ColumnDescriptor],
+        known_cfs: &[ColumnDescriptor],
```
Freyskeyd, line 56: I don't understand why we're doing this. the `finalize` method should recreate the DB with the expected cfs

Ekleog-Polygon: Doing in a follow-up

PR #1633, opened 2026-07-07
feat: add per-network pending/proven/settled height metrics

### LINKED ISSUE #1352: Introduce metrics for pending/proven/settled heights per rollup
This will allow for better alerting on stuck rollups

Example metrics:
```
agglayer_network_pending_height{rollup="14"} 35
agglayer_network_proven_height{rollup="14"} 35
agglayer_network_settled_height{rollup="14"} 34
```

Freyskeyd:
Better alerting on stuck rollups requires knowing each network's
certificate progress; this exposes it on the existing /metrics
endpoint.

Adds two observable gauges, labeled `network_id`:

- `agglayer_node_network_height{stage="pending"|"proven"|"settled"}`
- `agglayer_node_network_latest_certificate_in_error` (0/1)

Naming notes for alert authors: issue #1352's example used a `rollup`
label, an `agglayer_` name prefix and one metric name per stage; this
implements `network_id` (Agglayer terminology), the `agglayer_node_`
application prefix (agglayer is the ecosystem; future binaries such
as an indexer or client get their own prefix), and the stage as a
label so dashboards can aggregate `by (stage)` and future lifecycle
stages extend the label set without new metric names.

Design:

- Gauge callbacks read the per-network RocksDB pointer column
  families at scrape time, so values always match storage truth,
  including after restarts and admin force-edits.
- An absent pointer exports no series: height 0 is a valid height
  and is never used as a placeholder.
- Callbacks warn-and-skip on storage errors; a storage fault can
  never panic the /metrics endpoint.
- `in_error` mirrors `get_network_info`'s `NetworkStatus::Error`
  definition (latest pending certificate header is `InError`). It
  reports storage truth and deliberately does not consult the
  disabled-networks list (documented divergence).

Layering: `agglayer-storage` gains one all-network pointer scan
(`get_current_pending_heights`, mirroring the proven/settled
scans); `agglayer-telemetry` owns metric names and registration
via a named `NetworkStateSamplers` struct (storage-free);
`agglayer-node` bridges stores to samplers at node startup, after
the meter provider is installed.

Tested with unit tests at each layer, a prometheus-exporter
round-trip test, and an integration test that settles a
certificate against a real L1 and asserts every series from a
single /metrics scrape.

Closes #1352


Freyskeyd:
CI follow-up: the `Integration tests` failure (`storage_backups::restore_at_particular_level`, RocksDB corruption on the restored DB) was a real bug in this PR, not a flake.

Root cause: the observable-gauge closures held strong `Arc<PendingStore>`/`Arc<StateStore>` references inside the process-global meter provider. The provider outlives in-process node shutdowns, so the old node's RocksDB stayed open on the storage path while `BackupEngine::restore_at` rewrote that directory, corrupting the restored DB (confirmed by single-variable experiment: disabling only the registration call made the test pass).

Fix in d2eabcab: the callbacks now hold `Weak` references and upgrade per scrape, so a registration extends store lifetime by at most one in-flight collect; once the node drops its stores, the series simply disappear. A regression test pins the property (register → drop the `Arc`s → reopening the RocksDB paths must succeed; pre-fix it failed with `lock hold by current process`), with a positive-control assertion so dependency drift can't silently defuse it (1c12bd79).

Also merged latest `main` (794f427a). Locally re-verified on the merged result: workspace suite 836 passed, clippy `-D warnings` clean, and both `restore_at_particular_level` and the new metrics e2e pass.

crates/agglayer-telemetry/src/network.rs
```diff
@@ -0,0 +1,274 @@
...
+    );
+    register_height_gauge(
+        NETWORK_SETTLED_HEIGHT,
+        "Height of the latest settled certificate per network",
+        settled,
+    );
+
+    // The returned instrument handle is intentionally dropped: the callback
+    // registration lives in the meter provider, not in the handle.
+    let _ = global::meter(AGGLAYER_NODE_NETWORK_OTEL_SCOPE_NAME)
```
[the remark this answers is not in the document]
Freyskeyd: Not applicable to opentelemetry 0.31. `SdkMeter::create_observable_gauge` registers the callback into the meter provider's pipeline (`self.pipes.register_callback(...)`) and returns `ObservableGauge<T>`, which is `{ _marker: PhantomData<T> }`: a zero-sized handle that owns nothing. Dropping it can't unregister the callback, and this is the `.with_callback(...).build()` form the OTel docs recommend.

Two tests confirm this by gathering after the handles are dropped: the unit test `network_state_gauges_export_per_network_series`, and the integration test `per_network_height_metrics_are_exposed`, which scrapes `/metrics` on a real node and asserts all four `agglayer_network_*` series are present. Not a bug.

Freyskeyd:
Naming update in a7369dac: exported names now carry the application prefix, `agglayer_node_network_*` instead of `agglayer_network_*`. Rationale: agglayer is the ecosystem, agglayer-node is the application; future binaries (indexer, client, ...) get their own prefix instead of colliding inside a bare `agglayer_` namespace. Renamed now while no dashboard depends on the names. The e2e scrape test was re-run against a real node with the new names.

crates/agglayer-telemetry/src/network.rs
```diff
@@ -0,0 +1,274 @@
...
+const NETWORK_ID_LABEL: &str = "network_id";
+
+/// Gauge name: height of the latest pending certificate per network.
+pub const NETWORK_PENDING_HEIGHT: &str = "agglayer_node_network_pending_height";
+
+/// Gauge name: height of the latest proven certificate per network.
+pub const NETWORK_PROVEN_HEIGHT: &str = "agglayer_node_network_proven_height";
+
+/// Gauge name: height of the latest settled certificate per network.
+pub const NETWORK_SETTLED_HEIGHT: &str = "agglayer_node_network_settled_height";
```
hadjiszs: I think I prefer to have the stages by label/attribute instead because easier for the grafana dashboard with `by (stage)` and also less things to modify and adapt in case of having this set of stage to evolve (e.g., `Preconfirmed` ; `Aggregated` etc)

hadjiszs (dismissed):

crates/agglayer-telemetry/src/network.rs
Freyskeyd: Agreed, mostly for the evolution argument: with the settlement rework likely to add stages (Preconfirmed etc), one more label value beats minting a new metric name each time. Done in db7dae53: the three gauges collapsed into `agglayer_node_network_height{stage="pending"|"proven"|"settled"}`. The in-error flag stays a separate gauge, since that one is a boolean state rather than a stage of the same quantity.

Only query-side cost: cross-stage math now needs label matching, e.g. the backlog gap is `agglayer_node_network_height{stage="pending"} - ignoring(stage) agglayer_node_network_height{stage="settled"}`. Worth a recording rule if we alert on that. The e2e scrape test was re-run against a real node with the new shape.

hadjiszs (approved):

Issue #1318, opened 2026-01-09
Helper function to build a settlement tx for a given job with a given nonce and gas parameters


Ekleog-Polygon:
If there is already a function, list it here.
If not, write one and list it here.

Don't send the settlement tx yet, just build something that we can record to rocksdb before submission

Ekleog-Polygon:
```rust

    fn build_next_attempt_with_new_nonce(
        &self,
    ) -> (Address, Nonce, SettlementAttemptNumber, TxEnvelope) {
        // TODO: Build the next attempt with correct gas and other params. Use https://docs.rs/alloy/latest/alloy/rpc/types/struct.TransactionRequest.html#method.build
        // XREF: https://github.com/agglayer/agglayer/issues/1318
        todo!()
    }
```

PR #1634, opened 2026-07-07
docs(settlement): split the admin api design in two

Ekleog-Polygon:
## What

Splits the settlement admin API design document from #1562 into two:

- **New:** `docs/knowledge-base/src/ai-agents/specs/2026-07-07-settlement-admin-mutations-design.md`
  — the **mutation surface**: insert-attempt, mark-attempt-definitely-failed,
  remove-attempt-result, force-remove-job-result, plus the JSON-RPC exposure of the
  abort/reload task controls. Written as-built: this half is implemented
  (implementation PR to follow separately), and the document records where the
  implementation deliberately deviates from the #1562 draft — append-only insert
  instead of upsert, two new "undo" operations (remove-result, force-remove),
  L1-resolved attempt fields, the `admin` namespace, no separate
  `SettlementAdminWriter` trait / `settlement-admin` cargo feature, no tower
  `AdminCommand` indirection yet, and no audit rows yet.
- **Rewritten:** `2026-06-08-settlement-admin-api-design.md` — now titled
  "Settlement service pause, observability, and audit" and covering only the
  remaining, not-yet-implemented surface: pause/resume with full quiesce and
  trustworthy drain reporting, the observability read surface, and the durable
  admin audit log. Mutation content is replaced by pointers to the mutations
  document (the audit log will cover the implemented mutations when it lands),
  and the background/current-state and phasing sections are refreshed: node
  integration (#1393), startup recovery/reload (#1230), and the execution
  primitives (#1318/#1319/#1321) have all landed since the draft, so the
  remaining phases are unblocked.

The first commit is the mechanical split; the second is the rewrite of the
remainder document, so the two steps can be reviewed separately.

## Why

I'm just cutting the mutations bit off of #1562 to be able to focus on its
implementation.

**Note for reviewers:** I have not fully reviewed the "remainder" half (the
rewritten 2026-06-08 document) myself — treat it as the same draft-for-review it
was in #1562, restated to cover only what is left to build.

Refs: #1254

### DIFF
--- a/docs/knowledge-base/src/ai-agents/specs/2026-07-07-settlement-admin-mutations-design.md
+++ b/docs/knowledge-base/src/ai-agents/specs/2026-07-07-settlement-admin-mutations-design.md
@@ -1,0 +1,311 @@
+# Settlement admin mutations (issue 1254) — design
+
+- Issue: <https://github.com/agglayer/agglayer/issues/1254>
+- Related: PR 1562 (full settlement admin and observability API design).
+  That design is being split in two:
+  this document records the **mutation surface**, which is implemented;
+  the [PR 1562 document](2026-06-08-settlement-admin-api-design.md)
+  keeps the parts that are not implemented yet
+  (pause/resume with full quiesce, observability reads, durable audit log).
+- Status: implemented
+- Date: 2026-07-07
+
+## Context
+
+The settlement service (`crates/agglayer-settlement-service`) drives one L1
+contract call per `SettlementJob` until it lands, retrying across nonces and
+wallets.
+A job is intentionally **infallible**: it has no terminal-failure state,
+so the certificate orchestrator can safely iterate on subsequent certificates.
+Job state is derived from row presence in storage:
+a job is pending while it has no `SettlementJobResult` row,
+and completed once that row is written.
+Every pending job has one live `SettlementTask` driving it
+(spawned at intake or by startup recovery).
+
+Operators need a small set of manual interventions on stored settlement state:
+
+- **Register an attempt the service does not know about** — a settlement
+  transaction submitted out-of-band, or ported from the legacy settlement path
+  (a pre-existing settlement tx hash recorded before the settlement service
+  owned submission).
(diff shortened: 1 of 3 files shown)


Ekleog-Polygon:
Note: "which is implemented" is the AI being a bit over-hyped — I have a first AI-written draft that I need to review much more in-depth because it's certainly not ready yet — but I think it makes sense to leave the text as-is, because the PR will be up hopefully by tomorrow EOD.

docs/knowledge-base/src/ai-agents/specs/2026-07-07-settlement-admin-mutations-design.md
```diff
@@ -0,0 +1,327 @@
...
+  (pause/resume with full quiesce, observability reads, durable audit log).
+- Status: implemented
+- Date: 2026-07-07
+
+## Context
+
+The settlement service (`crates/agglayer-settlement-service`) drives one L1
+contract call per `SettlementJob` until it lands, retrying across nonces and
+wallets.
+A job is intentionally **infallible**: it has no terminal-failure state,
```
Freyskeyd, line 18: Are we sure about that? In production defining an action with an outbound interaction as infallible isn't safe.

How should we recover if a SettlementJob is created with a calldata A, it isn't settled due to RPC issue or any other transient error, and then those calldata becomes invalid. It means that the certificate orchestrator will have proceed and proven certificates based on an invalid settlement.

Maybe you can explain how you imagine handling that?

Ekleog-Polygon: Calldata becoming invalid would mean a terminal revert, wouldn't it? Which is a success from the settlement task's point of view, because the nonce was eventually consumed and we have a terminal result on L1.

Freyskeyd: yeah but the assertion of `so the certificate orchestrator can safely iterate on subsequent certificates.` is not right then

Ekleog-Polygon:
Close like #1562

PR #1635, opened 2026-07-07
refactor(storage): reopen the database with declared column options in finalize

Ekleog-Polygon:
Follow-up to #1630 (now merged) — simplifies how declared column options get applied on existing databases.

## Design

`Builder::finalize()` already receives the authoritative expected schema (previously unused). Instead of threading a `known_cfs` catalog into `Builder::open` so that existing column families open with their declared options, `finalize()` now **unconditionally closes the migration-time handle and reopens the database** with the declared options looked up from `expected_schema` (+ the migration-record CF; RocksDB defaults for on-disk CFs it does not list).

The migration steps themselves run on a plainly-opened handle with default options, which is harmless: CF options only affect the live handle, and the handle the node actually runs with is now always opened with declared options in exactly one place, regardless of which path (fresh create, existing open, mid-migration CF creation) produced the CFs. Cost: one extra close+reopen per store per boot, negligible at startup.

## Reverted from #1630 (no longer needed)

- `known_cfs` parameter on `DB::builder` / `Builder::open` / `open_rocksdb_existing` (back to their original signatures; existing DBs open by name again for the migration-time handle)
- The catalog argument at the four stores' `init_db` call sites, the migration test sample helpers, and the state legacy-schema test

## Kept from #1630

- `DB::descriptor_for_existing` (now used by `finalize` and the read-only path) with its per-CF declared-vs-default debug logs
- `open_cf_readonly` opening with declared options (read-only opens don't go through `finalize`)
- Option logging on CF creation during migrations, and the settlement-task hydration debug logs
- Both reopen regression tests — `list_settlement_attempts_stays_within_job_after_reopen` and `reopened_database_applies_declared_prefix_extractor` — which pass **unchanged**, now guarded by the finalize reopen alone

## Verification

`cargo make generate-proto` (no diffs), `cargo check --workspace --tests --all-features`, `cargo make ci-all`, `cargo nextest run --workspace --no-fail-fast` (829 tests), and `cargo make pp-check-vkey-change` all pass; the two reopen regression tests re-verified green after rebasing onto main (#1630 squash-merged mid-verification). One unrelated flake surfaced along the way: `agglayer-clock block::tests::test_block_clock_starting_with_genesis_already_passed` takes ~12.2s against its 13s rstest timeout even on an idle machine and timed out once under full parallel load; it passes in isolation and on the re-run. Worth loosening that timeout separately.

### DIFF
--- a/crates/agglayer-storage/src/storage/migration/mod.rs
+++ b/crates/agglayer-storage/src/storage/migration/mod.rs
@@ -45,18 +45,12 @@ impl Builder {
     /// initial schema (v0). It automatically sets up migration tracking and
     /// validates the database schema.
     ///
-    /// `known_cfs` catalogs every column family this binary declares across
-    /// all schema versions. Column family options only apply when passed at
-    /// open time, so an existing database is opened with the declared options
-    /// looked up there (RocksDB defaults for on-disk column families it does
-    /// not list).
-    pub fn open(
-        path: &Path,
-        cfs_v0: &[ColumnDescriptor],
-        known_cfs: &[ColumnDescriptor],
-    ) -> Result<Self, DBOpenError> {
+    /// The handle opened here only serves the migration steps; existing
+    /// column families are opened with default options. [`Self::finalize`]
+    /// reopens the database with the declared column options before handing
+    /// it out.
+    pub fn open(path: &Path, cfs_v0: &[ColumnDescriptor]) -> Result<Self, DBOpenError> {
         debug!("Preparing database for initialization and migration");
-        debug!(?known_cfs, "Column families known to this binary");
         let desc_v0: Vec<_> = cfs_v0.iter().map(DB::descriptor).collect();
         let cfs_v0: BTreeSet<_> = desc_v0.iter().map(|d| d.name()).collect();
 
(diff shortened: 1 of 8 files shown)


hadjiszs (approved):

crates/agglayer-storage/src/storage/migration/mod.rs
```diff
@@ -267,16 +247,48 @@ impl Builder {
...
+            .iter()
+            .cloned()
+            .chain([ColumnDescriptor::new::<MigrationRecordColumn>()])
+            .collect();
+        let descriptors: Vec<_> = rocksdb::DB::list_cf(&rocksdb::Options::default(), &path)
+            .map_err(DBError::from)?
+            .into_iter()
+            .filter(|name| name != rocksdb::DEFAULT_COLUMN_FAMILY_NAME)
+            .map(|name| DB::descriptor_for_existing(&name, &known_cfs))
+            .collect();
```
Freyskeyd, line 281: Why do we need this? It would re-open all the outdated CFs from previous migration

Ekleog-Polygon: It runs list_cf, and runs it after the migration happened, so AFAIU it should be correct. Unless I'm missing something?

PR #1663, opened 2026-07-08
feat(settlement): add admin mutations for settlement attempts and job results

Ekleog-Polygon:
> [!WARNING]
> **WIP — published for review and/or takeover.** I have not had the time to
> review all of this yet, and I'm only ~80% confident it's actually good.
> Feel free to review it as-is, request changes, or take the branch over
> entirely.

## What this adds

The minimum settlement admin surface, on the private admin JSON-RPC listener
(`AdminAgglayerImpl`), implementing the mutation slice of the settlement
admin design. The rest of that design (pause/quiesce, observability reads,
audit log) stays with the design-doc split from #1562.

| Method | Semantics |
|---|---|
| `admin_insertSettlementAttempt(job_id, attempt, force?)` | Append one new attempt (insert-only, store-assigned attempt number). Only `txHash` is mandatory: sender/nonce/fees are resolved from L1 when omitted. For porting externally-submitted settlement txs. |
| `admin_markSettlementAttemptDefinitelyFailed(job_id, attempt_number, reason, force?)` | Overwrite the attempt's result with an operator assertion (new `ClientErrorType::AbandonedByAdmin`), bypassing the upgrade-only rule. Terminal for the attempt, never the job; the reloaded task re-drives elsewhere. |
| `admin_removeSettlementAttemptResult(job_id, attempt_number, force?)` | Delete an attempt result so the task re-derives it from L1 (the undo of the above). |
| `admin_forceRemoveSettlementJobResult(job_id)` | Un-complete a job: remove its terminal result, drop the stale watcher, respawn a fresh task. |
| `admin_abortSettlementTask` / `admin_reloadSettlementTask` | Expose the existing task controls. |

Key semantics:

- **`force=true`** (string literal, `admin_forceEditCertificate` style) lets
  the attempt mutations edit a job that already has a terminal result — needed
  to correct attempt results *before* `admin_forceRemoveSettlementJobResult`,
  since the removal immediately respawns the task, which could re-derive and
  re-record the old result from uncorrected attempts.
- Mutations are declarative over stored state (per-job lock in storage), then
  the live task is told to reload; responses report
  `liveTask: notified | absent | notify-failed`.
- The regular result writer keeps an admin-abandoned result over weaker
  client-side notes (on-chain evidence still supersedes) so a
  not-yet-reloaded task can't wedge on the conflict (cf. #1617).

## Known limitations / open points

- **Insert race**: between an admin insert and the live task observing its
  reload, the task can collide with the admin-assigned attempt number and
  panic (job recovers on reload/restart). Documented; the clean fix is the
  pause mechanism from the remaining design.
- Double-settle safety of mark-definitely-failed and force-remove rests on
  the settlement contract's replay protection.
- No audit log; `reason` only survives inside the abandoned result message.

## Design doc

The matching design document (with per-decision rationale) is on
`design/split-settlement-admin-mutations`, splitting the #1562 design in two.

## Verification (as of writing; not re-run since)

`cargo +nightly fmt --check`, clippy (no warnings) on the five touched
crates, `cargo test` for agglayer-types / agglayer-storage /
agglayer-settlement-service, `cargo nextest run -p agglayer-jsonrpc-api` —
all green. Tests for touched files moved from inline `mod tests {}` to
sibling `tests.rs` files.

Refs #1254


Ekleog-Polygon:
Superseded by #1675

Issue #1675, opened 2026-07-10
Settlement Task - Minimal admin API surface


Freyskeyd:
### Associated PRs

- https://github.com/agglayer/agglayer/pull/1694
- https://github.com/agglayer/agglayer/pull/1695
- https://github.com/agglayer/agglayer/pull/1703
- https://github.com/agglayer/agglayer/pull/1705
- https://github.com/agglayer/agglayer/pull/1706
- https://github.com/agglayer/agglayer/pull/1711
- https://github.com/agglayer/agglayer/pull/1713
- https://github.com/agglayer/agglayer/pull/1716
- https://github.com/agglayer/agglayer/pull/1721
- https://github.com/agglayer/agglayer/pull/1722
- https://github.com/agglayer/agglayer/pull/1724
- https://github.com/agglayer/agglayer/pull/1725
- https://github.com/agglayer/agglayer/pull/1726
- https://github.com/agglayer/agglayer/pull/1727
- https://github.com/agglayer/agglayer/pull/1729
- https://github.com/agglayer/agglayer/pull/1730
- https://github.com/agglayer/agglayer/pull/1731

# Issue text

The settlement service runs jobs that sign and submit L1 transactions, track attempts per wallet and nonce, and record terminal results. When something goes wrong on a live network, an operator currently has no way to see or influence a running job short of grepping logs and restarting the node.

## What operators need

The failure modes we expect while stabilizing the service:

1. A job looks stuck. The operator must find it and understand it: which certificate, which wallet, what nonce, how many attempts, what the last error was.
2. A job is wedged for a transient reason (RPC flake, stale cached state). The operator wants to reload it from storage and restart it.
3. A job is blocking a wallet's nonce pipeline and needs to be stopped now.
4. A transaction was handled outside the node: someone bumped gas by hand, or the transaction was replaced and the node still waits on the old hash. The node needs to learn about the external transaction, or be told an attempt will never land so the job re-drives on a fresh nonce.
5. A bug or an operator mistake left a wrong result recorded, for an attempt or for the whole job. Without an undo, the certificate stays wedged until someone edits the database by hand.

## Proposed surface, on the private admin JSON-RPC listener only

Reads carry the highest value, since every other call needs a job id:

- `admin_listSettlementJobs`: job id, certificate id, wallet, state, nonce, attempt count, last error
- `admin_getSettlementJob(job_id)`: full detail including attempts

Task controls; the service already has internal abort and reload-and-restart hooks, this exposes them:

[...]

PR #1681, opened 2026-07-13
feat(settlement): admin groundwork for task controls and job reads

Freyskeyd:
## What

Service- and storage-layer groundwork for the settlement admin floor of #1675 (reads + task controls). No RPC changes yet: this is PR 1 of 3 (PR 2 wires the admin RPC task controls, PR 3 adds the reads). The design doc with per-decision rationale lives on the `design/settlement-admin-floor` branch (same convention as the #1663 design split), not in this PR.

## Behavior changes

1. **Typed admin errors** (`SettlementAdminError`): `admin_abort_task` / `admin_reload_and_restart_task` now classify failures against storage (`JobNotFound` / `JobCompleted` / `NoLiveTask` / `TaskNotResponding` / `ReloadFailed` / `Storage`) instead of returning opaque `eyre` errors, so the upcoming RPC layer can answer operators precisely (distinct error codes and machine-readable tags instead of string matching).
2. **Respawn-capable reload-and-restart**: a pending job whose task is dead (admin abort, failed in-task reload, or a task panic) is now revived by `admin_reload_and_restart_task`. It loads the job from storage and spawns a fresh task, which makes abort → inspect/fix → reload a complete unstick cycle. Concurrent respawns are serialized by a new service-level admin operation lock; a panicked task's stale control registration is detected via its closed command channel and respawned over. The run loop's reload arm now registers the replacement handle only after a successful load, so a failed in-task reload can no longer expose a closed-but-registered handle to a racing admin call. New `has_live_task` accessor exposes task liveness (advisory).
3. **Settlement job → certificate reverse link**: `insert_certificate_settlement_job_id` now populates the (previously never-written) `certificate_id_per_settlement_job_id_cf` column atomically with the forward link in one `WriteBatch`, rejects relinking a job to a second certificate, and the new `SettlementReader::get_settlement_job_certificate_id` exposes the reverse direction.

Service tests moved from the inline `mod tests` to a sibling `settlement_service/tests.rs` (pure move).

## Known limitation

Jobs linked **before** this change lack the reverse row, so `get_settlement_job_certificate_id` returns `None` for them (documented on the reader; the forward column could seed a backfill later if needed).

## Follow-up candidates surfaced during review (pre-existing, not addressed here)

- The run-loop's internal reload arm parents its replacement cancellation token on the **service** token, so an abort racing an in-task reload can be silently ignored (the task keeps running under the fresh handle).
- An RAII deregistration guard for the spawned task future would remove panicked tasks' stale registrations eagerly; with this PR they are recoverable via `admin_reload_and_restart_task`, so the guard is hygiene rather than a gap.
- One-line Display reformat of `agglayer-storage`'s `Unexpected` error, which leaks source indentation into rendered messages.

## Verification

- `cargo check --workspace --tests --all-features` — clean
- `cargo +nightly fmt --check` — clean
- `cargo clippy -p agglayer-settlement-service -p agglayer-storage --all-targets -- -D warnings` — clean (stable toolchain, matching CI)
- `cargo nextest run -p agglayer-settlement-service -p agglayer-storage` — 356/356 passed

Refs #1675


crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -224,9 +229,18 @@ impl StateWriter for StateStore {
...
-            .put::<SettlementJobIdPerCertificateIdColumn>(certificate_id, settlement_job_id)?)
+        let mut batch = WriteBatch::default();
+        self.db
+            .multi_insert_batch::<SettlementJobIdPerCertificateIdColumn>(
+                [(certificate_id, settlement_job_id)],
+                &mut batch,
+            )?;
+        self.db
+            .multi_insert_batch::<CertificateIdPerSettlementJobIdColumn>(
+                [(settlement_job_id, certificate_id)],
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 7f005736: the insert now rejects an existing reverse row before building the batch, so relinking a job to a second certificate fails with `UnprocessedAction` and both directions keep describing the first link. Pinned by `insert_certificate_settlement_job_id_rejects_relinking_a_job`.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -229,43 +236,114 @@ impl<
...
+
+        // Fast path: a live task processes the reload itself.
+        let control = self.task_controls.lock().await.get(&job_id).cloned();
+        if let Some(control) = control {
+            return control
+                .try_send(TaskAdminCommand::ReloadAndRestart)
+                .map_err(|error| SettlementAdminError::TaskNotResponding {
+                    job_id,
+                    reason: error.to_string(),
+                });
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 274b7be2 plus 7b72bad3, with a variant of the suggestion: blindly respawning on a `Closed` `try_send` would double-spawn, because a handle cloned just before the run loop's reload-arm swap reports `Closed` while the task lives on under the fresh handle. Instead the reload checks `is_closed()` on the handle currently registered in the map, under the admin operation lock: a live task always owns its map-handle's receiver, so closed-while-registered means the task died without cleanup, and the respawn path takes over the stale entries. 7b72bad3 closes the one gap in that invariant by registering replacement handles only after a successful in-task reload. Pinned by `admin_reload_and_restart_respawns_after_task_panic`.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -224,9 +229,28 @@ impl StateWriter for StateStore {
...
+                &mut batch,
+            )?;
+        Ok(self.db.write_batch(batch)?)
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 0391ec0b: the check-then-write is now serialized on a store-level lock (`certificate_settlement_link_lock`, same style and poison handling as `with_settlement_write_lock`), so at most one link reservation wins and the reverse index keeps the at-most-once invariant. A 50-round two-thread stress test (`concurrent_link_reservations_for_one_certificate_keep_directions_consistent`) reproduced the double-win deterministically before the fix and pins exactly-one-winner with both directions agreeing after it.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +253,135 @@ impl<
...
+                self.spawn_settlement_task(job_id, task, task_control_handle)
+                    .await;
+                info!(
+                    ?job_id,
+                    "Respawned settlement task via admin reload-and-restart"
+                );
+                Ok(())
+            }
+            Ok(StoredSettlementJob::Completed(_, _)) => {
+                Err(SettlementAdminError::JobCompleted(job_id))
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 358cefac. The `Ok(Completed)` fall-through arm of `admin_reload_and_restart_task` now removes the stale `task_controls` and `result_watchers` entries before returning `JobCompleted`, so `retrieve_settlement_result` falls through to the terminal result in storage instead of serving the panicked task's `None` watcher. Republishing to the existing watcher is not an option: its sender was owned by the dead task and is already dropped, so removing the entry and reading storage is the only path that yields the completed result. The arm is only reachable under `admin_operation_lock` for a closed or absent handle (a live task returns at the open-handle fast path), so no live task is evicted. Pinned by `admin_reload_over_panicked_completed_job_clears_stale_pending_watcher`, which stages the panic-mid-completion state and asserts the post-reload read returns `Completed`.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +253,143 @@ impl<
...
+            self.tx_config.clone(),
+            self.provider.clone(),
+            self.store.clone(),
+            self.wallet_nonce_locks.clone(),
+            task_control,
+        )
+        .await
+        {
+            Ok(StoredSettlementJob::Pending(task)) => {
+                self.spawn_settlement_task(job_id, task, task_control_handle)
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 25f3e854: `request_new_settlement` now holds `admin_operation_lock` across both `SettlementTask::create` (which writes the job row and the certificate→job link) and `spawn_settlement_task`, closing the window where a concurrent `admin_reload_and_restart_task` could observe the job as pending-with-no-live-task and spawn a second task from storage. The reload path already takes the same lock, so respawn and creation are now mutually exclusive. No deadlock: neither `create` nor `spawn_settlement_task` re-takes the lock. Pinned by `request_new_settlement_serializes_against_admin_lock`.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -208,6 +217,10 @@ impl<
                         .await
                         {
                             Ok(StoredSettlementJob::Pending(reloaded_task)) => {
+                                task_controls
+                                    .lock()
+                                    .await
+                                    .insert(job_id, task_control_handle);
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 6e121438: the run loop's reload arm now parents the replacement token on the old task's cancellation token (`task.cancellation_token()`) instead of the service token, and fast-checks that token right after the reloaded task is built. A `child_token()` of an already-cancelled token is born cancelled, so an abort that lands during hydration propagates to the reloaded task at its first control check regardless of timing (before token creation: child born cancelled; after: the fast-check runs the Cancelled cleanup and exits). The old handle stays open during the load (the 7b72bad3 invariant), so the abort still targets the live token. Pinned by `abort_during_in_task_reload_exits_the_reloaded_task`.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +253,135 @@ impl<
...
+            self.provider.clone(),
+            self.store.clone(),
+            self.wallet_nonce_locks.clone(),
+            task_control,
+        )
+        .await
+        {
+            Ok(StoredSettlementJob::Pending(task)) => {
+                self.spawn_settlement_task(job_id, task, task_control_handle)
+                    .await;
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in fc6b229d by making the result channel persist across respawn. The service now stores the `watch::Sender` (not a receiver) per job; `spawn_settlement_task` reuses the registered sender on a respawn instead of minting a new channel, and waiters get their receiver via `sender.subscribe()`. A certificate already in `wait_for_settlement` therefore keeps its receiver connected across an admin abort + respawn and resolves when the respawned task completes, instead of seeing the old sender drop and being mapped to a settlement error. Ordering is safe: the task persists the terminal result to storage before returning `Completed`, the `Completed` arm sends to the watcher then removes the sender, and a late subscriber (after the drop) falls through to the already-written storage result. Pinned by `waiter_before_respawn_still_receives_result` and `retrieve_reads_completed_pending_and_storage_fallback`. The orchestrator and `SettlementServiceTrait` are unchanged.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -153,50 +164,74 @@ impl<
...
         tokio::task::spawn(async move {
             loop {
                 match task.run().await {
                     SettlementTaskRunResult::Completed(result) => {
+                        // Publish the result, then drop the registered sender.
+                        // Waiters using `wait_for(Option::is_some)` observe the
+                        // value set here before the channel closes; a late
+                        // subscriber finds no sender and falls through to the
+                        // completed result in storage.
                         if let Err(error) = result_sender.send(Some(result)) {
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in e96cf4bd: both terminal-publish sites in `spawn_settlement_task` (the `Completed` arm and the reload-arm `Ok(Completed)` sub-arm) now use `watch::Sender::send_replace` instead of `send`. `send_replace` sets the stored value unconditionally (it delegates to `send_modify`, which mutates regardless of receiver count), so a startup-resumed job whose spawn receiver was discarded still records its terminal result in the channel. A `retrieve_settlement_result` that subscribes in the window before the sender is removed now sees `Some(result)` and reports `Completed`; one that subscribes after removal falls through to storage, which was written before the publish. Pinned by `completion_publish_sets_value_with_zero_receivers`, which is RED with `send` (value stays `None`) and green with `send_replace`.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +297,143 @@ impl<
...
+            Ok(StoredSettlementJob::Pending(task)) => {
+                self.spawn_settlement_task(job_id, task, task_control_handle)
+                    .await;
+                info!(
+                    ?job_id,
+                    "Respawned settlement task via admin reload-and-restart"
+                );
+                Ok(())
+            }
+            Ok(StoredSettlementJob::Completed(_, _)) => {
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 4e9c13f6: the fall-through `Ok(Completed)` arm now binds the loaded result and publishes it with `send_replace` on the registered sender before removing it, so a certificate waiter that subscribed before the owning task panicked observes the terminal result and resolves, instead of seeing the channel close and being persisted as a settlement error. The publish and the removal share one synchronous lock scope (no await held across the lock). The arm still returns `JobCompleted` and still clears the stale `task_controls`/`result_senders` entries (the 358cefac guarantee), so a later retrieve with no live waiter falls through to storage. Pinned by `admin_reload_over_panicked_completed_job_delivers_result_to_waiter` (RED without the publish: the waiter's channel closes and `wait_for_result` errors); the existing `admin_reload_over_panicked_completed_job_clears_stale_pending_watcher` still asserts the storage-fallthrough path for a fresh retrieve.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -285,6 +455,18 @@ impl<
...
+        // visible in storage as pending before `spawn_settlement_task`
+        // registers the task; without this lock a concurrent
+        // `admin_reload_and_restart_task` (which takes the same lock) could
+        // observe the pending job with no live task in that window and spawn
+        // a second task from storage, racing two tasks over one job.
+        // XREF: bot r3589631500 on PR #1681.
+        //
+        // No deadlock: `request_new_settlement` calls no admin method that
+        // re-takes this lock, and `spawn_settlement_task` never takes it.
+        let _admin_op = self.admin_operation_lock.lock().await;
```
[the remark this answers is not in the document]
Freyskeyd: Addressed in 6cabc4f8: gas estimation now runs before the lock. `SettlementTask::create` no longer resolves the gas limit internally; `request_new_settlement` resolves it via `resolve_settlement_gas_limit` (on the task's own cancellation token, so an abort/shutdown during estimation still cancels) and only then takes `admin_operation_lock`, holding it across the reserve/save + spawn window alone. An L1 outage that stalls a new submission's `eth_estimateGas` retries therefore no longer blocks `admin_reload_and_restart_task` for other jobs. The 25f3e854 create-vs-respawn guarantee is preserved: the job only becomes storage-visible (reserve/save) under the lock, through task registration, so a concurrent respawn still cannot double-spawn. Pinned by `request_new_settlement_estimates_gas_outside_admin_lock` (RED without the fix: the estimateGas response is never consumed while the lock is held), and `request_new_settlement_serializes_against_admin_lock` now also asserts estimation completed before the request parks on the lock.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -153,50 +164,72 @@ impl<
...
+                        // XREF: bot r3589964568 on PR #1681.
+                        let _ = result_sender.send_replace(Some(result));
+                        result_senders.lock().await.remove(&job_id);
                         task_controls.lock().await.remove(&job_id);
                         break;
                     }
                     SettlementTaskRunResult::Cancelled => {
                         info!(?job_id, "Settlement task cancelled");
-                        result_watchers.lock().await.remove(&job_id);
+                        result_senders.lock().await.remove(&job_id);
```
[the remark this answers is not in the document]
Freyskeyd: The observation is correct: the `Cancelled` arm removes the sender, so a certificate waiting in `wait_for_settlement` sees its watcher close and the orchestrator records a settlement error. This is intended behavior for abort rather than a bug, and it's distinct from the respawn thread (r3589571032): that one was about `admin_reloadAndRestartSettlementTask` preserving waiters, which it does — a reload of a live task reloads it in place and keeps the waiter connected via the reused sender, so the certificate resolves once the reloaded task settles.

`admin_abortSettlementTask` is the "stop this task now, regardless of the waiter" tool; it deliberately makes no resumption promise. Keeping the sender alive on cancel (the suggested fix) would trade the spurious error for a worse failure: a bare abort with no follow-up reload would leave the certificate waiter parked forever, since nothing republishes to that sender (unlike respawn, abort has no guaranteed follow-up). Making abort certificate-preserving properly needs an explicit "aborted" job result that the orchestrator treats as retryable-but-not-terminal, which the settlement service doc explicitly defers ("until the admin API grows an explicit aborted result") and which reaches into the certificate lifecycle this PR keeps untouched.

So the resolution here is documentation, not a code change: the operational guidance is to use `admin_reloadAndRestartSettlementTask` directly to bounce a wedged task without disturbing a waiting certificate, and to use `admin_abortSettlementTask` only when the task must stop regardless. Clarified in the settlement-operations runbook (ffc5eeed): the "A job must stop now" section now states that abort errors an in-flight certificate, that recovering the job does not un-error the certificate, and that reload-directly is the waiter-preserving path. An explicit aborted-result outcome remains a tracked follow-up if the team later wants abort itself to be certificate-preserving.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +297,156 @@ impl<
...
+            }
+            warn!(
+                ?job_id,
+                "Settlement task control channel is closed but still registered; respawning over \
+                 the stale entry"
+            );
+        }
+
+        // No live task: respawn from storage if the job is still pending.
+        let (task_control_handle, task_control) = TaskControlHandle::new(&self.cancellation_token);
```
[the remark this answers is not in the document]
Freyskeyd: Correct, and a real sibling of the in-task-reload fix (6e121438): that one covered the run-loop reload arm where the old handle is live, so parenting the replacement on the old task token propagated the abort. It did not cover this admin-respawn path, where the stale handle is closed/dead, so an abort landing during the awaited `load` cancelled a dead token and was lost while the fresh service-token handle was installed uncancelled.

Fixed in d0cb1dc9 by serializing abort against reload/respawn: `admin_abort_task` now takes `admin_operation_lock` for its whole body, the same lock `admin_reload_and_restart_task` and `request_new_settlement` already hold. An abort therefore runs strictly before or strictly after a respawn's load+spawn window, never inside it: after a respawn it finds the freshly-spawned handle and cancels the real token. This is safe because the only previously-unbounded await under the lock, gas estimation, was moved outside it in 6cabc4f8 (r3590141585), so nothing under the lock can block aborts for long. Lock ordering matches the existing holders (`admin_operation_lock` then `task_controls`), with no re-entry and no cycle.

This also closes the class: the three state-mutating post-startup entry points (`admin_abort_task`, `admin_reload_and_restart_task`, `request_new_settlement`) now all hold `admin_operation_lock`, so no two can interleave their task_controls/result_senders mutation windows; startup recovery runs before the service is shared, and the read paths are read-only. The remaining abort-vs-run-loop-self-removal interaction is the pre-existing, documented, benign completion race (cancel is a no-op on a lingering handle), which cannot lose an abort against a live task. Pinned by `admin_abort_task_serializes_against_admin_lock` (RED without the lock: abort completes while the lock is held) and `admin_abort_after_respawn_cancels_the_respawned_task`.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -34,12 +35,21 @@ pub struct SettlementService<L1Provider, SettlementStore> {
...
     task_controls: Arc<Mutex<HashMap<SettlementJobId, TaskControlHandle>>>,
-    result_watchers:
-        Arc<Mutex<HashMap<SettlementJobId, watch::Receiver<Option<SettlementJobResult>>>>>,
+    /// Per-job result channel senders. Holding the sender (rather than a
+    /// receiver) keeps the channel alive across a task respawn: an admin
+    /// reload replaces the task but reuses the registered sender, so
+    /// certificates already awaiting through a `subscribe()`d receiver keep
+    /// receiving instead of seeing the channel close.
+    /// XREF: bot r3589571032 on PR #1681.
+    result_senders:
```
Ekleog-Polygon, line 44: **[high] A panicked task now leaves its result channel permanently open — waiters hang silently instead of erroring**

Holding `watch::Sender`s in `result_senders` means the map itself keeps every job's channel alive. A task that panics unwinds without running any of the run loop's teardown arms, so its sender stays registered forever: `wait_for_result`'s `wait_for(is_some)` never resolves and never errors, and every later `retrieve_settlement_result` hands out a never-resolving `Pending` watcher.

On main the loop owned the only sender, so a panic closed the channel and the orchestrator's waiter got a prompt "settlement job watcher closed before producing a result" → certificate visibly `InError`. At PR head the certificate stays stuck in `Candidate` and the network task blocks on `receiver.recv()` — a silent per-network settlement stall (`certificate_task.rs:450-456` has no timeout by design).

The trigger is designed-in, not exotic: the `retry!` macro panics on every non-transient error (settlement_task.rs:569-586 — e.g. any non-Tenderly error dialect from `get_transaction_by_sender_nonce`, cf. utils.rs:163-173), as do the result-write `unwrap_or_else(panic!)`s (settlement_task.rs:1801-1808). And the recovery path (`admin_reload_and_restart_task`) isn't wired to any RPC in this PR, so the only remediation today is a node restart.

Suggested fix: make the spawned future's teardown unwind-safe (an RAII guard removing/closing the map entries on drop — see my comment at L201), or watch the JoinHandle and deregister on panic. That restores main's fail-fast while keeping the sender-reuse design for the graceful respawn path.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +297,177 @@ impl<
...
+    ) -> Result<(), SettlementAdminError> {
+        let _admin_op = self.admin_operation_lock.lock().await;
+
+        // Fast path: a live task processes the reload itself. A closed
+        // handle still registered in the map means the owning task
+        // panicked without deregistering: fall through and respawn over
+        // the stale entries.
+        let control = self.task_controls.lock().await.get(&job_id).cloned();
+        if let Some(control) = control {
+            if !control.is_closed() {
```
Ekleog-Polygon, line 403: **[medium] `is_closed()` runs on a clone after the map lock is released — races the in-task reload swap into a duplicate task spawn**

`TaskControlHandle::is_closed`'s own doc states the invariant: "Handles cloned out of the map earlier can report closed while the task lives on under a swapped-in handle; **only the map's current handle supports this conclusion**" (settlement_task.rs:299-301). This call site violates it: the handle is cloned out at L401, the guard drops, and `is_closed()` runs with no lock held. The run loop never takes `admin_operation_lock`, so its reload arm can insert the replacement handle (L261-264) and drop the old receiver (`task = reloaded_task`, L265) in between: a live mid-reload task then reads as panicked, and the fall-through loads the still-pending job and spawns a second task over it. Two tasks settling one job get *different* nonces from `assign_next_nonce_for_wallet` (`l1_nonce.max(max_local+1)`) — duplicate L1 settlement txs — until the `Disconnected` self-stop (settlement_task.rs:846-853) reaps them, likely wedging the job.

The window is a few instructions wide (needs a reload retry racing the swap), so low probability — but the fix is free: evaluate `is_closed()` while still holding the `task_controls` lock (clone only if open). Under the lock the reload arm's insert is excluded, making the panicked-task conclusion sound per the documented invariant.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -224,9 +239,28 @@ impl StateWriter for StateStore {
             )));
         }
 
-        Ok(self
+        if self
             .db
-            .put::<SettlementJobIdPerCertificateIdColumn>(certificate_id, settlement_job_id)?)
+            .get::<CertificateIdPerSettlementJobIdColumn>(settlement_job_id)?
```
Ekleog-Polygon, line 244: **[medium] The job-side relink guard reads only the never-backfilled reverse CF — vacuous for legacy forward-only rows; a V3 backfill step removes both known gaps**

`certificate_id_per_settlement_job_id_cf` has existed since #1599 but nothing ever wrote it before this PR, so any deployment that ran a settlement-linking build between #1599 and this PR holds forward-only rows (C1→J, no reverse row). For those, this check passes vacuously: `insert_certificate_settlement_job_id(C2, J)` commits C2→J + J→C2, leaving both C1→J and C2→J in the forward CF with the two directions disagreeing — violating the contract this PR itself writes in `interfaces/writer.rs` ("must fail if `settlement_job_id` already has a stored certificate id … at most one link ever wins").

Latent today (the sole caller links fresh ULIDs) and not a regression — but it's a trap for exactly the PR-2/3 admin tooling this groundwork exists for, and PR 3's reads will faithfully return `None` for legacy jobs (the documented limitation on the reader).

The repo already has the exact mechanism to fix both at once: a recorded V3 step with `Builder::add_cfs(cfs, migrate_fn)` streaming `SettlementJobIdPerCertificateIdColumn` into the reverse CF (mechanical key/value swap; `stores/migration_helpers.rs` has the streaming precedent, and `DbAccess` already gates writes to the new CFs only). ~30 lines; removes the reader caveat and makes this guard sound for legacy data.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +297,177 @@ impl<
...
+                {
+                    let mut result_senders = self.result_senders.lock().await;
+                    if let Some(sender) = result_senders.get(&job_id) {
+                        sender.send_replace(Some(result));
+                    }
+                    result_senders.remove(&job_id);
+                }
+                Err(SettlementAdminError::JobCompleted(job_id))
+            }
+            Err(error) => match self.store.get_settlement_job(&job_id) {
```
Ekleog-Polygon, line 463: **[medium] The load-failure arm returns without clearing the stale entries it just diagnosed**

When the fast path found a closed-but-registered handle (the `warn!` at L411: "respawning over the stale entry") and the subsequent `load` fails, this arm returns `JobNotFound`/`ReloadFailed` leaving the stale control+sender in place — contrast the Completed arm above and the run-loop's own load-Err arm, which both clean up. While load keeps failing: `has_live_task` stays `true`, `admin_abort_task` returns `Ok(())` cancelling a dead token, `retrieve_settlement_result` hands out watchers nothing will ever publish to, and the abort doc's "the next `admin_reload_and_restart_task` clears it" (L337-339) is false.

Transient failures self-heal on the next retry (the respawn reuses the stale sender, so pre-panic waiters even get the eventual result — leaving the *sender* is arguably right for continuity). The stale *control handle* has no such justification: removing just it in this arm makes `has_live_task` accurate and abort return the correct `NoLiveTask`, while preserving waiter continuity. Also worth a test: `admin_reload_and_restart_load_failure_returns_reload_failed` starts with no stale entry, so the stale-handle + load-failure combination is untested.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -34,12 +35,21 @@ pub struct SettlementService<L1Provider, SettlementStore> {
     store: Arc<SettlementStore>,
     cancellation_token: CancellationToken,
     task_controls: Arc<Mutex<HashMap<SettlementJobId, TaskControlHandle>>>,
-    result_watchers:
-        Arc<Mutex<HashMap<SettlementJobId, watch::Receiver<Option<SettlementJobResult>>>>>,
+    /// Per-job result channel senders. Holding the sender (rather than a
+    /// receiver) keeps the channel alive across a task respawn: an admin
+    /// reload replaces the task but reuses the registered sender, so
+    /// certificates already awaiting through a `subscribe()`d receiver keep
+    /// receiving instead of seeing the channel close.
```
Ekleog-Polygon, line 42: **[medium] This comment's guarantee provably fails for the documented abort→reload recovery pairing**

"certificates already awaiting … keep receiving instead of seeing the channel close" holds only for respawn-over-a-panicked-task (the r3589571032 case `waiter_before_respawn_still_receives_result` models) and live in-task reload. For abort→reload — which `admin_reload_and_restart_task`'s doc calls "the recovery step after `admin_abort_task`" — it fails in **both** orderings: the Cancelled arm removes the sender (L208) before the control (L209), so the respawn path can only ever find an empty entry and `or_insert_with` a *fresh* channel; and if the reload wins the race it just queues into the still-live handle (no respawn at all) while the cancelled task strands the waiter anyway at `break`.

Same waiter outcome as main at abort time, so not a regression — but the consequence deserves a caveat rather than a promise: the pre-abort waiter errors with "watcher closed", the certificate is persisted `InError`, and the respawned task can then settle the job to `Success` on L1 — a cert-InError-vs-storage-Success divergence an operator reconciles manually. Suggest scoping this comment to the cases that hold, and noting the divergence next to the planned "explicit aborted result" (struct doc L26-29), which is the real fix.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -153,50 +164,72 @@ impl<
...
+                        // Publish the result, then drop the registered sender.
+                        // `send_replace` sets the value unconditionally (even
+                        // with zero receivers, as for a job resumed at startup
+                        // whose spawn receiver was discarded), unlike `send`
+                        // which errors and leaves the value `None`. A current
+                        // or late-but-pre-removal subscriber observes the value
+                        // set here; a subscriber after removal finds no sender
+                        // and falls through to the completed result in storage.
+                        // XREF: bot r3589964568 on PR #1681.
+                        let _ = result_sender.send_replace(Some(result));
```
Ekleog-Polygon, line 201: **[medium] Five hand-rolled teardown sites + zero unwind coverage — the RAII guard the PR description defers is cheap now, and it fixes the panic-path hang**

The publish+deregister sequence repeats at L201-203, 208-209, 257-258, 276-278, 288-289 (plus the admin arm at L453-459), with the send_replace-vs-send rationale comment duplicated verbatim (L193-200, 268-275, 444-452). None of it runs on panic — the crate's designed non-recoverable exit (see my comment at L44).

Feasibility is better than "follow-up hygiene": neither map's guard is ever held across an await (L449-451 even relies on that), so both maps can become `std::sync::Mutex`, and a plain Drop guard bound at the top of the spawned block then covers all five exits *and* the unwind path (tokio drops the future's state machine on panic, running Drop). `guard.publish(result)` in the two Completed arms preserves publish-before-remove; the redundant `is_cancelled` fast-exit (L246-250 — its own comment concedes the general mechanism already guarantees exit) folds into the Cancelled arm.

Two riders: the admin Completed arm isn't guard-shaped but shrinks to `if let Some(sender) = self.result_senders.lock().await.remove(&job_id) { sender.send_replace(Some(result)); }` (edition-2021 scrutinee-temporary keeps the guard across the send — same semantics, one lookup instead of get+remove); and with the guard in place, most of the stale-entry detection choreography (L397-416, 439-461) becomes dead code.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -34,12 +35,21 @@ pub struct SettlementService<L1Provider, SettlementStore> {
...
+    /// XREF: bot r3589571032 on PR #1681.
+    result_senders:
+        Arc<Mutex<HashMap<SettlementJobId, watch::Sender<Option<SettlementJobResult>>>>>,
     /// Per-wallet locks serializing the nonce read-to-save window across
     /// concurrent settlement tasks.
     /// XREF: https://github.com/agglayer/agglayer/issues/1597
     wallet_nonce_locks: Arc<WalletNonceLocks>,
+    /// Serializes admin respawn operations so two concurrent
+    /// reload-and-restart calls cannot spawn two tasks for one job.
+    admin_operation_lock: Arc<Mutex<()>>,
```
Ekleog-Polygon, line 52: **[medium, non-blocking] Every race this global lock guards is same-job-id — a per-job lock would make cross-job independence structural before PR 2 builds on this**

The boundedness engineering is real (gas hoist, three lock-blocking tests, the doc proofs) and everything under the lock today is ms-scale sync RocksDB I/O — no correctness issue. But create-vs-respawn, abort-vs-respawn and reload-vs-reload are all races on one `SettlementJobId`, and the repo already has both halves of the alternative: `wallet_nonce_locks`' keyed-async-lock shape and `settlement_write_locks`' `SettlementJobId` key. `create` can take the per-job lock on its freshly generated ULID *before* reserve/save/spawn (the id is minted in-memory before anything is storage-visible), which closes the same window the comment at L498-506 describes.

That deletes the cross-job coupling — a wedged job's reload hydration briefly blocking unrelated submissions, mid-incident, per the gas-hoist rationale that itself says "an unrelated job" (settlement_task.rs:401-406) — and un-load-bears the boundedness proofs, which otherwise need re-verification every time an await lands under any of the three lock holders. Fine as a follow-up before PR 2 wires RPC onto these methods; a per-job map needs an entry-cleanup story, but `settlement_write_locks` shows the project already accepts growth there.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +297,177 @@ impl<
...
+            warn!(
+                ?job_id,
+                "Settlement task control channel is closed but still registered; respawning over \
+                 the stale entry"
+            );
+        }
+
+        // No live task: respawn from storage if the job is still pending.
+        let (task_control_handle, task_control) = TaskControlHandle::new(&self.cancellation_token);
+        match SettlementTask::load(
```
Ekleog-Polygon, line 420: **[medium, cleanup] Third verbatim copy of the 6-arg `SettlementTask::load` invocation — extract `load_stored_job(&self, job_id, task_control)`**

Token-identical at resume (L131), the run-loop reload arm (L233), and here; the match arms differ deliberately, but the invocation doesn't. The "loop has no `&self`" objection doesn't hold: `SettlementService` is `Clone` (Educe over Arcs), so the spawned loop can capture one `self.clone()` instead of the six field clones at L182-187 — itself a simplification. Tests hold 5 more copies, and the fact that tests/admin.rs already grew its own `load_and_spawn_pending_task` helper is the tell. A `load` signature or hydration change currently lands in ~8 places.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -310,11 +533,11 @@ impl<
         &self,
         job_id: SettlementJobId,
     ) -> eyre::Result<RetrievedSettlementResult> {
-        if let Some(watcher) = self.result_watchers.lock().await.get(&job_id) {
-            return match watcher.borrow().as_ref() {
+        if let Some(sender) = self.result_senders.lock().await.get(&job_id) {
```
Ekleog-Polygon, line 536: **[low, rises with PR 2] Post-abort state is "expected + recoverable" on the admin path but still "invariant broken" here**

The fallback below (L565-569 — unchanged code, flagging it because this PR redefines the state) fires on exactly the storage state the PR makes routine: `no_live_task_error` classifies it `NoLiveTask` with a documented recovery, `admin_abort_task` deliberately produces it, and `has_live_task`'s doc names it "wedged and needs reload". The PR even ships tests asserting both readings (admin.rs `…_returns_no_live_task` vs the pre-existing `retrieve_fails_when_pending_job_has_no_running_task`). After an abort, any retrieval re-entry logs ERROR "Settlement service invariant broken" — alert noise that masks real invariant breaks.

Minimal fix: `warn!` + `Err(SettlementAdminError::NoLiveTask(job_id).into())` (or the tagged-eyre equivalent per the review-body discussion) — return type unchanged, no trait ripple, one test message to update.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +297,177 @@ impl<
...
-    #[tracing::instrument(skip_all)]
-    pub async fn admin_abort_task(&self, job_id: SettlementJobId) -> eyre::Result<()> {
-        self.task_control(job_id).await?.cancel();
-        Ok(())
+    /// Whether an in-memory task is currently registered for `job_id`.
+    ///
+    /// Advisory: the answer can race with task completion. A `pending`
+    /// job without a live task is wedged and needs
+    /// [`Self::admin_reload_and_restart_task`].
+    pub async fn has_live_task(&self, job_id: SettlementJobId) -> bool {
```
Ekleog-Polygon, line 368: **[low] `has_live_task` can't distinguish "live, tearing down" from "panicked, stale entry" — the abort doc's poll guidance never terminates for the latter**

For a panicked task the stale registration keeps this `true` indefinitely (nothing clears it but a successful respawn), while `admin_abort_task` on it returns the same `Ok(())` as a real abort. An operator composing the doc directives at L332-337 ("watch the actual teardown through `has_live_task` … retry once `false`") literally polls forever on a panicked wedge. The docs are reconcilable for a careful reader (the retry sentence targets the live-task race), but since this accessor is PR-2/3 groundwork: either document "stays `true` forever after a panic; bound the poll; reload is safe to call unconditionally", or land the RAII deregistration guard (my comment at L201) so the answer becomes truthful.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -240,43 +297,177 @@ impl<
...
+        // Fast path: a live task processes the reload itself. A closed
+        // handle still registered in the map means the owning task
+        // panicked without deregistering: fall through and respawn over
+        // the stale entries.
+        let control = self.task_controls.lock().await.get(&job_id).cloned();
+        if let Some(control) = control {
+            if !control.is_closed() {
+                return control
+                    .try_send(TaskAdminCommand::ReloadAndRestart)
+                    .map_err(|error| SettlementAdminError::TaskNotResponding {
```
Ekleog-Polygon, line 406: **[low] `TrySendError::Full` and `::Closed` collapse into one `TaskNotResponding { reason: String }` — PR 2 can't map them to different RPC codes without string-matching tokio's Display**

Full means "alive but parked" (commands drain only at control checks; the stuck-tx wait between drains defaults up to 600 s and is unbounded during an L1 outage, since `retry_callback_until_success` selects only on cancellation) → retry-later. Closed means "dead, respawn" (the TOCTOU the doc at L385-387 acknowledges) → reload-from-storage. Today both get the same documented reaction ("retrying is safe"), so no concrete failure — but the RPC layer will want RESOURCE_EXHAUSTED-vs-FAILED_PRECONDITION, and `reason: String` forecloses that. Keeping the two cases distinct (two variants, or preserving the mpsc error in an eyre chain per the review body) is cheap now.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -153,50 +164,72 @@ impl<
...
-        self.result_watchers
-            .lock()
-            .await
-            .insert(job_id, result_receiver.clone());
+        // Reuse the registered sender on respawn so waiters that subscribed
+        // before the respawn keep receiving; only create a fresh channel for
+        // a job with no sender yet.
+        let result_sender = {
+            let mut result_senders = self.result_senders.lock().await;
+            result_senders
```
Ekleog-Polygon, line 176: **[low] Registering the sender before the control handle shrinks a spurious-failure window for free**

Between the `task_controls` insert (L167-170) and this `or_insert_with`, a concurrent `retrieve_settlement_result` (which takes no admin lock) finds no sender while storage says pending → ERROR "invariant broken" + hard bail while a respawn is mid-flight. Sender-first makes that window yield a harmless `Pending` watcher instead; admin commands can't observe the reversed window because every spawn caller holds `admin_operation_lock`.

Honest bounds: the gap is tiny, the identical failure exists across the wider create-persist→spawn gap and the whole post-abort period (see my comment at L536), the ordering is inherited from main, and nothing races retrieval in shipped wiring today — cheap hardening, not a bug fix.

Ekleog-Polygon (commented):
Reviewed with my code-review tooling at max effort (10 finder angles → adversarial verification of every candidate → gap sweep). **15 findings posted inline**: 1 high (silent settlement-stall regression on task panic), 6 medium, 8 low. Two additional candidates were explicitly refuted during verification and are *not* posted (CancellationToken chain growth on repeated reloads — tokio-util's tree self-compacts, checked against the vendored 0.7.18 source; and reusing `with_settlement_write_lock` for the link lock — the new lock is deliberately a different, cross-keyed granularity).

---

**Design: use an `eyre` + node-wide `RpcErrorCode` context tag instead of `SettlementAdminError` (#838)**

This re-introduces the per-service typed-error pattern #838 moves us away from — and the enum is erased at this PR's own consumption boundary anyway: `tower::Service<AdminCommand>` has `Error = eyre::Error`, so PR 2 would have to `downcast_ref::<SettlementAdminError>()` and write a settlement-specific mapping. Every admin surface after this one repeats that with its own enum, and that's how RPC error codes drift apart between methods.

Proposal — same D6 rationale ("answer operators precisely"), but on the #838 design:
- One node-wide `RpcErrorCode` enum (semantic: `NotFound`, `AlreadyCompleted`, `NoLiveTask`, `Unavailable`, …) living in `agglayer-types`.
- Admin methods keep returning `eyre::Result<()>` and tag at classification points: `.context(RpcErrorCode::NotFound)`.
- PR 2's RPC layer retrieves the code with `agglayer_errors::match_err!` / `downcast_ref` and maps it to the wire code (`RESOURCE_NOT_FOUND` = −10008 already exists) — one generic mapping shared by every RPC method, jsonrpc and future gRPC alike.

I verified the mechanics (eyre 0.6.12 + agglayer-errors 0.18): the tag downcasts through later `wrap_err` layers, `match_err!` retrieves it, the outermost tag wins on re-tagging (the layer closest to the boundary overrides), and it works on fresh `eyre!` reports for the `Ok(None)` paths. Note the tag's `Display` renders inside the error chain, so keep it a readable fragment ("not found"), not a mnemonic.

Side benefit: `TaskNotResponding { reason: String }` / `ReloadFailed { reason: String }` currently flatten the inner error via `to_string()` — with eyre the chain and spantrace survive to the boundary, which is the other half of #838.

Tradeoff, accepted: no compiler-enforced exhaustiveness — an untagged path falls through to internal error at the boundary. That's fail-closed, and the per-path tests in this PR would assert the tag instead of the enum variant.

Two concrete instances from the inline findings that fall out of this design for free: `SettlementTask::load`'s stringly not-found (settlement_task.rs:1605) is what forces the classifying re-read in the reload Err arm (settlement_service.rs:463) — with tagging, `load` marks `RpcErrorCode::NotFound` at its `Ok(None)` and the re-read disappears; and the `TrySendError` Full/Closed conflation (settlement_service.rs:406) stops mattering once the mpsc error itself is preserved in the chain and tagged distinctly.

crates/agglayer-settlement-service/src/settlement_service/tests/admin.rs
```diff
@@ -0,0 +1,1202 @@
...
+    expect_empty_startup_recovery(&mut store);
+    let job_id = mk_job_id(42);
+    expect_pending_job_reads(&mut store, 42);
+
+    let cancellation_token = CancellationToken::new();
+    let service = mk_service_with_token(Arc::new(store), cancellation_token.clone()).await;
+
+    // Register a live task so the abort would reach the cancel path (rather
+    // than the no-live-task classification) once it acquires the lock.
+    // Determinism assumption: on the current_thread test runtime the spawned
```
Ekleog-Polygon, line 1096: **[low] This determinism comment is false — the spawned task panics against the dead provider during the 200 ms sleep, so the test pins the stale-handle no-op path, not abort-of-a-live-task**

While the abort is parked on the externally-held admin lock, the runtime polls the spawned settlement task; it reaches the `http://127.0.0.1:0` provider and alloy classifies the connect error as `TransportErrorKind::Custom` — retryable only on "429 Too Many Requests" — so the `retry!` macro panics and the task dies without deregistering. The later `admin_abort_task` then "succeeds" by cancelling the dead task's stale closed handle. A regression where abort fails to cancel a *live* task's token would still pass this test.

Fix: park the task instead of killing it (e.g. an `Asserter` with no queued response so the gas call pends), or assert liveness (`!control.is_closed()`) just before the abort, mirroring the epilogue the sibling abort tests use.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -409,6 +440,34 @@ impl<
...
+    /// Resolve the settlement gas limit for `job` before it is persisted.
+    ///
+    /// Thin wrapper over [`resolve_settlement_gas_limit`] exposed so the
+    /// service caller can resolve the gas limit OUTSIDE `admin_operation_lock`
+    /// and then pass the already-resolved job to [`Self::create`]. Keeping gas
+    /// estimation out of the lock means an L1 outage during a new submission
+    /// cannot block an admin respawn of an unrelated job. The retry loop
+    /// observes `cancellation_token` for prompt cancellation.
+    /// XREF: Codex bot r3590141585 on PR #1681.
+    pub(crate) async fn resolve_gas_limit(
```
Ekleog-Polygon, line 452: **[low] Pure pass-through wrapper — call `resolve_settlement_gas_limit` directly**

Argument-for-argument forward with a near-duplicate 7-line doc, and the single call site's turbofish (`SettlementTask::<L1Provider, SettlementStore>::resolve_gas_limit`, settlement_service.rs:490) even has to name a type parameter the wrapper never uses. The free fn is `pub(crate)` (this PR made it so) and the crate's own tests already call it directly; a direct call infers `P` with no turbofish. Delete the wrapper and update the one doc xref in `create` (L465).

crates/agglayer-settlement-service/src/settlement_service/tests/admin.rs
```diff
@@ -0,0 +1,1202 @@
...
+    load_and_spawn_pending_task(&service, job_id).await;
+    assert!(service.has_live_task(job_id).await);
+
+    service
+        .admin_abort_task(job_id)
+        .await
+        .expect("abort of a live task must succeed");
+
+    // The task observes the cancellation asynchronously.
+    tokio::time::timeout(std::time::Duration::from_secs(10), async {
```
Ekleog-Polygon, line 236: **[low] Test-helper consolidation across the new tests**

- This `timeout(10s, while cond { sleep(10ms) })` loop is hand-rolled five times (L236/816/919/1193 + settlement_service.rs:1101, three byte-identical) — a closure-based `wait_until(cond)` in the parent tests module fits all five and could absorb #1680's 30 s `wait_for_recorded_nonces`.
- The "panicked task" staging block (create handle → drop control → insert handle → insert orphan sender) appears 4× (~L588/660/722/1151) → `stage_panicked_task(&service, job_id) -> watch::Receiver<_>` keeps the panic footprint in one place for when an RAII guard changes it.
- `load_and_spawn_pending_task` almost covers the two remaining inline load+spawn copies (L893-915, settlement_service.rs:1072-1097) — parametrize it to take/return the handle and hoist it to the parent module.
- storage `state/tests/settlement.rs:260`: `setup_store()` sits inside the 50-round loop — 50 RocksDB open/close cycles for round-unique ids; hoisting it is a one-liner that saves seconds of CI per run.

Issue #1609, opened 2026-07-03
Settlement-job persistence & recovery hardening (reverse CF + boot skip-and-log)


hadjiszs:
Stack of PRs fixing this issue:
- #1682 
- #1712 

-----

Follow-up from PR #1393 review (reviewer priority per thread):
- **[P1]** https://github.com/agglayer/agglayer/pull/1393#discussion_r3519527758 (the #1599 reverse-mapping CF is never written)
- **[P1]** https://github.com/agglayer/agglayer/pull/1393#discussion_r3519527752 (don't fail node boot on one unloadable job)

Three durability/recovery gaps in the settlement-job persistence layer; items 1-2 fold naturally into the
settlement-service admin RPC work (#1254/#1259/#1408).

## 1. Write the #1599 reverse-mapping CF at link creation
`CertificateIdPerSettlementJobIdColumn` — the job→certificate reverse mapping added by the breaking
storage change in #1599 — has no writer or reader at this head: `insert_certificate_settlement_job_id`
writes only the forward `SettlementJobIdPerCertificateIdColumn`, and the only references to the reverse
CF are its definition and a schema-presence test. Every settlement job created by this PR is therefore
persisted without a reverse link, so whatever consumes the CF later (e.g. job→cert resolution for the
admin API, or incident tooling) finds it empty and needs a backfill migration.
Write the reverse link alongside the forward one at link-creation time (ideally in the same batch), or
record the backfill debt if it's deliberately deferred.

## 2. Don't fail node boot on a single unloadable settlement job
`resume_pending_settlement_jobs` scans every settlement job id ever written and `?`-propagates per-job
load errors from inside the loop (`settlement_service.rs:112-122`); the `await?` makes that a hard
precondition of node boot. One undecodable/corrupt row — the CF is insert-only with no pruning and the
schema is still young (#1599) — means the node can't start at all, not even read-only RPC, until manual
DB surgery. The in-code TODO covers the scan cost but not the fail-closed behavior.
Skip-and-log (with a metric) on per-job load failures, so an unloadable job becomes a loudly-reported
stuck job the admin API can inspect and repair rather than a node that won't boot. (The scan-size half
is tracked in #1592.)

## 3. Recover dangling certificate-to-job links (merged from #1603)
If the node crashes between writing the certificate-to-job-id mapping and saving the job, recovery finds no job and the certificate fails. On resume, recreate the missing job instead of failing.

## Origin
[...]

Issue #1577, opened 2026-06-15
Complete the settlement run-loop before integrating the service (broadcast-before-wait-path hazard)


Freyskeyd:
Tracking/integration-gate issue surfaced by PR #1569 review thread https://github.com/agglayer/agglayer/pull/1569#discussion_r3413659304.

## Context

`SettlementTask::run()` (`crates/agglayer-settlement-service/src/settlement_task.rs`) is a run-loop skeleton assembled across several issues. Now that **#1318 (build a new-nonce attempt)** and **#1321/#1570 (submit to L1)** have both landed, the new-nonce path can:

1. build a real signed EIP-1559 tx (`build_next_attempt_with_new_nonce`),
2. **broadcast it** via `save_attempt_to_db_and_submit_to_l1` → `submit_attempt_to_l1` (`settlement_task.rs:629`),
3. then fall through to `next_overall_deadline()` → `next_attempt_deadline_for_nonce()` which is still `todo!()` → **panic**.

So a real cross-chain settlement transaction can be broadcast and then the task panics **without monitoring or recording its outcome**. (Before #1318, the loop panicked at the `todo!()` build step — i.e. *before* broadcasting — so the merge of build+submit is what unlocks the broadcast-then-panic ordering.)

## Why it is not a live incident yet

The crate is `#![allow(dead_code)]` and there is no production caller of `run()`; existing tests call the builder directly, not `run()`. This issue is the gate to keep it that way until the loop is whole.

## Remaining run-loop leaves (still `todo!()`)

- `next_attempt_deadline_for_nonce` / `next_overall_deadline` — retry/deadline scheduling (no dedicated issue yet; needs one or to be done here).
- `wait_for_any_nonce_on_l1` — #1314.
- `current_result_on_l1_for` — #1382.
- `build_next_attempt_with_nonce` (gas bump for resubmits) — #1319.

## Requirement / acceptance

- The settlement service MUST NOT be wired into production (or exercised via `run()`) until the above land and the loop no longer contains `todo!()` panics.
- A settlement job can go build → submit → monitor → record terminal result end-to-end without panicking.
- An integration test exercises `run()` end-to-end (e.g. against Anvil).
- Remove `#![allow(dead_code)]` from the crate once integrated.

## References

#1318, #1321, #1570, #1314, #1382, #1319; PR #1569.

Ekleog-Polygon:
[...]

PR #1682, opened 2026-07-14
fix(settlement): write reverse job mapping and harden startup recovery

hadjiszs:
Part of #1609. Settlement-job persistence and recovery hardening from the #1393 review ([reverse CF](https://github.com/agglayer/agglayer/pull/1393#discussion_r3519527758), [boot resilience](https://github.com/agglayer/agglayer/pull/1393#discussion_r3519527752)), plus the interrupted-completion TODO folded in from #1577.

**Reverse mapping.** `insert_certificate_settlement_job_id` now writes the certificate-to-job and job-to-certificate mappings in one atomic batch. The settlement storage only exists on test networks so far; jobs written before this change simply lack the reverse link, and nothing reads it yet, so no migration is added.

**Boot recovery.** One unreadable settlement job no longer prevents the node from booting. Startup recovery skips it, logs an error, and counts it in a new `agglayer_node_settlement_recovery_skipped_jobs` metric, so the job can be found and repaired. The count is recorded from `agglayer-node` to keep the service crate's dependency graph unchanged (the isolated feature checks are sensitive to it). A failure to scan the store as a whole still stops the boot. A skipped job that is queried later fails loudly to its certificate task.

**Interrupted completion writes.** A crash between the attempt-result writes and the final job-result write leaves the job looking pending. On restart the run loop re-derives every nonce's outcome from L1 rather than trusting stored state, so the interrupted completion is eventually finished through the loop's normal checks (inclusion, reorg detection, confirmation wait) regardless of the order nonces are scanned in. A still-pending nonce may bump or rebroadcast before the loop reaches the confirmed one, but no completion path bypasses those checks. Interrupted revert completions re-derive from L1 the same way.

Tests: reverse-link storage tests, skip-and-continue startup recovery, interrupted-completion recovery test, metric export test.


crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -117,25 +128,33 @@ impl<
...
+                // An unloadable job must not prevent node boot: skip it and
+                // report loudly so it can be inspected and repaired.
+                Err(error) => {
+                    error!(
+                        ?error,
+                        %job_id,
+                        "Failed to load settlement job during startup recovery; skipping"
+                    );
+                    skipped_jobs += 1;
+                }
```
Freyskeyd: In which case are we falling into this branch?

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -523,6 +523,14 @@ impl<
                 return run_result;
             }
 
+            // A recorded successful attempt result means a previous run
+            // already confirmed settlement per the configured policy but
+            // stopped before finishing the completion writes; finish them
+            // before any submission logic can run.
+            if let Some(job_result) = self.try_complete_interrupted_success().await {
+                return SettlementTaskRunResult::Completed(job_result);
```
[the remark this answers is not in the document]
hadjiszs: Reworked in d04509154. The special resume path is gone. The run loop now simply processes the nonce that has a recorded success first, and completion always goes through the loop's normal checks: transaction on L1, reorg check, confirmation wait. A stale stored receipt can no longer complete a job, because no code completes a job from a stored receipt anymore.

crates/agglayer-settlement-service/src/settlement_service.rs
hadjiszs: This branch runs when a settlement job cannot be read back from the database at startup: the row is corrupt or cannot be decoded. This should never happen in normal operation; realistic causes are a damaged database or manual database edits.

Before this PR, one broken row prevented the whole node from booting. Now the node boots, skips that job, logs an error, and counts it in the `agglayer_node_settlement_recovery_skipped_jobs` metric so it can be found and repaired. This is also explained in a comment on the branch now.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -1692,17 +1692,40 @@ impl<
...
+    /// successful attempt result come first.
+    ///
+    /// Such a result is only written after settlement-policy confirmation,
+    /// so at most the terminal job-result write is missing (a crash between
+    /// the completion writes). Handling that nonce first lets the loop finish
+    /// the interrupted completion through its normal L1 checks before any
+    /// other nonce can submit a new transaction; re-recorded completion
+    /// writes are idempotent no-ops. Interrupted revert completions need no
+    /// ordering: they re-derive from L1 wherever the loop starts.
+    fn nonces_in_processing_order(&self) -> Vec<(Address, Nonce)> {
```
Ekleog-Polygon: **P1 simplification: drop the ordering — the loop already finishes interrupted completions unordered, in the same pass**

Nothing processed before the winning nonce can complete the job or block the loop, so the winner early-returns `Completed` later in the same pass even without this ordering. All the ordering buys is that other pending nonces may skip one extra gas bump each, and only when the process died exactly between the attempt-result write and the job-result write. That corner case doesn't justify a dedicated function plus the order-sensitive test machinery.

Concretely: iterate `all_used_nonces()` as before, delete this function and its unit test, and replace the removed TODO at `write_job_result_to_db` with a one-line note that the normal run-loop re-derivation finishes interrupted completions. Keep a slimmed `run_finishes_interrupted_completion_before_other_nonces` (mock both nonces, don't assert order) — that's the real regression guard for deleting the TODO.

crates/agglayer-storage/src/stores/interfaces/writer.rs
```diff
@@ -70,7 +70,8 @@ pub trait StateWriter: Send + Sync {
 
     fn remove_settlement_tx_hash(&self, certificate_id: &CertificateId) -> Result<(), Error>;
 
-    /// Inserts the settlement job id associated with `certificate_id`.
+    /// Inserts the settlement job id associated with `certificate_id`,
+    /// writing the certificate→job and job→certificate mappings atomically.
```
Ekleog-Polygon, line 74: **P1 followup: "Closes #1609" also closes item 3, which this PR doesn't implement**

Issue #1609 item 3 (recover dangling certificate→job links when the crash lands between this link write and the job save) is untouched here: `list_settlement_job_ids` reads the jobs-CF keys, so a dangling link never enters startup recovery, and the sentence below this hunk — "startup recovery can recreate a missing job from the certificate-side id" — describes behavior that doesn't exist yet. Please split item 3 into its own issue before merging (or change `Closes` to `Part of`), and point this doc sentence at it so it reads as intent, not as implemented behavior.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -42,6 +42,7 @@ pub struct SettlementService<L1Provider, SettlementStore> {
     /// concurrent settlement tasks.
     /// XREF: https://github.com/agglayer/agglayer/issues/1597
     wallet_nonce_locks: Arc<WalletNonceLocks>,
+    recovery_skipped_jobs: u64,
```
Ekleog-Polygon: **P3 nit: return the count from `start()` instead of storing it on the service**

This is one-shot boot state carried forever in a `Clone`-able long-lived struct, read exactly once by `agglayer-node` right after `start()`. Having `start()` return `(Self, u64)` keeps boot statistics out of steady-state service state and drops the `let mut this` reassignment, the getter, and the duplicated doc line on `resume_pending_settlement_jobs`.

crates/agglayer-telemetry/src/settlement/tests.rs
```diff
@@ -140,3 +140,28 @@ fn settlement_metrics_export_the_issue_1676_series() {
         Some(1.0),
     );
 }
+
+#[test]
+fn recovery_skipped_jobs_counter_exports() {
+    let harness = MetricsHarness::install();
+
+    record_settlement_recovery_skipped_jobs(0);
```
Ekleog-Polygon: **P3 test: the zero-export claim is never actually asserted**

The helper's doc says "a zero count still exports the series" — the property that lets alerting distinguish "scan ran, nothing skipped" from "no data". But this test only asserts the sum after `add(0)` + `add(2)`, which passes even if `add(0)` alone exported nothing. Assert the series exists with value `0.0` after the first call (a second `harness.gather()` before the `add(2)`), so the claim is pinned.

Ekleog-Polygon (commented):
Reviewed at a9e8afab (CI green).

The reverse-mapping write and the boot skip-and-log both look right and are well-tested, and I like that interrupted completions now resume through the run loop's normal L1 checks instead of trusting stored state.

Four comments, none blocking:
- **P1 simplification**: the `nonces_in_processing_order` mechanism can go entirely — the loop already finishes interrupted completions unordered, in the same pass.
- **P1 followup**: `Closes #1609` would also close item 3 (dangling cert→job link recovery), which this PR doesn't implement — please split it out first.
- **P3 nit**: return the recovery-skip count from `start()` instead of storing it on the service.
- **P3 test**: the "zero count still exports" claim isn't actually asserted.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: Agreed, removed it. The loop iterates `all_used_nonces()` again, and the function + its unit test are gone. The old TODO at `write_job_result_to_db` is now a one-line note that the run loop re-derives interrupted completions from L1. Kept a slimmed `run_finishes_interrupted_completion_before_other_nonces`: the competing nonce is scanned first, doesn't block, then the winner completes — order no longer asserted.

crates/agglayer-storage/src/stores/interfaces/writer.rs
hadjiszs: You're right, item 3 (dangling cert→job link recovery) isn't implemented here. Reworded the doc so it reads as future work. Rather than split it out, I'll keep #1609 open and change this PR to `Part of #1609` — item 3 lands as a follow-up PR that closes it.

crates/agglayer-settlement-service/src/settlement_service.rs
hadjiszs: Done. `start()` now returns `(Self, u64)`; dropped the field, the getter, the `let mut this`, and the duplicated doc line. `agglayer-node` takes the count from the return value.

crates/agglayer-telemetry/src/settlement/tests.rs
hadjiszs: Good catch. The test now gathers right after `add(0)` and asserts the series is `0.0`, before `add(2)` — so the zero-export claim is actually pinned.

crates/agglayer-storage/src/stores/interfaces/writer.rs
```diff
@@ -70,12 +70,14 @@ pub trait StateWriter: Send + Sync {
...
     ///
     /// This is an insert-only operation and must fail if `certificate_id`
     /// already has a stored settlement job id. The settlement job may be
-    /// created after this link, so startup recovery can recreate a missing job
-    /// from the certificate-side id.
+    /// created after this link; recovering a dangling link whose job save was
+    /// interrupted is intended future work, not implemented here (see #1609
```
Ekleog-Polygon, line 79: P1 issue: Keep #1609 open for dangling-link recovery

Thanks, the documentation now correctly identifies item 3 as future work. The current PR description still starts with `Closes #1609`, though, so merging would close the issue while dangling certificate→job recovery remains unimplemented. Please change it to `Part of #1609`, as proposed in your reply.

crates/agglayer-settlement-service/src/settlement_task/tests.rs
```diff
@@ -904,6 +904,188 @@ async fn write_nonce_revert_replaces_previous_client_error_for_finalized_attempt
...
+    }
+}
+
+#[tokio::test]
+async fn run_finishes_interrupted_completion_before_other_nonces() {
+    // Regression guard for dropping the resumption TODO: a completion
+    // interrupted between the attempt-result write and the job-result write
+    // (the winning nonce carries a recorded success, but the terminal job
+    // result never landed) is finished by the run loop re-deriving it from L1.
+    // A competing pending nonce is scanned first, yet nothing before the
```
Ekleog-Polygon, line 999: P2 doc: Update the interrupted-completion claim after removing ordering

The code now intentionally scans nonces in their normal order, and this test makes the earlier nonce’s wallet unknown so it cannot submit. With a known signer and an expired retry deadline, that nonce may still bump or rebroadcast before the recorded success is reached—the tradeoff accepted in the prior thread. The PR description still says successful nonces run first and no other nonce can submit anything; please update that paragraph to promise eventual completion through the normal L1 checks instead.

Ekleog-Polygon (dismissed):
Reviewed at `19b87d0c` (CI green).

The previous code concerns are addressed: the dedicated recovery ordering is gone, the one-shot recovery count is returned instead of retained in service state, and zero-valued metric export is now tested. The full diff looks sound: reverse mappings are written atomically, unreadable per-job state no longer blocks node boot, and interrupted completions still pass through the normal live L1 checks.

Two PR-description corrections remain: keep #1609 open for dangling-link recovery, and remove the outdated claim that recorded successes are processed before any other nonce can submit.

No code blockers from me.

Ekleog-Polygon (approved):

PR #1683, opened 2026-07-14
feat(config): warn on ignored legacy [outbound.rpc] settlement config

### LINKED ISSUE #1602: Reject/warn on ignored legacy [outbound.rpc.settle-cert] settlement config
Follow-up from PR #1393 review: https://github.com/agglayer/agglayer/pull/1393#discussion_r3508368057

## Context
PR #1393 switched certificate settlement to `config.settlement.pessimistic_proof_tx_config`. The
legacy `[outbound.rpc.settle-cert]` block is still accepted and deserialized
(`crates/agglayer-config/src/outbound.rs`), so operators who tuned confirmations, retry intervals,
or gas/fee settings there now silently run the new settlement service with defaults — changing L1
finality and transaction behaviour with no signal.

## Proposed change
On startup, emit a logged error (or fail config validation) when `[outbound.rpc.settle-cert]` is set
but would be ignored by the new settlement service.

## Origin
Introduced by PR #1393.

hadjiszs:
Closes #1602

### DIFF
--- a/crates/agglayer-config/src/lib.rs
+++ b/crates/agglayer-config/src/lib.rs
@@ -78,7 +78,7 @@ pub struct Config {
     pub rate_limiting: RateLimitingConfig,
 
     /// The configuration for every outbound network component.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "is_default")]
     pub outbound: OutboundConfig,
 
     /// The L1 configuration.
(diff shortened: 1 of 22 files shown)


crates/agglayer-config/src/lib.rs
```diff
@@ -78,7 +78,7 @@ pub struct Config {
     pub rate_limiting: RateLimitingConfig,
 
     /// The configuration for every outbound network component.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "is_default")]
     pub outbound: OutboundConfig,
```
Freyskeyd: Shouldn't this be an `Option` ?

Freyskeyd (approved):

Issue #1632, opened 2026-07-07
Disable sendTx rpc


Ekleog-Polygon:
Have it return an error upon receiving a transaction.

PR #1692, opened 2026-07-21
feat(rpc): remove sendTx dead code and disable getTxStatus

hadjiszs:
Clone https://github.com/agglayer/agglayer/pull/1654
Meant to land after v0.6.0.

Follow-up to disabling interop_sendTx: delete the code that only
existed for the legacy proof-settlement flow, and disable
interop_getTxStatus the same way (error -10009). With sendTx gone
there are no new settlement transactions left to track.

Removed:

- the jsonrpc-api Kernel, the V0 AgglayerService, the SignedTx types
  and the zkevm node client; interop_sendTx now takes an arbitrary
  JSON value and every call gets the -10009 "method disabled" error
- the agglayer-rate-limiting crate (the kernel was its last production
  consumer) and the RPC error codes -10001..-10005 and -10007: their
  `RpcErrorCode` variants (agglayer-types) are deleted and the range
  is documented as retired, never to be reused
- the dedicated tx-settlement signer: node wiring, ConfiguredSigners,
  and the GCP KMS tx key parameters with their env vars
  (GOOGLE_KEY_NAME_TX_SETTLEMENT, GOOGLE_KEY_VERSION_TX_SETTLEMENT)
- the kernel-only telemetry counters verify_zkp, verify_signature,
  check_tx, execute and settle; the send_tx counter stays so leftover
  callers still show up in metrics
- the dead `Settler` trait and its only impl on `L1RpcClient`.
  Certificate settlement goes through the settlement service, which
  encodes calldata with `verify_pessimistic_trusted_aggregator_calldata`
  (kept) and owns gas, fee and nonce handling itself. With the trait go
  the `L1RpcClient` gas fields, `GasPriceParams`, `adjust_gas_estimate`
  and the node wiring that read `[outbound.rpc.settle-cert]` — the last
  read of `[outbound]`, which is now fully ignored, exactly as its
  deprecation warning (#1683) already claims
- `RollupContract::get_event_filter_block_range` and the `L1RpcClient`
  field behind it; its last callers were the settlement-client event
  scans deleted in #1709
- the `L1RpcError` variants UnableToGetTransaction,
  UnableToParseAggchainVkey and FailedToQueryEvents, which lost their
  last constructors with the flows above

None of the removed error variants can appear in storage: `RpcErrorCode`
is serialize-only and did not exist before #1695, `L1RpcError` has no
serde support at all, and the persisted error chain
(`CertificateStatusError`, settlement protos) is untouched.

Part of #1632

CONFIG-CHANGE: The top-level `full-node-rpcs`, `l2` and
`rate-limiting` sections no longer exist; a config file that still
contains them fails to parse at startup. The KMS
`tx-settlement-key-name`/`tx-settlement-key-version` keys, the whole
`[outbound]` section and `[l1] event-filter-block-range` are still
accepted but ignored. Local auth only uses the first private key.

BREAKING-CHANGE: `interop_getTxStatus` no longer reports settlement
status; every call returns JSON-RPC error -10009 ("The
interop_getTxStatus method is disabled").


crates/agglayer-jsonrpc-api/src/kernel/tests.rs
```diff
@@ -1,355 +0,0 @@
...
-#[tokio::test]
-#[ignore = "TODO: Implement full mock setup for alloy provider calls"]
-async fn interop_executor_verify_tx_signature() {}
-
-/// Test that checks if the verify_tx_signature method works with proof signer.
-#[tokio::test]
-#[ignore = "TODO: Implement full mock setup for alloy provider calls"]
-async fn interop_executor_verify_tx_signature_proof_signer() {}
-
-/// Basic tests for the verify_cert_signature method
```
Ekleog-Polygon, line 142: **P2 test: this deletes the repo's only tests that a bad certificate signature is rejected**

The `verify_cert_signature` test here covered PP `Certificate` signatures, not the sendTx flow: a valid signature with a wrong `proof_signers` entry, and corrupted certificate content, both rejected via `Certificate::verify_legacy_ecdsa` — the same method the live certificate-submission path calls (`agglayer_rpc::AgglayerService::verify_cert_signature`, `agglayer-rpc/src/lib.rs:940`). No other test covers rejection: agglayer-types only asserts a valid signature verifies, and neither agglayer-rpc nor grpc-api tests signatures at all.

Deleting the `#[cfg(test)]`-only `Kernel` duplicate itself is right — but please port the two rejection cases to a test on `agglayer_rpc::AgglayerService::verify_cert_signature`, so signature rejection keeps coverage on the live path.

crates/agglayer-jsonrpc-api/src/lib.rs
```diff
@@ -193,40 +181,35 @@ where
...
-                rollup_id.to_string(),
-            )],
-        );
-        warn!(
-            rollup_id,
-            "Rejected interop_sendTx call: method is disabled"
-        );
+        // "method not found". The payload is deliberately not parsed anymore.
+        agglayer_telemetry::SEND_TX.add(1, &[]);
+        warn!("Rejected interop_sendTx call: method is disabled");
```
Ekleog-Polygon: **P2 issue: disabled `interop_sendTx` no longer reports which network is still calling**

The `SEND_TX` counter is kept "to observe residual callers", but this change drops `rollup_id` from the metric and the warn log (main's disabled handler still records it) — so we learn that someone still calls, not *who*.

Suggestion: best-effort extract it from the raw payload for the log line, without reviving `SignedTx`:

```rust
let rollup_id = tx.pointer("/tx/RollupID").and_then(|v| v.as_u64());
warn!(?rollup_id, "Rejected interop_sendTx call: method is disabled");
```

`/tx/RollupID` is the legacy `ProofManifest` wire field, so real leftover clients hit the happy path; garbage payloads just log `None`. I'd keep the metric itself unlabeled — a caller-controlled label is an unbounded-cardinality risk.

crates/agglayer-node/src/node.rs
```diff
@@ -237,15 +206,6 @@ impl Node {
...
                 config.l1.polygon_zkevm_global_exit_root_v2_contract.into(),
-                settle_cert_config.gas_multiplier_factor,
-                {
-                    let gas_config = &settle_cert_config.gas_price;
-                    agglayer_contracts::GasPriceParams::new(
-                        gas_config.multiplier.as_u64_per_1000(),
-                        gas_config.floor..=gas_config.ceiling,
-                    )?
-                },
-                config.l1.event_filter_block_range.get(),
```
Ekleog-Polygon, line 248: **P2 followup: `[l1] event-filter-block-range` is dead — suggest deleting the field in a follow-up PR**

This removes the field's last read, yet in agglayer-config it still looks fully live: no doc marking it ignored, a `10000` default, and it still appears in serialized `Config` output (visible in the updated snapshots).

Since `L1` doesn't use `deny_unknown_fields`, deleting the field is painless: old configs carrying the key keep parsing — it just becomes an ignored unknown key, preserving this PR's "accepted but ignored" contract — while the misleading knob disappears from the struct and its serialized output.

crates/agglayer-config/src/auth.rs
```diff
@@ -77,26 +77,18 @@ pub struct GcpKmsConfig {
...
     /// The key version for PP certificate settlement.
+    ///
+    /// Falls back to the deprecated `key-version` if not specified.
     #[serde(default)]
     pub pp_settlement_key_version: Option<u64>,
-
-    /// The key name for Tx certificate settlement.
-    #[serde_as(as = "NoneAsEmptyString")]
-    #[serde(default)]
-    pub tx_settlement_key_name: Option<String>,
```
Ekleog-Polygon, line 95: **P3 issue: stale KMS tx-settlement settings are silently ignored**

Extra local private keys now produce a warning, but the equivalent KMS settings (`tx-settlement-key-*` and `GOOGLE_KEY_*_TX_SETTLEMENT`) are silently ignored. This can leave operators believing a configured key is still used.

Please emit the same kind of startup warning whenever these obsolete KMS settings are present.

crates/agglayer-config/tests/validate_deserialize.rs
```diff
@@ -4,12 +4,18 @@ use agglayer_config::{assert_toml_snapshot, Config};
 use pretty_assertions::assert_eq;
 
 #[test]
-fn empty_rpcs() {
-    let input = "./tests/fixtures/valide_config/empty_rpcs.toml";
-
-    let config = Config::try_load(Path::new(input)).unwrap();
-
-    assert_toml_snapshot!(config);
+fn removed_sendtx_options_are_rejected() {
```
Ekleog-Polygon: **P3 test: remove this generic unknown-field test**

This only exercises top-level `deny_unknown_fields` against the fixture's first key; it does not meaningfully test the removed sendTx configuration.

Please remove this test and its fixture rather than pinning generic serde behavior.

crates/agglayer-telemetry/src/lib.rs
```diff
@@ -33,35 +33,12 @@ pub use opentelemetry::KeyValue;
 lazy_static! {
     // Backward compatibility with the old metrics from agglayer go implementation
     // Those metrics are not linked to any registry
+    /// Number of calls to the disabled `interop_sendTx` method, kept to
+    /// observe residual callers.
     pub static ref SEND_TX: opentelemetry::metrics::Counter<u64> = global::meter(AGGLAYER_RPC_OTEL_SCOPE_NAME)
         .u64_counter("send_tx")
         .with_description("Number of transactions received on the RPC")
```
Ekleog-Polygon: **P3 nit: update this metric description to match what it counts**

The handler now accepts arbitrary payloads and rejects every call, so this no longer counts "transactions received." Please describe it as the number of calls to the disabled `interop_sendTx` method.

Ekleog-Polygon (dismissed):
Overall, the dead-code removal looks sound and I found no merge-blocking correctness issues. I left several comments around preserving certificate-signature rejection coverage, residual-caller observability, and removing or clearly reporting obsolete configuration surfaces.

crates/agglayer-jsonrpc-api/src/kernel/tests.rs
hadjiszs: Done. Ported the rejection cases to a new test on the live path, `AgglayerService::verify_cert_signature`, in `agglayer-rpc/src/tests/cert_signature.rs`: wrong `proof_signers` entry, corrupted certificate content, plus the valid-signature baseline.

crates/agglayer-jsonrpc-api/src/lib.rs
hadjiszs: Done as suggested: the rollup id is extracted best-effort from `/tx/RollupID` and logged in the warning. The metric stays unlabeled to keep its cardinality bounded.

crates/agglayer-node/src/node.rs
hadjiszs: Done in this PR rather than a follow-up: the field, its default and its serialized output are gone. Since `L1` accepts unknown keys, old configs carrying it keep parsing.

crates/agglayer-config/src/auth.rs
hadjiszs: Done. Both the `tx-settlement-key-*` TOML keys and the `GOOGLE_KEY_*_TX_SETTLEMENT` env vars now emit a startup warning, same wording as the extra-local-keys one.

crates/agglayer-config/tests/validate_deserialize.rs
hadjiszs: Done, test and fixture removed.

crates/agglayer-telemetry/src/lib.rs
hadjiszs: Done: "Number of calls to the disabled interop_sendTx method".

Ekleog-Polygon:
@hadjiszs I didn't review this yet, but FWIW CI is currently failing :smiling_face_with_tear:

PR #1705, opened 2026-07-24
feat(settlement): expose abort/reload task controls on the admin RPC

Ekleog-Polygon:
Expose the existing settlement task controls on the private admin
JSON-RPC listener.

| Method | Semantics |
| --- | --- |
| `admin_abortSettlementTask` | Abort only the live task. |
| `admin_reloadSettlementTask` | Reload live state from storage. |

Abort leaves the stored job pending. Recovery requires a node restart,
which respawns pending jobs. If a live waiter observes the abort, the
certificate becomes `InError` even though the job may settle after
restart; operators must reconcile that divergence manually.

Reload requires a live task and makes it discard in-memory state before
reading storage again. It is the recovery path when later mutations
report different `liveTask` and `notified` outcomes.

Admin errors are mapped generically by downcasting `RpcErrorCode`. Each
variant owns its wire code, with no mapping table or catch-all code.
Untagged reports fail closed as internal errors. The kebab-case error
name is serialized in `data`.

Wire payloads are snapshot-pinned. `AdminAgglayerImpl` carries the
settlement service, and the node clones the service before the
orchestrator consumes its `Arc`.

Part of #1675.


crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -799,4 +847,22 @@ where
...
+            .await
+            .map_err(map_admin_error)
+    }
+
+    #[instrument(skip(self))]
+    async fn reload_settlement_task(&self, job_id: SettlementJobId) -> RpcResult<()> {
+        warn!("(ADMIN) Reloading settlement task for job {job_id}");
+        self.settlement_service
+            .admin_reload_and_restart_task(job_id)
+            .await
```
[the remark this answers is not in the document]
Ekleog-Polygon: This is out of scope for this PR.

crates/agglayer-jsonrpc-api/src/settlement_admin.rs
```diff
@@ -0,0 +1,20 @@
...
+/// contract.
+pub(crate) fn map_admin_error(report: eyre::Report) -> Error {
+    match report.downcast_ref::<RpcErrorCode>() {
+        Some(&code) => Error::Classified {
+            code,
+            message: format!("{report:?}"),
+        },
+        None => {
+            tracing::error!(?report, "Admin operation failed with unclassified error");
+            Error::internal(format!("{report:?}"))
```
hadjiszs, line 17:
```suggestion
            code,
            message: format!("{report:#}"),
        },
        None => {
            tracing::error!(?report, "Admin operation failed with unclassified error");
            Error::internal(format!("{report:#}"))
```

in order to avoid the line number in the snapshot files that will fail as soon as the module get modified

```
      "message": "no live task\n\nCaused by:\n    no live settlement task for pending job 00000000000000000000000005\n\nLocation:\n    crates/agglayer-settlement-service/src/settlement_service.rs:267:25"
```

hadjiszs (dismissed):

crates/agglayer-jsonrpc-api/src/settlement_admin.rs
Ekleog-Polygon: Thanks! I’d rather keep the richer `{:?}` rendering for operators and accept a little more test-only code than lose source locations from the debugging experience. I’ve pushed a test-only normalizer that preserves the full runtime report and replaces only snapshot line/column coordinates with `<line>:<column>`. Does that tradeoff make sense to you too?

hadjiszs (approved):

PR #1711, opened 2026-07-28
feat(storage): admin mutations for settlement attempts

Ekleog-Polygon:
Add storage-only admin methods to append settlement attempts, override
attempt results, and remove attempt results. Keep mutations under the
per-job write lock and gate completed jobs with EditEvenIfCompleted.

Use structured errors so later service and RPC layers can classify
failures without parsing strings. Preserve AbandonedByAdmin over stale
client-side notes while still allowing L1 evidence to replace it,
avoiding the unresolved conflict behavior addressed by #1617.

Document the known attempt-number race with a live task. No service or
RPC consumers are included yet; they follow in the remaining #1675
stack.

Part of #1675.


crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -45,6 +46,90 @@ impl StateStore {
...
+    fn write_settlement_attempt_locked(
+        &self,
+        settlement_job_id: &SettlementJobId,
+        attempt_sequence_number: u64,
+        settlement_attempt: &SettlementAttempt,
+    ) -> Result<(), Error> {
+        let key = SettlementAttemptKey {
+            settlement_job_id: *settlement_job_id,
+            attempt_sequence_number,
+        };
```
hadjiszs, line 63: **🔴 P0: the "never overwrites" guarantee has no guard**

`multi_insert_batch` below is an unconditional put, so the insert-only rule rests entirely on the reverse scan that produced `attempt_sequence_number`. Enforcing it here rather than in the caller means a wrong scan is loud instead of destructive, and a future `admin_*` writer can't forget it.

```suggestion
    /// Writes a settlement attempt and its per-wallet index entry in one
    /// batch, refusing to overwrite an existing attempt. The caller must hold
    /// the job's settlement write lock and have checked its other
    /// preconditions.
    fn write_settlement_attempt_locked(
        &self,
        settlement_job_id: &SettlementJobId,
        attempt_sequence_number: u64,
        settlement_attempt: &SettlementAttempt,
    ) -> Result<(), Error> {
        let key = SettlementAttemptKey {
            settlement_job_id: *settlement_job_id,
            attempt_sequence_number,
        };

        if self.db.get::<SettlementAttemptsColumn>(&key)?.is_some() {
            return Err(Error::UnprocessedAction(format!(
                "Settlement attempt already exists for job {settlement_job_id} and attempt \
                 sequence number {attempt_sequence_number}"
            )));
        }
```

Error string copied from `insert_settlement_attempt`, so its duplicate test still passes. That method's own check at 288-299 becomes redundant: delete it if you like, harmless either way.

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -278,6 +333,30 @@ impl SettlementWriter for StateStore {
...
+                    {
+                        warn!(
+                            %settlement_job_id,
+                            attempt_sequence_number,
+                            ?tx_result,
+                            "Kept admin-abandoned settlement attempt result, dropped weaker \
+                             client-side write"
+                        );
+                        return Ok(());
+                    }
```
hadjiszs, line 354: **🟠 P1: `admin_override_*` can still panic a live task**

The check looks at the *stored* value, but the trouble comes from the *incoming* one.

Realistic case, straight out of the "transaction submitted outside the node" flow: attempt 3 (nonce 42) is stuck, and that nonce was already settled by a transaction the node never sent. The operator records the truth on attempt 3, `ClientError(SettlementSucceededElsewhere)`. The task hasn't reloaded, so it still thinks attempt 3 is pending; its watcher times out and writes `ClientError(Unknown)`. The store won't replace "settled elsewhere" with "don't know", returns `Err`, and `settlement_task.rs:1799` panics on it. The task dies, its command channel with it, so every admin call returns `NoLiveTask` until the node restarts.

The task already guards the mirror image of this in memory (`settlement_task.rs:1775-1786`: "Panicking here would repeat on every restart and wedge the job forever"). The store path has no equivalent, so today only `AbandonedByAdmin` is safe to pass here even though the doc says any result.

```suggestion
                    // A client-side note never overwrites a stronger stored
                    // result: it is derived from a view the writer may have
                    // taken before an admin override or an L1 confirmation
                    // landed. Dropping it (and reporting success) keeps a task
                    // that has not reloaded yet from wedging on a conflict it
                    // cannot resolve (cf. #1617). Two contradictory on-chain
                    // results remain a hard error.
                    if matches!(tx_result, SettlementAttemptResult::ClientError(_)) {
                        warn!(
                            %settlement_job_id,
                            attempt_sequence_number,
                            ?stored_result,
                            ?tx_result,
                            "Kept stored settlement attempt result, dropped weaker client-side \
                             write"
                        );
                        return Ok(());
                    }
```

Knock-ons: drop `ClientError, ClientErrorType` from the import, and `tests/settlement.rs:1091` now gets `Ok`, so assert the stored `contract_call` is unchanged instead.

Or keep it narrow and add one doc line: anything other than `AbandonedByAdmin` needs the task aborted first.

crates/agglayer-storage/src/stores/interfaces/writer/settlement_writer.rs
```diff
@@ -48,11 +69,69 @@ pub trait SettlementWriter: Send + Sync {
...
+    /// This never overwrites an existing attempt. It fails with
+    /// [`Error::SettlementJobNotFound`] if the job does not exist, or with
+    /// [`Error::SettlementJobAlreadyCompleted`] if it already has a terminal
+    /// result and `edit_even_if_completed` is [`EditEvenIfCompleted::No`].
+    ///
+    /// Known limitation: before the live task observes its reload, it may
+    /// collide with the store-assigned attempt number and panic; the job
+    /// recovers on reload or restart. The per-job storage lock serializes
+    /// admin writers, but not admin assignment against task-side numbering.
+    /// Fully closing this race requires the out-of-scope pause mechanism.
```
hadjiszs, line 103: **🟠 P1: does the job really recover on reload?**

I tried to trace the reload path and couldn't get it to work: `TaskControl` owns the admin-command receiver, so it dies with the panicked task and `admin_reload_and_restart_task` gets `Closed` -> `NoLiveTask`. That would leave a node restart as the only thing that re-drives the job, and the planned RPC guidance ("retry with an explicit task reload, or abort the task") would fail on both counts. Am I missing a path that revives it?

If not, does this wording match what you intended?

```suggestion
    /// Known limitation: before the live task observes its reload, it may
    /// collide with the store-assigned attempt number and panic. That panic
    /// kills the task for good: its admin-command channel dies with it, so a
    /// task reload cannot revive it and only a node restart re-drives the job.
    /// The per-job storage lock serializes admin writers, but not admin
    /// assignment against task-side numbering. Fully closing this race
    /// requires the out-of-scope pause mechanism; until then, callers should
    /// abort the live task before inserting into a job that has one.
```

hadjiszs (changes requested):

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Ekleog-Polygon: Fixed in 99aa4742. The duplicate-key guard now lives in write_settlement_attempt_locked, so both the regular and admin insert paths enforce the append-only invariant at the shared write boundary. I removed the now-redundant caller-side check; the existing duplicate-insert test still pins the no-overwrite behavior.

crates/agglayer-storage/src/stores/state/settlement/mod.rs
Ekleog-Polygon: I am keeping this narrow to AbandonedByAdmin, as specified by D7. The intended orchestration for every admin mutation—including AbandonedByAdmin—is abort -> edit -> reload. This keep-rule is only a last-resort safety net for an already-racing task or stale client-side write; it is not permission to edit a live task. Generalizing it to every stored result would silently relax the existing conflict semantics beyond D7. I will make that orchestration requirement explicit in the admin writer documentation.

crates/agglayer-storage/src/stores/interfaces/writer/settlement_writer.rs
Ekleog-Polygon: Your trace is correct for the current intermediate stack: if this race panics the task today, the receiver dies and reload cannot revive it. I am intentionally leaving the wording unchanged in this incremental PR because #1675 already plans the recovery mechanism six PRs later; once the full stack lands, reload will re-drive the job and this documented end-state will be true. Changing it here would create temporary wording that we would immediately reverse. If the remaining #1675 plan changes as the stack evolves, I will revisit this limitation and its documentation.

crates/agglayer-storage/src/stores/interfaces/writer/settlement_writer.rs
```diff
@@ -4,11 +4,32 @@ use agglayer_types::{
...
 
+/// Whether an `admin_*` attempt mutation may touch a job that already has a
+/// terminal result.
+///
+/// Editing a completed job's attempts is normally refused: the job is never
+/// re-driven, so the edit could only create inconsistencies. The exception is
+/// preparing the removal of a wrong terminal result: corrections to attempt
+/// results must land *before* `admin_force_remove_settlement_job_result`,
+/// because the removal immediately respawns the task, which could re-derive
+/// and re-record the job result from the uncorrected attempts.
```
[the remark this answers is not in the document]
Ekleog-Polygon: This is coming in #1713

crates/agglayer-jsonrpc-api/src/error.rs
```diff
@@ -145,6 +149,7 @@ impl Error {
             Self::SendCertificate { .. } => RpcErrorCode::SendCertificate.code(),
             Self::RateLimited { .. } => RpcErrorCode::RateLimited.code(),
             Self::MethodDisabled { .. } => RpcErrorCode::MethodDisabled.code(),
+            Self::Classified { code, .. } => code.code(),
```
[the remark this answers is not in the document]
Ekleog-Polygon: Considering CI passes I'm pretty sure you're wrong

hadjiszs (approved):

Issue #1603, opened 2026-07-03
Recover dangling certificate -> settlement-job links (+ optional job_id field)


hadjiszs:
Follow-up from PR #1393 review:
- https://github.com/agglayer/agglayer/pull/1393#discussion_r3508368059 (recover dangling links)
- https://github.com/agglayer/agglayer/pull/1393#discussion_r3518581826 (carry `job_id` as a field)

## Context
`SettlementTask::reserve_settlement_job_id()` writes the certificate -> job-id mapping *before*
`save_settlement_job_to_db()` writes the job. A crash or write failure between the two leaves a
certificate mapped to a non-existent settlement job. On restart, settlement-service recovery finds
no job to resume, while the certificate task sees the mapping, moves to `Candidate`, and
`wait_for_settlement()` fails with "No settlement job found".

Related: the certificate task looks the job-id up from storage twice on the resume path —
`process_impl` (to decide to resume) and again in `process_from_candidate` (to wait on it).

## Proposed change
- Teach certificate recovery to handle the dangling-link case — recreate the missing job in storage,
  then wait on it — instead of failing the certificate, aligning this window with the broader RocksDB
  non-atomicity model already used elsewhere.
- While reworking this certificate ↔ job-id linkage, consider carrying the job-id as an optional
  field on the certificate task/state (per the review nit) so the resume path reads it once instead
  of querying storage twice. This fits the storage refactoring the code already defers (the TODO
  around the `process_impl` Proven-reboot block in `certificate_task.rs`).

## Notes
Same non-atomicity class as the residual crash window between epoch assignment and
`write_local_network_state` that the P1 finalization-ordering fix intentionally left open. Consider
covering both here.

## Origin
Pre-existing storage/settlement-service atomicity gap; PR #1393 newly documents and depends on this
invariant from the certificate recovery path.

Ekleog-Polygon:
Follow-up triage (settlement end-game restructure, per Leo): tagged Project = Phase 1 - Stable, Priority = P1. Data-consistency gap (dangling cert -> settlement-job links) — high priority within the stable phase.

hadjiszs:
Merged into #1609 (recovery hardening).

PR #1712, opened 2026-07-29
fix(settlement): persist settlement job and certificate link atomically

### LINKED ISSUE #1609: Settlement-job persistence & recovery hardening (reverse CF + boot skip-and-log)

hadjiszs:
Closes #1609 — the last remaining item (dangling certificate→job link recovery, merged in from #1603). Items 1–2 landed in #1682, which this stacks on.

## Problem
`SettlementTask::create` wrote the certificate→job-id link and then saved the settlement job as two separate writes. A crash in between left a certificate pointing at a job that was never saved; on restart the certificate task reads the link, moves to `Candidate`, and settlement fails with "no settlement job found".

## Approach — close the window instead of recovering from it
A new `insert_settlement_job_with_certificate` writes the job and both link directions (certificate→job and job→certificate) in one atomic batch, insert-only. `create` uses it on the certificate path; the no-certificate path keeps the plain `insert_settlement_job`. The intermediate dangling state can no longer exist.

#1603 originally proposed *recovering* the dangling link on resume, to stay consistent with other tolerated non-atomic windows. This instead *eliminates* the window: the job and both links share one store transaction, so prevention is cheaper than resume-path recovery and removes the failure mode rather than tolerating it (in the spirit of item 1's atomic link batch). The settlement store only exists on test networks so far, so there are no pre-existing dangling links to recover. Happy to switch to the recovery pattern if you'd prefer uniformity.

## Cleanup
Removes `insert_certificate_settlement_job_id` (the link-only writer): with atomic creation it has no callers left and could only ever produce the exact dangling state this prevents. The reader `get_certificate_settlement_job_id` is unchanged.

## Tests
- Storage: the atomic write persists the job + both links; an already-linked certificate is rejected with nothing partially written.
- Settlement service: `create` routes the certificate path through the atomic write.
- Certificate orchestrator: the settlement test simulation now mirrors the single atomic call.


crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -425,7 +425,17 @@ impl<
...
         )
         .await?;
-        let id = Self::reserve_settlement_job_id(store.as_ref(), certificate_id).await?;
+        let id = Self::generate_settlement_job_id().await;
+        // Persist the job and its certificate links in a single atomic write so a
+        // crash can never leave a certificate pointing at a job that was never
+        // saved (issue #1609 item 3).
+        match certificate_id {
+            Some(certificate_id) => {
+                store.insert_settlement_job_with_certificate(&id, &job, &certificate_id)
```
Ekleog-Polygon, line 435: **P2 cleanup: the `StateReader + StateWriter` bounds on `SettlementStore` are now dead**

`insert_certificate_settlement_job_id` was this crate’s last `StateWriter` call; with it gone, no `StateReader`/`StateWriter` method is called anywhere in the crate. The bounds survive at `settlement_task.rs:409`, `settlement_service.rs:83/410/437/464`, and `settlement_service_trait.rs:49`, plus the two imports. Dropping them finishes the cleanup this PR already did for `mock_settlement_persisting`, and types the service as touching only settlement column families.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -425,7 +425,17 @@ impl<
...
+            }
+            None => store.insert_settlement_job(&id, &job),
```
Ekleog-Polygon, line 437: **P2 cleanup: drop the `None` path — take `CertificateId` directly and delete `insert_settlement_job`**

The `None` arm is production-dead: `submit_settlement_job` always passes `Some`, and the only other route, tower’s `Service<RequestNewSettlement>`, has no constructors in the repo. Only tests still use the plain insert (`same_wallet_nonce_race.rs`, storage fixtures).

Please make `certificate_id` non-optional through `request_new_settlement`/`RequestNewSettlement`/`create` and delete `SettlementWriter::insert_settlement_job`, porting the test fixtures to the atomic variant — that leaves exactly one way to persist a settlement job. An immediate follow-up PR is fine if the test churn is too much here.

crates/agglayer-storage/src/stores/interfaces/writer/settlement_writer.rs
```diff
@@ -20,6 +21,21 @@ pub trait SettlementWriter: Send + Sync {
         settlement_job: &SettlementJob,
     ) -> Result<(), Error>;
 
+    /// Inserts a settlement job together with its certificate links, in one
+    /// atomic batch: the job, the certificate→job-id forward link, and the
+    /// job-id→certificate reverse link.
+    ///
+    /// Writing all three together means a crash can never leave a certificate
+    /// pointing at a settlement job that was never saved. Insert-only: fails if
+    /// `settlement_job_id` already exists or `certificate_id` already has a
```
Ekleog-Polygon, line 58: **P2 doc: document the single-writer requirement**

The uniqueness check is serialized by job id, not certificate id. This is safe because production calls it only from the certificate task, but the API does not state that assumption. Please document that this method must not be called concurrently with the same `certificate_id`.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -425,7 +425,17 @@ impl<
             &control.cancellation_token,
         )
         .await?;
-        let id = Self::reserve_settlement_job_id(store.as_ref(), certificate_id).await?;
+        let id = Self::generate_settlement_job_id().await;
+        // Persist the job and its certificate links in a single atomic write so a
+        // crash can never leave a certificate pointing at a job that was never
+        // saved (issue #1609 item 3).
```
Ekleog-Polygon: **P3 doc: remove the issue reference**

Please drop `(issue #1609 item 3)`. The explanation is already self-contained, while the issue reference will become stale once closed.

Ekleog-Polygon (dismissed):
Nice direction overall. Persisting the settlement job and both certificate mappings in one atomic batch removes the dangling-link crash window at the storage boundary. I prefer this to adding recovery for a state the store should no longer create.

The remaining comments are non-blocking API cleanup and documentation items, so approving.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: Done — dropped `(issue #1609 item 3)`.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: Done — removed the `StateReader`/`StateWriter` bounds and both imports across the crate; the service now types as touching only settlement column families.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: Agreed — taking you up on the follow-up. Deleting `insert_settlement_job` (and making `certificate_id` non-optional) means porting ~50 call sites, mostly storage-test fixtures — too much churn for this PR. I'll do it in a dedicated follow-up.

crates/agglayer-storage/src/stores/interfaces/writer/settlement_writer.rs
hadjiszs: Done — documented that it must not be called concurrently with the same `certificate_id` (the uniqueness check is serialized per job id, and production only calls it from the single per-certificate task).

crates/agglayer-storage/src/stores/state/settlement/mod.rs
```diff
@@ -150,6 +152,53 @@ impl SettlementWriter for StateStore {
...
 
+    fn insert_settlement_job_with_certificate(
+        &self,
+        settlement_job_id: &SettlementJobId,
+        settlement_job: &SettlementJob,
+        certificate_id: &CertificateId,
+    ) -> Result<(), Error> {
+        let settlement_job: v0::SettlementJob = settlement_job.into();
+
+        self.with_settlement_write_lock(settlement_job_id, || {
```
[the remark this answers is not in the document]
hadjiszs: Real concern in the abstract, but it can't occur in production: there's exactly one settlement task per certificate, so this is never called concurrently for the same `certificate_id`. That single-writer requirement is now documented on the method. Keeping storage job-keyed rather than adding certificate-keyed locking avoids extra lock-ordering surface for a case the caller invariant already prevents.

Ekleog-Polygon (approved):
Re-reviewed the current HEAD (`7eb770c`). The review follow-ups are addressed, and the atomic storage and service-routing checks pass locally. The remaining single persistence-path cleanup is follow-up-only, so approving.

crates/agglayer-settlement-service/src/settlement_task.rs
Ekleog-Polygon: Thanks for scoping this out. Please link the follow-up issue or PR here once opened so the single persistence-path cleanup remains tracked.

PR #1714, opened 2026-07-29
fix(storage): back up the state DB when a certificate is proven

### LINKED ISSUE #1704: Keep making enough on-demand backups with the new settlement service
Before settlement-service, we were making on-demand backups pretty often.

Now, we're only making epoch backups when editing an epoch, and state backups only when triggering write_local_network_state. Which means that if a certificate comes in and gets submitted to L1, then the agglayer-node crashes before settlement / write_local_network_state, we could be in a hard-to-recover situation.

To mitigate that, we should reintroduce backups triggered either just before creating the settlement job, or just after receiving a certificate.

My personal opinion would be to backup just after receiving any certificate (whose signature is valid), so that we're sure we always have all certificates in at least one backup; but I haven't thought about this in-depth yet.

(Noticed by @hadjiszs, thank you!)

hadjiszs:
A certificate can reach L1 and then be missing from every backup: the
settlement job is submitted, the node crashes, and the next backup only
happens once the certificate settles.

This is a regression. The happy path used to reach `Candidate` through
`update_settlement_tx_hash`, which backs up; the settlement service
moved the transition to `update_certificate_header_status`, which does
not.

Back up when a header moves to `Proven`. Settlement is submitted from a
spawned task right after, so this is the last write still ordered ahead
of the certificate reaching L1, and the proof is already on disk by
then. Restoring such a backup can produce a duplicate settlement job,
but at most one of the two txs succeeds and the admin API can repair the
header — recoverable, unlike an L1 tx for a certificate that is in no
backup at all.

Cost: a settled certificate now triggers three state backups instead of
two.

Closes #1704


crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -317,6 +307,14 @@ impl StateWriter for StateStore {
...
                 )?;
             }
+
+            // A `Candidate` certificate has a settlement job that may already have
+            // reached L1, so this is the last point at which a backup can still capture
+            // the certificate, its proof and its job before the settlement outcome is
+            // known.
+            if let CertificateStatus::Candidate = status {
+                self.request_backup(certificate_id);
+            }
```
Ekleog-Polygon: **P1 issue: request the backup before `submit_settlement_job`, not after it**

This request is issued only after `submit_settlement_job` has returned, and the service broadcasts from a spawned task, so the L1 tx can land before the request is even queued. The worst disaster-restore case is then a settlement tx on L1 for a certificate present in no backup at all — the #1704 scenario. A duplicate settlement job after restoring an older snapshot is recoverable by comparison: at most one of the two txs succeeds, and the admin API can repair the header.

Two placements fix the ordering (both keep 3 backups per settled certificate; the request stays fire-and-forget, which is fine given the added margin):
- Gate this trigger on `Proven` instead of `Candidate`. The certifier persists the proof before that status write, so the backup still captures the proof — as late as possible while still ahead of submission.
- Trigger in `insert_certificate_header` instead. Widest margin (the whole proving duration) and every received certificate lands in at least one backup, but the proof isn't captured yet.

crates/agglayer-storage/src/stores/state/mod.rs
```diff
@@ -431,6 +429,17 @@ impl StateWriter for StateStore {
 }
 
 impl StateStore {
+    /// Requests a best-effort backup of the state and pending databases after a
+    /// certificate-scoped write.
+    fn request_backup(&self, certificate_id: &CertificateId) {
```
Ekleog-Polygon: **P1 nit: fold the fourth backup site into the helper**

`write_local_network_state` still has the inline copy of this block, presumably because it lacks a certificate id for the log field. `update_settlement_tx_hash` and `remove_settlement_tx_hash` are `#[instrument(skip(self))]` so their spans already carry the id; if `update_certificate_header_status` got the same attribute, `request_backup` could drop its parameter and all four sites could share it.

crates/agglayer-storage/src/stores/state/tests/backup.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        assert!(
+            backups.try_recv().is_err(),
+            "Moving a certificate to {status} should not request a backup"
+        );
+    }
+}
+
+#[test]
+fn candidate_status_from_settlement_tx_hash_triggers_a_state_backup() {
```
Ekleog-Polygon: **P3 test: consider deleting this test**

The backup it asserts is unconditional in `update_settlement_tx_hash` — the call sits outside the `UpdateStatusToCandidate` branch, so the test passes identically with `::No`, and the name wrongly suggests Candidate-gating logic exists on this path. What it does pin (tx-hash writes request a backup) is pre-existing behavior, already covered by the integration tests' three-backups-per-certificate counts. Suggest dropping it — or, if you'd rather keep unit-level coverage of the tx-hash trigger, rename it to `recording_a_settlement_tx_hash_triggers_a_state_backup`.

Ekleog-Polygon (changes requested):
I reviewed 367463f2. The mechanics check out: settlement jobs, job results and both cert↔job links live in the state DB, the proof in the pending DB, and every backup request snapshots both DBs together — so a backup at `Candidate` does capture the certificate, its proof, and its job. The knowledge-base trigger list matches the four `backup_client.backup` call sites exactly, and the 2→3 count updates in the integration tests are right, including `restore_at(…, 3)`.

The one substantive concern is ordering: the `Candidate` backup is requested only after `submit_settlement_job` has returned, and the broadcast happens on a spawned task, so the worst case of #1704 — a settlement tx on L1 for a certificate present in no backup — is still reachable. See the inline P1 for two placements that close it: after proving (still captures the proof) or at receipt (widest margin).

Thanks for tracking down the regression — the observable `BackupClient` test harness and the knowledge-base addition are both welcome.

crates/agglayer-storage/src/stores/state/mod.rs
hadjiszs: Agreed — the broadcast runs on a spawned task, so the `Candidate` request
could be queued after the tx hit L1. Moved to `Proven`.

Proof is still captured: `insert_generated_proof` runs inside `certify()`,
which returns before `set_status(Proven)`.

Not `insert_certificate_header`: it backs up on every received certificate
(the `InError` storms I wanted to avoid) and the proof doesn't exist yet.
Per-certificate count and order are unchanged, so the two-certificate
integration tests still expect 6 backups and `restore_at(…, 3)` still lands
on certificate1's settled snapshot.

crates/agglayer-storage/src/stores/state/mod.rs
hadjiszs: Done — `#[instrument(skip(self))]` added, `request_backup` takes no argument,
all four sites share it.

crates/agglayer-storage/src/stores/state/tests/backup.rs
hadjiszs: Renamed to `recording_a_settlement_tx_hash_triggers_a_state_backup` and
switched to `::No`, so it states plainly that the trigger is unconditional.
Kept rather than deleted: it's the only unit guard on `request_backup`, now
shared by four sites.

tests/integrations/tests/storage_backups/main.rs
```diff
@@ -93,11 +93,11 @@ async fn recover_with_backup(#[case] state: Forest) {
...
 
-    // Each settled certificate produces two backups (tx-hash known, then
-    // settled). Wait for both so the restore captures the settled state rather
-    // than the earlier tx-hash snapshot, which would leave the certificate
+    // Each settled certificate produces three backups (proven, tx-hash known,
+    // then settled). Wait for all of them so the restore captures the settled
+    // state rather than an earlier snapshot, which would leave the certificate
     // non-Settled after restart.
-    wait_for_backup_counts(&backup_dir.path, 2, 2).await;
+    wait_for_backup_counts(&backup_dir.path, 3, 3).await;
```
Ekleog-Polygon, line 100: **P1 test followup: restore the Proven snapshot itself**

The unit test only proves that a request was enqueued, while these integration changes only count three backups and restore the latest `Settled` snapshot. Neither verifies #1704’s load-bearing invariant: backup 1 contains both the `Proven` header in state and its generated proof in pending before settlement submission.

Please restore backup ID 1 from both `state/` and `pending/` into temporary DBs and assert that the header is `Proven` and `get_proof(certificate_id)` returns the generated proof. This pins the cross-DB ordering the fix relies on.

Ekleog-Polygon (approved):
Reviewed `1870ef6c`. The previous ordering concern is addressed: the backup is now requested at `Proven`, after proof persistence and before settlement submission. The helper and test follow-ups from the prior round are also resolved, and CI is green.

I left one non-blocking test follow-up to pin the restored contents of the new pre-settlement snapshot. Otherwise, this looks good.

tests/integrations/tests/storage_backups/main.rs
hadjiszs: Will tackle that in https://github.com/agglayer/agglayer/pull/1715

PR #1715, opened 2026-07-29
fix(settlement): error-log job build failures, deprecate set-settlement-tx-hash

### LINKED ISSUE #1610: Certificate settlement-path cleanups (verifier-type error log + retire set-settlement-tx-hash)
Follow-up from PR #1393 review (reviewer priority per thread):
- **[P2]** https://github.com/agglayer/agglayer/pull/1393#discussion_r3519527764 (verifier-type failures lost their error-level log)
- **[P2]** https://github.com/agglayer/agglayer/pull/1393#discussion_r3519527767 (remove the now-inert set-settlement-tx-hash admin op)

Two small cleanups on the certificate settlement path.

## 1. Restore error-level logging for verifier-type failures
At the base commit, a failing verifier-type lookup surfaced as `Error::UnableToGetVerifierType` →
`CertificateStatusError::InternalError`, which `process()` logs with
`error!("Internal error in certificate processing")`. The rewrap into `SettlementError` routes it to
the `_ => debug!` arm instead — so an L1 RPC endpoint that's down or misconfigured at settlement-build
time sends every proven certificate to InError with no error-level log line, exactly when a whole
network stops settling. Tooling that branches on the InError kind from `getCertificateHeader` also sees
the classification change.
Either keep this as `InternalError` (it is an infrastructure failure, not a settlement outcome), or add
an explicit `error!` at this site so ops alerting keyed on error-level logs still fires.

## 2. Remove or deprecate the now-inert set-settlement-tx-hash admin op
With settlement resolved exclusively through the cert→job-id link, the Candidate path no longer reads
`header.settlement_tx_hash`, so `admin_forceEditCertificate`'s `set-settlement-tx-hash` operation
(`admin.rs:591`) no longer drives Candidate settlement recovery: a Candidate cert with no job link goes
straight back to InError regardless of the supplied hash, and one with a link ignores it in favor of
the job's terminal result. (Not a complete no-op — the op still writes the header, and the hash still
feeds the Proven arm's recompute block-pinning — but the incident-runbook use is gone.) The endpoint
still accepts it without warning, and the pre-existing `admin_fixup_tx_hash` test
(`send_certificate.rs:399`) only runs with `process-now = false` and asserts the header field, so
nothing surfaces the change.
Recovery tooling for stuck settlements is being handled by the settlement-service admin API under design
(#1254/#1259/#1408), so no replacement is needed here — remove or deprecate `set-settlement-tx-hash`
(e.g. reject it for certificates whose settlement is job-managed) so operators aren't led into a no-op
runbook during an incident.

## Origin
Both introduced by this PR: the error rewrap (#1) and the job-id-driven Candidate path (#2).

hadjiszs:
Closes #1610. Two cleanups from the #1393 review.

1. Settlement-job build failures now log at error level. They all returned
`SettlementError`, which only reached `debug!`, so a bad L1 endpoint sent every
proven certificate to `InError` with nothing for ops alerting to catch. Job
submission failures are classified as `InternalError` instead, so `process`
logs them at error level after its cancellation check rather than at the call
site — a graceful shutdown cancels submission, and that expected error should
not page anyone.

2. `admin_forceEditCertificate`'s `set-settlement-tx-hash` no longer pretends to
recover settlements. Settlement resolves through the certificate to job link, so
the operation could not fix a stuck settlement, but it still wrote the header and
looked like it worked. It is now rejected for every certificate, and marked
retired in the rustdoc. The rejection happens while parsing the operation, before
the certificate header is read, so having no settlement job does not exempt it.

Tests: new `admin_set_settlement_tx_hash_is_retired`, covering both setting and
clearing the hash. fmt, clippy, and the suites of both touched crates pass.

BREAKING-CHANGE: `admin_forceEditCertificate` rejects `set-settlement-tx-hash`
  with `INVALID_PARAMS` for every certificate, instead of writing the header
  field. This includes certificates with no settlement job, where the operation
  previously pinned the L1 block used when reprocessing a `Proven` certificate;
  there is no replacement for that use.


crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -151,9 +151,21 @@ pub(crate) trait AdminAgglayer {
     /// `null` to represent the absence of a hash.  Any other value must be a
     /// hex-encoded [`SettlementTxHash`] (a 32-byte keccak digest).
     ///
+    /// **Deprecated as a settlement-recovery tool.**  Settlement is resolved
+    /// through the certificate -> settlement-job link, not through this header
+    /// field, so editing the hash can neither restart nor redirect an in-flight
+    /// settlement; it is rejected for a certificate that has a settlement job
+    /// (see [Guards](#guards)).  Recovering a stuck settlement is a matter of
+    /// acting on the job, not on this field.  What the operation still does,
+    /// for a certificate with no job, is header bookkeeping plus pinning
```
Ekleog-Polygon: **P1 issue: Reject the operation when there is no job too**

The documented no-job use case appears unreachable. `CertificateTask::process_impl` handles every `Proven` certificate before its status match: with a job it resumes through `Candidate`, while without a job it resets the certificate to `Pending` and removes its proof. Consequently, the later `Proven` arm that passes `header.settlement_tx_hash` to `recompute_state` cannot use this value.

Allowing this operation without a job therefore preserves the no-op recovery path that #1610 aims to retire. Please reject `set-settlement-tx-hash` for every certificate rather than retaining the no-job exception.

Ekleog-Polygon (commented):
Reviewed at `20e48637`.

#1610 moves the settlement path in the right direction, but two operational gaps remain: the no-job exception preserves an obsolete recovery operation whose value is no longer consumed, and settlement-job submission failures still bypass error-level alerting.

Recovery should go through the settlement admin API, while both job construction and submission should use the shared `log_err` helper for consistent observability.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -322,7 +322,13 @@ where
...
         // atomically when creating the job and rejects duplicates, so the
         // orchestrator does not persist the link itself.
-        let job = self.build_settlement_job().await?;
+
+        // Infrastructure failures (L1 unreachable, missing proof) that stop a whole
+        // network from settling: `SettlementError` only reaches `debug!` in `process`.
+        let job = self
+            .build_settlement_job()
+            .await
+            .inspect_err(|error| error!(?error, "Failed to build the settlement job"))?;
```
Ekleog-Polygon: **P2 issue: Log settlement-job submission failures at error level too**

The same alerting gap remains after job construction. `submit_settlement_job` performs L1 gas estimation and persists the certificate-to-job link and job; failures are converted to `SettlementError`, which `process` logs only at `debug!`. An unavailable L1 endpoint or storage failure here can therefore stop settlement without an error-level signal.

Please log submission failures too. `agglayer_errors::ResultExt::log_err` can cover both this path and the new build path without the verbose `inspect_err` closure.

crates/agglayer-jsonrpc-api/src/admin.rs
hadjiszs: Agreed. Rejected for every certificate now, and the variant with its precondition and apply arms is gone.

crates/agglayer-certificate-orchestrator/src/certificate_task.rs
```diff
@@ -322,11 +323,15 @@ where
...
-        let job = self.build_settlement_job().await?;
+        let job = self
+            .build_settlement_job()
+            .await
+            .log_err("Failed to build the settlement job")?;
         let job_id = self
             .settlement_service
             .submit_settlement_job(certificate_id, job)
             .await
+            .log_err("Failed to submit the settlement job")
```
Ekleog-Polygon: **P2 followup: Preserve shutdown cancellation suppression**

`submit_settlement_job` returns `"settlement gas estimation cancelled before completion"` when the settlement service’s global cancellation token fires. This call logs that expected shutdown error before `process()` can apply its deliberate `cancellation_token.is_cancelled()` suppression, potentially generating operational alerts during graceful shutdown.

Please remove the call-site log and classify submission infrastructure failures as `InternalError`, allowing `process()` to log them only after its cancellation check.

crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -638,34 +630,16 @@ where
                         from: parse_status(from_status)?,
                         to: parse_status(to_status)?,
                     })
-                } else if let Some(operation) =
-                    operation.strip_prefix("set-settlement-tx-hash,from=")
+                } else if operation
+                    .strip_prefix("set-settlement-tx-hash")
```
Ekleog-Polygon, line 764: **P3 doc: Align the PR description with unconditional retirement**

This parser now rejects `set-settlement-tx-hash` for every certificate, before even loading its header. The PR description still says it is rejected only for job-managed certificates and remains allowed without a job; its test summary also names the removed job-managed-only test. The `BREAKING-CHANGE` section therefore understates which clients are affected.

Please update the PR description, test summary, and `BREAKING-CHANGE` section to say that this operation is always rejected.

Ekleog-Polygon (dismissed):
Reviewed at `21b2f75b`.

The earlier review points are addressed: `set-settlement-tx-hash` is now retired unconditionally, and settlement-job submission failures receive error-level visibility. The remaining comments are non-blocking follow-ups concerning graceful-shutdown logging and the outdated PR description.

CI is green for this revision. GitHub currently reports a merge conflict with `main`.

Ekleog-Polygon (approved):
Reviewed at `a2029f67`.

The previous feedback is addressed: settlement-job submission failures now receive error-level visibility after the graceful-shutdown cancellation gate, and the PR description accurately documents unconditional retirement of `set-settlement-tx-hash`.

The follow-up from #1714 now restores backup 1 from both state and pending storage and verifies that it contains a recoverable certificate header and generated proof. Accepting either `Proven` or `Candidate` is consistent with the intentionally fire-and-forget backup request.

CI is green for this revision.

Issue #1321, opened 2026-01-09
Helper function to submit a given settlement attempt to L1


Ekleog-Polygon:
If there is already a function, list it here.
If not, write one and list it here.

This will not wait for settlement in any way, just submit it (and maybe just retry in case of network failure?)

Ekleog-Polygon:
```rust
    async fn submit_attempt_to_l1(&self, _tx: TxEnvelope) -> eyre::Result<()> {
        // TODO: Submit attempt to L1. Use https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.send_tx_envelope
        // XREF: https://github.com/agglayer/agglayer/issues/1321
        todo!()
    }
```

PR #1717, opened 2026-07-30
feat(settlement): emit attempt and duration metrics, add the jobs gauge

### LINKED ISSUE #1676: Settlement Task - Minimal metrics surface
### Related PR

- https://github.com/agglayer/agglayer/pull/1717

----
The settlement service emits no metrics of its own. Certificate-level timing survived the settlement rework (`agglayer_certificate_stage_duration_seconds` and the end-to-end duration still emit when a certificate settles), but the service running the jobs is a black box: no way to see job states, error rates, or transaction churn without reading logs.

## What QA and ops must be able to answer from metrics alone

1. Is anything stuck right now? How many jobs sit in each state, per wallet, and is that number moving?
2. Are transactions landing cleanly or churning? How often do we submit, bump gas, or replace a transaction?
3. Is the nonce pipeline healthy? Concurrent jobs on one wallet are new in this design, and nonce contention is its main failure mode. A rising nonce-error rate must be visible the minute it starts, not after log archaeology.
4. How long do settlements take? The certificate stage histogram already approximates this, so the service-level answer only needs to add what the certificate view cannot see, such as the wallet dimension.

## Proposed instruments, in value order

A new `settlement` module in `agglayer-telemetry`, following the conventions of the existing `certificate` module.

The floor, not negotiable:

- `agglayer_node_settlement_jobs`: gauge of jobs by `state` and `wallet`. Answers question 1.
- `agglayer_node_settlement_attempt_errors_total`: counter by `kind` (`nonce_too_low`, `underpriced`, `rpc`, `other`) and `wallet`. Answers question 3.

Next, cheap once the module exists:

- `agglayer_node_settlement_attempts_total`: counter by `kind` (`submission`, `gas_bump`, `replacement`). Answers question 2.

First cut if scope must shrink, since the certificate stage histogram approximates it:

- `agglayer_node_settlement_job_duration_seconds`: histogram by `outcome` and `wallet`, job creation to terminal state. Answers question 4 with the wallet dimension.

Plus one continuity check that costs nothing: an integration assertion that the certificate stage metrics still emit on the settlement-service path, so a future refactor cannot orphan them silently.

## Out of scope

Dashboards and alert rules, storage and runtime stall instrumentation (a separate concern with its own scope), storage-metric isolation refactors, audit or event logs, tracing changes.

## Done when

- An operator watching metrics can spot a stuck job, a nonce-error spike, and abnormal transaction churn without opening logs.
- The instruments live in `agglayer-telemetry`, are emitted by the settlement service, and are covered by tests.
- An integration test asserts the certificate stage histogram still emits when a certificate settles through the settlement service.

### LINKED ISSUE #1687: Settlement metrics: attempt-kind and error classification, richer gauge states
Follow-up to #1676, capturing the settlement-metrics work deliberately deferred from the clean-slice iteration (jobs gauge + submission/rpc/other/duration emission).

## Context

The first iteration (design: `docs/knowledge-base/src/ai-agents/specs/2026-07-16-settlement-metrics-design.md`) ships:

- `agglayer_node_settlement_jobs{state="running"}` live-occupancy gauge.
- Emission for `record_settlement_attempt(Submission)`, `record_settlement_attempt_error(Rpc | Other, wallet)`, and `record_settlement_job_duration(outcome, wallet, seconds)`.

Three pieces were left out because they need extra work beyond following the certificate-task emission pattern.

### 1. `gas_bump` and `replacement` attempt kinds

`record_settlement_attempt` accepts `SettlementAttemptKind::{Submission, GasBump, Replacement}`, but only `Submission` is emitted today. The bump-versus-replacement distinction exists inside `build_next_attempt_with_nonce` (bump when live tx fees are present, replacement when not), but is not surfaced to the broadcast call site `save_attempt_to_db_and_submit_to_l1`. Emitting these requires threading a `kind` value out of the build functions to the submission funnel.

Answers issue #1676 question 2 (transaction churn) in full.

### 2. `nonce_too_low` and `underpriced` error classification

`record_settlement_attempt_error` accepts `SettlementAttemptErrorKind::{NonceTooLow, Underpriced, Rpc, Other}`, but only `Rpc`/`Other` are emitted. No RPC error-string or error-code inspection exists today; `ClientErrorType` has no such variants.

Answers issue #1676 question 3 (nonce pipeline health), the issue's headline failure mode.

### 3. Richer live gauge states

The gauge reports a single live state (`running`) today. Finer live states (`pending`, `awaiting_inclusion`) live only transiently inside `SettlementTask::run` and are not exposed on `TaskControlHandle` or in the service map. Surfacing them requires instrumenting the run-loop phase machine with a shared status field, which carries drift risk on future run-loop changes. Only add these if operators find the coarse `running` count insufficient to distinguish a healthy in-flight job from a wedged one.

## Done when

- Settlement submissions emit `gas_bump` and `replacement` kinds where applicable.
- Nonce and gas-price errors are classified as `nonce_too_low` / `underpriced` rather than collapsing to `rpc`/`other`.
- (Optional, demand-driven) the jobs gauge exposes finer live states with run-loop instrumentation.
- New emission paths are covered by tests.

## Out of scope

Same as #1676: dashboards, alert rules, storage/runtime stall instrumentation, storage-metric isolation, audit or event logs, tracing changes.

hadjiszs:
The settlement service declared these instruments in #1679 but nothing ever
called them. This wires the emission and adds the missing jobs gauge.

| Series | Labels |
| --- | --- |
| `agglayer_node_settlement_jobs` | `state`, `wallet` |
| `agglayer_node_settlement_attempts_total` | `kind` |
| `agglayer_node_settlement_attempt_errors_total` | `kind`, `wallet` |
| `agglayer_node_settlement_job_duration_seconds` | `outcome`, `wallet` |

`state` is `building` (nothing broadcast yet) or `submitted` (a transaction of
this job is live on L1) — a stuck job in each means a different incident.

The gauge is sampled from the live task registry on every scrape rather than
tallied as jobs change state, so it cannot drift when a task exits early.

Closes #1676
Closes #1687


crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -177,6 +178,54 @@ pub enum RetrievedSettlementResult {
...
+    /// Derived from the live task registry on every call rather than from a
+    /// tally kept across transitions, so it cannot drift away from the tasks
+    /// that actually exist. `TaskControlRegistrationGuard` keeps that registry
+    /// honest on every exit path, including an unwind.
+    ///
+    /// Every live job is labeled with the current default signer: that is the
+    /// wallet each of them will use for its next attempt, whatever wallet any
+    /// earlier attempt of theirs went out on.
+    pub fn job_state_samples(&self) -> Vec<SettlementJobStateSample> {
+        let wallet = self.provider.default_signer_address().to_string();
```
Ekleog-Polygon: P1 issue: Remove the wallet label from the jobs gauge

A settlement job can contain attempts from multiple wallets after signer rotation, but this attributes every live job to the current default signer. The resulting per-wallet counts are incorrect, and there is no single truthful wallet to assign to a job that has attempted settlement through multiple signers.

Please remove `wallet` from `SETTLEMENT_JOBS` and `SettlementJobStateSample`. If wallet-level monitoring is needed, it belongs on attempt-level instruments, where each observation has an unambiguous sender.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -183,6 +187,36 @@ fn submission_outcome(
...
+/// attempt. Nonce and fee rejections are split out from the generic `rpc`
+/// bucket so nonce contention is legible on its own, which is the failure
+/// mode concurrent same-wallet jobs are most exposed to.
+///
+/// The node's error strings are matched case-insensitively on the substrings
+/// geth and its derivatives use (`nonce too low`, `... underpriced`); an
+/// unrecognized JSON-RPC error response is `rpc`, and anything that is not a
+/// transport error at all is `other`.
+fn classify_attempt_error(error: &eyre::Error) -> SettlementAttemptErrorKind {
+    let Some(transport_error) = error.downcast_ref::<TransportError>() else {
```
Ekleog-Polygon: P1 issue: Remove the lossy submission-outcome wrapper

`submission_outcome` maps `RetryCallbackError<TransportError>` into a second private enum and boxes the transport error in `eyre::Error`. This classifier must then immediately downcast it to recover the original type; the failed-downcast/`Other` branch cannot occur in production.

Please remove `submission_outcome`, return `RetryCallbackError<TransportError>` from `submit_attempt_to_l1`, and match `Cancelled`/`Error` at the caller so the `TransportError` can be classified directly. Only format or wrap it at the logging/storage boundary.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -669,6 +754,18 @@ impl<
                     if deadline > SystemTime::now() {
                         continue 'nonces; // wait for deadline to be reached
                     }
+                    // Mirrors the branch `build_next_attempt_with_nonce` takes
+                    // on the same read: a live tx in the mempool is out-bid
+                    // (a gas bump), while a nonce whose every attempt errored
+                    // on broadcast is re-broadcast at fresh base fees.
+                    let attempt_kind = if self
```
Ekleog-Polygon: P1 issue: Return the attempt kind from the builder

This duplicates the decision inside `build_next_attempt_with_nonce`, which independently reads the pending fees to choose between bumping and rebuilding at base fees. That creates a second source of truth: a future change could build one kind of transaction while reporting the other kind in metrics.

Please return `SettlementAttemptKind` from `build_next_attempt_with_nonce` alongside the attempt number and transaction, based on its existing `live_tx_fees` decision.

crates/agglayer-telemetry/src/settlement.rs
```diff
@@ -88,6 +91,68 @@ pub enum SettlementJobOutcome {
...
+/// The state of a live settlement job, rendered as the `state` label value
+/// on [`SETTLEMENT_JOBS`].
+///
+/// The two states separate the incidents an operator must tell apart: a job
+/// stuck in [`Building`](Self::Building) never reached L1 (gas estimation,
+/// nonce assignment, or signing is wedged), while a job stuck in
+/// [`Submitted`](Self::Submitted) is broadcast and waiting on inclusion.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, strum_macros::Display)]
+#[strum(serialize_all = "snake_case")]
+pub enum SettlementJobState {
```
Ekleog-Polygon: P1 issue: Keep the jobs gauge to live occupancy

#1687 described richer live states as optional and demand-driven. `building` is not a reliable incident signal here: gas estimation happens before the task is registered, a non-recoverable build failure terminates and removes the task, and a healthy transition can easily occur between scrapes. The phase split therefore cannot reliably expose the failures its documentation promises.

Please remove `Building`/`Submitted` and the `state` label, and export a single live-job count derived from the task registry. Build and broadcast failures are better diagnosed through the attempt-error metrics and logs.

crates/agglayer-node/src/node.rs
```diff
@@ -285,6 +285,7 @@ impl Node {
         agglayer_telemetry::settlement::record_settlement_recovery_skipped_jobs(
             recovery_skipped_jobs,
         );
+        crate::metrics::register_settlement_job_metrics(&settlement_service);
```
Ekleog-Polygon, line 288: P2 test: Cover certificate-stage metric continuity

This PR closes #1676, whose completion criteria explicitly require an integration assertion that `agglayer_certificate_stage_duration_seconds` still emits when a certificate settles through the settlement-service path. I could not find that assertion in the current revision; the added tests stop at the settlement task, service, and telemetry helpers.

Please add the required certificate-orchestrator/settlement-service integration test before closing #1676.

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -183,6 +187,36 @@ fn submission_outcome(
...
 
+/// Buckets a failed broadcast for the attempt-error counter.
+///
+/// Only non-transient failures reach here: transient ones are retried inside
+/// [`SettlementTask::submit_attempt_to_l1`] and never surface as a failed
+/// attempt. Nonce and fee rejections are split out from the generic `rpc`
+/// bucket so nonce contention is legible on its own, which is the failure
+/// mode concurrent same-wallet jobs are most exposed to.
+///
+/// The node's error strings are matched case-insensitively on the substrings
```
Ekleog-Polygon: P2 doc: Declare the classifier best-effort and metrics-only

This relies on non-standard English error text from specific client dialects and will silently classify unfamiliar wording as generic RPC failure. That is acceptable for a coarse observability label, but it is not production-grade semantic classification and would be unsafe for retry or settlement behavior.

Please document explicitly that this classifier is best-effort and metrics-only, record the supported message assumptions, and warn against using its result for control or persistence logic until structured error codes or types are available.

crates/agglayer-settlement-service/src/settlement_task/tests.rs
```diff
@@ -2551,3 +2558,142 @@ fn latest_pending_attempt_fees_for_nonce_ignores_errored_and_unknown() {
...
+// Both the plain and the replacement phrasing must land in the fee bucket.
+#[case("transaction underpriced", SettlementAttemptErrorKind::Underpriced)]
+#[case(
+    "replacement transaction underpriced",
+    SettlementAttemptErrorKind::Underpriced
+)]
+// An error response the classifier does not recognize stays generic rather
+// than being mislabeled as a nonce or fee problem.
+#[case("execution reverted", SettlementAttemptErrorKind::Rpc)]
+fn classify_attempt_error_buckets_node_error_responses(
```
Ekleog-Polygon, line 2569: P2 test: Exercise metrics through the production call sites

The added settlement-task tests exercise the classifier and restored state, while the telemetry tests call the recording helpers directly. No test drives the new `record_settlement_attempt`, `record_settlement_attempt_error`, or `record_settlement_job_duration` call sites while collecting metrics. Deleting one of those calls or passing the wrong label would therefore leave the suite green, although #1687 explicitly requires the new emission paths to be covered.

Please add a task-level integration test that drives submission failure and terminal completion through these production paths and asserts the exported kind, outcome, and wallet labels.

Ekleog-Polygon (changes requested):
Thanks—the metric surface is useful, but I think the gauge and error plumbing should be simplified before merge.

The jobs gauge cannot truthfully attach one wallet to a job with attempts from multiple signers, and its `Building`/`Submitted` split does not reliably correspond to observable incidents. The attempt-kind decision should also have one source of truth, while submission errors should remain structured rather than being boxed and subsequently downcast.

I also left follow-up comments for production-path metrics coverage, the #1676 certificate-metric continuity test, and documenting the string classifier as best-effort and metrics-only.

crates/agglayer-settlement-service/src/settlement_service.rs
hadjiszs: Done. The `wallet` label is gone, and `job_state_samples()` is now
`live_job_count() -> u64`, read from the task registry.

crates/agglayer-telemetry/src/settlement.rs
hadjiszs: Done. `Building`/`Submitted` and the `state` label are gone; the gauge is now
one live-job count.

This also removed `mark_submitted`, `TaskControlHandle::state` and
`has_pending_attempt`, which only existed to serve that state. It moots the
older Codex comment on this file too: there is no `Building` any more.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: Done. `SubmitAttemptError` and `submission_outcome` are gone.
`submit_attempt_to_l1` returns `RetryCallbackError<TransportError>`, and
`classify_attempt_error` takes `&TransportError`. No boxing, no downcast.

Note: nothing can produce `SettlementAttemptErrorKind::Other` any more, so I
removed that variant too.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: Done. `build_next_attempt_with_nonce` now returns the kind, from its own
`live_tx_fees` read. The caller's duplicate branch is gone, and the three
builder tests assert the kind.

crates/agglayer-settlement-service/src/settlement_task.rs
hadjiszs: Done. The doc now says it is best-effort and metrics-only, lists the substrings
it matches, and warns against using it for retry, settlement or persistence.

crates/agglayer-settlement-service/src/settlement_task/tests.rs
hadjiszs: Done. Added `metrics_are_emitted_through_the_production_call_sites`. It drives a
failed broadcast and a successful completion through the real call sites, then
reads the exported series back.

Each assertion expects an exact value on an exact label set, so a deleted call
or a wrong label makes it fail.

crates/agglayer-node/src/node.rs
hadjiszs: This test already exists and predates this PR, which is why it is not in the
diff.

`settlement_path_metrics_are_exposed`
(`tests/integrations/tests/certificate_settlement/metrics.rs:133`) asserts
`agglayer_certificate_stage_duration_seconds_count >= 1` for `pending`,
`proven` and `candidate` after a certificate settles. It came from #1678
(20be4925d).

crates/agglayer-settlement-service/src/settlement_task.rs
```diff
@@ -815,16 +902,25 @@ impl<
...
         tx: TxEnvelope,
     ) -> Option<SettlementTaskRunResult> {
         self.save_attempt_to_db(wallet, nonce, attempt_number, &tx);
         // The nonce is recorded; other same-wallet tasks may now read it.
         drop(nonce_guard);
+        record_settlement_attempt(attempt_kind);
         match self.submit_attempt_to_l1(tx).await {
-            Ok(()) => None,
+            Ok(()) => {
+                self.control.mark_submitted();
```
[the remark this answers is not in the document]
hadjiszs: The `Building`/`Submitted` split is gone, so no job can be shown as
still building after a broadcast that may have landed.

The lost-response problem itself is unchanged, still tracked by #1321.

crates/agglayer-node/src/node.rs
Ekleog-Polygon: Thanks, you’re right — I missed the existing integration assertion in #1678.

Ekleog-Polygon (approved):
Re-reviewed the latest changes; all prior comments are addressed. LGTM.

PR #804, opened 2025-05-16
resolving issue with logging of websocket url with sensitive api key

obynonwane:
Fixes #803
### Description

When a user runs the agglayer-node, the logs may include the full WebSocket connection URL — including an embedded secret or token as the last path segment. This becomes a security/privacy risk when users share logs to support team for troubleshooting and help or even to community channel like discord for other developers to help resolve issue, unaware that the secret is being exposed.

### Example

Logs currently show something like: {"message":"Starting BlockClock with provider: wss://websocket-url/secret-key"}

This can leak sensitive credentials unintentionally.

### Expected Behavior

Sensitive tokens in URLs should be redacted in logs. For example:  {"message":"Starting BlockClock with provider: wss://websocket-url/xxxx..."}

### Recommendation

Use a sanitizer (e.g., `sanitize_ws_url`) before printing any WebSocket URLs to the console or log. Ensure this is applied consistently across all modules that output connection details.

### Impact

- Leaks secrets in public/shared logs
- Users may unknowingly compromise their node setup
- Affects troubleshooting and security hygiene

### Additional Context

This can be mitigated by adopting a standard log-sanitizing wrapper or utility for sensitive data.

PR #1718, opened 2026-07-30
fix: redact configured L1 RPC endpoints

### LINKED ISSUE #803: Sensitive WebSocket token exposed in logs: secret included in connection URL

1sgtpepper:
Prevent configured L1 endpoint credentials from entering operational logs through first-party formatting and pinned transport diagnostics, complementing the Alloy error redaction from #1432.

Store endpoints in a Serde-transparent type whose `Display` and `Debug` are always redacted. Current Agglayer runtime code exposes raw values explicitly only at its two transport constructors. Reject pinned endpoint-bearing dependency records at their affected metadata levels, including Hyper legacy-client `WARN` records. Because dependency metadata does not distinguish every sensitive and benign same-target record, some adjacent records at the pinned levels are intentionally suppressed.

Logging initialization owns the process-wide redaction policy: the first successful configuration wins, pre-existing process-global tracing or `log` ownership is rejected, and embedding callers must not race those one-shot initializers. Agglayer claims tracing before installing its `log` bridge so a tracing conflict leaves the host logger untouched; if the bridge is already owned, startup fails closed and Agglayer's tracing policy remains installed. The owned dispatch remains active while the public node entrypoint polls startup and shutdown futures.

Validation:

- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- Local builds and tests were intentionally not run; focused and full CI are pending.

BREAKING-CHANGE: `agglayer_config::L1::{node_url, ws_node_url}` now use `L1RpcUrl` instead of `url::Url`. Their TOML representation is unchanged.
BREAKING-CHANGE: `agglayer_node::main` now requires ownership of the process-global tracing dispatcher and `log` logger. Embedding callers that preinstall either global receive a startup error. If the `log` logger is already owned, Agglayer's tracing policy remains installed because the tracing and `log` globals cannot be updated transactionally. Embedding callers must leave those one-shot globals available and must not race initialization; a thread-local scoped subscriber remains supported.

CONFIG-CHANGE: None. Existing L1 TOML configuration remains unchanged.

Fixes #803.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -508,6 +508,7 @@ dependencies = [
  "tower 0.4.13",
  "tower-http",
  "tracing",
+ "tracing-log",
  "tracing-subscriber",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5719,7 +5719,6 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
- "test-log",
  "thiserror 2.0.18",
  "tokio",
  "tokio-stream",
--- a/crates/agglayer-config/src/l1.rs
+++ b/crates/agglayer-config/src/l1.rs
@@ -8,6 +8,8 @@ use url::Url;
 ///
 /// Formatting never exposes the endpoint because any URL component may contain
 /// credentials. Use [`Self::expose_url`] only when constructing a transport.
+// Serde stays transparent so configuration files preserve the endpoint;
+// Display and Debug are the redaction boundary for formatted values.
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq)]
 #[serde(transparent)]
 pub struct L1RpcUrl(Url);
--- a/crates/agglayer-config/src/l1.rs
+++ b/crates/agglayer-config/src/l1.rs
@@ -7,9 +7,11 @@ use url::Url;
 /// A configured L1 RPC endpoint.
 ///
 /// Formatting never exposes the endpoint because any URL component may contain
-/// credentials. Use [`Self::expose_url`] only when constructing a transport.
-// Serde stays transparent so configuration files preserve the endpoint;
-// Display and Debug are the redaction boundary for formatted values.
+/// credentials. [`Display`](std::fmt::Display) and [`Debug`](std::fmt::Debug)
+/// are redacted; serialization intentionally preserves the raw endpoint for
+/// configuration compatibility and must be treated as secret-bearing. Use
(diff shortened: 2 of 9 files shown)


1sgtpepper:
@hadjiszs @Ekleog-Polygon @Freyskeyd — could you review the design before I mark this ready?

I used a Serde-transparent `L1RpcUrl` so `Debug` and `Display` redact by default, while TOML remains unchanged and raw URLs are exposed only when building transports. This follows the wrapper direction from #804 and extends #1432’s Alloy error redaction to the Clock path.

The filter also covers dependency diagnostics, since removing first-party URL logs alone does not prevent endpoints from being emitted. Logging setup fails closed if an incompatible logger is already installed.

Does this scope—and the public `L1RpcUrl` and logger-ownership changes—fit #803, or should I narrow or split it? Is the HTTP/WS/WSS coverage sufficient, and should I rebase around #1692 first?

hadjiszs:
@1sgtpepper thank you for your contributions 👍 

can we just remove the log lines which let appear the URLs as suggested there https://github.com/agglayer/agglayer/issues/803#issuecomment-3004120060

1sgtpepper:
Thanks! Just to confirm, would you prefer me to narrow this PR to removing the direct Agglayer log statements that expose URLs, including the BlockClock provider log, and drop the `L1RpcUrl` wrapper and broader logging changes?

Should URL-bearing logs emitted by dependencies remain out of scope for #803? Also, should I rebase on #1692 first?

PR #1727, opened 2026-08-05
feat(settlement): respawn-capable reload and task liveness accessor

Ekleog-Polygon:
Complete the settlement admin unstick cycle by letting reload revive a
pending job after abort, task panic, or failed in-task reload.
Operators can now abort, inspect and fix stored state, then reload
without restarting the node.

A live task still receives the existing queued reload command. When
no task is registered, reload classifies storage, refuses unknown and
completed jobs, loads pending state, and spawns a fresh task and
watcher.

The liveness decision and `try_send` happen while holding the current
`task_controls` map lock, addressing finding 2 from the #1681 review.
The admin-operation lock now serializes both reload and force-remove
so two spawn-capable operations cannot create duplicate tasks.

This deliberately keeps task-owned result senders. Unlike #1681's
service-held sender map, a panic closes the channel so existing waiters
fail fast instead of hanging indefinitely:
https://github.com/agglayer/agglayer/pull/1681#pullrequestreview-4713084371

A respawn therefore creates a fresh watcher. Pre-death waiters have
already errored with "watcher closed", and their certificate may be
`InError` even if the respawned job later settles successfully on L1.
The operator must reconcile that certificate-status versus
storage-result divergence.

A failed storage load registers no fresh task or watcher and clears
any dead watcher left by the previous task, addressing review finding 4.

Add `has_live_task` as an advisory accessor for the reads follow-up.
This branch includes the teardown guard, so a panicking task
deregisters and `false` reliably means no task is registered. The
result can still become stale immediately after it is read.

Reviewer question: now that abort routinely creates a pending job
without a task and reload heals it, should `retrieve_settlement_result`
downgrade its pending-without-task log from `error!` to `warn!`? This PR
leaves it at `error!`, matching the outcome of PR 2 review.

Part of #1675.


crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -406,39 +405,112 @@ impl<
...
+        // clearing it here also leaves a load failure with no stale control or
+        // watcher for this job.
+        self.result_watchers.lock().await.remove(&job_id);
+
+        let (task_control_handle, task_control) = TaskControlHandle::new(&self.cancellation_token);
+        let stored_job = self
+            .load_stored_job(job_id, task_control)
             .await
+            .wrap_err(RpcErrorCode::Unavailable)
+            .wrap_err_with(|| format!("Failed to reload pending settlement job {job_id}"))?;
```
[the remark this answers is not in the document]
Ekleog-Polygon: Confirmed: this requires two admin RPCs for the same job to overlap in a narrow order. The respawn-capable reload must read the pending job, then an attempt mutation must commit and run its best-effort notification before the fresh task registration is installed. The documented operator flow is sequential (abort → inspect/fix → reload), so it cannot hit this window. These calls are rare and operator-driven; the mutation remains durable, and another reload reconciles the task with storage if operators do invoke them concurrently. We are accepting that recoverable concurrency risk here rather than expanding the global admin lock to every mutation in this PR.

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -406,39 +405,112 @@ impl<
...
+                if task_control.is_cancelled() {
+                    return Err(eyre::eyre!(
+                        "Settlement task {job_id} teardown after cancellation is in progress; \
+                         retry"
+                    )
+                    .wrap_err(RpcErrorCode::Unavailable));
+                }
+
+                match task_control.try_send(TaskAdminCommand::ReloadAndRestart) {
+                    Ok(()) => return Ok(()),
```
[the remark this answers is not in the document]
Ekleog-Polygon: This is only happening when two admin operations race, it's fine.

docs/knowledge-base/src/settlement-operations.md
```diff
@@ -24,8 +24,8 @@ ID and attempt number from existing logs or storage inspection.
 | Failure scenario | Method | Effect and recovery |
 |---|---|---|
 | 1. A job looks stuck and must be inspected | Not shipped: `admin_listSettlementJobs` and `admin_getSettlementJob` | Use existing logs or storage inspection until the read phase adds job discovery and detail. |
-| 2. A live task is wedged by transient or stale state | `admin_reloadSettlementTask(job_id)` | Queues a reload from storage for an existing task. It cannot respawn a missing task; restart the node in that case. |
-| 3. A job blocks a wallet's nonce pipeline | `admin_abortSettlementTask(job_id)` | Stops the in-memory task without changing stored state. Restart the node; startup recovery respawns pending jobs. |
+| 2. A task is wedged or missing | `admin_reloadSettlementTask(job_id)` | Reloads a live task from storage or respawns a missing task for a pending job. Retry if task teardown is still in progress. |
+| 3. A job blocks a wallet's nonce pipeline | `admin_abortSettlementTask(job_id)` | Stops the in-memory task without changing stored state. Inspect or fix the job, then reload it to spawn a fresh task. |
```
hadjiszs, line 28: nit: I see two 4 and two 5 right below

crates/agglayer-jsonrpc-api/src/tests/settlement_admin.rs
```diff
@@ -243,20 +317,6 @@ async fn admin_reload_settlement_task_errors_are_classified() {
...
-        .insert_settlement_job(&pending_job_id, &settlement_job())
-        .unwrap();
-    let error = call_error(
-        context
-            .admin_client
-            .request("admin_reloadSettlementTask", rpc_params![pending_job_id])
-            .await,
-        RpcErrorCode::NoLiveTask,
-    );
-    insta::assert_snapshot!("admin_reload_settlement_task__no_live_task", error);
```
hadjiszs, line 258: nit: this snapshot file can be removed then

crates/agglayer-settlement-service/src/settlement_service.rs
```diff
@@ -406,39 +405,112 @@ impl<
...
+                // will never drain the command.
+                if task_control.is_cancelled() {
+                    return Err(eyre::eyre!(
+                        "Settlement task {job_id} teardown after cancellation is in progress; \
+                         retry"
+                    )
+                    .wrap_err(RpcErrorCode::Unavailable));
+                }
+
+                match task_control.try_send(TaskAdminCommand::ReloadAndRestart) {
```
hadjiszs, line 452: Don't we need to propagate this same way of handling the cancellation also in the function `notify_live_task_of_admin_edit()` below?

hadjiszs (approved):

Ekleog-Polygon:
I'll handle your comments in a PR following #1729 ; thank you for these!

PR #1729, opened 2026-08-07
feat(settlement): admin RPC reads for settlement jobs

Ekleog-Polygon:
Add admin_listSettlementJobs and admin_getSettlementJob so operators can
discover job IDs and inspect settlement state over RPC.

Summaries include the certificate link, status, task liveness, attempt
count, latest attempt, and latest error. Details include job parameters,
all attempts and their recorded results, and the terminal result.

Keep serialization at the JSON-RPC boundary with internally tagged
attempt results and stable camelCase client-error tags. Domain settlement
types remain serde-free.

Reads are point-in-time rather than transactional. A pending job with
hasLiveTask false needs admin_reloadSettlementTask. Listing performs a
full settlement-jobs column-family scan, which is acceptable at the
current operator scale.

This completes the RPC unstick cycle: list, inspect, abort, verify gone,
reload, and verify back.

Stacked on #1727.
Part of #1675.


crates/agglayer-jsonrpc-api/src/admin.rs
```diff
@@ -456,6 +490,86 @@ where
...
+                self.read_settlement_job(job_id)?
+                    .map(|job| SettlementJobSummary::from(&job))
+                    .ok_or_else(|| {
+                        eyre::eyre!(
+                            "Settlement job {job_id} disappeared after listing its storage key"
+                        )
+                    })
+            })
+            .collect()
+    }
```
hadjiszs, line 569: One unreadable job hides every other job from the list.

Startup recovery scans this same column family and skips a job it can't decode, logging it, so one bad row can't block boot. Here `collect()` fails the whole response instead.

```suggestion
    fn read_settlement_job_summaries(&self) -> eyre::Result<Vec<SettlementJobSummary>> {
        Ok(self
            .state
            .list_settlement_job_ids()
            .wrap_err("Failed to scan settlement job ids")?
            .into_iter()
            // A job whose stored rows cannot be read back must not hide every
            // other job from the operator. Startup recovery skips and reports
            // such a job for the same reason.
            .filter_map(|job_id| match self.read_settlement_job(job_id) {
                Ok(job) => job.as_ref().map(SettlementJobSummary::from),
                Err(error) => {
                    error!(
                        ?error,
                        %job_id,
                        "Failed to read settlement job while listing; skipping"
                    );
                    None
                }
            })
            .collect())
    }
```

hadjiszs: Also, `docs/knowledge-base/src/settlement-operations.md` still documents `admin_listSettlementJobs` and `admin_getSettlementJob` as not shipped (lines 26, 36-39, 250).

Ekleog-Polygon: For the knowledge-base, I'm editing it in the last PR of the list!

hadjiszs (approved):

PR #1741, opened 2026-08-14
fix(settlement): retry request timeout errors

### LINKED ISSUE #1742: Fix panic in settlement task
An L1 RPC returning HTTP 408 is not yet handled as a transient error. This currently leads to a panic, which is easily recoverable but shouldn't happen in the first place.

Ekleog-Polygon:
Alloy's default policy does not retry HTTP 408 responses, allowing
provider request timeouts to reach the non-recoverable settlement path.

Retry standardized HTTP 408 responses. For direct JSON-RPC errors,
require both the implementation-defined -32009 code and tight timeout
wording so unrelated permanent errors remain non-transient. Keep
task-level restart behavior unchanged.

Move utility tests into a dedicated module and cover the captured
error shape plus false-positive cases.

The following verification passed:

- `cargo check --workspace --tests --all-features`
- `mdbook build docs/knowledge-base/`
- `cargo make ci-all`
- `cargo nextest run --workspace` (950 tests)

Fixes #1742


hadjiszs (approved):
maybe we should include the 504 response handling too

PR #1785, opened 2026-09-02
feat(storage): add backup metrics

hadjiszs:
Closes https://github.com/agglayer/issues/issues/5

### DIFF
--- a/crates/agglayer-telemetry/src/backup.rs
+++ b/crates/agglayer-telemetry/src/backup.rs
@@ -3,11 +3,11 @@
 //!
 //! Two queues, state and epoch. A write path raises a request, it waits,
 //! then it runs against the RocksDB backup engines.
+//!
+//! Every instrument is pushed at the event, so nothing here reads node state
+//! on a scrape.
 
-use std::{
-    sync::{Arc, Mutex},
-    time::{Instant, SystemTime, UNIX_EPOCH},
-};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use lazy_static::lazy_static;
 use opentelemetry::{global, metrics::*, KeyValue};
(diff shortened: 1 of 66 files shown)


crates/agglayer-telemetry/src/backup.rs
```diff
@@ -0,0 +1,338 @@
...
+    Failure,
+}
+
+/// Histogram buckets in seconds.
+///
+/// Not [`crate::certificate::DURATION_BUCKETS_SECONDS`], which starts at
+/// 0.5 s: an idle engine serves a request almost immediately, so that set
+/// would report every healthy queue wait in one bucket.
+const BACKUP_DURATION_BUCKETS_SECONDS: &[f64] = &[
+    0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0, 30.0, 60.0, 120.0, 300.0, 600.0, 1800.0,
```
Ekleog-Polygon: Considering our current duration estimated at ~9min, I think we should have buckets that are a bit more relaxed on the lower end (I don't really see the point of the 50ms bucket for example). And on the contrary we should probably have one bucket per minute up to 15min of waiting time.

Also, this seems to be reused for both queue wait and backup duration, but it might make sense to separate the two — though this would be less pressing, if we have the suggested 1s / 30s / 1min / 2min / 3min / etc. / 15min / 20min / 25min / 30min / more buckets above.

crates/agglayer-telemetry/src/backup.rs
```diff
@@ -0,0 +1,338 @@
...
+/// global meter provider is installed (see [`crate::ServerBuilder`]), and at
+/// most once per provider.
+pub fn register_backup_metrics(metrics: &Arc<BackupMetrics>) {
+    let metrics = Arc::downgrade(metrics);
+
+    // The instrument handle is intentionally dropped: the callback
+    // registration lives in the meter provider, not in the handle.
+    let _ = global::meter(AGGLAYER_NODE_BACKUP_OTEL_SCOPE_NAME)
+        .f64_observable_gauge(BACKUP_OUTSTANDING_AGE_SECONDS)
+        .with_description("Age of the backup request currently being served, in seconds")
```
Ekleog-Polygon: It'd be simpler, and more prometheus-like, to just expose the enqueuing time of the backup request currently being served as a gauge. Then, dashboards can compute polling_time() - enqueuing_time() ; and it'd avoid having to write a custom metrics callback.

Ekleog-Polygon (commented):
Overall LGTM! I just submitted two comments to try and simplify the code — they should be very simple to handle 😄

Ekleog-Polygon (approved):
LGTM, thank you for the changes!

PR #1807, opened 2026-09-16
fix(storage): batch pending pointer and certificate

Bruce039:
The pending-store insert path currently writes the latest pointer first and the certificate body afterwards. If the body write fails, the pointer can be persisted on its own and point at a certificate that is not in the pending queue.

This changes the insert to stage both writes in the same RocksDB `WriteBatch` and publishes the pending metric only after the batch commits.

I also added a regression test that opens the pending DB without the proto certificate CF, forces the body write path to fail, and checks that the latest pointer is not persisted.

Validation:
- exact PR/issue duplicate search: no matching fix found
- regression coverage added for the partial-write failure path

CONFIG-CHANGE: none

BREAKING-CHANGE: none


Bruce039:
@codex review

PR #1810, opened 2026-09-16
fix(storage): propagate epoch startup iterator errors

Bruce039:
Epoch-store startup currently drops iterator errors in three places:

- start-checkpoint scan uses `filter_map(|v| v.ok())`
- end-checkpoint scan does the same
- the reverse certificate scan treats `Some(Err(_))` like an empty database

That lets corrupt/decode-failing rows be silently ignored and can make startup reconstruct incomplete state.

This changes all three paths to propagate the iterator error instead.

Regression coverage writes invalid raw bytes into each affected column family and verifies that reopening the epoch store fails instead of continuing with partial state.

Validation:
- exact PR/issue duplicate search: no matching fix found
- covered start checkpoint, end checkpoint, and latest certificate reverse scan

CONFIG-CHANGE: none

BREAKING-CHANGE: none


Bruce039:
@codex review

PR #1813, opened 2026-09-16
fix(clock): prevent block height overflow wrap

Bruce039:
The runtime BlockClock increment currently does `fetch_add(1)` before checking whether the returned value can be incremented. At `u64::MAX`, the atomic has already wrapped to zero by the time `checked_add(1)` returns `None`.

This replaces that with a checked atomic update. A runtime overflow now returns `BlockHeightOverflow` without mutating the shared block-height counter.

The regression test covers both sides of the boundary:
- `MAX - 1 -> MAX` still succeeds
- the next increment returns an overflow error and leaves the atomic at `MAX`

I also checked the historical clock overflow work in #28; that PR was for the TimeClock path and does not fix this BlockClock runtime increment.

Validation:
- exact current PR/issue duplicate search: no matching BlockClock fix found
- deterministic unit regression coverage added

CONFIG-CHANGE: none

BREAKING-CHANGE: none


Bruce039:
@codex review

